PR #2, opened 2019-10-22
Create a new rails application with the govuk design system

lozette:
## Changes in this PR

Add govuk-frontent to the application with npm. Add some basic strings and the most basic homepage. 

*Note:* This PR is merging into `develop`, not `master` as per the previous project PR. 

## Screenshots of UI changes

Screenshot 2019-10-22 at 10 24 01


.env.example
```diff
@@ -9,5 +9,4 @@
 ROLLBAR_ACCESS_TOKEN=ROLLBAR_ACCESS_TOKEN
 ROLLBAR_ENV=development
 
-# TODO: Replace `rails-template` with the name of the app.
-DATABASE_URL=postgres://postgres@localhost:5432/rails-template-development
+DATABASE_URL=postgres://postgres@localhost:5432/roda-development
```
tahb: A benefit of rebasing this branch on top of develop (rather than merging) means that this other unrelated work will not show in this PR. Making self contained and easier to review. 

I think we can fix this by:

1.  doing an interactive rebase and deleting this merge commit `f2dbe75d5b6069f3f91628d49b76d0c8dc39be94`
2. rebasing this new work on top of the existing code on develop `git rebase origin/develop`
3. updating this pull request with the new history `git push --force-with-lease`

config/database.yml
```diff
@@ -7,6 +7,9 @@ default: &default
 
 development:
   <<: *default
+  username: postgres
```
tahb: I would have hoped that these values would be configured through DATABASE_URL in `.env.development.local`, was this not the case? If we need it, this feels like a separate commit.

lozette: No this is my misunderstanding! Will fix

app/views/visitors/index.html.erb
```diff
@@ -1 +1,3 @@
-<h3>Welcome</h3>
+<%= content_for :page_title_prefix, I18n.t('page_title.home') %>
+
+<h3>BEIS - Report Overseas Development Assistance</h3>
```
tahb: We could use our new local file to supply this content consistently across the app with `I18n.t(app.title)`.

app/views/visitors/index.html.erb
```diff
@@ -1 +1,3 @@
-<h3>Welcome</h3>
+<%= content_for :page_title_prefix, I18n.t('page_title.home') %>
```
tahb: This is the fault of the rails-template for having an `erb` file where we currently have `haml` as our default for Rails apps, can we switch over to haml?

lozette: I did wonder!!

README.md
```diff
@@ -18,6 +16,16 @@ TODO: Add a summary of who the application is for and what it will do.
 
   Our intention is that the example should include enough to get the application started quickly. If this is not the case, please ask another developer for a copy of their `/.env.development.local` file.
 
+  ```bash
```
lozette: This looks a bit confused now as the changes to the README on master are not reflected

tahb: I see. We're looking to only have your new addition showing here at the moment. When it is ready, we can `git rebase origin/develop` to pull in my changes to the README. Then it should make sense.

lozette: Right, this PR only has the bit I added now 👍

tahb: Looks good. Now the rebase should add my bits as existing changes, and it should ask you to resolve the conflict and slot it into the right place.

PR #3, opened 2019-10-22
Use Travis for continuous integration

tahb:
## Changes in this PR

- Use Travis as BEIS's tool of choice
- We have no platform to deploy to so this only runs tests on pull requests and pushes. Soon we will be able to configure it to deploy on pushes to master and develop to GPaaS


tahb:
I don't know why 2 builds run here. The push passes okay but this `pr` build seems to fail for odd reasons that look like dependency issues. No idea why it works for one, and fails for another.

PR #4, opened 2019-10-25
Dockerize application

lozette:
## Changes in this PR

Trello: https://trello.com/c/QgNqQ0cq/39-dockerize-new-rails-application

This PR Dockerizes the application. The app will now run with either:

`docker-compose up` to use docker-compose

or

`docker build . -t roda:latest && docker run -p 3000 roda:latest` to not use docker-compose, as required by the ticket

*Note:* I'm not sure if `yarn.lock` should have been committed? Happy to remove if it's not needed.


lozette:
Also note this PR should fix the Travis build for https://github.com/UKGovernmentBEIS/beis-report-overseas-development-assistance/pull/3

Gemfile
```diff
@@ -20,40 +20,34 @@ gem "sass-rails", "~> 6.0"
...
 gem "uglifier", ">= 1.3.0"
+gem "dotenv-rails"
 
 group :development do
   gem "listen", ">= 3.0.5", "< 3.2"
   gem "spring"
   gem "spring-watcher-listen", "~> 2.0.0"
   gem "web-console", ">= 3.3.0"
+  gem "better_errors"
+  gem "html2haml"
```
leeky: There are gems in here that don't seem to be related to the PR..?

Gemfile
```diff
@@ -20,40 +20,34 @@ gem "sass-rails", "~> 6.0"
 gem "turbolinks", "~> 5"
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby]
 gem "uglifier", ">= 1.3.0"
+gem "dotenv-rails"
```
leeky: Should `dotenv-rails` be left in the development / test group? Production usually has environment variables set directly, rather than using `.env` files.

Gemfile
lozette: Good point, I had a bit of a re-org of the gemfile and it ended up in here, I think I'll yank it

PR #5, opened 2019-10-28
(43) Use Heroku temporarily

leeky:
Remove yarn.lock to allow the Heroku node.js buildpack to correctly detect the correct lock file in order to install the `govuk-frontend` package. 

NB: I haven't added an ADR to document the decision to use Heroku as it is only being use temporarily, until we are able to access and deploy to [GPaaS](https://www.cloud.service.gov.uk/).


tahb:
@leeky I do think we should still create ADR for decisions like this even when temporary. We hope to move to GPaaS but we can't predict the future. Whilst we intend to move to GPaaS, we don't know when and we haven't crossed the various bridges in getting set up. Any of which may force us to remain on Heroku for an unexpected length of time.

PR #10, opened 2019-10-30
Add a CHANGELOG.md and reference it in the README

lozette:
## Changes in this PR

Trello: https://trello.com/c/aabjqNFM/46-create-a-changelog

Add a blank CHANGELOG.md and details of how to use it in README

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,2 @@
+# Change Log
+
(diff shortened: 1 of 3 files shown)


tahb:
I think we should record this decision with an ADR. As it is not yet merged into our rails-template, we may be able to take a copy of [the proposed ADR record](https://github.com/dxw/rails-template/pull/26/files#diff-ee2f6bd44d86af3083ba6767c8e6bfd4).

lozette:
ADR added

PR #12, opened 2019-10-31
Deploy to staging with containers

lozette:
## Changes in this PR

Trello: https://trello.com/c/TnhcILEI/45-deploy-to-staging-with-containers

Build & deploy the app on Heroku with Docker


tahb:
From memory I think we will nee to change the deployment instructions for Travis to deploy with the container mechanism, rather than the existing build pack one. On Report a Defect (and Circle CI) I had to do something like this: https://github.com/LBHackney-IT/report-a-defect/blob/develop/.circleci/config.yml#L73.

lozette:
I've run `heroku stack:set container -a beis-roda-staging` which allegedly changes the deployment instructions (according to https://devcenter.heroku.com/articles/build-docker-images-heroku-yml) - do you think this is enough?

tahb:
It may have changed but I definitely fell down this hole myself. Perhaps the problem is that we will be pushing to our github remote rather than a heroku one, which the last documented step suggests to do: `git push heroku master`.

I think I then tried to push to this remote https://github.com/LBHackney-IT/report-a-defect/blob/develop/.circleci/config.yml#L69 but found it didn't deploy new containers, only build packs.

Is there a way we can test this without having to open another pull request if it needs more work?

tahb:
I've sent you a DM with an idea for how we could achieve this @lozette.

tahb:
We planned at standup that I would recreate this PR with the same intent.

PR #17, opened 2019-11-04
Add Google Tag Manager

tahb:
## Changes in this PR

- Google Tag Manager used as a middleman to achieve data being reported to Google Analytics. This enables greater flexibility of analytic configuration as it would not require code changes, instead changes through the Google Tag Manager. This would also allow other sources of analytics to be configured such as Google Conversion Tracker, or other third party tools that may or may not already be used by BEIS.
- add Javascript snippets to enable Google Tag Manager to report on events, removing GA code added via the template. This should have a change pushed downstream as this approach looks superior and is already being used on 2 other projects
- I have configured a single 'tag' within the tag manager to report on simple page views for both a staging and production environment. The tag configuration looks at the environment name value in order to work out what GA code to use. This is done through a lookup list variable following [this guide](https://marketlytics.com/blog/google-tag-manager-environments)
- everyone on the team has been sent an invite to both tools, which the reviewer can also use to verify the configuration
- the instruction is to copy the Google JS code snippet directly. There is no warning against exposing the `gtm_auth` value neither directly nor on further searching. I have looked at other people using this value and have found 2 (of many) examples that follow the direction from Google: https://github.com/alphagov/govuk-puppet/pull/8254/files (note this is being removed for other reasons), https://github.com/uktrade/export-opportunities/commit/3715c2a65279e29042eb0a187a53b8fbf7adfc2a. I have used environment variables to enable us to configure the different environments, rather than to obscure the values. The values will be calculated into the users browser on read in any case.

## Screenshots

Screenshot 2019-11-04 at 18 57 32

## Next steps

- [x] add the new environment variable to staging and production applications
- [ ] deploy to staging and verify staging page views are being collected


CHANGELOG.md
```diff
@@ -1,2 +1,5 @@
 # Change Log
 
+## Unreleased changes
```
lozette: Woo! 👍

PR #22, opened 2019-11-07
(55) Create a minimal organisation

lozette:
Trello: https://trello.com/c/32LBQ5EX/55-create-a-minimal-organisation

## Changes in this PR

Create a basic Organisation model, with `index`, `show` and `create` views. 

The organisation type, language code and default currency are taken from the IATI standards. The data lists have been downloaded from https://andylolz.github.io/iatistandard-multilingual/en/203/introduction/

## Screenshots of UI changes

Screenshot 2019-11-07 at 13 47 27

Screenshot 2019-11-07 at 13 47 42

Screenshot 2019-11-07 at 13 47 53


spec/features/organisations_can_be_created_spec.rb
```diff
@@ -0,0 +1,28 @@
+RSpec.feature "Users can create organisations" do
+  scenario "successfully creating an organisation" do
```
tahb: This is a really good end-to-end test 👍

config/locales/en.yml
```diff
@@ -8,16 +8,36 @@ en:
...
+    submit: "Submit"
+    name:
+      label: "Name"
+      hint: "The name of the organisation"
+    organisation_type:
+      label: "Organisation type"
+      hint: "IATI Organisation type"
+    default_currency:
+      label: "Default currency"
+      hint: "IATI default currency"
```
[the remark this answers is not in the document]
lozette: I'll amend it slightly to remove "code", as we're not displaying the code, we're displaying the display name

app/views/organisations/show.html.haml
```diff
@@ -18,12 +18,12 @@
         %dt
           = t("organisation.type")
         %dd
-          = @organisation.organisation_type
+          = t("organisation.organisation_type.#{@organisation.organisation_type}")
         %dt
           = t("organisation.language_code")
         %dd
-          = @organisation.language_code
+          = t("organisation.language_code.#{@organisation.language_code.downcase}")
```
tahb: I think downcasing is a very good idea to make sure this works reliably. It prompts to we think we should probably have some test coverage for it too.

From there I think it might make sense to have this set on a global getter for an organisation record rather than in the views, so they don't have to remember _every_ single time. This then avoids us having to test this low level edge case with features, and can use faster unit tests for it too!

We could add a spec to the existing model but then I think that data logic be the same as view logic, they should be decoupled so that how we process data isn't affected by the presentational logic. It should be able to be capitalised in our database, with no affect on the views.

This would lead me to create the  first presenter object for an organisation, with 2 small unit tests.

```
# app/presenters
class OrganisationPresenter < SimpleDelegator
  def language_code
    super.downcase
  end

  def default_currency
    super.downcase
  end
end

app/organisation_controller.rb
@organisation = OrganisationPresenter.new(organisation)
```

app/views/organisations/show.html.haml
```diff
@@ -1,29 +1,29 @@
...
         %dt
           = t("organisation.language_code")
         %dd
-          = t("organisation.language_code.#{@organisation.language_code.downcase}")
+          = t("organisation.language_code.#{@organisation_presenter.language_code}")
         %dt
           = t("organisation.default_currency")
         %dd
-          = t("organisation.default_currency.#{@organisation.default_currency.downcase}")
+          = t("organisation.default_currency.#{@organisation_presenter.default_currency.downcase}")
```
tahb: I think we can remove the `.downcase` now, whoop.

PR #28, opened 2019-11-11
Feature/64 authenticate all controllers by default

tahb:
This is another and more detailed attempt based on the feedback of https://github.com/UKGovernmentBEIS/beis-report-overseas-development-assistance/pull/25. Hopefully we now have a clearer picture of why I've arrived here.

To enable all staff controllers to be easily authenticated, a design decision has been made to group all controllers by the three known categories of authentication `api`, `public` and `staff`. It is imagined that within `staff`, Pundit will later be use to control authorisation. This allows `include Secured` to be added to `Staff::BaseController` without having its before_actions spill into other controllers.

I recognise the same _could_ be achieved with a flat structure with the use and skipping of before_actions and I hope to describe why I think this decision has other advantages.

I think the folder structure provides better conceptual clarity. It is easier to understand the makeup of the application at a glance, it's easier to navigate and easier to know where you need to make changes, with confidence that we have protection from the relevant base class. 

The structure is also preferable as the use 'before actions' can quickly get out of hand. With a flat controller structure, the only way to work out what influences what and in what order is to either go through it each time or hold it in your head. These both get exponentially harder the more the pattern is used in my opinion. There is the added complexity when we need to have more complex before_action chaining, having to safely understand and maintain `before_action :foo, :bar, logged_in_using_omniauth?, :baz` is hard to do, for little return.

I believe grouping controllers in such a way is at least a somewhat established pattern. I do appreciate that in the following examples not all controllers are grouped and some remain at root level, I think that as nearly everything will be behind auth there will be no public pages.

- https://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing
- https://github.com/dxw/mytvh-web/tree/develop/app/controllers
- https://github.com/dxw/DataSubmissionServiceAPI/tree/develop/app/controllers

As I said I do see the other way to achieve this. I'm open to learning about disadvantages to the chosen approach as I know I might have overlooked something.

## Changes in this PR

- authenticate all staff controllers by default
- public controllers/pages are explicitly grouped as so
- empty api folder has been added as a placeholder to provide future direction
- delete unused about page
- omniauth test setup refactored into a single support file, and after test cleanup added to restore a blank state between tests

## Alternative

To document the alternative and to illustrate the problem, this is the combination of before actions and skipping should do the same thing *for now* until the application grows in complexity.

```
app/controllers/application_controller
app/controllers/auth0_controller
app/controllers/visitors_controller
```

```
class ApplicationController < ActionController::Base
  include Secured
  skip_before_action logged_in_using_omniauth?, only: :health_check 
end

class Auth0Controller < ApplicationController
  skip_before_action logged_in_using_omniauth?
  def callback

class VisitorsController < ApplicationController
  skip_before_action logged_in_using_omniauth?
end
```


app/controllers/application_controller.rb
```diff
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  helper_method :current_user
+  include Secured
```
leeky: Is it worth extracting the code out of the `Secured` module here so there's an explicit `before_action` that gets cancelled out by a `skip_before_action`?

config/routes.rb
```diff
@@ -1,14 +1,17 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
-  get "health_check" => "application#health_check"
+  scope module: "public" do
```
leeky: Enforcing clear namespacing which doesn't add cruft to the user-facing URLs :+1:

spec/support/authentication_helpers.rb
```diff
@@ -9,4 +9,8 @@ def mock_successful_authentication(uid: "12345", name: "Alex", email: "alex@exam
       }
     )
   end
+
+  def stub_authenticated_session(name: "Alex", email: "alex@example.com")
```
leeky: I'm unsure what the difference between `mock_successful_authentication` and `stub_authenticated_session` is here? Do we need both? Could they be combined?

tahb: As I understand it `mock_successful_authentication` will allow us to test the omniauth callback phase. This essentially fakes the request phase and makes our fake response to our callback endpoint where we set a session and redirect.

I think this is useful to test in this detail on our authentication feature specs 👍 

So long as we have those feature specs providing this coverage, I think we can remove this overhead from every other test to make things run a bit quicker. These may mount up quickly when we have hundreds of test. In order to do test everything else I think we can stop testing the omniauth handshake each time by setting up a trusted session ourselves using `stuf_authenticated_session`. This is then checked by `logged_in_using_omniauth?` when we visit a particular page in a feature test and allowed through.

app/controllers/application_controller.rb
tahb: This change was committed by accident as a transitional change that wasn't squashed away. I have now done that and moved this to `Staff::BaseController`. As it's no longer in `ApplicationController` I think the likely hood that it will need to be skipped goes way down to a point where I'm thinking that we can make it more explicit if the included module named `Secured` is not enough.

PR #32, opened 2019-11-13
(67) Missing I18n translations cause test failure

leeky:
Set up the `i18n-tasks` gem and associated RSpec test, to warn us of references to translations in our code, where we haven't set the associated key in `config/locales/*.yml`.

It also warns if the locale files aren't normalised (i.e. sorted keys etc). If this particular change is unpopular, we can remove the associated test in future.


leeky:
@lozette Actually whilst the IATI codelist didn't need normalising, I'd forgotten to include that folder in the list of I18n file that were scanned. 

Have now fixed this, and also renamed `iati.yml` to `iati.en.yml` as this is the English version of the codelists.

config/i18n-tasks.yml
```diff
@@ -0,0 +1,123 @@
...
+# Read and write translations.
+data:
+  ## Translations are read from the file system. Supported format: YAML, JSON.
+  ## Provide a custom adapter:
+  # adapter: I18n::Tasks::Data::FileSystem
+
+  # Locale files or `File.find` patterns where translations are read from:
+  read:
+    - config/locales/%{locale}.yml
+    - config/locales/**/*.%{locale}.yml
```
lozette: 👍

PR #34, opened 2019-11-14
(84) Application uses Rails 6 defaults

leeky:
Remove the Rails 5.2 defaults that were still being used, because dxw's rails template was originally built from this version.


lozette:
Wonder if we need to push something upstream for rails-template for this?

PR #35, opened 2019-11-14
(52) When users sign out of the application, they are also signed out of Auth0

lozette:
## Changes in this PR

1. Move Sign Out link to a new navigation in the header
1. Log users out of Auth0 when they log out of the application (https://auth0.com/docs/quickstart/webapp/rails#logout)
1. Remove the Sign out feature test as it has been superseded by a request spec. As we are now going outside of the application to log a user out, it makes sense to test this with a request spec. 

## Screenshots of UI changes

Screenshot 2019-11-14 at 15 02 57

## Next steps

- [x] Add `Allowed Logout URLs` to Auth0 development
- [x] Add `Allowed Logout URLs` to Auth0 staging
- [ ] Add `Allowed Logout URLs` to Auth0 production


app/controllers/application_controller.rb
```diff
@@ -1,2 +1,23 @@
 class ApplicationController < ActionController::Base
+  def sign_out
```
leeky: This doesn't feel like it belongs in `ApplicationController`, particularly as some controllers that inherit from this don't have any authentication. How about moving this to `Staff::SignoutController` or similiar?

.env.test
```diff
@@ -6,3 +6,5 @@
 # Reference: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
 
 DATABASE_URL=postgres://postgres@localhost:5432/roda_test
+AUTH0_CLIENT_ID=123456
```
leeky: Is it worth stubbing `AUTH0_CLIENT_ID` and `AUTH0_DOMAIN` in the tests that require them rather than setting them globally here?

app/views/shared/_navigation.html.haml
```diff
@@ -0,0 +1,6 @@
+%nav
+  %ul.govuk-header__navigation
+    -if defined?(current_user) && current_user
```
leeky: It's a pity that `current_user` doesn't exist everywhere so this could be `if current_user` instead, but this works 👍

spec/support/authentication_helpers.rb
```diff
@@ -10,11 +10,6 @@ def mock_successful_authentication(uid: "12345", name: "Alex", email: "alex@exam
     )
   end
 
-  def log_in
```
leeky: :+1:

app/controllers/application_controller.rb
lozette: I don't think they belong in `Auth0Controller` as they log users out of the application as well as Auth0, but I'll see if a `Staff::` controller works

leeky: There's definitely something odd happening here, so I'm accepting this for now!

PR #36, opened 2019-11-14
Add some reminders to the PR template about ADRs, Changelog and envir…

lozette:
…onment vars

## Changes in this PR

Add:

- [ ] Is an ADR required?
- [ ] Is a changelog entry required?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,3 +7,7 @@
 ### After
 
 ## Next steps
+
+- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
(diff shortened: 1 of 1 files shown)


lozette:
Should I add an ADR about this change? 🤔

tahb:
That's a good point. We could include some definitions here to help us know when to add them.

I'd say that:

- ADR = Any change to the architecture. That could be changes to the infrastructure like using containers or adding a Sidekiq process or inclusions of dependencies such as Auth0
- Changelog = Everything should have an update, apart from small tweaks like this that have no impact outside of the dev team

lozette:
Updated!

PR #38, opened 2019-11-19
(61) Create users with organisations

tahb:
## Changes in this PR

- refactoring to organisation locale use to make it easier to add a new user form
- handle the scenario where a local user record does not match Auth0 and the need for a "authenticated, but not authorised" happens. The 401 error is intended to be reusable. We don't have a support email so the call to action is limited to asking the to contact the person who invited them. Not ideal but can be replaced later.
- add climate control
- add Auth0 API, influenced by CCS RMI which had a few iterations 
- create users locally with a new form
- user creation requires the creation of the user within Auth0, database transactions are used to ensure synchronisation between local and external sources of user information
- users can be associated with multiple organisations, or none as it is not mandatory yet until we know more. I imagine it wouldn't make sense for admins to belong to an organisation.
- I have added Auth0 permissions to enable the same set of environment variables to create and read grants, which is required for the token creation which happens before every API request

## Screenshots of UI changes

Screenshot 2019-11-19 at 12 18 15
Screenshot 2019-11-18 at 18 12 53
Screenshot 2019-11-18 at 18 13 05
Screenshot 2019-11-19 at 12 52 30

## Next steps

### DIFF
--- a/app/assets/stylesheets/forms.css.scss
+++ b/app/assets/stylesheets/forms.css.scss
@@ -1,0 +1,6 @@
+.govuk-checkboxes__label::after {
+  top: 11px;
+  left: 7px;
+  width: 27px;
+  height: 12px;
+}
--- a/app/controllers/staff/base_controller.rb
+++ b/app/controllers/staff/base_controller.rb
@@ -3,7 +3,7 @@ class Staff::BaseController < ApplicationController
 
   helper_method :current_user
   def current_user
-    @current_user ||= User.where(identifier: signed_in_user_identifier).first_or_create! { |user|
+    @current_user ||= User.find_by!(identifier: signed_in_user_identifier) { |user|
       user.name = session.dig(:userinfo, "info", "name")
       user.email = session.dig(:userinfo, "info", "email")
     }
--- a/app/controllers/staff/base_controller.rb
+++ b/app/controllers/staff/base_controller.rb
@@ -1,12 +1,22 @@
+class UserNotAuthorised < StandardError; end
 class Staff::BaseController < ApplicationController
   include Secured
 
+  rescue_from(UserNotAuthorised) do
+    render "pages/errors/not_authorised", status: 401
+  end
+
   helper_method :current_user
   def current_user
-    @current_user ||= User.find_by!(identifier: signed_in_user_identifier) { |user|
-      user.name = session.dig(:userinfo, "info", "name")
-      user.email = session.dig(:userinfo, "info", "email")
-    }
+    @current_user ||= if session.dig(:userinfo)
+      User.find_by!(identifier: signed_in_user_identifier) do |user|
+        user.name = session.dig(:userinfo, "info", "name")
+        user.email = session.dig(:userinfo, "info", "email")
+      end
+    end
+  rescue ActiveRecord::RecordNotFound
+    repudiate!
+    raise UserNotAuthorised
   end
 
   def sign_out
@@ -16,5 +26,10 @@ class Staff::BaseController < ApplicationController
 
   def signed_in_user_identifier
     session.dig(:userinfo, "uid")
+  end
+
+  private def repudiate!
+    session.delete(:userinfo)
+    @current_user = nil
   end
 end
--- a/app/controllers/staff/base_controller.rb
+++ b/app/controllers/staff/base_controller.rb
@@ -1,35 +1,4 @@
-class UserNotAuthorised < StandardError; end
 class Staff::BaseController < ApplicationController
   include Secured
-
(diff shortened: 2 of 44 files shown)


spec/support/auth0_helpers.rb
```diff
@@ -0,0 +1,18 @@
+module Auth0Helpers
```
tahb: These methods were all lifted from CCS RMI so I am commenting them in as I need them but they do not end up as commented out.

app/models/result.rb
```diff
@@ -0,0 +1,9 @@
+Result = Struct.new(:success) {
```
tahb: This isn't modelling any data so there is an argument that it shouldn't be in models. Perhaps it should be in services/ or lib/ ?

app/services/create_user.rb
```diff
@@ -0,0 +1,33 @@
+require "./lib/auth0_api"
+
+class CreateUser
```
tahb: This service is refactored in a later commit to split the responsibilities of creating local versus creating in auth0.

config/locales/en.yml
```diff
@@ -22,6 +22,7 @@ en:
       submit: Submit
     user:
       create:
+        failed: The service is experiencing issues creating new users and the team has been alerted to the problem.
```
tahb: There is nothing the user can do correct the issue if this scenario happens. The best we can do at this point (without a support desk) is to reassure them that we have noticed the Rollbar event and will be dealing with it.

app/controllers/staff/base_controller.rb
```diff
@@ -1,12 +1,22 @@
+class UserNotAuthorised < StandardError; end
```
tahb: Base controller is not becoming quite heavy with auth logic. Should we move it to an authentication concern?

app/controllers/staff/base_controller.rb
```diff
@@ -1,12 +1,22 @@
...
-      user.email = session.dig(:userinfo, "info", "email")
-    }
+    @current_user ||= if session.dig(:userinfo)
+      User.find_by!(identifier: signed_in_user_identifier) do |user|
+        user.name = session.dig(:userinfo, "info", "name")
+        user.email = session.dig(:userinfo, "info", "email")
+      end
+    end
+  rescue ActiveRecord::RecordNotFound
+    repudiate!
```
tahb: Repudiating the session isn't strictly required as a bad session will always cause an error in this method. It will however allow the before action in Secured to trigger earlier, and remove the need to look up the database a second time. We want each authentication attempt to have a fresh session and a clean slate to pick up any updates made in between.

app/assets/stylesheets/forms.css.scss
```diff
@@ -0,0 +1,6 @@
+.govuk-checkboxes__label::after {
```
tahb: Maybe there is a better way to fix this, or the cause for this is something I have not foreseen?

app/views/staff/organisations/new.html.haml
```diff
@@ -17,10 +17,10 @@
           = f.input :language_code,
                     wrapper: 'select',
                     collection: yaml_to_options("language_code"),
-                    hint: I18n.t("create_organisation.language_code.hint")
+                    hint: I18n.t("form.organisation.language_code.hint")
```
lozette: Do you think these would be better in `config/locales/simple_form.en.yml` or is that overkill?

spec/features/staff/users_can_invite_new_users_spec.rb
```diff
@@ -43,6 +43,16 @@
 
   context "when the name and email are not provided" do
     it "shows the user validation errors instead" do
+      visit new_user_path
+
+      expect(page).to have_content(I18n.t("page_title.users.new"))
+      fill_in "user[name]", with: "" # deliberatly omit a value
```
lozette: `deliberately`

app/controllers/staff/base_controller.rb
lozette: I think yes

app/views/staff/organisations/new.html.haml
tahb: That sounds like a reasonable way to escalate if this `form` grouping proves to be painful and not fix anything.

tahb:
Thanks for looking. I've addressed the feedback now. In doing so I noticed that 20 new specs have added 9 seconds onto the test suite, this is a bad smell. It could feels like it's related to something outside of this PR though, given the existing 21 specs run in 9 seconds on `develop`.

We could merge to unblock your work and I look at this as a new PR?

lozette:
Works for me!

PR #40, opened 2019-11-20
(65) Users can create funds and activities

lozette:
## Changes in this PR

Trello: https://trello.com/c/Z51XjHBK/65-users-can-create-funds-with-a-one-page-form

Introduce `fund` and `activity` records. The fund is a basic object with a name, and it belongs to an organisation.

The activity belongs to a fund, and has various attributes, some of which are populated by IATI codelists.

## Next steps


db/migrate/20191118150004_create_activities.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.string :identifier
+      t.string :sector
+      t.string :title
+      t.text :description
+      t.string :status
+      t.date :planned_start_date
+      t.date :planned_end_date
+      t.date :actual_start_date
+      t.date :actual_end_date
+      t.string :recipient_region, default: "998"
```
tahb: These default options from the tickets were intended (my bad) to be default form options rather than database defaults. I think those default values wouldn't make sense when being filled out from the perspective of a grant level activity, hopefully we can use form helpers to provide sensible defaults dynamically to help the user.

I think these 3 defaults can be taken out 👍

db/migrate/20191118150004_create_activities.rb
```diff
@@ -0,0 +1,22 @@
+class CreateActivities < ActiveRecord::Migration[6.0]
+  def change
+    create_table :activities, id: :uuid do |t|
+      t.string :identifier
```
tahb: We should _probably_ add a uniqueness constraint here as it's quite important that it's universally unique for IATI.

spec/factories/activity.rb
```diff
@@ -0,0 +1,15 @@
+FactoryBot.define do
+  factory :activity do
+    title { Faker::Lorem.sentence }
+    identifier { "GB-GOV-13-GCRF" }
```
tahb: We'll be making a lot of activities so worth making this dynamic so they are unique, something like:

```
sequence(:identifier) {|n| "GB-GOV-13-GCRF-#{n}" }
```

app/helpers/codelist_helper.rb
```diff
@@ -1,13 +1,13 @@
...
+module CodelistHelper
+  def yaml_to_options(entity, type)
+    yaml = load_yaml(entity, type)
     data = yaml["data"]
     data.collect { |item| [item["name"], item["code"]] }
   end
 
-  def load_yaml(type)
-    YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/IATI/#{IATI_VERSION}/organisation/#{type}.yml"))
+  def load_yaml(entity, type)
```
tahb: 👍

tahb: This is also a really well focused commit, perfect.

app/controllers/staff/activities_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      flash[:notice] = I18n.t("form.activity.create.success")
+      redirect_to activity_path(@activity)
+    else
+      render :new
+    end
+  end
+
+  private
+
+  # This can/should be made better in future iterations!
```
tahb: Sounds like it could be reused by other controllers to come. The method itself isn't used in the commit that adds it, was it meant to be left in?

tahb: Just seen the commit is `WIP`

app/controllers/staff/organisations_controller.rb
```diff
@@ -4,7 +4,7 @@ class Staff::OrganisationsController < Staff::BaseController
   include Secured
 
   def index
-    @organisations = Organisation.all
+    @organisations = current_user.organisations
```
tahb: If we make this change every user can only see those that they themselves are associated with. I think from dicussions so far that `administrators` and `fund managers` will be able to manage all users and organisations.

Can we do this in such a way that the future administrator role can see all organisations?

app/helpers/codelist_helper.rb
lozette: Thanks!

app/controllers/staff/activities_controller.rb
lozette: Yeah, I've decided to take it out as it's out of scope. I've left the tests in (skipped) to remind us to do it later!

app/controllers/staff/organisations_controller.rb
lozette: I have take nit out for now but left the tests in (skipped). I think this should be handled with roles?

leeky: Yeah, this can be done with a Pundit scope.

spec/features/staff/users_can_create_a_fund_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  context "when the user is not logged in" do
+    it "redirects the user to the root path" do
+      page.set_rack_session(userinfo: nil)
+      visit funds_path
+      expect(current_path).to eq(root_path)
+    end
+  end
+
+  scenario "successfully creating a fund" do
+    visit new_fund_path
```
tahb: I think it is valuable to structure feature tests in such a way that they simulate exactly what the user would do, end-to-end. By starting from the landing page (dashboard_path) and clicking everything in order just as a real user would until they have finished the journey, our test will give us confidence that that feature is available and working.

With the current approach of visiting `new_fund_path` directly in all our tests we have skipped a step that should e responsible for driving us provide navigation.

So long as we have one happy path test that gives us this confidence, I think it is then reasonable to skip the step in subsequent tests for things like validation for efficiency sake.

spec/features/staff/users_can_create_an_activity_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  context "when the user is not logged in" do
+    it "redirects the user to the root path" do
+      page.set_rack_session(userinfo: nil)
+      visit new_activity_path
+      expect(current_path).to eq(root_path)
+    end
+  end
+
+  scenario "successfully creating an activity" do
+    visit new_activity_path
```
tahb: Same feedback as on the approach to funds.

spec/features/staff/users_can_create_an_activity_spec.rb
```diff
@@ -0,0 +1,35 @@
+RSpec.feature "Users can create an activity" do
```
tahb: The activity form is the core function of the service so I think it is worth thoroughly testing. We can add a missing scenario for validating the ID and fund/parent, I think it would be a good idea to test the validation errors appear and then correct the errors and resubmit in the same test. That would have caught an error I observed when testing where the fund value gets set wrong and prevents the form being resubmitted when the ID is omitted.

spec/features/staff/users_can_view_a_fund_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it "redirects the user to the root path" do
+      page.set_rack_session(userinfo: nil)
+      visit fund_path(viewable_fund)
+      expect(current_path).to eq(root_path)
+    end
+  end
+
+  context "when the fund belongs to the user's organisation" do
+    scenario "the user can view the fund" do
+      visit fund_path(viewable_fund)
```
tahb: Again, I think we should have at least one test that tests this end-to-end and ensures we have navigation in place.

app/views/staff/funds/new.html.haml
```diff
@@ -0,0 +1,19 @@
...
+  = link_to t("generic.link.back"), funds_path, class: "govuk-back-link"
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %h1.govuk-heading-xl
+        = t("page_title.fund.new")
+
+      = simple_form_for @fund do |f|
+        .govuk-form-group
+          = f.input :organisation_id,
```
tahb: When my user doesn't have an organisation I get an empty select box and no call to action to fix it. Users don't require an organisation at the moment because of administrators. Perhaps it would be simpler to change that now so that admins always belong to BEIS?

If we don't do the above, then it might make sense to hide the form and show a call to action message insteaed. Which might be that they need to be invited to at least one organisation first and to contact a memeber of the BEIS team.

Either way we should have test coverage for this.

db/schema.rb
```diff
@@ -10,12 +10,41 @@
...
+    t.date "actual_end_date"
+    t.string "recipient_region"
+    t.string "flow"
+    t.string "finance"
+    t.string "aid_type"
+    t.string "tied_status"
+    t.uuid "fund_id"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["fund_id"], name: "index_activities_on_fund_id"
```
tahb: On reflection I think that we should use a polymorphic association here to relate an activity to anyone of our 4 known parent types, rather than fund specifically. It seems to be the right tool for what we want to achieve here.

```
activity.hierachy => fund
activity.hierachy => programme
activity.hierachy => project
activity.hierachy => grant
```

I'm not sure if `hierarchy` is a good name but it is loosely related due to IATI terms. Is there a better one?

Hopefully by using a polymorphic association we will make these associations easier to manage, especially helping us if we need to add, rename or remove a given type.

app/views/staff/activities/new.html.haml
```diff
@@ -0,0 +1,55 @@
...
+        .govuk-form-group
+          = f.input :fund_id,
+                    wrapper: "select",
+                    collection: @funds
+          = f.input :identifier
+          = f.input :sector,
+                    wrapper: "select",
+                    collection: yaml_to_options("activity", "sector")
+          = f.input :title
+          = f.input :description
```
tahb: This should be a textarea rather than the single rowed text input as we are asking for more information here and examples are good chunky paragraphs.

db/schema.rb
lozette: The only downside of this is I can't seem to spell `hierarchy` consistently 😆

spec/features/staff/users_can_create_an_activity_spec.rb
lozette: As I've changed the routes so that you can't get to the Activity#new view without a fund ID, I think this is OK now - but please check!

lozette:
This is ready for re-review. I have one comment/TODO, and that's that at the moment I'm assuming all hierarchy objects on the Activities are Funds. I assume as we add more hierarchy types we will refactor to reflect the other types.

config/routes.rb
```diff
@@ -9,7 +9,13 @@
   scope module: "staff" do
     resource :dashboard, only: :show
     resources :users
-    resources :organisations
+    resources :organisations do
+      resources :funds, only: [:new, :create]
+    end
+    resources :funds, only: [:index, :show] do
```
tahb: From an initial click through and another visit to our wireframes I think we need to make a tweak here.

Since `fund#show` is not nested within an organisation the navigating back is jarringsince we go back to `fund#index` where we have to pick an organisation again before doing anything else. When in fact we came from `org/:id/fund/new` so we'd expect to go back to `org/:id`, I think.

I think the following routes would help with this:

```
resources :organisations do
  resources :funds, only: [:index, :show, :new, :create]
end

resources :funds do
  resources :activities, only: [:new, :create, :show]
end
```

This removes the route level `resources :activities` for showing the records, and indexing (which I don't think was used). I think we will be able to add different ways to route to activities later when we get to reporting. This is my bad in being a bit premature on the wireframe.

Have a click through with and without these changes and see if it makes sense to you. I'll attach a WIP commit as discussed as it will probably be more clear.

spec/features/staff/users_can_create_an_activity_spec.rb
tahb: We've removed the fund ID which is good and removes this unexpected scenario.

As well as fixing it, I was getting at extending our test coverage a little so that we catch any future regressions. Given that we no longer have a validations for this so that we can do the following, it can be added later:

```
1. fill out form with errors
2. submit invalid form
3. assert that validations are shown
4. correct errors
5. submit form again
6. assert that everything works
```

PR #41, opened 2019-11-21
(54) Users have basic roles

leeky:
## Changes in this PR

- Users can have one of two basic `role`, `administrator` or `delivery_partner` (the default)
- Add `Pundit` to only allow access to parts of the sites appropriate for the role of the user accessing it

Initially, users with the `administrator` role can access everything, whereas the `delivery_partner` role can only view, not edit.

NB: I've added two recommended checks in `Staff::BaseController` that ensure that include the following in our controllers:

- A call to Pundit's `authorize` method, which rejects users with an error message when they access a page they don't have access to.
- Use Pundit's scopes when calling ActiveRecord (e.g. `User.find(params[:id])` becomes `policy_scope(User).find(params[:id])`). This currently won't change any behaviours, but will give us more control in future (e.g. only allow users to be added to your organisations, unless you're an admin).


app/controllers/staff/organisations_controller.rb
```diff
@@ -1,23 +1,26 @@
 # frozen_string_literal: true
 
 class Staff::OrganisationsController < Staff::BaseController
-  include Secured
-
   def index
-    @organisations = Organisation.all
+    @organisations = policy_scope(Organisation)
```
leeky: NB: You can actually write this as `@organisations = authorize policy_scope(Organisations)` but was worried about hitting Rubocop's line length rules.

doc/architecture/decisions/0009-use-pundit-to-manage-permissions.md
```diff
@@ -0,0 +1,22 @@
+# 9. Use Pundit to manage permissions
+
+Date: 2019-11-22
+
+## Status
+
+Accepted
+
+## Context
```
tahb: Is it worth documenting the option of using Auth0 of the master for roles and then why we don't think it is preferable?

app/models/user.rb
```diff
@@ -1,4 +1,11 @@
 class User < ApplicationRecord
   has_and_belongs_to_many :organisations
   validates_presence_of :name, :email
+
+  enum role: {
+    administrator: "administrator",
+    delivery_partner: "delivery_partner",
+  }
+
+  attribute :role, :string, default: "delivery_partner"
```
tahb: How does this decision impact adding multiple roles to a user in future?

spec/policies/organisation_policy_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    it { is_expected.to permit_action(:destroy) }
+  end
+
+  context "as a delivery partner" do
+    let(:user) { build_stubbed(:user) }
+
+    it { is_expected.to permit_action(:index) }
+    it { is_expected.to permit_action(:show) }
+
+    it { is_expected.to forbid_new_and_create_actions }
```
tahb: Very noice 👍

app/controllers/staff/organisations_controller.rb
```diff
@@ -1,23 +1,26 @@
 # frozen_string_literal: true
 
 class Staff::OrganisationsController < Staff::BaseController
-  include Secured
```
tahb: Good catch.

app/controllers/staff/base_controller.rb
```diff
@@ -1,8 +1,17 @@
 class Staff::BaseController < ApplicationController
+  rescue_from Pundit::NotAuthorizedError, with: :pundit_not_authorized
```
tahb: We recently created an `Authorisation` module to handle the case when the authenticated user doesn't exist locally. It catches and renders (rather than redirects) to a custom error page that explains the problem and sets some expectations about what to do if this is unexpected.

I think there is an opportunity to merge these 2 error handling flows together into that concern. Would it be possible to render an error page here too instead of abruptly redirecting the user all the way back to the dashboard. From the render they can at least go 'back' in the browser to get back to where they went wrong to prevent any lost progress or flow.

config/locales/pundit.en.yml
```diff
@@ -0,0 +1,4 @@
+---
+en:
```
tahb: 👍

app/models/user.rb
leeky: Particularly because we expect the total number of users not to be huge, I think it should be straightforward enough to migrate this over to a many-to-many association with a join table if needed in the future.

app/controllers/staff/base_controller.rb
leeky: :+1:

PR #43, opened 2019-11-22
(91) New users can are able to set their password and sign into the service

tahb:
## Changes in this PR

- add redis
- add sidekiq
- add mail-notify to provide an easier way to send emails to GOV.UK Notify
- this change requires that verification is auto enabled and all Auth0 email templates are disabled
- a welcome email will be sent by GOV.UK Notify that will include a password reset link that will be live for 5 days, allowing the user to use it to set a new password. The password rest process will end be redirecting the user back to the service's start page.

## Testing steps

- Visit manage.auth0.com and remove your user from the development tenant
- Delete that email address from your local database using a console
- Log in to your local app using the generic roda user
- Create a new user using your dxw email address (for it to work it must be the same user as invited to Notify with)
- Recieve an email to your dxw inbox from Notify
- Copy the link into an ingonito browser and follow the steps
- Use your work email and new password to sign into the local service

## Screenshots of UI changes

Screenshot 2019-11-22 at 16 26 37
Screenshot 2019-11-22 at 16 27 15
Screenshot 2019-11-22 at 16 27 42
Screenshot 2019-11-22 at 16 27 54
Screenshot 2019-11-22 at 16 28 10
Screenshot 2019-11-22 at 16 28 15
Screenshot 2019-11-22 at 16 28 31

## Next steps

I've added the following environment variables to staging and production: `DOMAIN`, `NOTIFY_KEY`, `NOTIFY_WELCOME_EMAIL_TEMPLATE`.

I've added a Redis resource to each environment.

We don't have a life Notify account yet so our API keys are all team only. Have requested Odeta put this live as she has the gov.uk email.


config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,23 @@
+redis_url = "#{ENV["REDIS_URL"]}/0"
```
leeky: I think `REDIS_URL` default to database `0` if not specified.  As the database number can be set as part of the environment variable, I suspect this might cause problems - i.e. `redis://example.com:6379/5/0`

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,23 @@
...
+
+options = {
+  concurrency: Integer(ENV.fetch("RAILS_MAX_THREADS") { 5 }),
+}
+
+Sidekiq.configure_server do |config|
+  config.options.merge!(options)
+  config.redis = {
+    url: redis_url,
+    size: config.options[:concurrency] + 5,
```
leeky: Worth linking to https://github.com/mperham/sidekiq/wiki/Using-Redis#complete-control here - otherwise that `concurrency + 5` seem arbitrary?

Gemfile
```diff
@@ -17,8 +17,11 @@ gem "mini_racer"
 gem "puma", "~> 4.3"
 gem "rollbar"
 gem "rails", "~> 6.0.1"
+gem "redis"
+gem "redis-namespace"
```
leeky: The Redis configuration below doesn't appear to use namespacing..?

.env.test
```diff
@@ -4,5 +4,7 @@
 # For managing sensitive values and overrides use `/.env.test.local`
 #
 # Reference: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
-
+DOMAIN=test.local
```
leeky: Should a different `REDIS_URL` be specified here so the test suite doesn't overwrite keys in the development Redis instance?

app/mailers/user_mailer.rb
```diff
@@ -0,0 +1,24 @@
+require "./lib/auth0_api"
+
+class UserMailer < ApplicationMailer
+  def welcome(user_id)
```
leeky: Any reason you're taking a `user_id` here instead of just sending in the `user` instance directly, as per the [Action Mailer docs](https://guides.rubyonrails.org/action_mailer_basics.html#calling-the-mailer) ?

app/services/create_user.rb
```diff
@@ -21,6 +20,7 @@ def call
       end
     end
 
+    SendWelcomeEmail.new(user: user).call
```
leeky: :+1:

spec/mailers/user_mailer_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+
+RSpec.describe UserMailer, type: :mailer do
+  let(:user) { create(:user) }
+  before(:each) do
+    stub_const("NOTIFY_WELCOME_EMAIL_TEMPLATE", "123")
```
leeky: FYI `ClimateControl` is available if you need to stub out environment variables.

spec/features/staff/users_can_invite_new_users_spec.rb
```diff
@@ -39,6 +40,9 @@
     fill_in "user[name]", with: new_user_name
     fill_in "user[email]", with: new_user_email
 
+    # Ensure a welcome email is enqueued
+    expect_welcome_email_delivery
```
leeky: Nice! 🎉

spec/services/send_welcome_email_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+RSpec.describe SendWelcomeEmail do
+  let(:user) { build_stubbed(:user, id: "123") }
```
leeky: Does `id` need to be set here?

spec/support/email_helpers.rb
```diff
@@ -0,0 +1,21 @@
...
+    allow(welcome_email_delivery).to receive(:deliver_later)
+  end
+
+  def expect_welcome_email_delivery(user: nil)
+    welcome_email_delivery = instance_double(ActionMailer::MessageDelivery)
+    if user
+      expect(UserMailer).to receive(:welcome)
+        .with(user.id)
+        .and_return(welcome_email_delivery)
+    else
```
leeky: When would we expect a welcome email without an attached user?

heroku.yml
```diff
@@ -7,4 +7,7 @@ release:
     - rake db:migrate
 run:
   web: bundle exec puma -C config/puma.rb
-
+  worker:
```
leeky: Is it worth adding a `Procfile` to spin both these up with Foreman in development?

.env.test
tahb: Good point. I think it would be good to use a Mock Redis.

app/mailers/user_mailer.rb
tahb: I was working on the old understanding that you couldn't send objects into Redis via Sidekiq. Can we do this given that Sidekiq is the Action Mailer queue here?

leeky: It'll use Active job and that'll handle the serialisation before it get sent to Sidekiq.

spec/support/email_helpers.rb
tahb: In our feature test where I want to expect a welcome email is sent but there is no `user` record set up to use in the expectation, since the submit button has not been clicked at that point.

Can you think of how we might intercept the User in a concise way? Stubbing the User creation seemed too convulted and messy. I do agree that it would be a better test if we could asset that an email was sent to a given email address.

Perhaps we could send the email address in as a parameter solely for the purposes of testing?
```
expect(UserMailer).to receive(:welcome)
  .with(user.id, foo@bar.com)

UserMailer
  def welcome(id, email)
    user = User.find id
    email = email # This would be better to use the user.email to avoid any weird state problem
```

spec/mailers/user_mailer_spec.rb
tahb: Thanks for reminding me, using that will make this more consistent.

config/initializers/sidekiq.rb
tahb: I'm happy to put it in the commit so that if someone wants to blame it it's available for reading.

config/initializers/sidekiq.rb
tahb: 👍

heroku.yml
tahb: It might be time to do that. I think it should be the responsibility of a separate PR rather than here though, we may have other opinions around it that may slow this mostly unrelated work down.

Gemfile
tahb: This was a blind copy and paste. I have used it now however to separate dev and test redis databases.

PR #54, opened 2019-11-28
(105) Refactor user creation form with "govuk_design_system_formbuilder" 

leeky:
## Changes in this PR

Use `govuk_design_system_formbuilder` gem to replace `simple_form` starting with the user creation form. This has been [written by DfE](https://github.com/DFE-Digital/govuk_design_system_formbuilder) to adapt the standard Rails form builder to generate markup that's 100% compliant with the [GOVUK Design System](https://design-system.service.gov.uk/)

This change should make it much easier to build forms without having to use lots of boilerplate code to make it work.

Because we are currently using the *Simple Form* form builder for the majority of our forms, the two can co-exist until we have to time to switch all forms over.

### Before

Screenshot 2019-11-28 at 16 52 11

### After

Screenshot 2019-11-28 at 16 53 42

## Next steps

- [X] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ ruby "2.6.3"
 gem "auth0", "~> 4.9"
 gem "bootsnap", ">= 1.1.0", require: false
 gem "coffee-rails", "~> 5.0"
+gem "govuk_design_system_formbuilder", github: "dxw/govuk_design_system_formbuilder", branch: "i18n"
 gem "haml-rails"
 gem "high_voltage"
 gem "jbuilder", "~> 2.5"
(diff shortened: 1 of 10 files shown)


Gemfile
```diff
@@ -8,6 +8,7 @@ gem "auth0", "~> 4.9"
 gem "bootsnap", ">= 1.1.0", require: false
 gem "bootstrap", ">= 4.3.1"
 gem "coffee-rails", "~> 5.0"
+gem "govuk_design_system_formbuilder", github: "DFE-Digital/govuk_design_system_formbuilder", branch: "i18n"
```
tahb: Depending on a branch rather than a version seems like a risk. Have we thought about the impact to us if that branch disappears? Are we happy with it? Have we read anything that gives us confidence that this is likely to be merged into a new version?

PR #55, opened 2019-11-28
Remove bootstrap and related files

deberny:
Using bootstrap adds unnessesary styling clutter and overrides some of the GOV.UK styling, like for example links.

## Changes in this PR

- Removed bootstrap from the gemfile and also all related files.

- Left the flash messages with a note to restyle them after

### Before

Screenshot 2019-11-28 at 17 49 31

### After

Screenshot 2019-11-28 at 17 49 00

## Next steps

- remember to re-style the flash messages


deberny:
@leeky ok thanks, will remove it, rebase and merge :)

PR #60, opened 2019-12-02
Add gov.uk styles to definition lists

deberny:
After removing bootstrap the definition lists reverted to browser’s default styling, which didn’t look good.
Added gov.uk classes to the definition list and now they are more clear and use ‘summary-list’ component.

### Before

Screenshot 2019-12-02 at 14 06 18

### After

Screenshot 2019-12-02 at 14 06 45

Screenshot 2019-12-04 at 13 37 38
Screenshot 2019-12-04 at 13 38 08
Screenshot 2019-12-04 at 13 38 16

### DIFF
--- a/app/views/staff/activities/show.html.haml
+++ b/app/views/staff/activities/show.html.haml
@@ -12,60 +12,74 @@
 
   .govuk-grid-row
     .govuk-grid-column-two-thirds
-      %dl
-        %dt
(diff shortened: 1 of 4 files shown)


deberny:
Thank you for spoting that, i'll work on it.

deberny:
@lozette can you have a second look at the changes?

PR #61, opened 2019-12-02
Replace erroneous sector data list with the actual sector codes

lozette:
## Changes in this PR

Trello: https://trello.com/c/jSULZV34/111-fix-collection-of-fund-sector-information

When adding the Sector codelists we accidentally added the SectorVocabulary list
(http://reference.iatistandard.org/203/codelists/SectorVocabulary/) instead of
the correct Sector list (http://reference.iatistandard.org/203/codelists/Sector/)

We are expecting that users on this system will only ever need the `OECD DAC CRS Purpose Codes (5 digit)`


lozette:
Done the sorting and the hint, let me know what you think

PR #62, opened 2019-12-02
GOV.UK images are visibile in production

tahb:
## Changes in this PR

* Tell GOV.UK CSS where to find the GOV.UK images when the assets are being precompiled. This works in local since the link is not to a precompiled asset (you can tell as the image path doesn't have a long hash on the end). When we are in a live environment though we make sure that precompiled assets are the only ones that are used, so this url no longer works. We have to tell GOV.UK CSS to use the `url` function so that it uses a link to an precompiled asset when there is one.
* Add the following to your config/environments/development.rb in order to simulate the problem locally.

```
config.assets.compile = false
config.assets.css_compressor = nil
config.consider_all_requests_local = false
config.action_controller.perform_caching = true
config.eager_load = true
config.cache_classes = true
```
### Before
Screenshot 2019-12-02 at 16 26 31

### After
Screenshot 2019-12-02 at 16 26 35


tahb:
By default yes, however we then manually include them with `@import "forms";`.

deberny:
Fonts are loading fine `think the fonts are fine, all of the elements are using `font-family: "GDS Transport", Arial, sans-serif;``

tahb:
We found the same favicon error on Agz's machine. We had to make sure we commented out all configuration settings in development.rb that were being copied in in my initial notes.

PR #66, opened 2019-12-04
(79) Start splitting up activity form

leeky:
## Changes in this PR

Split the activity form into multiple steps, which should help us:

- give additional help and guidance on each step, without making the form too big
- add more potentially optional fields for other parts of IATI reporting
- not overwhelming user with too many validation errors at once
- allow more complex form controls (e.g. searching and autocomplete)

Note to code reviewers: This is probably worth reviewing one commit at a time. There are several commits where we break the form apart, one per step. There are then a few commits that refactor from those changes, see commit messages for details.

- [X] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [X] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has to impact outside the development team.


config/locales/en.yml
```diff
@@ -40,7 +40,7 @@ en:
       status:
         hint:
           html: IATI activity status. <a href="http://reference.iatistandard.org/203/codelists/ActivityStatus/" target="_blank">See IATI for detailed descriptions.</a>
-      submit: Submit
+      submit: Continue
```
lozette: I would prefer it if the key was `continue` instead of `submit`

doc/architecture/decisions/0012-use-wicked-for-multi-step-forms.md
```diff
@@ -0,0 +1,24 @@
...
+
+Date: 2019-12-04
+
+## Status
+
+Accepted
+
+## Context
+
+The create activity form is large and complex, with arguable too much for a
```
tahb: Small thing: We could include the GOV.UK guidance on this as a viable pattern that gives us some degree of confidence that this approach will make the process more manageable. https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page.

tahb:
This looks great :+1: 

We still need to address the need for field validation. I'd like to suggest adding a new card rather than adding it ontop here as I presume that adding step-by-step validation isn't going to change this foundational later. What are your thoughts @leeky?

config/locales/en.yml
leeky: I don't agree - I'd much rather have consistency in the I18n keys for submit buttons across forms, i.e. `form.activity.submit` / `form.organisation.submit` etc. 

Plus, the new form builder uses the standard ActiveRecord I18n keys, and this generally sticks with `submit`.

leeky:
@tahb Yeah, I feel like validations should go in a separate card. I'd probably switch over to the new form builder at the same time, as that'll tidy up the inconsistency with the date params not being nested under `activity[]`.

config/locales/en.yml
lozette: Fair comment!

PR #76, opened 2019-12-06
104 Users can edit a fund

tahb:
**Important - this change is based on https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/75 so requires that to be agreed and merged first** 

## Changes in this PR

- remove fund level routing that didn't require an organisation
- refactor `simple_form` out of the form new form
- allow the `fund` record to be edited. Note this change does not include editing the associated `activity` record to keep this proposal small, however I imagine the edit journey will be very similar. The links for activity edit will go to `ActivityFormController` rather than `FundController`

## Screenshots of UI changes

Screenshot 2019-12-06 at 17 48 21
Screenshot 2019-12-06 at 17 48 30
Screenshot 2019-12-06 at 17 48 37

## Next steps

### DIFF
--- a/app/controllers/staff/funds_controller.rb
+++ b/app/controllers/staff/funds_controller.rb
@@ -37,6 +37,26 @@ class Staff::FundsController < Staff::BaseController
     end
   end
 
+  def edit
+    @fund = policy_scope(Fund).find(id)
+    authorize @fund
+  end
+
+  def update
+    @fund = policy_scope(Fund).find(id)
+    authorize @fund
+
+    @fund.assign_attributes(fund_params)
+
+    if @fund.valid?
+      @fund.save
+      flash[:notice] = I18n.t("form.fund.update.success")
+      redirect_to organisation_fund_path(@fund.organisation, @fund)
+    else
+      render :edit
+    end
+  end
+
   private
 
   def fund_params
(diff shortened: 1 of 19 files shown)


app/controllers/staff/funds_controller.rb
```diff
@@ -18,18 +18,20 @@ def show
 
   def new
     @fund = policy_scope(Fund).new
+    @organisation = Organisation.find(organisation_id)
```
lozette: Out of curiosity (doesn't necessarily affect mergability!) do we need to add the policy check on to the `Organisation#find` instances here, or is it safe(ish) to assume by this point we know the user can access this organisation? They might have manipulated the URL to get a different Organisation ID into the form/params maybe?

tahb: Wowza, yes if I change the URL I can hop between organisations unchecked. @leeky should we be authorizing every resource in the scope at each step or perhaps a before filter to authorise any parent resource?

```
Funds
before_filter authorize_org

def create
  …
  authorize @fund
  …
end
```

leeky: Well, `organisation_id` is user input, so we should probably apply the policy. If you got it through an association, then I'd say not.

leeky: NB: I wouldn't call `authorize` on the other model, as `Fund` is the main thing we're dealing with here.

tahb: I now think I understand why I could URL this.. our policy does take care of this concern but  as an `administrator` I don't get halted! https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/policies/fund_policy.rb#L27

PR #80, opened 2019-12-09
(100) Edit transactions

lozette:
## Changes in this PR

Trello: https://trello.com/c/FQXbKkVH/100-edit-transactions-on-a-fund

* Edit transactions associated to a fund.
* Refactor Transaction#new view to use a re-usable Transaction create/edit form

### DIFF
--- a/app/views/staff/transactions/_form.html.haml
+++ b/app/views/staff/transactions/_form.html.haml
@@ -1,0 +1,21 @@
+= f.govuk_error_summary
+= f.govuk_text_field :reference
+= f.govuk_text_area :description
+= f.govuk_collection_select :transaction_type,
+                               yaml_to_objects("transaction", "transaction_type"),
+                               :code,
+                               :name,
+                               hint_text: t("form.transaction.transaction_type.hint")
+= f.govuk_date_field :date,
+            legend: { text: t("form.transaction.date.label"), size: "s" }
+= f.govuk_collection_select :currency,
+                               currency_select_options,
+                               :code,
+                               :name
+= f.govuk_number_field :value
+= f.govuk_collection_select :disbursement_channel,
+                               yaml_to_objects("transaction", "disbursement_channel"),
+                               :code,
+                               :name,
+                               hint_text: t("form.transaction.disbursement_channel.hint")
+= f.govuk_submit t("generic.button.submit")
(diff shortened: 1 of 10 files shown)


app/views/staff/transactions/_form.html.haml
```diff
@@ -0,0 +1,24 @@
+= f.govuk_error_summary
+= f.govuk_text_field :reference
+= f.govuk_text_area :description
+= f.govuk_collection_select :transaction_type,
+                               yaml_to_objects("transaction", "transaction_type"),
+                               :code,
+                               :name,
+                               label: { text: t("form.transaction.transaction_type.label") },
```
tahb: Leeky left me some feedback today on how we can move these labels to the active record locale and get additional benefits:

- labels don't need to be passed in on each form
- error messages include consistent field names
- form locale variables are grouped up (our en.yml is getting busy)

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/77#discussion_r355410041

I ended up doing https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/config/locales/models_and_forms.en.yml#L6 which doesn't include hints as it's not nicely supported yet.

spec/features/staff/users_can_edit_a_transaction_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  let(:organisation) { create(:organisation) }
+  let!(:fund) { create(:fund, organisation: organisation) }
+  let!(:transaction) { create(:transaction, fund: fund) }
+  let(:user) { create(:user, organisations: [organisation]) }
+
+  context "when the user is not logged in" do
+    it "redirects the user to the root path" do
+      page.set_rack_session(userinfo: nil)
+      visit fund_path(fund)
```
tahb: I think this should be `edit_fund_path`

app/views/staff/transactions/edit.html.haml
```diff
@@ -0,0 +1,11 @@
+=content_for :page_title_prefix, t("page_title.transaction.edit")
+
+= link_to t("generic.link.back"), fund_path(@fund), class: "govuk-back-link"
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %h1.govuk-heading-xl
+        = t("page_title.transaction.edit")
+
+      = form_with model: @transaction, url: fund_transaction_path(@fund), method: :patch do |f|
```
tahb: I've learnt that we don't need the `url` or `method` fields for the new form builder so you should be able to remove these without impact 👍

lozette: If I remove them, I get an error ` undefined method `transactions_path'` - the path is `fund_transactions_path`, so I think it needs to stay in.

PR #90, opened 2019-12-16
(128) Activity form uses govuk form builder

lozette:
## Changes in this PR

Trello: https://trello.com/c/AvIykhu2/128-refactor-activity-form-to-use-new-form-builder

All forms in the application now use `govuk_design_system_formbuilder`. 

## Next steps


app/models/activity.rb
```diff
@@ -3,6 +3,10 @@ class Activity < ApplicationRecord
   validates_presence_of :identifier
   validates_uniqueness_of :identifier
 
+  attribute :recipient_region, :string, default: "998"
```
tahb: I think these defaults will work fine for fund activities but I wonder how we'll be able to support these being changed for programmes and projects, any ideas?

app/helpers/codelist_helper.rb
```diff
@@ -1,15 +1,6 @@
 # frozen_string_literal: true
 
 module CodelistHelper
-  # TODO: Remove `yaml_to_options` when all forms are migrated from simple form to the govuk form builder,
```
tahb: Good to see us closing the loop and binning the old version 👍

lozette:
I think the asterisk for a required field is not part of the govuk design standard: https://govuk-elements.herokuapp.com/form-elements/#form-optional-fields

app/models/activity.rb
lozette: Possibly a set of env variables? 

Or maybe a helper along the lines of ActivityHelper. Although if the values needed to change than env variables are probably more flexible?

tahb:
@lozette Ah brilliant 👌 Really what we'd need if anything would be to add `(optional)` in everywhere else however [this card should add validation to everything](https://trello.com/c/GiHeA5QU).

app/models/activity.rb
tahb: I hadn't thought of environment variables, how might that look?

I wonder if having some initializer helpers might help. Something like:

```
class Activity

def setup_hierarchy_defaults
  case hierachy.class.name
  when fund setup_fund_defaults
…
end

def setup_fund_defaults
  self.recipient_region = "998"
  …
end

def setup_programme_detaults
  self.recipient_region = "2"
  …
end
```

```
class Staff::ActivitiesController < Staff::BaseController
  include ActivityHelper
  def create 
    @activity.hierarchy = hierarchy
    @activity.set_up_hierachy_defaults
    @activity.save(validate: false)
…

lozette: I like the initialiser helpers more, env variables would be clunky and I don't think these values are going to change often (maybe?)

PR #91, opened 2019-12-16
151 Transactions capture provider and reciever organisation

tahb:
## Changes in this PR

- transactions require 2 new fields: provider and receiver
- these fields are currently required and we can make it optional if the users are unable to succeed. This ensures we have a chance of having pristine data from the start given that this data will help us score well on transparency
- this proposal includes a tricky patch for the govuk form builder to ensure error messages are shown ontop of association fields. I wonder if there is an easier way to do this? If not, I hope it is extendable for when we have to add more associations for participating organisations in future. More info on the commit
- there is no room to put this new information in the transactions table as a new horizontal column. Until we have redesigned this page or implemented an individual transaction#show page we are only storing this data internally

http://reference.iatistandard.org/202/activity-standard/iati-activities/iati-activity/transaction/
http://reference.iatistandard.org/202/activity-standard/iati-activities/iati-activity/transaction/provider-org/
http://reference.iatistandard.org/202/activity-standard/iati-activities/iati-activity/transaction/receiver-org/

## Screenshots of UI changes

Screenshot 2019-12-16 at 15 57 49
Screenshot 2019-12-16 at 15 57 58
Screenshot 2019-12-16 at 15 58 06
Screenshot 2019-12-16 at 15 58 15
Screenshot 2019-12-16 at 15 58 22
Screenshot 2019-12-16 at 15 58 31

## Next steps

### DIFF
--- a/app/controllers/staff/transactions_controller.rb
+++ b/app/controllers/staff/transactions_controller.rb
@@ -6,8 +6,9 @@ class Staff::TransactionsController < Staff::BaseController
 
   def new
     @transaction = policy_scope(Transaction).new
-    @fund = Fund.find(fund_id)
     authorize @transaction
+
+    @fund = Fund.find(fund_id)
   end
 
   def create
@@ -15,8 +16,11 @@ class Staff::TransactionsController < Staff::BaseController
     authorize @transaction
 
     @fund = Fund.find(fund_id)
-    @transaction.update(fund: @fund)
+
+    @transaction.fund = @fund
     @transaction.assign_attributes(transaction_params)
+    @transaction.provider = provider
+    @transaction.receiver = receiver
     @transaction.value = monetary_value
     @transaction.date = format_date(date)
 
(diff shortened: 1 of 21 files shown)


app/controllers/staff/transactions_controller.rb
```diff
@@ -30,23 +34,26 @@ def create
...
     authorize @transaction
+
+    @fund = Fund.find(fund_id)
   end
 
   def update
     @transaction = policy_scope(Transaction).find(id)
     authorize @transaction
 
+    @fund = Fund.find(fund_id)
```
tahb: The edit form needs this as an instance variable which is rendered when the edit form is invalid.

app/helpers/form_helper.rb
```diff
@@ -0,0 +1,6 @@
+module FormHelper
+  def list_of_organisations
```
tahb: I predict we'll have to do more complicated logic here in the future. Defaulting the `receiver` as BEIS when creating a fund and the `provider` when creating a project level activity.

app/views/staff/funds/edit.html.haml
```diff
@@ -7,5 +7,5 @@
       %h1.govuk-heading-xl
         = t("page_title.fund.edit")
 
-      = form_with model: @fund, url: organisation_fund_path(@fund.organisation), method: :put do |f|
```
tahb: When doing some debugging this didn't seem to be required.

lozette:
This build is still failing thanks to `standard`!

tahb:
@lozette finally green!

PR #93, opened 2019-12-17
[WIP] Update GitHub Actions to build and test with containers

lozette:
## Next steps


tahb:
We made this pull request to help us debug a problem on Travis. Whilst it was helpful and worked, I think we should stick with Travis as BEIS's CI of choice where we can.

PR #95, opened 2019-12-18
(122) Add validations to Activity multi-step form

lozette:
## Changes in this PR

Trello: https://trello.com/c/GiHeA5QU/122-add-validations-to-activity-multi-step-form

* Implement validations on the Activity form. All fields are required (except dates) but the user can add them in their own time, but only in one order.
* Add a helper which hides the "Edit" links for fields on the fund#show page which have not been filled in yet, *unless* those fields form part of the next Wizard step. This encourages the user to fill the form out from start to finish. 
* If a field has not been filled out yet, the CTA is "Add". If it's been filled in before, the CTA is "Edit", so the user knows the attributes can be edited.
* Default attributes are presented in the select boxes but are no longer automatically saved onto the record. We found that a consequence of this prior design that it appeared to the user that they had already provided a `flow`, `tied_status` and `recipient_region` when in fact they hadn't even got that far in the journey.

Screenshot 2019-12-18 at 15 51 37

## Next steps

Co-authored-by: @tahb

### DIFF
--- a/app/helpers/codelist_helper.rb
+++ b/app/helpers/codelist_helper.rb
@@ -28,6 +28,11 @@ module CodelistHelper
     objects.unshift(OpenStruct.new(name: "ODA", code: "10")).uniq
   end
 
+  def tied_status_select_options
+    objects = yaml_to_objects(entity: "activity", type: "tied_status", with_empty_item: false)
+    objects.unshift(OpenStruct.new(name: "Untied", code: "5")).uniq
+  end
+
   def load_yaml(entity:, type:)
     yaml = YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/IATI/#{IATI_VERSION}/#{entity}/#{type}.yml"))
     yaml["data"]
--- a/app/helpers/codelist_helper.rb
+++ b/app/helpers/codelist_helper.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 module CodelistHelper
-  def yaml_to_objects(entity, type, with_empty_item = true)
-    data = load_yaml(entity, type)
+  def yaml_to_objects(entity:, type:, with_empty_item: true)
+    data = load_yaml(entity: entity, type: type)
     return [] if data.empty?
 
     objects = data.collect { |item| OpenStruct.new(name: item["name"], code: item["code"]) }.sort_by(&:name)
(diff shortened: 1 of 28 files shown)


app/views/staff/activity_forms/country.html.haml
```diff
@@ -1,2 +1,2 @@
 = render layout: "wrapper" do |f|
-  = f.govuk_collection_select :recipient_region, yaml_to_objects("activity", "recipient_region"), :code, :name
+  = f.govuk_collection_select :recipient_region, region_select_options, :code, :name, options: { selected: f.object.recipient_region || region_select_options.first.code }
```
tahb: Ensuring we have the HTML attribute `option selected=selected` on our default option took this change which is worth sharing as we'll need to do it elsewhere.

app/helpers/codelist_helper.rb
```diff
@@ -14,11 +14,26 @@ def yaml_to_objects(entity, type, with_empty_item = true)
   end
 
   def currency_select_options
-    objects = yaml_to_objects("generic", "default_currency", false)
+    objects = yaml_to_objects(entity: "generic", type: "default_currency", with_empty_item: false)
     objects.unshift(OpenStruct.new(name: "Pound Sterling", code: "GBP")).uniq
   end
 
-  def load_yaml(entity, type)
+  def region_select_options
```
tahb: We were set to extend these activity options using a model method called `set_hierachy_defaults`. With this change I imagine we could later extend this to ingest a hierarchy object to use in a switch statement to deviate from the defaults where needed.

spec/models/activity_spec.rb
```diff
@@ -9,35 +9,81 @@
...
+      subject { build(:activity, finance: nil, wizard_status: :finance) }
+      it { should validate_presence_of(:finance) }
+    end
+
+    context "when tied_status is blank" do
+      subject { build(:activity, tied_status: nil, wizard_status: :tied_status) }
+      it { should validate_presence_of(:tied_status) }
+    end
+
+    context "when the wizard_status is complete" do
```
tahb: As I explained on this commit it's premature and not strictly needed. I felt more confident having test coverage in place from the off to ensure we have a mechanism for calling all validations at once. This is in the Wicked docs and will be helpful in any backfilling we have to do of old activities.

tahb:
@lozette I do an approval when you're happy so it becomes mergeable.

lozette:
Rebased and all looks great to me!

PR #100, opened 2019-12-19
Feature/143 add transaction information an activity xml

lozette:
## Changes in this PR

Trello: https://trello.com/c/J9fiyVbh/143-add-transaction-information-an-activity-xml

* Transaction information is now available as XML
* Refactored the XML views into partials, so the Activity + Transaction XML can be combined
* Added a new XML view for Funds, which renders all the Activity + Transaction information for a fund. We can add more to this XML as more information is added to Funds.

To view a Transaction or Fund as XML, add `.xml` to the end of their `show` URL.

## Next steps


app/views/staff/shared/xml/_transaction.xml.haml
```diff
@@ -0,0 +1,17 @@
...
+  %transaction-type{"code" => transaction.transaction_type}
+  %transaction-date{"iso-date" => transaction.date}
+  %value{"currency" => transaction.currency, "value-date" => transaction.date}=transaction.value
+  %description
+    %narrative=transaction.description
+  %provider-org{"provider-activity-id" => "", "type" => transaction.provider.organisation_type, "ref" => ""}
+    %narrative=transaction.provider.name
+  %receiver-org{"receiver-activity-id" => "", "type" => transaction.receiver.organisation_type, "ref" => ""}
+    %narrative=transaction.receiver.name
+  %disbursement-channel{"code" => transaction.disbursement_channel}
```
tahb: I don't believe we need to show these fields at the transaction level too, given we aren't allowing the user to override the defaults set by the parent activity.

On the Trello card we have the following that aren't needed within the transaction XML:

```
disbursement-channel
recipient-country
recipient-region
flow-type
finance-type
aid-type
tied-status
```

app/controllers/staff/transactions_controller.rb
```diff
@@ -59,6 +59,16 @@ def update
     end
   end
 
+  def show
+    @transaction = policy_scope(Transaction).find(id)
+    authorize @transaction
+
+    @activity = Activity.find_by(hierarchy_id: @transaction.fund)
```
tahb: The reference to `fund` here is a concern as it doesn't support other hierarchies without more logic.

Have we missed that transactions need a polymorphic association to the fund as well, just like activity? That would enable us to use `transaction.hierarchy` in our query instead. If this sounds like the right thing to do could you perhaps put something in Trello so we don't forget, then we can get this merged!

app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -0,0 +1,22 @@
+%iati-activity{"default-currency" => "#{activity.default_currency}"}
```
tahb: V nice to see we're partialling this up 👍

app/controllers/staff/transactions_controller.rb
lozette: Yep looks like we need a polymorphic association here: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/models/transaction.rb#L2

PR #106, opened 2019-12-20
Feature/restrict who can edit organisations

tahb:
## Changes in this PR

By default the delivery partner role cannot see or edit an organisation they are not associated with. This then fixes a problem where any role could see and create new funds, I've prevented fund creation however we probably need something more robust to prevent fund transactions from being created too with a specially crafted URLs. Any ideas appreciated, should we be checking the users organisation all the way down through organisation > fund > transaction?

- only use Pundit scopes on the querying of lists ie. index pages, after finding a problem with 404 errors not being raised if records don't exist, I re-read the documentation for the edits I've made which is on the commit
- only administrators or users who belong to an organisation can see and navigate to it
- only administrators or users who belong to an organisation can edit 
- only administrators or users who belong to an organisation can see associated funds

## Next steps

### DIFF
--- a/app/controllers/staff/activities_controller.rb
+++ b/app/controllers/staff/activities_controller.rb
@@ -4,7 +4,7 @@ class Staff::ActivitiesController < Staff::BaseController
   include ActivityHelper
 
   def show
-    @activity = policy_scope(Activity).find(id)
+    @activity = Activity.find(id)
     authorize @activity
 
     @activity_presenter = ActivityPresenter.new(@activity)
@@ -16,7 +16,7 @@ class Staff::ActivitiesController < Staff::BaseController
   end
 
   def create
-    @activity = policy_scope(Activity).new
+    @activity = Activity.new
     authorize @activity
 
     @activity.hierarchy = hierarchy
@@ -38,6 +38,6 @@ class Staff::ActivitiesController < Staff::BaseController
 
   def hierarchy
     # TODO: Add support for new hierarchies here and/or move to a service
-    @hierarchy = policy_scope(Fund).find(fund_id)
+    @hierarchy = Fund.find(fund_id)
   end
 end
--- a/app/policies/fund_policy.rb
+++ b/app/policies/fund_policy.rb
@@ -1,22 +1,22 @@
 class FundPolicy < ApplicationPolicy
   def index?
-    true
+    user.administrator? || user.organisations.include?(@record.organisation)
   end
 
   def show?
-    true
+    user.administrator? || user.organisations.include?(@record.organisation)
   end
 
   def create?
-    true
+    user.administrator?
   end
 
   def update?
-    true
+    user.administrator?
   end
 
   def destroy?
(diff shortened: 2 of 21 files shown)


app/controllers/staff/activities_controller.rb
```diff
@@ -4,7 +4,7 @@ class Staff::ActivitiesController < Staff::BaseController
   include ActivityHelper
 
   def show
-    @activity = policy_scope(Activity).find(id)
+    @activity = Activity.find(id)
```
leeky: Whilst I agree that `policy_scope` has probably been used in too many places, I don't think it should be removed everywhere.

In this example, if the user has a role that is allowed by `ActivityPolicy` to pass the `show?` check, then they can view *every* activity, even for organisations they don't belong to.

Not wrapping the query in `policy_scope` will skip applying the scope defined within the policy, which mostly defeats the object of using Pundit.

leeky: I assume the Pundit defaults go on the basis of "If I can't see in on the index page, I can't reach the show page", but this feels insecure to me.

app/policies/organisation_policy.rb
```diff
@@ -4,6 +4,20 @@ def index?
...
   def show?
-    true
+    user.administrator? || user.organisations.include?(@record)
+  end
+
+  def update?
+    user.administrator? || user.organisations.include?(@record)
+  end
+
+  class Scope < Scope
```
leeky: This will get ignored because of the removal of `policy_scope` in `Staff::OrganisationController`

spec/policies/organisation_policy_spec.rb
```diff
@@ -16,13 +16,21 @@
...
     it { is_expected.to forbid_edit_and_update_actions }
     it { is_expected.to forbid_action(:destroy) }
+
+    context "when the delivery_partner belongs to that organisation" do
+      let(:user) { build_stubbed(:delivery_partner, organisations: [organisation]) }
+      let(:organisation) { create(:organisation) }
+
+      it { is_expected.to permit_action(:index) }
+      it { is_expected.to permit_action(:show) }
+    end
```
tahb: This file is missing scope tests since `policy_scope` is called in the controller.

app/policies/fund_policy.rb
```diff
@@ -1,22 +1,22 @@
...
   end
 
   def show?
-    true
+    user.administrator? || user.organisations.include?(@record.organisation)
   end
 
   def create?
-    true
+    user.administrator?
```
tahb: We should also allow users who belong to that organisation to create funds. We can do this by setting the association in the method before calling `authorize`. Same for update and destroy.

tahb:
Having done some pairing with Leeky this proposal is being reorganised in a [new pull request](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/119) so that this branch remains as a fall back option.

PR #109, opened 2019-12-23
Link Transactions to Funds via polymorphic association "Hierarchy"

lozette:
## Changes in this PR

Trello: https://trello.com/c/XpMDq8O8/175-refactor-make-fund-transaction-association-polymorphic

Previously Transactions belonged to a Fund. Now they can belong to a Fund via the polymorphic association `hierarchy`, which means when we add Projects/Programmes etc they will also be able to have Transactions.

There are still some TODOs here which will need to be addressed when we add more `hierarchy` types

### DIFF
--- a/app/controllers/staff/transactions_controller.rb
+++ b/app/controllers/staff/transactions_controller.rb
@@ -3,21 +3,22 @@
 class Staff::TransactionsController < Staff::BaseController
   include Secured
   include DateHelper
+  include TransactionHelper
 
   def new
     @transaction = policy_scope(Transaction).new
     authorize @transaction
 
-    @fund = Fund.find(fund_id)
+    @hierarchy = hierarchy
   end
 
   def create
     @transaction = policy_scope(Transaction).new(transaction_params)
     authorize @transaction
 
-    @fund = Fund.find(fund_id)
+    @hierarchy = hierarchy
 
-    @transaction.fund = @fund
+    @transaction.hierarchy = @hierarchy
     @transaction.assign_attributes(transaction_params)
     @transaction.provider = provider
     @transaction.receiver = receiver
@@ -26,7 +27,7 @@ class Staff::TransactionsController < Staff::BaseController
 
     if @transaction.save
       flash[:notice] = I18n.t("form.transaction.create.success")
-      redirect_to organisation_fund_path(@fund.organisation, @fund)
+      redirect_to transaction_hierarchy_path_for(transaction: @transaction)
     else
       render :new
     end
@@ -36,14 +37,14 @@ class Staff::TransactionsController < Staff::BaseController
     @transaction = policy_scope(Transaction).find(id)
     authorize @transaction
 
-    @fund = Fund.find(fund_id)
+    @fund = hierarchy_for(transaction: @transaction)
   end
 
   def update
     @transaction = policy_scope(Transaction).find(id)
     authorize @transaction
 
-    @fund = Fund.find(fund_id)
+    @fund = hierarchy_for(transaction: @transaction)
 
     @transaction.assign_attributes(transaction_params)
     @transaction.provider = provider
@@ -53,7 +54,7 @@ class Staff::TransactionsController < Staff::BaseController
 
     if @transaction.save
       flash[:notice] = I18n.t("form.transaction.update.success")
-      redirect_to organisation_fund_path(@fund.organisation, @fund)
+      redirect_to transaction_hierarchy_path_for(transaction: @transaction)
     else
       render :edit
     end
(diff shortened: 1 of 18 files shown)


app/controllers/staff/transactions_controller.rb
```diff
@@ -26,7 +27,7 @@ def create
 
     if @transaction.save
       flash[:notice] = I18n.t("form.transaction.create.success")
-      redirect_to organisation_fund_path(@fund.organisation, @fund)
+      redirect_to transaction_hierarchy_path_for(transaction: @transaction)
```
tahb: Did we intetionally change the behaviour of this redirect?

app/views/staff/transactions/new.html.haml
```diff
@@ -1,11 +1,12 @@
...
+%p= link_to t("generic.link.back"), hierarchy_object_path(hierarchy: @hierarchy), class: "govuk-back-link"
 %main.govuk-main-wrapper#main-content{ role: "main" }
   .govuk-grid-row
     .govuk-grid-column-two-thirds
       %h1.govuk-heading-xl
         = t("page_title.transaction.new")
 
-      = form_with model: @transaction, url: fund_transactions_path(@fund) do |f|
+      =# TODO: replace fund_transactions_path with something hierarchy-agnostic
+      = form_with model: @transaction, url: fund_transactions_path(@hierarchy) do |f|
```
tahb: Did you run into problems with trying to make this path value dynamically?

spec/helpers/transaction_helper_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    context "when the hierarchy_type is Fund" do
+      it "returns the Fund" do
+        fund = create(:fund)
+        transaction = build(:transaction, hierarchy: fund)
+        expect(helper.hierarchy_for(transaction: transaction)).to eq(fund)
+      end
+    end
+  end
+
+  describe "#transaction_hierarchy_path_for" do
```
tahb: This method name confused me into thinking there was a change in behaviour where we'd be redirecting the transaction show page. I think this could be `organisation_hierarchy_path_for`?

app/controllers/staff/transactions_controller.rb
```diff
@@ -36,14 +37,14 @@ def edit
     @transaction = policy_scope(Transaction).find(id)
     authorize @transaction
 
-    @fund = Fund.find(fund_id)
+    @fund = hierarchy_for(transaction: @transaction)
```
tahb: Could we update the name of this variable with `@hierarchy` too?

app/controllers/staff/transactions_controller.rb
```diff
@@ -36,14 +37,14 @@ def edit
...
   end
 
   def update
     @transaction = policy_scope(Transaction).find(id)
     authorize @transaction
 
-    @fund = Fund.find(fund_id)
+    @fund = hierarchy_for(transaction: @transaction)
```
tahb: Could we update the name of this variable with `@hierarchy` too?

app/policies/transaction_policy.rb
```diff
@@ -24,9 +24,10 @@ def resolve
       if user.administrator?
         scope.all
       else
+        # TODO: when we add other hierarchy types, include them here somehow!
```
tahb: I've been mulling over how we could make this extendable to any type of hierarchy but I've not thought of a way to do this nicely. Ideally the policy scope could be called with additional context but we aren't able to pass this: 

```
transactions = policy_scope(Transaction, hierarchy: :fund).where(fund: @fund)
```

We could create new policies for each hierarchy `FundTransactionPolicy` and `ProgrammeTransactionPolicy` which inheritly knows what parent table to check.. creating new files just to carry state doesn't feel great.

@leeky do you have any ideas for this?

mec:
I haven't quite got my head around the modelling yet, but `hierarchy` as the name here is throwing me. Once we add Projects/Programmes are the transactions still about money, if so does something like `funded` work?

app/controllers/staff/transactions_controller.rb
```diff
@@ -108,7 +109,15 @@ def monetary_value
     end
   end
 
-  def fund_id
+  def hierarchy
```
leeky: I think this needs renaming - when we're calling `@transaction.hierarchy` above, it looks like we're reaching into the *association* of it.  When I was prying in the console to figure out what was happening it took me a while to realise why calling `@transaction.hierachy` was returning `nil` rather than an `ActiveRecord::Relation`.

app/models/fund.rb
```diff
@@ -2,4 +2,5 @@ class Fund < ApplicationRecord
   validates_presence_of :name
   belongs_to :organisation
   has_one :activity, as: :hierarchy
+  has_many :transactions, as: :hierarchy
```
leeky: I think this is going to confuse folks (and possibly Rails) - we now have *two* polymorphic associations (`Activity` and `Transaction`) with the same `as:` name.

lozette: Agree; I was adding hierarchy as requested but maybe we need another naming session

app/controllers/staff/transactions_controller.rb
lozette: Yeah, agree; I got myself in a big confused pickle while doing this bc of the naming. I think we might need another name for these.... concepts?

lozette:
Closing this as @leeky & I have started another branch after discussing different approaches

PR #119, opened 2020-01-06
Feature/164 restrict delivery partner access

tahb:
This is a pretty big pull request. What started as a simple restriction of delivery partners so that they couldn't create funds, ended up addressing the fact that delivery partners could do everything when they should in fact be able to do almost nothing.

I believe we've arrived here mainly because the default role of our User FactoryBot was `delivery_partner`. When we wrote create(:user) in our tests and built out all our features, we were testing that delivery partners _could_ do it.

Leeky and I went through the strategy of how we could use Pundit's `authorize` and `policy_scope` to control permissions. We decided that policy_scope should only be applied on index actions, where users are allowed to see the content but we want to filter that list down to only allowed entities. We'd use authorise everywhere else to check the user was allowed to interact with that record before punting them to a not authorised page. We'd adjust our Pundit before filters [to follow their guidance on using scopes](https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used) and ensure we had feature tests for our happy paths to ensure we had test coverage for these scenarios rather than _only_ unit policy tests.

This work forced me to think clearly about what each role could do and I've sought to write our feature tests accordingly:

- `administrator`s can do anything on the service (dxw/product team)
- `fund_manager` can do manage users, organisations, funds, fund activities, fund transactions
- `delivery_partner` can view and edit their organisation

Some feature tests have had their prefix renamed from `users_` to `fund_managers_` where fund managers are expected to be able to perform an action. Administrators can still perform this action but aren't the primary user so the tests do not concern themselves with it as administrators are only going to use their permission to test and debug the service.

We have a lot of tests that do a lot of work provide confidence that funds can be created, along with an associated activity and transactions. These tests all use funds as a subtle but crucial bit of context. When we come to add programmes we should write new tests to ensure they can be added in isolation from funds, and they can be only be added by the expected user. To this end I've added context blocks for who the user is. The other option is to create a separate spec file for `fund_manager_can_create_activity_spec.rb` and `programme_manager_can_create_activity_spec.rb` which helpers that help us reuse code. Whilst these look the same now I expect these user journeys *will* change to become more and more distinct.

## Changes in this PR

- prevent future tests being written without deliberatly choosing a user role, this should ensure that going forwards we don't have to come back and refactor our tests to restrict delivery partners which were the default role type
- restrict what delivery partners can do so that right now they can only view their organisation
- allow fund managers (and administrators) to perform CRUD actions on users and organisations
- allow fund managers (and administrators) to perform CRUD actions on funds, activities and transactions
- use policy actions methods (ie. new?) as well as only `policy_scope` to authorise our controller endpoints. Following the [Pundit guidance](https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used) we intend to use scopes to filter down a list of content and methods to block users visiting the page. This is preferable as it allows us access to the `record` object which can be used in more complicated rules, i.e if a record is a fund and the user is a fund_manager, they should be allowed. There is no way to access this in Scope, and there is no way to change behaviour per endpoint. I.e if the delivery partner should be able to update by not destroy, we cannot use the same `policy_scope`.

## Next steps


app/policies/fund_policy.rb
```diff
@@ -4,19 +4,23 @@ def index?
...
-    true
+    user.administrator? || associated_fund_manager?(user: user, record: record)
   end
 
   def destroy?
-    true
+    user.administrator? || associated_fund_manager?(user: user, record: record)
+  end
+
+  private def associated_fund_manager?(user:, record:)
```
leeky: Since `user` and `record` are globally available in the policy class, they don't need to be passed as parameters.

So you could simplify this to `user.administrator? || associated_fund_manager?`

app/policies/fund_policy.rb
```diff
@@ -4,19 +4,23 @@ def index?
...
+    user.fund_manager? && user.organisations.include?(record.organisation)
```
leeky: Not sure if it makes it less readable, but `user.organisation_ids.include?(record.organisation_id)` is likely going to be quicker.

app/policies/organisation_policy.rb
```diff
@@ -4,6 +4,28 @@ def index?
   end
 
   def show?
-    true
+    user.administrator? || user.organisations.include?(record)
```
leeky: Worth extracting `user.organisations.include...` into a separate method?

spec/features/staff/users_can_create_a_fund_spec.rb
```diff
@@ -1,43 +1,72 @@
...
-    authenticate!(user: user)
-  end
-
-  let(:organisation) { create(:organisation, name: "UKSA") }
-  let(:user) { create(:user, organisations: [organisation]) }
-
   context "when the user is not logged in" do
     it "redirects the user to the root path" do
+      organisation = create(:organisation)
+      authenticate!(user: build_stubbed(:user))
```
leeky: This reads strangely - why are we calling `authenticate` for a test case where the user is specifically *not* logged in?

spec/features/staff/users_can_view_a_fund_spec.rb
```diff
@@ -27,8 +27,8 @@
 
   context "when the fund belongs to another organisation" do
     scenario "the user cannot view the fund" do
-      expect { visit organisation_fund_path(organisation, forbidden_fund) }
-        .to raise_exception(ActiveRecord::RecordNotFound)
+      visit organisation_fund_path(organisation, forbidden_fund)
+      expect(page).to have_content(I18n.t("page_title.errors.not_authorised"))
```
leeky: Yeah, this makes much more sense!

spec/policies/organisation_policy_spec.rb
```diff
@@ -6,23 +6,53 @@
...
+    it { is_expected.to permit_action(:destroy) }
+
+    it "includes only the associated organisation in the resolved scope" do
+      expect(resolved_scope).to include(organisation)
+      expect(resolved_scope).not_to include(unassociated_organisation)
+    end
+
+    context "without an associated organisation" do
+      let(:unassociated_organisation) { create(:organisation) }
+      let(:user) { create(:user, organisations: []) }
```
leeky: :+1: For explicitly showing that organisations is empty

spec/requests/transaction_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe "Transaction", type: :request do
+  context "when the user's organisation is not associated with the fund" do
```
leeky: Perhaps this should mention "the user attempts to create a fund for an organisation they do not belong to" ?

app/policies/fund_policy.rb
tahb: That'd be much better! 💥

app/controllers/staff/transactions_controller.rb
```diff
@@ -4,20 +4,28 @@ class Staff::TransactionsController < Staff::BaseController
   include Secured
   include DateHelper
 
-  def new
-    @transaction = policy_scope(Transaction).new
+  def destroy
```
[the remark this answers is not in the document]
tahb: Yes leftover from our playing around 😬

app/controllers/staff/transactions_controller.rb
```diff
@@ -4,20 +4,28 @@ class Staff::TransactionsController < Staff::BaseController
...
+    @transaction = Transaction.find(id: @id)
+
     authorize @transaction
+  end
 
+  def new
     @fund = Fund.find(fund_id)
-  end
+    @transaction = Transaction.new(fund: @fund)
+    @transaction.fund = @fund
```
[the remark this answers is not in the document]
tahb: Will remove the duplicated part 👍 

`@fund` is used by the view so I think trying to refactor that query out might be a step too far given our focus here.

app/controllers/staff/funds_controller.rb
```diff
@@ -43,12 +45,12 @@ def create
   end
 
   def edit
-    @fund = policy_scope(Fund).find(id)
+    @fund = Fund.find(id)
```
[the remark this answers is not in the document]
tahb: I think this would make more sense if the style could be applied to all actions. As it is some actions will have it in one line whilst the others will have to have it on multi-line eg.

```
def create
    @fund = Fund.new(fund_params)
    @organisation = Organisation.find(organisation_id)
    @fund.organisation = @organisation

    authorize @fund
```

I'm in favour of the original which keeps it is slightly more verbose but keeps it consistent.

app/controllers/staff/transactions_controller.rb
```diff
@@ -4,20 +4,28 @@ class Staff::TransactionsController < Staff::BaseController
...
 
-  def create
-    @transaction = policy_scope(Transaction).new(transaction_params)
     authorize @transaction
+  end
 
+  def create
     @fund = Fund.find(fund_id)
-
+    @transaction = Transaction.new(transaction_params)
```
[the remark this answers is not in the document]
tahb: Agree 👍

tahb:
In addressing this feedback I noticed that we're not addressing user, organisation or activity policies yet. This highlights a large issue where all our tests have been written from the perspective of a `delivery_partner`. Now we are trying to limit what they can do all the tests are breaking so I'm going through and fixing these and will have a collection of new commits with each resource.

lozette:
`standard` is doing a complain on this :(

app/controllers/staff/activities_controller.rb
```diff
@@ -16,10 +16,10 @@ def show
   end
 
   def create
-    @activity = policy_scope(Activity).new
+    @activity = Activity.new
+    @activity.hierarchy = hierarchy
```
tahb: Assigning this before we call `authorize` method in all controllers allows Pundit to use this state inside it's authorisation methods, eg. `record.hierarchy` will now return the hierarchy. The other way around it will return `nil`.

app/controllers/staff/organisations_controller.rb
```diff
@@ -3,24 +3,23 @@
...
   end
 
   def show
-    organisation = policy_scope(Organisation).find(params[:id])
+    organisation = Organisation.find(id)
     authorize organisation
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
-    @funds = Fund.includes(:organisation).where(organisation: organisation)
+    @funds = policy_scope(Fund).includes(:organisation).where(organisation: organisation)
```
tahb: Given that funds are listed on the organisation show page, we need to make sure we only show funds for a user who's a fund manager. Delivery partners shouldn't see any.

app/controllers/staff/users_controller.rb
```diff
@@ -1,24 +1,24 @@
 class Staff::UsersController < Staff::BaseController
   def index
+    authorize :user, :index?
```
tahb: Only fund_managers and administrators should be able to see this page so we call `authorize` first before spending time querying.

app/policies/organisation_policy.rb
```diff
@@ -4,6 +4,32 @@ def index?
   end
 
   def show?
-    true
+    user.administrator? || user.fund_manager? || associated_user?
```
tahb: Users who are associated with an organisation should be able to view it.

app/policies/organisation_policy.rb
```diff
@@ -4,6 +4,32 @@ def index?
...
+  end
+
+  def create?
+    user.administrator? || user.fund_manager?
+  end
+
+  def update?
+    user.administrator? || user.fund_manager? || associated_user?
```
tahb: Users who are associated with an organisation should be able to update the information.

PR #122, opened 2020-01-07
Use 'pry' when debugging on the rails console

leeky:
Swap the plain 'pry' gem for 'pry-rails', so it's used instead of IRB for the Rails console by default, which makes debugging much nicer.


peteryates:
[This snippet](https://coderwall.com/p/hlbana/indent-and-colorize-html-strings-in-pry) in your `.pryrc` is worth its weight in gold btw

PR #125, opened 2020-01-08
(81) Admins can assign roles to users

mec:
## Changes in this PR

- Shows the users role on the index and show pages
- Allows a user with the correct permission to choose the users role when creating or updating a user

https://trello.com/c/UoCQvLew

### Before

Screenshot_2020-01-08 Edit user — Report Official Development Assistance

Screenshot_2020-01-08 Dashboard — Report Official Development Assistance

Screenshot_2020-01-08 User — Report Official Development Assistance

### After

Screenshot_2020-01-08 Dashboard — Report Official Development Assistance

Screenshot_2020-01-08 Edit user — Report Official Development Assistance

Screenshot_2020-01-08 User — Report Official Development Assistance

## Next steps

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -3,4 +3,10 @@ module FormHelper
     @list_of_organisations ||=
       [OpenStruct.new(name: "", id: ""), Organisation.sorted_by_name].flatten
   end
+
+  def list_of_user_roles
+    @list_of_user_roles ||= begin
+      User.roles.map { |id, name| OpenStruct.new(id: id, name: I18n.t("activerecord.attributes.user.roles.#{name}")) }
+    end
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,4 +9,8 @@ class User < ApplicationRecord
   }
 
   attribute :role, :string, default: "delivery_partner"
+
+  def role_name
+    I18n.t("activerecord.attributes.user.roles.#{role}")
+  end
 end
--- a/app/views/staff/users/_form.html.haml
+++ b/app/views/staff/users/_form.html.haml
@@ -3,9 +3,18 @@
 
   = f.govuk_text_field :name
   = f.govuk_email_field :email
+  = f.govuk_collection_radio_buttons :role,
+      list_of_user_roles,
+      :id,
+      :name,
+      legend: { tag: :h2 }
 
   - if @organisations.any?
-    = f.govuk_collection_check_boxes :organisation_ids, @organisations, :id, :name
+    = f.govuk_collection_check_boxes :organisation_ids,
+        @organisations,
+        :id,
+        :name,
+        legend: { tag: :h2 }
   - else
     .govuk-inset-text
       = succeed "." do
--- a/spec/features/staff/users_can_edit_a_user_spec.rb
+++ b/spec/features/staff/users_can_edit_a_user_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+
+RSpec.feature "Editing a user" do
+  let!(:user) { create(:user) }
+
+  before do
+    stub_auth0_token_request
+    authenticate!(user: build_stubbed(:user, role: :administrator))
+  end
+
+  scenario "the role can be changed" do
+    visit dashboard_path
+    click_on "Manage users"
+
+    expect(page).to have_content(user.name)
+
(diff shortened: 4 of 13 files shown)


app/helpers/form_helper.rb
```diff
@@ -3,4 +3,17 @@ def list_of_organisations
...
   end
+
+  def list_of_user_roles
+    @list_of_user_roles ||= build_user_roles_list
+  end
+
+  private
+
+  def build_user_roles_list
+    user_roles = User.roles.values
```
leeky: You can write this as:

```ruby
User.roles.map{ |id, name| OpenStruct.new(id: id, name: name) }
```

`map` loops over each `User.roles` in turn. As it is a `Hash` the block accepts a `key` and `value`, that I've renamed to `id` and `name` for clarity.

app/models/user.rb
```diff
@@ -9,4 +9,8 @@ class User < ApplicationRecord
   }
 
   attribute :role, :string, default: "delivery_partner"
+
+  def role_name
```
leeky: This should probably use i18n instead, but [enums are weird](https://www.kostolansky.sk/posts/localize-rails-enums/) and don't support it out of the box.

app/views/staff/users/_form.html.haml
```diff
@@ -3,6 +3,12 @@
 
   = f.govuk_text_field :name
   = f.govuk_email_field :email
+  = f.govuk_collection_radio_buttons :role,
+      list_of_user_roles,
+      :id,
+      :name,
+      legend: { text: "Role" },
```
leeky: `legend` and `hint_text` should be translated (they go in `config/locales/models_and_forms.yml`)

The [form builder documentation](https://govuk-form-builder.netlify.com/building-blocks/localisation/) explains how to do this.

spec/features/staff/users_can_create_a_user_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+
+RSpec.feature "Creating a user" do
+  let(:user) { build(:user) }
```
leeky: How about renaming this `user` to something like `target_user`, to differentiate between the admin user who is creating the user *AND* the user being created?

spec/features/staff/users_can_edit_a_user_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    authenticate!(user: build_stubbed(:user, role: :administrator))
+  end
+
+  scenario "the role can be changed" do
+    visit dashboard_path
+    click_on "Manage users"
+
+    expect(page).to have_content(user.name)
+
+    find("tr", text: user.name).click_link("Edit")
```
leeky: :+1:

spec/features/staff/users_can_create_a_user_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+
+RSpec.feature "Creating a user" do
```
leeky: There's actually a spec for this already - `users_can_invite_new_users_spec.rb`

spec/helpers/form_helper_spec.rb
```diff
@@ -7,4 +7,19 @@
       helper.list_of_organisations
     end
   end
+
+  describe "#list_of_user_roles" do
```
leeky: I'm not sure if this test is necessary..?

app/helpers/form_helper.rb
```diff
@@ -3,4 +3,17 @@ def list_of_organisations
     @list_of_organisations ||=
       [OpenStruct.new(name: "", id: ""), Organisation.sorted_by_name].flatten
   end
+
+  def list_of_user_roles
+    @list_of_user_roles ||= build_user_roles_list
```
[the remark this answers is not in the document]
mec: It was originally one method, but I felt like it made the memoize redunant so I extracted the method. I've never seen `begin` before - looks nice!

app/models/user.rb
mec: Okay cool, how about:

```ruby
def role_name
  t("activerecord.attributes.user.roles.#{role}")
end
````

spec/helpers/form_helper_spec.rb
mec: This is something I find difficult to make calls on, so I can learn, would you mind expanding on why you feel like this isn't necessary?

PR #126, opened 2020-01-08
Chore/175 refactor make fund transaction association polymorphic (part 2)

lozette:
## Next steps


tahb:
Closing since we've move the card back to planned. This PR is referenced to that trello card and  can reopened or just referenced in future if we need it. Removing clears the decks of open stuff that the team cannot progress right now.

PR #131, opened 2020-01-09
Feature/110 active link nav

deberny:
## Changes in this PR

We use a helper method `navigation_item_class` to check if a menu item 'path' is a curent page. If this is true we apply the active class to the `li` element.

### Before

Screenshot 2020-01-09 at 16 00 05

### After

Screenshot 2020-01-09 at 15 58 55

## Next steps

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -8,6 +8,6 @@ module ApplicationHelper
   def navigation_item_class(path)
     classes = ["govuk-header__navigation-item"]
     classes << "govuk-header__navigation-item--active" if current_page?(path)
-    return classes.join(" ")
+    classes.join(" ")
   end
 end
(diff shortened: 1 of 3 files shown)


lozette:
This PR looks fine except for `standard` complaining about syntax, if you fix that up I will approve!

app/helpers/application_helper.rb
```diff
@@ -4,4 +4,10 @@ module ApplicationHelper
   def l(object, options = {})
     super(object, options) if object
   end
+
+  def navigation_item_class(path)
+    classes = ["govuk-header__navigation-item"]
+    classes << "govuk-header__navigation-item--active" if current_page?(path)
+    return classes.join(" ")
```
lozette: If you remove this `return` standard will pass

deberny:
ah, I just pushed a force change thinking that'll fix it

app/helpers/application_helper.rb
deberny: brilliant, thank you!

deberny:
@lozette fix is pushed, waiting for checks 😄

PR #134, opened 2020-01-13
Chore/Update Auth0 ADR to include Classic views

mec:
To cover the decision to use the 'Classic' login views from Auth0 which
offer the best balance between user experience and technical debt in the
service.


doc/architecture/decisions/0006-use-auth0-for-authentication.md
```diff
@@ -14,11 +14,21 @@ In order to implement this quickly, we'll use Auth0 to manage this.
 As Auth0's authentication uses OAuth2, it should be straightforward to migrate
 to another service, if BEIS have a preference for something else.
 
+Auth0 provides views for authentication that we can use in our user journeys.
+There are two versions of these views 'Classic' and 'New'.
+
+Classic uses JavaScript and is not progressively enhanced. New uses language
+that result in a poor user experience.
```
lozette: `result` should be `results` (results in a poor user experience)

PR #142, opened 2020-01-17
Fund managers can create a programme activity

lozette:
## Changes in this PR

Trello: https://trello.com/c/Y2x07ws7/202-fund-managers-can-create-a-programme-activity

Allow Fund Managers to add additional information to a Programme via an attached Activity (and activity form)

The activity form journey for a Programme is exactly the same as that for a Fund. This will need to be looked at in future, as some of the form defaults for Programmes may not be the same as for Funds.

Additionally, there are some TODOs in this PR around how we handle polymorphic links between Funds -> Organisations, and Programmes -> Funds, for example in `activity_helper.rb`

## Next steps

### DIFF
--- a/app/policies/activity_policy.rb
+++ b/app/policies/activity_policy.rb
@@ -4,27 +4,23 @@ class ActivityPolicy < ApplicationPolicy
   end
 
   def show?
-    user.administrator? ||
-      fund? && user.fund_manager?
+    user.administrator? || (user.fund_manager? && hierarchy?)
   end
 
   def create?
-    user.administrator? ||
-      fund? && user.fund_manager?
+    user.administrator? || (user.fund_manager? && hierarchy?)
   end
 
   def update?
-    user.administrator? ||
-      fund? && user.fund_manager?
+    user.administrator? || (user.fund_manager? && hierarchy?)
   end
 
   def destroy?
-    user.administrator? ||
-      fund? && user.fund_manager?
+    user.administrator? || (user.fund_manager? && hierarchy?)
   end
 
-  private def fund?
-    record.hierarchy.is_a?(Fund)
+  private def hierarchy?
+    record.hierarchy.is_a?(Programme) || record.hierarchy.is_a?(Fund)
   end
 
   class Scope < Scope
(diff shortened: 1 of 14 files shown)


lozette:
Opening this for review, hoping that someone can assist in the polymorphism fun in `activity_helper`

config/routes.rb
```diff
@@ -24,10 +24,10 @@
     end
 
     resources :funds, only: [], concerns: [:activity, :transactionable] do
-      resources :programmes, only: [:new, :create, :show]
+      resources :programmes, except: [:destroy]
     end
-    # TODO: Extend with more hierarchies using this format
-    # resources :programmes, only: [], concerns: [:activity, :transactionable]
+
+    resources :programmes, only: [], concerns: [:activity, :transactionable]
```
leeky: Should we keep the comment in place until we have all four levels implemented..?

app/policies/activity_policy.rb
```diff
@@ -5,35 +5,54 @@ def index?
...
+      fund_and_fund_manager? ||
+      programme_and_fund_manager?
   end
 
   private def fund?
     record.hierarchy.is_a?(Fund)
   end
 
+  private def programme?
+    record.hierarchy.is_a?(Programme)
```
leeky: As we're not applying specific permissions only on either funds and programmes, how about replacing these with:

```ruby
private def hierarchy?
  record.hierarchy.is_a?(Programme)  || record.hierarchy.is_a?(Fund)
end
```

and then doing:

`user.adminstrator? || (user.fund_manager? && hierarchy?)`

app/policies/activity_policy.rb
```diff
@@ -5,35 +5,54 @@ def index?
...
     def resolve
       if user.administrator?
         scope.all
       else
+        # TODO: This will eventually become unsustainable,
+        # we need a better way to do this
         funds = Fund.where(organisation_id: user.organisation)
-        scope.where(hierarchy: funds)
+        programmes = Programme.where(fund_id: funds)
+        scope.where(hierarchy: [funds + programmes])
```
leeky: Good point. Never considered that this scope will include all four levels, so we need to avoid a horrible query!

PR #153, opened 2020-01-27
Feature/214 users land on their organisation show instead of dashboard

deberny:
## Changes in this PR

 Trello: https://trello.com/c/S1w9OyXP/214-all-user-roles-are-taken-to-the-organisation-show-page-rather-than-a-dashboard

When a user longs in, they land on their "home" organisation show page, instead of a dashboard.

If the user is a fund manager or administrator, they can view the organisation index page and the show pages for other organisations.

If the user is a delivery partner, they will only be able to view the show page for their own organisation.

### Fund manager view after log in and viewing other's organisation

Land on org page - FM A

### Delivery partner view after log in, they cannot view organisations or edit their own

land on org page - DP

## Next steps


app/controllers/staff/organisations_controller.rb
```diff
@@ -3,6 +3,7 @@
 class Staff::OrganisationsController < Staff::BaseController
   def index
     @organisations = policy_scope(Organisation)
+    authorize @organisations
```
mec: Do we need to call `authorize` here? We already have the scope of organisations?

app/views/shared/_navigation.html.haml
```diff
@@ -2,6 +2,6 @@
   %ul{class: 'govuk-header__navigation', aria: { label: "Top Level Navigation"}}
     -if defined?(current_user) && current_user
 
-      %li{ class: navigation_item_class(dashboard_path) }
-        %a{ href: dashboard_path, class: 'govuk-header__link' }
-          Dashboard
+      %li{ class: navigation_item_class(organisation_path(current_user.organisation)) }
+        %a{ href: organisation_path(current_user.organisation), class: 'govuk-header__link' }
+          = I18n.t("page_title.dashboard")
```
LornaHarwood: Do we need this `I18n`?
I noticed we don't use it in the application layout, do we need it here?

app/views/staff/organisations/show.html.haml
```diff
@@ -1,16 +1,19 @@
...
+    .govuk-grid-column-full
       %h1.govuk-heading-xl
         = t("page_title.organisation.show", name: @organisation_presenter.name)
 
   .govuk-grid-row
     .govuk-grid-column-two-thirds
       %h2.govuk-heading-m Details
-      = link_to t("page_content.organisation.button.edit"), edit_organisation_path(@organisation_presenter), class: "govuk-button"
+
+      - if policy(Organisation).edit?
```
mec: I notice we used `edit?` here, I was expecting `update?` which threw me for a bit, is there a reason to use edit instaed?

config/locales/en.yml
```diff
@@ -203,7 +203,7 @@ en:
         sector: Sector
         status: Status
         tied_status: Tied status
-    dashboard: Dashboard
+    dashboard: Home
```
mec: Should we change this to something else as we have removed the concept of a 'dashboard' entirely? Maybe something like `home_navigation_item` or just `home`?

spec/features/staff/fund_managers_can_edit_a_fund_spec.rb
```diff
@@ -12,14 +12,14 @@
 
   context "when the user is a fund_manager" do
     before do
-      authenticate!(user: build_stubbed(:fund_manager, organisation: organisation))
+      authenticate!(user: create(:fund_manager, organisation: organisation))
```
mec: We used a `let` in spec/features/staff/fund_managers_can_create_an_organisation_spec.rb but it was in the `before` for this one, which I like better – could we do one for consistancy?

spec/features/staff/fund_managers_can_edit_an_organisation_spec.rb
```diff
@@ -1,27 +1,28 @@
...
     end
   end
 
-  context "when the user is allowed to edit an organisation" do
+  context "when the fund manager is allowed to edit an organisation" do
     scenario "successfully editing an organisation" do
-      authenticate!(user: build_stubbed(:fund_manager))
+      authenticate!(user: create(:fund_manager, organisation: beis_organisation))
 
       successfully_edit_an_organisation
```
mec: It looks like `successfully_edit_an_organisation` is only used here and it relies on the two organisations, I wonder if just including it here would make the spec easier to read?

spec/features/staff/fund_managers_can_invite_new_users_spec.rb
```diff
@@ -86,7 +90,9 @@
           click_button I18n.t("generic.button.submit")
 
           expect(page).to have_content(I18n.t("form.user.create.failed"))
-          expect(User.count).to eq(0)
+
+          expect(User.count).to eq(1) # This is the fund_manager
```
mec: Is there another way to express this expectation, this feels like it relies on prior knowledge of there only being one `User` that seems a bit tricky to follow? Can we expect the number of Users to not change in some way?

Perhaps we do not need the count check at all if we are confident that seeing the error means no User was created?

spec/features/staff/fund_managers_can_view_other_users_spec.rb
```diff
@@ -46,16 +50,20 @@
   end
 
   context "when the user is a delivery_partner" do
-    before { authenticate!(user: build_stubbed(:delivery_partner)) }
-
     scenario "cannot view the user" do
```
mec: Could we re-word this to better express what this test is doing? Something like 'does not see the mange users button on their organisation page'?

spec/helpers/organisation_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      let(:user) { create(:delivery_partner, organisation: organisation) }
+
+      context "the user is on their own organisation's show page" do
+        it "does not show a back link" do
+          params = {controller: "staff/organisations", action: "show", id: organisation.id}
+          expect(helper.organisation_page_back_link(user, params)).to eq(nil)
+        end
+      end
+
+      context "the user is on another organisation's show page" do
```
mec: Is this a situation we could actually be in? A delivery partner cannot see an organisation that is not theirs, do we need the coverage?

app/views/staff/organisations/show.html.haml
lozette: No, just an error I think!

spec/features/staff/fund_managers_can_invite_new_users_spec.rb
lozette: Agree, I really hate this check but I couldn't think of a better way to do it!

PR #154, opened 2020-01-28
Fund creation refactored to use a shareable Activity model

tahb:
This is an enormous pull request that has been paired on by most of the team. The commits are large due to the nature of these brutal changes that broke almost everything. We do however have commits that should each result in a green test suite.

After spiking this in a separate branch we split the work up into 5 phases:

1. Revert back to a recent point in time before programmes were added to make it easier
2. Merge all the `Activity` fields, tests and journeys into the existing equivalant for `Fund` 
3. Delete all occurrences of `Activity` now that the functionality has been duplicated
4. Rename all occurrences of `Fund`  back to `Activity` to reclaim the desired name
5. Store the level/hierarchy of an `Activity` on a new `level` field and ensure the fund manager still sees meaningful language to them around the "creation of funds" rather than "creation of activities". Activity should only be used as a term under the hood.

## Changes in this PR

- we have decided to iterate the current data modelling to simplify the structure so that:
  * adding new levels/hierarchy is faster due to fewer moving parts
  * the codebase is easier to work on as it takes less time to understand
- fund creation now only includes a single step, the previous step to add a fund name has been completely removed. Instead the user goes straight to the multi-step form
- the previous implementation stored the hierarchy of an activity through a dedicated model such as `Fund`. This has been replaced by an enum on the Activity record that is called `level`. This could later change to be a polymorphic association to new tables to help us store fields only required for specific levels of activity
- given this change has gone back in time to a [particular commit](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/4b89ddf155b468ce44070f7e3f3769d583531780) we will need to organise ourselves to readd
  * programme creation
  * programme editing (still in progress)
  * budgets
  * users can only belong to a single organisation

## Next steps


tahb:
[] - make a plan to rescue work that needs to be redone then force develop back to commit https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/4b89ddf155b468ce44070f7e3f3769d583531780 so that this can be merged free of conflicts
[] - write an ADR explaining what the intention is for our data model so that is committed as code

PR #155, opened 2020-01-29
Dates updated to have human readable format as default

deberny:
## Changes in this PR

We have set  a default date format in translation locale. Added presenters for all the fund activity dates and included tests checking if those specific dates are in a human readable format.

We used ruby travel_to helper method in our tests to ensure that they will continue to pass without being dependent on a current date.  Especially becaue we know that activity dates are validated to be within a certain time frame.

We thought about using strftime in our helper, but then we realised it would’ve been duplicated in a lot of places. Therefore setting the date format in locale file means we don’t need to worry about using locale in the views for an activity and it will be easier to change our time formats in the future.

Co-authored-by: Tom <tomhipkin@gmail.com>

### Before

Screenshot 2020-01-28 at 12 13 00

### After

Screenshot 2020-01-29 at 15 40 36

## Next steps

### DIFF
--- a/app/presenters/activity_presenter.rb
+++ b/app/presenters/activity_presenter.rb
@@ -16,6 +16,26 @@ class ActivityPresenter < SimpleDelegator
     I18n.t("activity.status.#{super}")
   end
 
+  def planned_start_date
+    return if super.blank?
+    I18n.l(super)
+  end
+
+  def planned_end_date
+    return if super.blank?
+    I18n.l(super)
+  end
+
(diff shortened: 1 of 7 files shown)


deberny:
@mec Thank you for your comments, these are valid points. I'd be happy to address them and refactor the presenter spec to be morte robust, would be great to pair up on this. I have an upcoming card to work on date formats for transaction, so maybe we can do that work then?

Let me know if I can merge and attach this PR's comment in the new card.

PR #163, opened 2020-01-30
(Feature/257) Re-introduce programmes

mec:
## Changes in this PR
Now the work to simplify the activity model is complete we can re-introduce the programme level.

At this point the aim is to just be able to add programme level activities, not associate them to funds which will be the next step. By adding programme level activities now we can un-block work that relies on an assoication to programme level activities.

We thought about showing programme level activities on the organisation show page but decided not to for two reasons:

1. It felt more natural for them to appear on the associated fund level activity show page and we do not have that associations as yet
2. It would make the scope feature tests more complicated

We wondered about calling the instance variable in `ProgrammeController` `@programme` but settled on `@activity` we may want to think about this as more activity levels are added to the service?

## Screenshots of UI changes

The 'Create programme' button:
Screenshot_2020-01-30 Organisation Department for Business, Energy Industrial Strategy — Report Official Development Assist

A successfully created programme level activity:
Screenshot_2020-01-30 Activity Another Programme — Report Official Development Assistance

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,8 +27,8 @@
 - Restrict delivery partners so they can only view and edit their own organisation
 - Fund managers can manage users, organisations, funds, fund activites and fund transactions
 - Transaction and Activity dates are restricted to 10 years in the past or 25 years in the future at most
-- Fund managers can create basic programmes
 - Provide a way to flag an organisation as BEIS
 - User email addresses must be valid emails
 - Users are only associated with one organisation
 - Users land on their organisation#show page when they log in, instead of a "dashboard"
+- Fund manager can add a programme level activity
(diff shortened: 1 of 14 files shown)


mec:
@tahb 

1. Yep, I believe so - I'll drop something in Slack to the team to be sure.  

2. Agreed, if we had users this would be a different conversation! Having rubber ducked the association this morning I feel confident that a simple `has_many - belongs_to` is all that is needed for this stage but I am reluctant to make that call alone…   
This work feels like the place to add the association, I'd be very happy to try and get to the decision on the association and then go back over this to make it more 'complete'?  

3. Oops! Thanks! fixed.

mec:
We've settled on an approach for making associations between activities, see https://trello.com/c/zqHR7vCB. So we'll roll that work into this PR as this feels like the right place to introduce it.

I've closed this PR and will append the association work to the end and also add and display programmes in the appropriate place i.e. to a fund.

PR #164, opened 2020-01-30
Feature/214 users land on their organisation show instead of dashboard

lozette:
A new version of https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/153 rebased over the "New World" `develop`

## Changes in this PR

Trello: https://trello.com/c/S1w9OyXP/214-all-user-roles-are-taken-to-the-organisation-show-page-rather-than-a-dashboard

When a user logs in, they land on their "home" organisation show page, instead of a dashboard.

If the user is a fund manager or administrator, they can view the organisation index page and the show pages for other organisations.

If the user is a delivery partner, they will only be able to view the show page for their own organisation.

### Fund manager view after log in and viewing other's organisation

Land on org page - FM A

### Delivery partner view after log in, they cannot view organisations or edit their own

land on org page - DP

## Next steps


app/controllers/auth0_controller.rb
```diff
@@ -5,7 +5,16 @@ def callback
     session[:userinfo] = request.env["omniauth.auth"]
 
     # Redirect to the URL you want after successful auth
-    redirect_to "/dashboard"
+
+    user = User.find_by(email: session[:userinfo][:info][:email])
```
mec: Could we use `current_user` and possibly save the call to the database?

lozette: Apparently we can!

app/views/staff/organisations/show.html.haml
```diff
@@ -3,26 +3,15 @@
...
-      - if policy(Organisation).index?
-        %h3.govuk-heading-m
-          = t("page_content.dashboard.header.manage_organisations")
-        = link_to t("page_content.dashboard.button.manage_organisations"), organisations_path, class: "govuk-button"
-
     .govuk-grid-column-two-thirds
       %h2.govuk-heading-m Details
 
-      - if policy(Organisation).edit?
+      - if policy(Organisation).update?
```
mec: 👍

PR #165, opened 2020-01-30
(204) Editing a complete activity form returns to the activity

leeky:
@mec says:

I've picked this up whilst @leeky was on a support sprint.

The work changed once we refactored the model and was only about returning the user to the appropriate place when editing an activity.

- when the activity is complete, the user edits an attribute and returns to the activity summary view
- when the activity is incomplete, the user edits or adds an attribute they continue the stepped form flow.

I also found some of the Actvity factory traits weren't reflecting the behaviour of the model so took the opportunity to update them:

Becuase an Activity can now have a nil identifier attribute when the `wizard_step` is 'identifier' some specs were updated to use the 'purpose' step as this better reflects the state of the activity under test i.e. has a 'identifier' value set.

I've left @leeky comment as it is mostly his work:

----
Allow editing a programme activity.

This introduces a `wizard_complete?` method on an activity (completed wizards set the `wizard_status` to 'complete'). This allows edits to individual form steps to return to the summary page afterwards, instead of continuing on as per the initial run through of the wizard.

### DIFF
--- a/spec/features/staff/fund_managers_can_edit_an_activity_spec.rb
+++ b/spec/features/staff/fund_managers_can_edit_an_activity_spec.rb
@@ -6,23 +6,6 @@ RSpec.feature "Fund managers can edit a fund level activity" do
   context "when the user is a fund_manager" do
     before { authenticate!(user: create(:fund_manager, organisation: organisation)) }
 
-    scenario "clicking edit starts the ActivityForm journey from that step" do
-      activity = create(:activity, organisation: organisation)
-
-      visit organisation_activity_path(activity.organisation, activity)
-
-      # Click the first edit link that opens the form on step 1
-      within(".identifier") do
-        click_on(I18n.t("generic.link.edit"))
-      end
-
-      # This helper fills in the form from step 1
-      fill_in_activity_form
-
-      # TODO: On Edit we shouldn't show a message that says "successfully created"
-      expect(page).to have_content I18n.t("form.fund.create.success")
-    end
-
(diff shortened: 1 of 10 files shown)


spec/features/staff/fund_managers_can_edit_an_activity_spec.rb
```diff
@@ -45,6 +28,22 @@
     end
 
     context "when the activity is complete" do
+      it "editing and saving a step returns the user to the activity page" do
+        activity = create(:activity, organisation: organisation)
+        identifier = "AB-CDE-1234"
+        visit organisation_activity_path(activity.organisation, activity)
+
+        within(".identifier") do
+          click_on("Edit")
```
lozette: This feels so fussy but I'd rather this was `I18n.t("generic.link.edit")` than the string Edit, to keep it consistent!

spec/features/staff/fund_managers_can_edit_an_activity_spec.rb
```diff
@@ -54,6 +53,25 @@
...
 
+    context "when the activity is incomplete" do
+      it "editing and saving a step goes to the next step in the form" do
+        activity = create(:activity, :at_country_step, organisation: organisation)
+        recipient_region = "Oceania, regional"
+
+        visit organisation_activity_path(activity.organisation, activity)
+
+        within(".recipient_region") do
+          click_on("Edit")
```
lozette: As above!

PR #173, opened 2020-02-03
(227) Activities collect information about participating organisations that provide the funds

lozette:
## Changes in this PR

Trello: https://trello.com/c/LZjMGZPu/227-activities-collect-information-about-the-participating-organisations-that-provides-the-funds-funder

Add the Funding Organisation information to an Activity. For now, this is a hard-coded set of information, which differs depending on whether the Activity is a programme or fund activity. 

Eventually we will make this information something the Fund managers can choose, but for now it is static. 

## Next steps

### DIFF
--- a/app/controllers/staff/funds_controller.rb
+++ b/app/controllers/staff/funds_controller.rb
@@ -8,6 +8,9 @@ class Staff::FundsController < Staff::ActivitiesController
 
     @activity.wizard_status = "identifier"
     @activity.level = :fund
+    @activity.funding_organisation_name = "HM Treasury"
+    @activity.funding_organisation_reference = "GB-GOV-2"
+    @activity.funding_organisation_type = "10"
     @activity.save(validate: false)
 
     redirect_to activity_step_path(@activity.id, @activity.wizard_status)
--- a/app/controllers/staff/funds_controller.rb
+++ b/app/controllers/staff/funds_controller.rb
@@ -2,16 +2,8 @@
 
 class Staff::FundsController < Staff::ActivitiesController
   def create
-    @activity = Activity.new
-    @activity.organisation = Organisation.find(organisation_id)
+    @activity = CreateFundActivity.new(organisation_id: organisation_id).call
     authorize @activity
-
-    @activity.wizard_status = "identifier"
-    @activity.level = :fund
-    @activity.funding_organisation_name = "HM Treasury"
-    @activity.funding_organisation_reference = "GB-GOV-2"
-    @activity.funding_organisation_type = "10"
-    @activity.save(validate: false)
 
     redirect_to activity_step_path(@activity.id, @activity.wizard_status)
   end
--- a/app/services/create_fund_activity.rb
+++ b/app/services/create_fund_activity.rb
@@ -1,0 +1,20 @@
+class CreateFundActivity
+  attr_accessor :organisation_id
+
+  def initialize(organisation_id:)
+    self.organisation_id = organisation_id
+  end
+
+  def call
+    activity = Activity.new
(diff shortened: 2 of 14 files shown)


app/controllers/staff/funds_controller.rb
```diff
@@ -8,6 +8,9 @@ def create
 
     @activity.wizard_status = "identifier"
     @activity.level = :fund
+    @activity.funding_organisation_name = "HM Treasury"
+    @activity.funding_organisation_reference = "GB-GOV-2"
+    @activity.funding_organisation_type = 10
```
tahb: We're storing this field as a string but setting an integer. Presumably ActiveRecord is figuring it out for us? It might be clearer to the reader to set this as a string instead.

spec/features/staff/fund_managers_can_create_a_fund_level_activity_spec.rb
```diff
@@ -39,6 +39,18 @@
       expect(page.find("option[@selected = 'selected']").text).to eq activity_presenter.tied_status
     end
 
+    scenario "the activity has the appropriate funding organisation defaults" do
+      visit organisation_path(organisation)
+      click_on(I18n.t("page_content.organisation.button.create_fund"))
+
+      fill_in_activity_form
+
+      activity = Activity.where(level: :fund).last
```
tahb: As I think we only want to query for 1 `activity` record we could change this `where` to a `find_by`. It would be more helpful to get a test failure saying `RecordNotFound` than a  more ambigious error for calling `.last` on an empty array, or if `.last` someday returned a different activity record entirely! Something like this may help:

```
fill_in_activity_form(identifier: 'a-fund-has-a-funding-organisation')

activity = Activity.find_by(identifier: 'a-fund-has-a-funding-organisation')
```

I think we could do this on both feature tests here?

spec/factories/activity.rb
```diff
@@ -22,12 +22,18 @@
...
 
-    factory :programme_activity do
-      level { :programme }
-    end
-
     factory :fund_activity do
       level { :fund }
+      funding_organisation_name { "HM Treasury" }
+      funding_organisation_reference { "GB-GOV-2" }
+      funding_organisation_type { 10 }
```
tahb: Same point about using an integer for a string type in the factories. Think it would be clearer to change this to "10"?

lozette:
Good call on the service!! Ready for another look @tahb

spec/services/create_fund_activity_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+RSpec.describe CreateFundActivity do
+  let(:organisation) { create(:organisation) }
+
+  describe "#call" do
+    it "successfully creates a Fund Activity" do
+      result = described_class.new(organisation_id: organisation.id).call
+
+      expect(result.organisation).to eq(organisation)
```
tahb: I think there is value in splitting these assertions out into their own isolated unit tests here. Doing so can help us document what each part is responsible for which in turn helps us know where to make future changes with confidence, without affecting other parts. Individual tests also allow us to wrap them in context blocks for unhappy paths. 

```
it "associates the new activity to the organisation passed in"
it "sets the form wizard to the first step"
it "sets the level as type 'fund'"
it "stores information about the organisation that funds this activity"
```

app/services/create_fund_activity.rb
```diff
@@ -0,0 +1,20 @@
+class CreateFundActivity
```
tahb: This LGTM 🎉

tahb:
I'd love it if we could separate our new unit tests, then we're good to merge!

PR #175, opened 2020-02-04
(261) Activities collect information about accountable participating organisations

lozette:
## Changes in this PR

Trello: https://trello.com/c/9gobZxRi/261-activities-collect-information-about-the-participating-organisations-that-are-accountable

Add Accountable Organisation details to Activity records. These are `participating-org` organisations with a `@role` of 2, according to IATI.

Currently these values are hardcoded for MVP, but eventually they will be selectable. 

## Next steps

### DIFF
--- a/spec/features/staff/fund_managers_can_view_an_activity_as_xml_spec.rb
+++ b/spec/features/staff/fund_managers_can_view_an_activity_as_xml_spec.rb
@@ -19,10 +19,14 @@ RSpec.feature "Fund managers can view an activity as XML" do
         expect(xml.at("iati-activity/iati-identifier").text).to eq(activity.identifier)
 
         # The funding organisation XML is present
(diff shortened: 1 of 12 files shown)


spec/features/staff/fund_managers_can_view_an_activity_as_xml_spec.rb
```diff
@@ -19,10 +19,14 @@
         expect(xml.at("iati-activity/iati-identifier").text).to eq(activity.identifier)
 
         # The funding organisation XML is present
-        expect(xml.at("iati-activity/participating-org/@ref").text).to eq(activity.funding_organisation_reference)
```
mec: Is it worth mentioning in the commit message why this changed?

lozette: Good call, I will update the commit message

PR #180, opened 2020-02-06
Chore/299 signing in with auth0 does not require js

tahb:
## Changes in this PR

- Users can sign into the service both when JS is enabled **and** disabled in the browser
- Fix the sign in button to work without JS after an unexpected error occurred with where Rails was unable to render the `data` attribute and convert the link into a POST request
- Auth0 has been updated to switch to the "new experience" - this shouldn't affect the development team unless they happen to have JS disabled

## Next steps

### DIFF
--- a/doc/architecture/decisions/0015-disable-auth0s-requirement-for-javascript-during-the-authentication-journey.md
+++ b/doc/architecture/decisions/0015-disable-auth0s-requirement-for-javascript-during-the-authentication-journey.md
@@ -1,0 +1,28 @@
+# 16. Disable Auth0's requirement for Javascript during the authentication journey
+
+Date: 2020-02-05
+
+## Status
+
+Accepted
+
+Partial supersede of [6. Use Auth0 for authentication](0006-use-auth0-for-authentication.md)
+
+## Context
+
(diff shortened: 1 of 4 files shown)


doc/architecture/decisions/0015-disable-auth0s-requirement-for-javascript-during-the-authentication-journey.md
```diff
@@ -0,0 +1,28 @@
...
+## Context
+
+The welcome and the sign in journey currently bounce our user to pages controlled by Auth0. Auth0 give us 2 options called 'experiences' in the admin area under 'Universal Login':
+
+1. Classic - requires Javascript to be enabled in the browser
+2. New - no Javascript required
+
+https://auth0.com/docs/universal-login/new
+
+The [service standard does not explicitly mention failing for JS](https://www.gov.uk/service-manual/service-assessments/pre-july-2019-digital-service-standard) however  there is [guidance in the service manual advises that we use progressive enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement) so that when a user doesn't have JS enabled, the service always remains functional.
```
mec: Double space after 'however'?

PR #184, opened 2020-02-12
(270) Users can create a project (level C) activity

lozette:
Needs https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/182/commits/033fd5e436ef6dfdb3de47f301286094aaa1a3db to be merged before tests will pass!

## Changes in this PR

Trello: https://trello.com/c/kOKYJ1UE/270-users-can-create-a-project-level-c-activity

Project (level C) activities can be added.

This PR required a lot of changes to the things Delivery partners are allowed to see. Previously, DPs could not view fund or programme activities, but they need to be able to navigate through those activities to create a project "under" a programme. 

DPs can now see a "read only" view of funds and programmes, and cannot add budgets or transactions to those activities. But they can create & update project level activities, and add budgets & transactions to projects.

As a side-effect of doing this work, budgets can now be added to fund level activities (https://trello.com/c/RsLTbcF2/280-users-can-add-a-budget-to-a-fund)

## Next steps

### DIFF
--- a/app/controllers/staff/organisations_controller.rb
+++ b/app/controllers/staff/organisations_controller.rb
@@ -11,7 +11,7 @@ class Staff::OrganisationsController < Staff::BaseController
     authorize organisation
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
-    activities = policy_scope(Activity.funds).includes(:organisation).where(organisation: organisation)
+    activities = policy_scope(Activity.fund).includes(:organisation).where(organisation: organisation)
     @activities = activities.map { |activity| ActivityPresenter.new(activity) }
   end
 
--- a/app/helpers/activity_helper.rb
+++ b/app/helpers/activity_helper.rb
@@ -13,6 +13,6 @@ module ActivityHelper
 
   def activity_back_path(activity)
     return organisation_path(activity.organisation) if activity.is_fund_level?
-    return organisation_activity_path(activity.parent_activity.organisation, activity.parent_activity) if activity.is_programme_level?
+    organisation_activity_path(activity.parent_activity.organisation, activity.parent_activity)
   end
 end
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -20,10 +20,12 @@ class Activity < ApplicationRecord
   enum level: {
     fund: "fund",
     programme: "programme",
+    project: "project",
   }
 
   scope :funds, -> { where(level: :fund) }
   scope :programmes, -> { where(level: :programme) }
+  scope :projects, -> { where(level: :project) }
 
   def identifier_step?
     wizard_status == "identifier" || wizard_complete?
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -23,10 +23,6 @@ class Activity < ApplicationRecord
     project: "project",
   }
 
-  scope :funds, -> { where(level: :fund) }
-  scope :programmes, -> { where(level: :programme) }
-  scope :projects, -> { where(level: :project) }
-
   def identifier_step?
     wizard_status == "identifier" || wizard_complete?
   end
--- a/app/policies/activity_policy.rb
+++ b/app/policies/activity_policy.rb
@@ -19,18 +19,20 @@ class ActivityPolicy < ApplicationPolicy
     user.administrator? || user.fund_manager?
   end
 
-  private def associated_user?
-    user.organisation.id.eql?(record.organisation_id)
+  private def associated_delivery_partner?
+    user.delivery_partner? && user.organisation.id.eql?(record.organisation_id)
   end
 
   private def user_can_access?
     return true if user.administrator? || user.fund_manager?
-    associated_user?
+    associated_delivery_partner?
   end
 
   private def user_can_create_and_update?
-    return true if user.administrator? || user.fund_manager?
-    associated_user? && record.level.eql?("project")
+
+    return true if user.administrator?
+    return true if user.fund_manager? && !record.level.eql?("project")
+    associated_delivery_partner? && record.level.eql?("project")
   end
 
   class Scope < Scope
--- a/app/policies/activity_policy.rb
+++ b/app/policies/activity_policy.rb
@@ -1,22 +1,36 @@
 class ActivityPolicy < ApplicationPolicy
   def index?
-    user.administrator? || user.fund_manager?
+    true
   end
 
   def show?
-    user.administrator? || user.fund_manager?
+    user_can_access?
   end
 
   def create?
-    user.administrator? || user.fund_manager?
+    user_can_create_and_update?
   end
 
   def update?
-    create?
+    user_can_create_and_update?
   end
 
   def destroy?
     user.administrator? || user.fund_manager?
+  end
+
+  private def associated_user?
+    user.organisation.id.eql?(record.organisation_id)
+  end
+
+  private def user_can_access?
+    return true if user.administrator? || user.fund_manager?
+    associated_user?
+  end
+
+  private def user_can_create_and_update?
+    return true if user.administrator? || user.fund_manager?
+    associated_user? && record.level.eql?("project")
   end
 
   class Scope < Scope
@@ -24,7 +38,7 @@ class ActivityPolicy < ApplicationPolicy
       if user.administrator? || user.fund_manager?
         scope.all
       else
-        scope.none
+        scope.where(organisation: user.organisation)
       end
     end
   end
--- a/app/policies/transaction_policy.rb
+++ b/app/policies/transaction_policy.rb
@@ -5,17 +5,20 @@ class TransactionPolicy < ApplicationPolicy
 
   def show?
     user.administrator? ||
-      activity? && user.fund_manager?
+      activity? && user.fund_manager? ||
+      activity_is_project_level? && user.delivery_partner?
   end
 
   def create?
     user.administrator? ||
-      activity? && user.fund_manager?
+      activity? && user.fund_manager? ||
+      activity_is_project_level? && user.delivery_partner?
   end
 
   def update?
     user.administrator? ||
-      activity? && user.fund_manager?
+      activity? && user.fund_manager? ||
+      activity_is_project_level? && user.delivery_partner?
   end
 
   def destroy?
@@ -25,6 +28,12 @@ class TransactionPolicy < ApplicationPolicy
 
   private def activity?
     record.activity_id.present?
+  end
+
+  private def activity_is_project_level?
+    return unless activity?
+    activity = Activity.find(record.activity_id)
+    activity.project?
   end
 
   class Scope < Scope
--- a/app/services/create_project_activity.rb
+++ b/app/services/create_project_activity.rb
@@ -1,9 +1,10 @@
 class CreateProjectActivity
-  attr_accessor :organisation_id, :programme_id
+  attr_accessor :organisation_id, :programme_id, :user
 
-  def initialize(organisation_id:, programme_id:)
+  def initialize(organisation_id:, programme_id:, user:)
     self.organisation_id = organisation_id
     self.programme_id = programme_id
+    self.user = user
   end
 
   def call
@@ -22,6 +23,8 @@ class CreateProjectActivity
     activity.accountable_organisation_name = "Department for Business, Energy and Industrial Strategy"
     activity.accountable_organisation_reference = "GB-GOV-13"
     activity.accountable_organisation_type = "10"
+
+    activity.extending_organisation = user.delivery_partner? ? user.organisation : nil
 
     activity.save(validate: false)
     activity
--- a/app/services/create_project_activity.rb
+++ b/app/services/create_project_activity.rb
@@ -1,0 +1,29 @@
+class CreateProjectActivity
+  attr_accessor :organisation_id, :programme_id
+
+  def initialize(organisation_id:, programme_id:)
+    self.organisation_id = organisation_id
+    self.programme_id = programme_id
+  end
+
+  def call
+    activity = Activity.new
+    activity.organisation = Organisation.find(organisation_id)
+    programme = Activity.find(programme_id)
+    programme.activities << activity
+
+    activity.wizard_status = "identifier"
+    activity.level = :project
+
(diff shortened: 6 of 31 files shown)


app/policies/activity_policy.rb
```diff
@@ -1,30 +1,46 @@
...
     user.administrator? || user.fund_manager?
   end
 
+  private def associated_user?
+    user.organisation.id.eql?(record.organisation_id)
+  end
+
+  private def user_can_access?
+    return true if user.administrator? || user.fund_manager?
+    return true if record.class.name.eql?("Class")
```
lozette: This is a bit horrible. Is there a better way of telling if a record has actually been created? `record.persisted?` doesn't work here

tahb: Perhaps `exists?`?

app/models/activity.rb
```diff
@@ -20,10 +20,12 @@ class Activity < ApplicationRecord
   enum level: {
     fund: "fund",
     programme: "programme",
+    project: "project",
   }
 
   scope :funds, -> { where(level: :fund) }
   scope :programmes, -> { where(level: :programme) }
+  scope :projects, -> { where(level: :project) }
```
tahb: As we have `level` as an enum I think we get this (and the 2 pre-existing) scopes for free?

https://api.rubyonrails.org/v5.2.4.1/classes/ActiveRecord/Enum.html

app/policies/activity_policy.rb
tahb: From reading the commit and tests around this private method I'm unsure when or how we get into the pickle where a record we're trying to view doesn't exist. What scenario prompted this guard clause?

app/helpers/activity_helper.rb
```diff
@@ -15,4 +15,9 @@ def activity_back_path(activity)
     return organisation_path(activity.organisation) if activity.is_fund_level?
     return organisation_activity_path(activity.parent_activity.organisation, activity.parent_activity) if activity.is_programme_level?
   end
+
+  def user_can_edit?(user, activity)
```
tahb: This method looks like it might be doing the same authorisation that use Pundit for elsewhere.

If this is the case then I wonder if this could be replaced with something like `policy(@activity).update?` in the view? This would keep all the authorisation logic together in Pundit. We use this quite a lot on [the organisation#show view](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/views/staff/organisations/show.html.haml#L15).

app/views/staff/activities/show.html.haml
```diff
@@ -16,8 +16,9 @@
       .govuk-grid-column-two-thirds
         %h2.govuk-heading-m
           = t("page_content.organisation.programmes")
-        = form_tag(organisation_fund_programmes_path(@activity.organisation, @activity), method: "post") do
-          = submit_tag t("page_content.organisation.button.create_programme"), class: "govuk-button"
+        - if current_user.administrator? || current_user.fund_manager?
```
tahb: I wonder if we can add the concept of `level` into the Pundit policies so that when we ask say `policy(@activity).create?` the `create?` check will take the level into account. 

```
# activity_policy.rb
def create?
  if record.fund? && (current_user.administrator? || current_user.fund_manager?) ||
     record.programme? && (current_user.administrator? || current_user.fund_manager?) ||
     record.project? && (current_user.administrator? || current_user.delivery_partner?)  
```

This would benefit us by keeping our views clean and our authorisation logic DRY with great test coverage.

app/views/staff/activities/show.html.haml
```diff
@@ -40,14 +41,29 @@
       %h2.govuk-heading-m
         = t("page_content.activity.transactions")
 
-      = link_to(I18n.t("page_content.transactions.button.create"), new_activity_transaction_path(@activity), class: "govuk-button")
-      = render partial: '/staff/shared/transactions/table', locals: { transactions: @transaction_presenters }
+        = link_to(I18n.t("page_content.transactions.button.create"), new_activity_transaction_path(@activity), class: "govuk-button")
```
tahb: This indentation change looks like it's moved the create link within the header 2 element. I had a look at another of our create buttons and that has it outside of the preceeding header so perhaps this was an accident one?

app/views/staff/shared/activities/_activity.html.haml
```diff
@@ -12,7 +12,7 @@
     %dd.govuk-summary-list__value
       = activity_presenter.identifier
     %dd.govuk-summary-list__actions
-      - if step_is_complete_or_next?(activity: activity_presenter, step: :identifier)
+      - if step_is_complete_or_next?(activity: activity_presenter, step: :identifier) && user_can_edit?(current_user, activity_presenter)
```
tahb: I left a comment on an earlier commit that added `user_can_edit?` about the potential to move this into Pundit.

spec/features/staff/users_can_edit_an_activity_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.feature "Fund managers can edit a fund level activity" do
+RSpec.feature "Users can edit an activity" do
```
tahb: 👍

app/views/staff/activities/show.html.haml
```diff
@@ -11,59 +11,7 @@
...
-        %h2.govuk-heading-m
-          = t("page_content.organisation.programmes")
-        - if current_user.administrator? || current_user.fund_manager?
-          = form_tag(organisation_fund_programmes_path(@activity.organisation, @activity), method: "post") do
-            = submit_tag t("page_content.organisation.button.create_programme"), class: "govuk-button"
-        %ul.govuk-list.govuk-list--bullet
-          - @activities.each do |activity|
-            %li
-              = link_to activity.display_title, organisation_activity_path(activity.organisation, activity)
+    = render partial: '/staff/shared/activities/activity_activities', locals: { activity: @activity, activities: @activities }
```
tahb: This is a great tidy up, spot on 👍

app/helpers/activity_helper.rb
```diff
@@ -20,4 +20,9 @@ def user_can_edit?(user, activity)
     return true if user.administrator? || user.fund_manager?
     user.delivery_partner? && activity.is_project_level?
   end
+
+  def user_can_add_budget?(user, activity)
```
tahb: Could we merge this logic into the budget policies and ask the Budget policy for this instead with a `policy(Budget).create?`?

app/views/staff/shared/activities/_activity_activities.html.haml
```diff
@@ -0,0 +1,37 @@
...
+          %li
+            = link_to sub_activity.display_title, organisation_activity_path(sub_activity.organisation, sub_activity)
+
+- if activity.is_programme_level?
+  .govuk-grid-row
+    .govuk-grid-column-full
+      %h2.govuk-heading-m
+        = t("page_content.activity.extending_organisation.heading")
+
+      -if @activity.extending_organisation.present?
```
tahb: Looks like we're missing space after the hyphen here.

app/services/create_project_activity.rb
```diff
@@ -23,6 +24,8 @@ def call
     activity.accountable_organisation_reference = "GB-GOV-13"
     activity.accountable_organisation_type = "10"
 
+    activity.extending_organisation = user.delivery_partner? ? user.organisation : nil
```
tahb: Providing our authorisation code is working as intended, is there a scenario where a user other than a delivery partner is allowed to create a project?

app/policies/activity_policy.rb
lozette: You know what I can't even remember! I will remove it.

app/helpers/activity_helper.rb
lozette: Thanks! I still struggle to get my head around Pundit so this is helpful!

app/views/staff/activities/show.html.haml
lozette: So this is where I stumble on Pundit stuff, because if you're checking if a user can create an Activity, you don't yet know what level the Activity is at. If you pry the Activity at this point, you get something like this:

`Activity(id: uuid, organisation_id: uuid, created_at: datetime, updated_at: datetime, identifier: string, sector: string, title: string, description: text, status: string, planned_start_date: date, planned_end_date: date, actual_start_date: date, actual_end_date: date, recipient_region: string, flow: string, finance: string, aid_type: string, tied_status: string, wizard_status: string, level: string, activity_id: uuid, funding_organisation_name: string, funding_organisation_reference: string, funding_organisation_type: string, accountable_organisation_name: string, accountable_organisation_reference: string, accountable_organisation_type: string, extending_organisation_id: uuid)`

There's no way of knowing what level it's at because it's not been created yet.

app/views/staff/activities/show.html.haml
lozette: This change was superseded when I refactored the transaction partial

app/helpers/activity_helper.rb
lozette: Again I don't know how to check the level of the activity at this point, because the budget does not yet exist. If I try to check the parent activity's level on the budget, none exists because the budget has not been created. If you pry the `record` at this point, you get:

`Budget(id: uuid, activity_id: uuid, budget_type: string, status: string, period_start_date: date, period_end_date: date, value: decimal)
`

app/services/create_project_activity.rb
lozette: Yes, I think all user roles can create projects. The tests in "Users can create a project level activity" show all roles can create projects - is this not supposed to be the case?

tahb: I was expecting that only delivery partners would ever create projects, modelling what happens now as they're the main reporter.

My thinking has been that BEIS should be able to create funds and programmes whilst DP can only create projects and sub-projects. There's no harm in delivery partners being able to view the hierarchies above for extra context but there's not a clear need for it.

app/controllers/staff/organisations_controller.rb
```diff
@@ -11,7 +11,7 @@ def show
     authorize organisation
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
-    activities = policy_scope(Activity.funds).includes(:organisation).where(organisation: organisation)
+    activities = policy_scope(Activity.fund).includes(:organisation).where(organisation: organisation)
```
tahb: I'm surprised this works without being plural!

lozette: Plural didn't work at all

app/services/create_project_activity.rb
lozette: OK shall I refactor it do only DPs can create projects? Can other roles view projects?

I'm cognizant that this needs to be done for the project board and I'm off on Monday!

app/models/activity.rb
```diff
@@ -23,10 +23,6 @@ class Activity < ApplicationRecord
     project: "project",
   }
 
-  scope :funds, -> { where(level: :fund) }
```
tahb: 👍

app/helpers/activity_helper.rb
```diff
@@ -30,14 +30,4 @@ def user_can_add_transaction?(user, activity)
     return true if user.delivery_partner? && activity.project?
     user.administrator? || user.fund_manager?
   end
-
```
tahb: I think we could squash the earlier commit (f3d5c89a0f2df1ae1da2f326e9643c04d3fc829b) in which these methods were added and remove the need for a new commit here

app/services/create_project_activity.rb
tahb: I think the story is to enable delivery partners so yes I think so.

> Can other roles view projects?

There is no express need for other roles to view them at the moment. Approval processes will come eventually but that's for later. That said there's probably no impact if they can. I'd go for the simpler and faster option around who can view.

app/helpers/activity_helper.rb
lozette: I tried so many times, I'm normally good with git hygiene but these two methods keep coming back! I'll keep trying

app/helpers/activity_helper.rb
tahb: Ok. Could we make the assumption that if a user can create an activity for a given level (using `policy(Activity).create?`) then that user can also create budgets and transactions?

I'm not aware that we'll have user groups that can create particular parts of activities and not others.

lozette: OK I will try that

app/policies/transaction_policy.rb
```diff
@@ -27,6 +30,12 @@ def destroy?
     record.activity_id.present?
   end
 
+  private def activity_is_project_level?
```
tahb: Wondering if we can replace this method by using `record.project?` in the public methods instead?

app/helpers/activity_helper.rb
```diff
@@ -30,9 +30,4 @@ def user_can_add_transaction?(user, activity)
...
-  def user_can_add_budget?(user, activity)
```
tahb: Could we squash this into 7f21cdfd9ad4a59eabaf7a7e2c762aed0ee56b07 which added this method in the commit before to clear up the git history?

app/views/staff/activities/show.html.haml
tahb: That makes sense. Hmm. I'll have another go and see if I can think of another way to make this less reptitive.

app/helpers/activity_helper.rb
lozette: This commit has been dropped completely

app/policies/transaction_policy.rb
lozette: `record` at this point is a transaction, not an activity. For Transaction create/update we need to get the Activity and then see if it's a project

app/views/staff/activities/show.html.haml
tahb: When I check out this branch this logic isn't in this view so I'm finding it hard to get myself into the scenario to test this.

lozette: It's an outdated commit - this logic is now in `app/views/staff/shared/activities/_activity_activities.html.haml:6`

mec:
Superceeded by #198

PR #188, opened 2020-02-13
(281) Add Budgets to XML

lozette:
## Changes in this PR

Trello: https://trello.com/c/o0YNJlMt/281-add-budgets-to-xml

Budgets are added to the Activity XML as per the IATI standard: http://reference.iatistandard.org/203/activity-standard/iati-activities/iati-activity/budget/

The budget XML passes validation (see https://test-validator.iatistandard.org/view/dqf/files/ba982acc-8187-4885-989c-50bff12c4fd6?isTestfiles=true) but there are some errors in the Activity XML which are not related to budgets.

See also https://trello.com/c/anTHm0iW/301-budget-periods-cannot-be-more-than-1-year

## Next steps

### DIFF
--- a/app/presenters/budget_xml_presenter.rb
+++ b/app/presenters/budget_xml_presenter.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+class BudgetXmlPresenter < SimpleDelegator
+  XML_BUDGET_TYPES = {"original" => 1, "updated" => 2}
+  XML_STATUSES = {"indicative" => 1, "committed" => 2}
+
+  def period_start_date
+    return if super.blank?
+    I18n.l(super, format: :iati)
+  end
+
+  def period_end_date
+    return if super.blank?
+    I18n.l(super, format: :iati)
+  end
+
+  def value
+    super.to_s
+  end
+
+  def budget_type
+    return if super.blank?
+    XML_BUDGET_TYPES[super]
+  end
+
+  def status
+    return if super.blank?
+    XML_STATUSES[super]
+  end
+end
--- a/app/views/staff/shared/xml/_activity.xml.haml
+++ b/app/views/staff/shared/xml/_activity.xml.haml
@@ -26,6 +26,10 @@
   %default-finance-type{"code" => "#{activity.finance}"}/
   %default-aid-type{"code" => "#{activity.aid_type}"}/
   %default-tied-status{"code" => "#{activity.tied_status}"}/
+  - if budgets
+    - budgets.each do |budget|
+      = render partial: "staff/shared/xml/budget", locals: { budget: budget }
   - if transactions
     - transactions.each do |transaction|
       = render partial: "staff/shared/xml/transaction", locals: { transaction: transaction }
+
--- a/app/views/staff/shared/xml/_activity.xml.haml
+++ b/app/views/staff/shared/xml/_activity.xml.haml
@@ -28,7 +28,7 @@
   %default-tied-status{"code" => activity.tied_status}/
   - if budgets
     - budgets.each do |budget|
-      = render partial: "staff/shared/xml/budget", locals: { budget: budget }
+      = render partial: "staff/shared/xml/budget", locals: { budget: BudgetXmlPresenter.new(budget) }
   - if transactions
     - transactions.each do |transaction|
       = render partial: "staff/shared/xml/transaction", locals: { transaction: transaction }
--- a/app/views/staff/shared/xml/_activity.xml.haml
+++ b/app/views/staff/shared/xml/_activity.xml.haml
@@ -1,10 +1,10 @@
-%iati-activity{"default-currency" => "#{activity.default_currency}", "xml:lang" => activity.organisation.language_code }
+%iati-activity{"default-currency" => activity.default_currency, "xml:lang" => activity.organisation.language_code }
   %iati-identifier= activity.identifier
   %reporting-org{"type" => activity.organisation.organisation_type, "ref" => activity.organisation.iati_reference }
     %narrative= activity.organisation.name
   %title
     %narrative= activity.title
-  %description{type: "1"}
+  %description{"type" => "1"}
     %narrative= activity.description
   - if activity.has_funding_organisation?
     %participating-org{"ref" => activity.funding_organisation_reference, "type" => activity.funding_organisation_type, "role" => 1}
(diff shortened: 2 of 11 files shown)


app/views/staff/shared/xml/_budget.xml.haml
```diff
@@ -0,0 +1,5 @@
+%budget{"type" => budget.budget_type, "status" => budget.status}
+  %period-start{"iso-date" => budget.period_start_date.strftime("%Y-%m-%d")}/
```
mec: Is it worth declaring the date format in the locale file so we can use it as a formatter, we could then do something like:

`l(budget.period_start_date, format: :iati_date)`

and manage the date format in a single place?

spec/factories/budget.rb
```diff
@@ -1,7 +1,7 @@
 FactoryBot.define do
   factory :budget do
-    budget_type { "original" }
```
mec: It feels like a shame to loose the clarity in the code of what 1 and 2 represent, do you think we can use a presenter for the xml that renders the values as 1 and 2 but we keep seeing 'original' and 'indicative'?

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -1,69 +1,76 @@
-RSpec.feature "Users can download an activity as XML" do
```
mec: So do we still need the `fund_managers_can_view_an_activity_as_xml_spec.rb` version?

app/views/staff/shared/xml/_budget.xml.haml
mec: I guess we could use a presenter as a solution as well? There is another attribute where a presenter might help, I've not used them before so it may not be a good fit, I'd been keen to learn.

app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -16,16 +16,16 @@
...
-  %activity-date{"iso-date" => "#{activity.planned_end_date}", type: "2"}/
-  %activity-date{"iso-date" => "#{activity.actual_start_date}", type: "3"}/
-  %activity-date{"iso-date" => "#{activity.actual_end_date}", type: "4"}/
-  %recipient-region{"code" => "#{activity.recipient_region}"}/
-  %sector{"vocabulary" => "1", "code" => "#{activity.sector}"}/
-  %default-flow-type{"code" => "#{activity.flow}"}/
-  %default-finance-type{"code" => "#{activity.finance}"}/
-  %default-aid-type{"code" => "#{activity.aid_type}"}/
-  %default-tied-status{"code" => "#{activity.tied_status}"}/
+  %activity-date{"iso-date" => activity.planned_start_date, type: "1"}/
```
mec: Is the xml vaild with `"iso-date => ,`?

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -37,15 +36,26 @@
...
+        expect(xml.at("iati-activity/participating-org[@role = '3']/narrative").text).to eq(activity.extending_organisation.name)
       end
 
       it "contains the transaction XML" do
+        transaction = create(:transaction, activity: activity)
         visit organisation_activity_path(organisation, activity, format: :xml)
         expect(xml.at("iati-activity/transaction/@ref").text).to eq(transaction.reference)
       end
+
+      it "contains the budget XML" do
```
mec: Given the duplication of the tests we perform I was thinking of a generic activity spec that test how the pieces fit together and then in the contexts for each level we could test what makes that level different from the generic, what do you think?

app/views/staff/shared/xml/_budget.xml.haml
lozette: I've added the localisation as I'm not sure how to get it working via a presenter, I might revisit later!

app/views/staff/shared/xml/_activity.xml.haml
lozette: If the value is blank then the attribute name won't show up. Before it was adding `iso-date=""` to the XML which was being parsed by the validator as an invalid date.

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
lozette: No I just forgot to update this one properly!

spec/factories/budget.rb
lozette: I've done it with i18n as I'm not sure how to have it in a presenter and not mess up the view elsewhere. I'm not super hot on presenters but I think they need to have method names the same as the attribute name, so having something like `BudgetPresenter.iati_status` wouldn't work (I think??)

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
lozette: Good call, there is tons of duplication here!

mec:
@tahb I was wondering about adding a presenter that renders the values for the XML here - does that sound like a good approach to handle diffrent views of the data - feels like what presenters are made for but I would like to check before trying.

config/locales/en.yml
```diff
@@ -5,6 +5,7 @@ en:
   date:
     formats:
       default: "%-d %b %Y"
+      iati: "%Y-%m-%d"
```
mec: It would be nice to mention this in the commit, although I can see the format here, making supet clear why there is a date format might help future us? Only needs to be: 

> Note that dates need to be in xs:Date format (YYYY-mm-dd) so we have added a new `iati` date format to the locales file

Or something like that?

app/presenters/budget_xml_presenter.rb
```diff
@@ -0,0 +1,27 @@
...
+  def period_end_date
+    return if super.blank?
+    I18n.l(super, format: :iati)
+  end
+
+  def value
+    super.to_s
+  end
+
+  def budget_type
```
mec: The presenter is cool - I like it! 

I did expect to to see code here to translate the type and status into the numeric value, it might be that I'm not loving all the locale stuff but it feels more like the presenters job to transalate the type into its value rather than the locale file, the XML is the only place it needs to be numeric, right?

What do you think about something like:

````ruby
BUDGET_TYPES = {
"original": "1",
"updated": "2"
}

def budget_type
return if super.blank?
BUDGET_TYPES[super]
end

File: spec/factories/activity.rb [hunk 8]
```diff
@@ -42,6 +42,8 @@
       accountable_organisation_name { "Department for Business, Energy and Industrial Strategy" }
       accountable_organisation_reference { "GB-GOV-13" }
       accountable_organisation_type { "10" }
+
+      association :extending_organisation, factory: :beis_organisation
```
mec: I am not sure this should be here?

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -10,88 +10,14 @@
...
-        it "contains the budget XML" do
-          budget = create(:budget, activity: activity)
-          visit organisation_activity_path(organisation, activity, format: :xml)
-          expect(xml.at("iati-activity/budget/@type").text).to eq("1")
-          expect(xml.at("iati-activity/budget/@status").text).to eq("1")
-          expect(xml.at("iati-activity/budget/value").text).to eq(budget.value.to_s)
-          expect(xml.at("iati-activity/budget/period-start/@iso-date").text).to eq(budget.period_start_date.strftime("%Y-%m-%d"))
-          expect(xml.at("iati-activity/budget/period-end/@iso-date").text).to eq(budget.period_end_date.strftime("%Y-%m-%d"))
-        end
+        it_behaves_like "valid activity XML"
```
mec: I've just read about this - pretty clever! 

I wonder if the file that contains it could be called `behaves_like_valid_activity_xml_spec.rb', just so I can understand the link if I have not seen the pattern before - or is that not the convention for this?

app/presenters/budget_xml_presenter.rb
lozette: Good point, I have made this needlessly complex!! 🤦‍♀

spec/factories/activity.rb
lozette: Me either! Where did that come from??

lozette: Aha seems the XML isn't valid without it, but I will add it as a separate commit because it doesn't belong with this test refactor

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
lozette: Agree

lozette:
Ready for another look!

PR #189, opened 2020-02-13
(245) transactions can store arbitrary reciever and providing orgs

tahb:
## Changes in this PR

- Refactor how users add providing and receiving organisation details when creating transactions
- Ask the user for the organisational references for use in IATI and output the result of this value into IATI if present
- Organisation `references` for providers and receivers are optional. IATI requires either `name` or `ref` and our hypothesis is that of the 2, our users should be able to more easily provide a name of the company they are sending money to rather than their full IATI reference which is a challenging thing to work out. This was confirmed with users when we tested it on our google prototype.
- The `(optional)` string is included in the form headers in line with the [GDS design system guidance  that suggests instead of marking mandatory fields, mark the ones that are not important.](https://govuk-elements.herokuapp.com/form-elements/#form-optional-fields)
- I did think about setting some defaults for the `type` field but it felt slightly premature to jump to 'private sector' without testing with users first

### Before

Screenshot 2020-02-13 at 16 59 41

### After

Screenshot 2020-02-13 at 17 00 47

Screenshot 2020-02-13 at 17 30 42

## Next steps

### DIFF
--- a/app/views/staff/shared/transactions/_table.html.haml
+++ b/app/views/staff/shared/transactions/_table.html.haml
@@ -3,23 +3,23 @@
     %thead.govuk-table__head
       %tr.govuk-table__row
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.reference")
+          = t("activerecord.attributes.transaction.reference")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.description")
+          = t("activerecord.attributes.transaction.description")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.transaction_type")
+          = t("activerecord.attributes.transaction.transaction_type")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.date")
+          = t("activerecord.attributes.transaction.date")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.currency")
+          = t("activerecord.attributes.transaction.currency")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.value")
+          = t("activerecord.attributes.transaction.value")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.disbursement_channel")
+          = t("activerecord.attributes.transaction.disbursement_channel")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.provider_id")
+          = t("page_content.transactions.table.headers.providing_organisation")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.receiver_id")
+          = t("page_content.transactions.table.headers.receiving_organisation")
         %th.govuk-table__header
 
     %tbody.govuk-table__body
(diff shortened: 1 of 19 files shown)


app/views/staff/shared/transactions/_table.html.haml
```diff
@@ -17,9 +17,9 @@
         %th.govuk-table__header
           =t("activerecord.attributes.transaction.disbursement_channel")
         %th.govuk-table__header
-          =t("activerecord.attributes.transaction.provider_id")
+          = "Provider"
```
mec: This project uses locales a great deal of the time - how do you come to a decision about when to use the locale and when not to as here?

db/migrate/20200213152249_remove_association_between_transaction_and_organisation.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveAssociationBetweenTransactionAndOrganisation < ActiveRecord::Migration[6.0]
```
mec: I like how these are two seperate migrations! 👍

spec/support/form_helpers.rb
```diff
@@ -138,8 +138,8 @@ def fill_in_transaction_form(expectations: true,
     value: "1000.01",
     disbursement_channel: "Money is disbursed through central Ministry of Finance or Treasury",
     currency: "Pound Sterling",
-    provider_organisation: Organisation.first,
-    receiver_organisation: Organisation.first)
+    providing_organisation: OpenStruct.new(name: "Example provider", type: "Government"),
+    receiving_organisation: OpenStruct.new(name: "Example receiver", type: "Private Sector"))
```
mec: This did throw me for a second! When you use it later it looks just like the work you are refactoring! :)

mec:
The commit messge for 1c7f4b9ba603d9a4c6b06f32f3ffc0b14fa97f6d looks like it's been duplicated a few times, I was impressed at the length of the message and then I realised what had happened! 😄

mec:
You mentioned `reference` vs `iati_reference` – have we captured the need to get a handle on that somewhere, I agree we should aim to keep the api for all the organisations the same if we can.

app/views/staff/shared/transactions/_table.html.haml
tahb: Good point. I try to think of using locale by default whereever possible. When I came to this I couldn't figure out where a table header should be organised in the locale file and forgot to make a note to come back to it! I will have another crack 👍

spec/support/form_helpers.rb
tahb: We could split the object down into 3 fields for each provider and receiver. I felt like there are many parameters allowed and grouping related fields into objects wouldn't be the worst idea, which in turn made the rest of the test require minimal changes.

tahb:
We had a chat about the choice of `iati_reference` or `reference` to make things consistent going forwards. We decided to stick with `reference` as whilst "reference" could mean different types of reference we mean IATI as we do with every other field we collect and store.

Meyric also noted that a better name in the future might be `.identifier` that we give to IATI in the context of a reference to connect elements together.

We both agreed that finding the perfect name was less important than using a consistent term since any future rename would be made much easier.

tahb:
@mec I thought this feature would benefit from an example placeholders for IATI references so added a new commit on the end. Easy to remove and come back to later if this premature?

Screenshot 2020-02-14 at 10 56 24

mec:
@tahb re the placeholder: Love the intention, wonder if it's more GOVUK like to include it in the hint (see @deberny work on date formats), something like:

> For example, GB-COH-12345. To lookup codes of for more infomation see <link>

or similar?

tahb:
Good suggestion, that would keep us consistent which would be important. 

Now I think of it I do remember seeing guidance on "For example," over "Eg." but not about not using placeholders. Interestingly the govukformbuilder gem supports them!

New version:
Screenshot 2020-02-14 at 11 12 05

PR #191, opened 2020-02-14
Pull request links to IATI validator documentation

tahb:
## Changes in this PR

Before this change, when clicking the link from a GitHub pull request it broke.

The link path was:
https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/doc/xml-validation.md

Notice the `/pull/doc/`? I assume the relative syntax we've got works from the context of the app but doesn't work from the context of a pull request?

Here's an old pull request where you can test the bug https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/189

An absolute link is prone to breaking if we change our master branch or move the file. In which case a pointer to documentation that exists is still useful even if the link breaks.

## Next steps


tahb:
You should be able to click the XML Validation link in this PR where I've applied these changes.

PR #192, opened 2020-02-14
Activity dates have correct roles

tahb:
#Changes in this PR 

When validating an activity that only had planned dates there was an error that stated the activity start date could not be in the future.

I believe this was because we had rendered the planned_end_date with the role intended for the actual_start_date. This meant actual_start_date was being run incorrectly.

IATI activity-date roles defined here http://reference.iatistandard.org/202/activity-standard/iati-activities/iati-activity/activity-date/


tahb:
With this change (and a few others I'm also proposing in separate PRs) this goes green https://test-validator.iatistandard.org/view/dqf/files/38abc561-38bf-44dc-9a50-48f319b801f7?isTestfiles=true

PR #197, opened 2020-02-18
All users can do everything within the service

tahb:
## Changes in this PR

During recent work to allow users to create projects we ran into significant complexity in the test suite. We were using users role AND organisation they were assigned to AND the activities level to decide what users could and couldn't do.

Our feature tests were over detailed and hard to work with as they were setting up contexts with lots of these different variations. It was hard to conceptualise what the service did at a glance which made adding new work hard to do with confidence.

In order to make things better we made a plan of which this is the first step:

1. Allow everyone to do everything. Using the abstract role of administrator as a tool to remove the fund manager and delivery partner role. This would allow projects to be added in a simpler way with less complexity. Given we're in private beta and we will have more control over what select fund managers and delivery partners click on, we can
2. Remove the concept of roles and rely on the organisation the user belongs to
3. Add read/write roles later when we need them

This change is a large commit that was driven by the removal of the 2 roles and doing a find and replace for the words `fund_manager` and `delivery_partner` and replacing them with `administrator`. In doing so I consolidated test contexts that were referencing both into 1.

This change also renames feature tests where the user role is no longer relevant to comply with our new naming directive of users_*

There will be a follow up pull request for step 2.

## Next steps


app/policies/organisation_policy.rb
```diff
@@ -1,22 +1,22 @@
 class OrganisationPolicy < ApplicationPolicy
   def index?
-    user.administrator? || user.fund_manager?
+    user.administrator?
   end
 
   def show?
-    user.administrator? || user.fund_manager? || associated_user?
+    user.administrator? || associated_user?
```
lozette: Do we need to keep the `associated_user?` check, as theoretically all users are admins and therefore associated users now. I've tried removing it and the tests still pass.

app/policies/transaction_policy.rb
```diff
@@ -5,22 +5,22 @@ def index?
 
   def show?
     user.administrator? ||
-      activity? && user.fund_manager?
+      activity?
```
lozette: The `activity?` check can be removed here, it's not needed as everyone can do everything now (unless you want to keep it for if/when we re-introduce restrictions on what users can do?)

app/policies/organisation_policy.rb
tahb: You're right this doesn't make sense anymore. In the next pull request I switch the first part back to being a `beis_user?` so hopefully all is right again https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/199/files#diff-172620c2e4ee916f15c9384c3b1257abR7

app/policies/transaction_policy.rb
tahb: Yes it could be and yes I would love to keep it and let the next PR remove it. Actually it sets it to a simpler `true` in all cases and defers to the activity policy: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/199/files#diff-fb8862afbee289b19cc09ac511fc0da4R3

lozette: I'm just looking at the next PR so feel free to leave these in!

PR #198, opened 2020-02-18
(270) Users can report project level activities

mec:
## Changes in this PR
In #184 although @lozette amazing work and determination got us a working feature, the team were concerned about the debt we were building up around user roles and authorisation.

In order to deliver a this feature we decided to re-scope the work and remove the authorisation altogether. By doing so any user signed in to the service can be thought of as an 'Administator' for the moment and can perform create, update and edit actions on all activity levels.

We know we have work to do to authorise users and activities, but we can deliver this feature in a managable chunk of work that raises little debt and can be reviewed easily.

All the work here is taken from @lozette previous work.

This PR enables users to:

- create any number of projects as children of a programme
- view the details of the projects
- once viewing the project, edit the details
- add transactions and budgets to the projects
- view the transations and budgets of the projects

## Screenshots of UI changes
A programme level activity with no projects
Screenshot_2020-02-18 Activity A Programme — Report Official Development Assistance

The same programme with a project added
Screenshot_2020-02-18 Activity A Programme — Report Official Development Assistance(1)

The project level activity view showing add budgets and transations
Screenshot_2020-02-18 Activity Project 01 — Report Official Development Assistance

The project level activity with transation and budget added
Screenshot_2020-02-18 Activity Project 01 — Report Official Development Assistance(1)

## Next steps

- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [x] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.
- [ ] Do any environment variables need amending or adding?
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation](doc/xml-validation.md)

PR #199, opened 2020-02-18
Chore/314 show and hide features based on user organisation

tahb:
Based on https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/197.

## Changes in this PR

- Only users that belong to BEIS can manage other users and organisations
- Only users that belong to BEIS can create funds and programmes, and associated budgets, transactions and extending organisations (budgets was only allowed on programmes prior to this)
- Users that belong to a delivery partner cannot see or interact with funds
- Users that belong to a delivery partner can see a list of programmes on their landing page/organisation#show
- Users that belong to a delivery partner can view but cannot edit programmes
- Users that belong to a delivery partner can click back from a programme and land on their organisation page (and not see a 401 not authorised when being sent to the fund page)

This work uses new policies files called `FundPolicy` and `ProgrammePolicy` are used to help show and hide content on the organisation#show page. This continues giving Pundit the responsibility for managing authorisation.

### As a BEIS user
Screenshot 2020-02-18 at 11 49 17

### As a DP user
Screenshot 2020-02-18 at 11 48 52
Screenshot 2020-02-18 at 16 56 40
Screenshot 2020-02-18 at 16 59 31
Screenshot 2020-02-18 at 16 59 45

## Next steps


lozette:
Couple of standard errors:
`  spec/features/staff/users_can_create_a_project_spec.rb:17:1: Layout/TrailingWhitespace: Trailing whitespace detected.
  spec/policies/budget_policy_spec.rb:9:1: Layout/TrailingWhitespace: Trailing whitespace detected.`

spec/features/staff/users_can_create_a_budget_spec.rb
```diff
@@ -1,40 +1,44 @@
 RSpec.describe "Users can create a budget" do
-  before { authenticate!(user: user) }
+  context "when the user is n
```
lozette: I realise this change has been superseded by another but I am not 100% happy with this malformed code being in a commit 😳

app/policies/user_policy.rb
```diff
@@ -1,21 +1,27 @@
 class UserPolicy < ApplicationPolicy
   def index?
-    user.administrator?
+    beis_user?
```
lozette: 👍 Much clearer!

spec/support/activity_helpers.rb
```diff
@@ -0,0 +1,18 @@
+module ActivityHelpers
+  def page_displays_an_activity(activity_presenter:)
```
lozette: We've got a mixture of helpers and shared examples now, do you think this is OK? Or that we should go with one pattern or the other? I'm easy either way.

app/models/activity.rb
```diff
@@ -21,7 +21,7 @@ class Activity < ApplicationRecord
     fund: "fund",
     programme: "programme",
     project: "project",
-  }, _prefix: :is, _suffix: :level
+  }
```
lozette: Aww no @mec will be sad to see this go 😆

spec/features/staff/users_can_create_a_budget_spec.rb
tahb: Yes, gosh what have I managed to get myself into here!

spec/support/activity_helpers.rb
tahb: I think helpers are good we need to reuse the way we are asserting across different (but similar) contexts. I think shared examples are useful when you can defer the responsibility for the whole setup, execution and assertions.

In the past I've tried to send variables into shared_examples for the setup and exectute steps to make the examples reusable but never ended up feeling happy that they've made things better.

For example in this use of the helper the way the user arrives on that page changes based on the context setup by the feature https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/95bcd0fd1b794adb8111382ea70c2bc1470c84f0/spec/features/staff/users_can_view_programme_level_activity_spec.rb#L21 I'd like to use shared examples if we can apply them in a case like this?

app/models/activity.rb
tahb: I noticed the differences of approaches during a conflict nightmare last night! Given I had used the `fund?`, `programme?` in favour of using the enums which I thought was the plan, then I rebased ontop of `project?` and it all went wrong and I had to pick. I chose the one that following the enum conventions and removed the extra tests. 

I can change it back as a follow up refactor commit if this way provides more clarity?

spec/support/activity_helpers.rb
lozette: Yes that makes sense to me!

app/models/activity.rb
lozette: I think using `activity.project?` etc is clearer myself (and less code 😆 )

tahb:
Oh my the conflicts are back in force. I will deal with these after sprint ceremonies.

app/models/activity.rb
mec: 😄 I like the wordy version but happy to go without! @lozette you should definately call me out on things you have a preference for!

lozette: I thought the prefix + suffix thing was pretty cool tho!

spec/features/staff/users_can_create_a_budget_spec.rb
tahb: I think I've got rid of this. It all went wrong after conflict resolution with the parent pull request. I dropped the shared origin commit which included this issue and problem solved.

tahb:
Rebased on top of https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/198 and updated to authorise project endpoints too.

PR #203, opened 2020-02-20
(269) Activities collect one or more implementing organisations

lozette:
##  Changes in this PR

Trello: https://trello.com/c/VwXXIspE/269-activities-collect-one-or-more-implementing-organisations

@mec did 99% of this PR, I'm just finishing it for him!

Project-level activities can now have multiple implementing organisations. These are expressed in the IATI XML as `participating-org[@role='4']`

Implementing organisations attributes are free-text fields. Because implementers can be any organisation anywhere in the world (not necessarily registered on this platform) we rely on the delivery partners filling in these details themselves.

Validated here: https://test-validator.iatistandard.org/view/dqf/files/4e82520b-38b5-462f-86d8-f020af58c1e2?isTestfiles=true

## Next steps

### DIFF
--- a/app/models/implementing_organisation.rb
+++ b/app/models/implementing_organisation.rb
@@ -1,0 +1,5 @@
+class ImplementingOrganisation < ApplicationRecord
+  validates_presence_of :name, :organisation_type
+
+  belongs_to :activity
+end
--- a/app/views/staff/activities/show.html.haml
+++ b/app/views/staff/activities/show.html.haml
@@ -27,3 +27,6 @@
   - if @activity.project? && policy(:project).create?
     = render partial: "staff/shared/activities/transactions", locals: { activity: @activity, transaction_presenters: @transaction_presenters }
     = render partial: "staff/shared/activities/budgets", locals: { activity: @activity, budget_presenters: @budget_presenters }
+
+  - if @activity.project?
+    = render partial: "staff/shared/activities/implementing_organisations", locals: { activity: @activity, implementing_organisation_presenters: @implementing_organisation_presenters }
--- a/app/views/staff/implementing_organisations/new.html.haml
+++ b/app/views/staff/implementing_organisations/new.html.haml
@@ -1,0 +1,19 @@
+= content_for :page_title_prefix, t("page_title.activity.implementing_organisation.new", name: @activity.title)
+
+= link_to t("generic.link.back"), organisation_activity_path(@activity.organisation, @activity), class: "govuk-back-link"
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %h1.govuk-heading-xl
+        = t("page_title.activity.implementing_organisation.new")
+
+      = form_with model: @implementing_organisation, url: activity_implementing_organisations_path(@activity) do |f|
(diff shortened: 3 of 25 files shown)


app/views/staff/activities/show.html.haml
```diff
@@ -27,3 +27,14 @@
   - if @activity.project? && policy(:project).create?
     = render partial: "staff/shared/activities/transactions", locals: { activity: @activity, transaction_presenters: @transaction_presenters }
     = render partial: "staff/shared/activities/budgets", locals: { activity: @activity, budget_presenters: @budget_presenters }
+
+  - if @activity.project?
+    .govuk-grid-row
```
tahb: After the recent changes to this page I reckon we can partial this up in.

If only only users who belong to delivery partners should be able to create and edit these then a test should help drive this into the above block that checks `@activity.project? && policy(:project).create?` and within the partial a further check for `:edit?`

We might be able to also show the list of implementers to everyone so it'd be separate?

app/models/implementing_organisation.rb
```diff
@@ -0,0 +1,5 @@
+class ImplementingOrganisation < ApplicationRecord
```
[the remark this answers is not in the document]
mec: Updated!

spec/features/staff/users_can_manage_implementing_organisations_spec.rb
```diff
@@ -31,5 +31,30 @@
       expect(page).to have_content other_public_sector_organisation.name
       expect(page).to have_content other_public_sector_organisation.reference
     end
+
+    scenario "they can edit an implementing organisation" do
+      fund = create(:fund_activity, organisation: beis)
```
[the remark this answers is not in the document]
mec: Yeah absolutely! I may have been waiting for the factories to be sorted and then had to leave this work – I planned to sort this but it was lost in the hand over.

There was a reason for the BEIS thing, but I cannot remember it! Apologies – it's fixed now.

app/views/staff/implementing_organisations/new.html.haml
```diff
@@ -0,0 +1,19 @@
...
+= link_to t("generic.link.back"), organisation_activity_path(@activity.organisation, @activity), class: "govuk-back-link"
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %h1.govuk-heading-xl
+        = t("page_title.activity.implementing_organisation.new")
+
+      = form_with model: @implementing_organisation, url: activity_implementing_organisations_path(@activity) do |f|
+        = f.govuk_error_summary
+        = f.hidden_field :activity_id, value: @activity.id
```
tahb: I got confused for a second here as `participating-org` also has a field called `activity-id` that is used to allow us to point to a related activity down the tree! http://reference.iatistandard.org/202/activity-standard/iati-activities/iati-activity/participating-org/#example-usage

I don't think we can avoid this for the moment unless we rename `activity` to be more explicit in `parent_activity`. We aren't putting this value into the XML which was the main worry I had 👍

mec: Yeah - I regret not calling it 'parent_activity' - we should definately do that refactor, we have too much 'activity' already!

PR #204, opened 2020-02-20
Feature/145 budgets can be edited

lozette:
## Changes in this PR

Trello: https://trello.com/c/S16pkVHn/145-budgets-can-be-edited

Budgets can be edited. Much like transactions, budgets are displayed in a table with an `Edit` link. The link takes the user to a form where all attributes of the budget can be edited. 

We have also added a small formatter for the budget value. Previously, values were being shown to one decimal place if there were no fractions (e.g. 20 became 20.0); as these are monetary values it makes sense to display them to two decimal places.

## Next steps

### DIFF
--- a/app/controllers/staff/budgets_controller.rb
+++ b/app/controllers/staff/budgets_controller.rb
@@ -28,7 +28,37 @@ class Staff::BudgetsController < Staff::BaseController
     end
   end
 
+  def edit
+    @budget = Budget.find(id)
+    authorize @budget
+
+    @activity = Activity.find(activity_id)
+  end
+
+  def update
+    @budget = Budget.find(id)
+    @activity = Activity.find(activity_id)
+    @budget.activity = @activity
+    authorize @budget
+
+    @budget.assign_attributes(budget_params)
+    @budget.value = monetary_value
+    @budget.period_start_date = format_date(period_start_date)
+    @budget.period_end_date = format_date(period_end_date)
+
+    if @budget.save
+      flash[:notice] = I18n.t("form.budget.update.success")
+      redirect_to organisation_activity_path(@activity.organisation, @activity)
+    else
+      render :edit
+    end
+  end
+
   private
+
+  def id
+    params[:id]
+  end
 
   def activity_id
     params[:activity_id]
--- a/app/presenters/budget_presenter.rb
+++ b/app/presenters/budget_presenter.rb
@@ -18,4 +18,8 @@ class BudgetPresenter < SimpleDelegator
     return if super.blank?
     I18n.l(super)
   end
+
+  def value
+    ActionController::Base.helpers.number_to_currency(super, unit: "£")
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,4 +48,6 @@
 - Users who belong to delivery partners see a list of programmes on their organisation page
 - Users are authorised based on the organisations they belong to
 - Users can add budgets to activities at all levels
+- Users can edit a budget
+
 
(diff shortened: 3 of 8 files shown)


app/presenters/budget_presenter.rb
```diff
@@ -18,4 +18,8 @@ def period_end_date
     return if super.blank?
     I18n.l(super)
   end
+
+  def value
+    "%.2f" % super
```
mec: I am interesteted to know what led you to choose this over something like `number_to_currency`? 

I think I agree that this is nice and simple and money always scares me a little bit. When I first saw this I knew what the result was but I had no idea what was going on – do you think it is too much wrap it in a method for simple folk like me, `number_to_two_decimal_places` or something - what do you think?

lozette: I didn't know number_to_currency existed!

I can use it but that means taking it out of the presenter & using the method in the view, if you think that's OK?

app/controllers/staff/budgets_controller.rb
```diff
@@ -28,8 +28,38 @@ def create
...
+  end
+
+  def update
+    @budget = Budget.find(id)
+    @activity = Activity.find(activity_id)
+    @budget.activity = @activity
+    authorize @budget
+
+    @budget.assign_attributes(budget_params)
+    @budget.value = monetary_value
```
mec: This has no bearing on your work, but I am wondering if we should do something in `monetary_value` to format the display? I work on the principle that the value is stored in the lowest possible unit, so pence for us and converted for display – I've not looked at what we do but the `to_f` in `monetary_value` feels a bit brutal.

app/presenters/budget_presenter.rb
mec: I noticed later we have a `monetary_value` that uses `to_f` I wonder if that is actually the place where we make sure it is 2 decimal places – I don't have the confidence to say right now as I haven't looked at how we handle money as yet!

lozette: I'm a bit out of my depth as well. I know we store the value with monetary_value but the problem I'm trying to solve in the presenter is that if the value is `X` it will be presented as `X.0` but if it's `X.01` it will be presented as `X.01`.

Maybe @tahb can weigh in on the budget presenter? The issue I'm trying to solve is to stop values like `21.0` appearing on the front end.

tahb: Thinking aloud.. given we are working with money we only care about a value to 2 decimal places: 

"0.2" is 20p - this is the case we are thinking of here
"0.23" is 23p
"0.238" the decimal type will round this to 0.24p

We are using the `decimal` data type when storing this information in the database with a scale of 2, which means when we persist the data some rounding can happen (we could address more than 2 decimals by validating the number of decimals in the value users provide on the way in and preventing them).

If this were any other type of data I'd suggest we modify the setter to store values the rest of the system can use without having to start 0 padding everywhere however when we look at the `value` field in Ruby we see the class is `BigDecimal` and the value is `0.2e0` and this is the same whether we store a trailing 0 or not. There doesn't seem to be any value add by changing the setter so I think we should solve this at the presentational layer instead. `number_to_currency` looks the perfect tool for the job:

```
number_to_currency(BigDecimal("0.2"), unit: "£") => "£0.20"
number_to_currency(BigDecimal("0.20"), unit: "£") => "£0.20"
```

I notice that transaction values don't have their values presented and could benefit from this change too!

lozette:
This is approved but @mec or @tabh can you have another look over 1ecf4a861f905982271732f5287c217135727a64 thanks!

app/controllers/staff/budgets_controller.rb
tahb: The `to_f` is brutal and I think it's actually redundant as the database decimal type will do a more precise conversion for us since it has `scale: 2`. I think it was intended to deal with the fact that users could provide more than 2 decimal places but this should probably be validated rather than chopped!

lozette: Should we open a refactor card to address this?

tahb: I'll add a card. I suspect it will go to the icebox to die but it's there if this does become an issue as a prompt.

PR #207, opened 2020-02-24
Extract env vars from VCAP_SERVICES

rjw1:
## Changes in this PR

GPaaS provides a VCAP_SERVICES environment variable which contains JSON
with information about services bound to the application.

From the JSON we need to extract the REDIS_URL and any other environment
variables we have put into a user provided service.

This is based on the work we had done for CCS RMI
https://github.com/dxw/DataSubmissionServiceAPI/pull/376 and others

### DIFF
--- a/config/initializers/vcap_services.rb
+++ b/config/initializers/vcap_services.rb
@@ -1,0 +1,2 @@
+require "vcap_parser"
+VcapParser.load_service_environment_variables!
--- a/lib/vcap_parser.rb
+++ b/lib/vcap_parser.rb
@@ -1,0 +1,28 @@
+# Class to parse the VCAP_SERVICES environment variable.
+# Cloud Foundry provides an environment variable called VCAP_SERVICES which
+# contains JSON.
+# The JSON provides details of services bound to the application.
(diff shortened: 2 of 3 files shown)


config/initializers/vcap_services.rb
```diff
@@ -0,0 +1,2 @@
+require "vcap_parser"
+VcapParser.load_service_environment_variables!
```
tahb: It would be great but not essential to have a comment here that explains what a `vcap` is to those that are new to it, and when the reader needs to worry about it.

Something like: 

> VCAP services is a Cloud Foundry mechanism that helps provision application environment variables to allow the app to run on Government Platform as a Service (GPaaS).

rjw1:
Its easy to add tests when they have already been written elsewhere :)

config/initializers/vcap_services.rb
rjw1: sure.

rjw1:
@tahb updated to add some comments

tahb:
Oh no I pressed the wrong button.

tahb:
LGTM. Rebasing so that we can merge.

PR #208, opened 2020-02-24
(316) Transactions cannot be set to a future date

lozette:
## Changes in this PR

Trello: https://trello.com/c/6XaRLaol/316-transactions-cannot-be-set-to-a-future-date

When testing with the IATI validator we realised that Transactions with a date in the future are not valid according to the IATI standard. This PR adds validations to the Transaction model which means a transaction cannot be saved if its date is in the future.

## Screenshots of UI changes

Screenshot 2020-02-24 at 12 23 21

## Next steps

### DIFF
--- a/spec/validators/date_not_in_future_validator_spec.rb
+++ b/spec/validators/date_not_in_future_validator_spec.rb
@@ -1,0 +1,57 @@
+require "rails_helper"
+
+RSpec.describe DateNotInFutureValidator do
+  subject do
+    Class.new {
+      include ActiveModel::Validations
+      attr_accessor :date
+      validates :date, date_not_in_future: true
+
+      def self.name
+        "Transaction"
+      end
+    }.new
+  end
+
+  describe "#valid?" do
+    context "when date is nil" do
+      it "is valid" do
+        subject.date = nil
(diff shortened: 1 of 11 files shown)


spec/validators/date_not_in_future_validator_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    }.new
+  end
+
+  describe "when date is nil" do
+    it "is valid" do
+      subject.date = nil
+      expect(subject.valid?).to be true
+    end
+  end
+
```
tahb: I think having a test that checks what the behaviour is when the date is _today_ would be good to have. We'd expect it to be true, I think?

spec/validators/date_not_in_future_validator_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      attr_accessor :date
+      validates :date, date_not_in_future: true
+
+      def self.name
+        "Transaction"
+      end
+    }.new
+  end
+
+  describe "when date is nil" do
```
tahb: As these blocks are adding new context I think we could rename them to `context`. I suppose the method we are describing is `valid?`?

PR #211, opened 2020-02-24
Chore/refactor how we sanitise money values

tahb:
## Changes in this PR

Transaction and Budget building has been refactored into a set of services, each with a new set of unit tests. The motivation behind this was to consolidate the way we are building these objects given they have shared behaviour around the sanitising of monetary values.

Services provide advantages of removing logic and complexity from the controllers, keeping them small and focused. Services will give us a natural place to extend and reuses these processes outside of controllers later, bulk uploading is a good example of this.

The new services look rather similar at the moment but they do require a different dependency. Creating the pair require an activity, but updating the pair require the original record. Instead of forcing these into a single service to do both, I've kept each class with a single responsibility.

The way we were formatting dates with special controller methods has dissappeared. I think this occurred when we added this using TDD and used a `Hash` in our tests instead of `ActionController::Parameters`. The Parameters object does magic behind the scenes and does the translation of 'day', 'month' and 'year' and sets it to our field. A refactor here.

`Monetize.parse(value)` is no longer called with `.to_f`. Since we are using the database `decimal` type with the scaling of 2 this change becomes another refactor as it is still being rounded when more than 2 decimal places are provided.

I could do further and move the `Monetary.parse` into a sharable library but it seemed like a lot of plumbing for a small line of code. I'm on the fence with it so I think it can wait until later.

## Next steps

### DIFF
--- a/app/services/update_transaction.rb
+++ b/app/services/update_transaction.rb
@@ -1,0 +1,26 @@
+class UpdateTransaction
+  attr_accessor :transaction
+
+  def initialize(transaction:)
+    self.transaction = transaction
+  end
+
+  def call(attributes: {})
+    transaction.assign_attributes(attributes)
+    transaction.value = sanitize_monetary_string(value: attributes[:value])
+
+    result = if transaction.valid?
+      Result.new(transaction.save, transaction)
+    else
+      Result.new(false, transaction)
+    end
+
+    result
+  end
+
+  private
+
+  def sanitize_monetary_string(value:)
+    Monetize.parse(value)
+  end
+end
--- a/app/controllers/staff/budgets_controller.rb
+++ b/app/controllers/staff/budgets_controller.rb
@@ -34,16 +34,13 @@ class Staff::BudgetsController < Staff::BaseController
 
   def update
     @budget = Budget.find(id)
-    @activity = Activity.find(activity_id)
-    @budget.activity = @activity
     authorize @budget
 
-    @budget.assign_attributes(budget_params)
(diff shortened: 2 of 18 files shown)


app/services/update_transaction.rb
```diff
@@ -0,0 +1,27 @@
...
+  def initialize(transaction:)
+    self.transaction = transaction
+  end
+
+  def call(attributes: {})
+    transaction.assign_attributes(attributes)
+    transaction.value = sanitize_monetary_string(value: attributes[:value])
+
+    if transaction.valid?
+      transaction.save
```
mec: I often wonder about `save` like this. What happens if for some reason the save fails, do we just trust it works?

tahb: This is a fair concern.

In thinking about it more I wonder if we should change it around a bit to:

```
result = Result.new(transaction.save, transaction)
```

Since `transaction.save` _can_ return false theoretically. The other way might be to `transaction.save!` and terminate the execution chain and fail loudly. What do you think?

tahb: I've made sure if transaction.save returns false unexpectedly, we return a falsey Result.

mec: To be honest, we trust `save` all over the place - I was interested in what your take was. I think because `save` calls `valid?` I get hung up on the duplication, but I like the way the code feels with the explicit call to `vaid?` – it tells the story nicely.

I don't think we should worry here - this was a great piece of work - maybe chat about `save` over a meyricano™ sometime...

PR #216, opened 2020-02-27
(318) Activitity XML includes reporting org in identifier

tahb:
## Changes in this PR

The `iati-identifier` field that is output in the Activity XML has had 2 updates.

1. It now includes the reporting organisation reference to that it passes the validator. This change involved an important decision on how we deal with changes to organisation iati references against the requirement by IATI for these to always remain unchanged. Feedback would be great on this, more info in the commit!
2. It can include up 2 to more prefixes to the activity identifier part itself. This maps our internal hierarchy and how we ask for identifiers at the different levels back to what BEIS currently publish to IATI. This change is more controversial as it's not required to produce valid IATI but it is required to keep it consistent with BEIS publishing. I've left more info on the commit that added this.

A new `ActivityXmlPresenter` was added along the way to make it easier to add and test the presentational logic around identifiers only in the context of where we show it in XML. 

## Screenshots of UI changes

Screenshot 2020-02-27 at 13 26 00

Screenshot 2020-02-27 at 13 26 00

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,3 +55,4 @@
 - Users can view implementing organisations in the Activity XML
 - Planned start and end dates are mandatory
 - Actual start and end dates must not be in the future
+- Activity XML includes 'iati-identifier' which is includes the reporting organisation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,3 +56,4 @@
 - Planned start and end dates are mandatory
 - Actual start and end dates must not be in the future
 - Activity XML includes 'iati-identifier' which is includes the reporting organisation
+- Activity XML includes 'iati-identifier' with an identifier that consists of all present levels of hierarchy: fund and/or programme
--- a/app/controllers/staff/activities_controller.rb
+++ b/app/controllers/staff/activities_controller.rb
@@ -22,7 +22,9 @@ class Staff::ActivitiesController < Staff::BaseController
         @budget_presenters = @budgets.map { |budget| BudgetPresenter.new(budget) }
         @implementing_organisation_presenters = @activity.implementing_organisations.map { |implementing_organisation| ImplementingOrganisationPresenter.new(implementing_organisation) }
       end
-      format.xml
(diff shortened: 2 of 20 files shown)


tahb:
The date I used was invalid but otherwise XML is good 👍  https://test-validator.iatistandard.org/view/dqf/files/4efe69a4-1503-409a-a613-e5bca0d17ac3?isTestfiles=true

tahb:
I've run this by Odeta now she's back from leave and fortunantly the inclusion of fund and programme activitiy identifiers as we have makes sense and is what BEIS would like their iati-identifiers to include.

Perhaps I can share some of that context?

As I understand it the IATI requirement is for an activity to have some kind of identifier, and the identifier is for the reporting organisation to define. As the reporting organisation for government delivery partners, it is for BEIS to define their own structure. This is where the IATI rules and BEIS requirements meet, IATI say identify without setting strict rules and BEIS have defined that activities should include certain pieces of information. To date BEIS have been constructing these based on what we now think of as `fund`, `programme`, `project` and `sub-project` without strong consistency and structure and is something we're hoping to help with.

I don't believe these composite identifiers are used for anything else other than publishing to IATI. That said we can be relatively confident that the component parts/identifiers are used in these  are used through BEIS and delivery partners both within and outside of this new service. For example, fund identifiers like "GCRF" are used and understood by a variety of users. Users care about answers to questions such as: 

- how much money was spent on a fund with the "GCRF" identifier
- how many active programmes are under "GCRF" at the moment
- how much underspend is there on a fund with the "Newton" identifier

I don't think we know what delivery partners are using for their project identifiers. I imagine a combination of what's in their allocation letter or what they identify the work as internally. Hopefully user research will help guide us in one way or another. 

When we know more about the allocation process we may be able to spot another opportunity to automate the creation of the project identifiers to make them preset and fixed - I can see that this would save time for users and improve data integrity even more! For now I'm happy that we take small steps towards that by making the flexible spreadsheets redundant.

PR #220, opened 2020-02-28
(301) Budget dates are validated

mec:
## Changes in this PR

The IATI standard requires that budget start and end dates be validated
as follows:

- start date must not be after end date
- end date not more than one year after start date

http://reference.iatistandard.org/203/activity-standard/iati-activities/iati-activity/budget/

We spent some time attempting to make a generic validator for these two
validations but in the end decided that a validator specifically for
budget was the simplest and most elegant solution until we need to reuse
this validation criteria.

We decided that as the validation requirements are clearly defined by
IATI, by wrapping them up in a validator we keep all of the elements to
carry out the validation together in one place.

The interface and associated challenges that went with a generic solution felt quite messy to us and we were not happy with the result. we much preferred this simple and explicit implementation.

We check that the object being validated has the interface required (duck typing?) but we were not sure if/how we should test this. We also considered if this validation should live inside the budget model but decided it encapsulated enough descreet behaviour to be it's own class.

We had some interesting thoughts about how changes to our objects interfaces could allow this class to validate other types - as the dates represent a period that has a start and end, when other objects have the same requirement, simply using the same column names would result in a reusable class, the validator doesn't care that the object is a budget, simply that it has a start and end date.

## Screenshots of UI changes
Screenshot_2020-02-28 Create budget — Report Official Development Assistance(1)

Screenshot_2020-02-28 Create budget — Report Official Development Assistance

## Next steps

### DIFF
--- a/app/validators/budget_dates_validator.rb
+++ b/app/validators/budget_dates_validator.rb
@@ -1,0 +1,22 @@
+class BudgetDatesValidator < ActiveModel::Validator
+  def validate(record)
+    unless dates_not_more_than_365_days_apart?(record.period_start_date, record.period_end_date)
+      record.errors.add :period_end_date,
+        I18n.t("activerecord.errors.models.budget.attributes.period_end_date.within_365_days_of_start_date")
+    end
+
+    unless start_date_not_after_end_date?(record.period_start_date, record.period_end_date)
+      record.errors.add :period_start_date, I18n.t("activerecord.errors.models.budget.attributes.period_start_date.not_after_end_date")
+    end
+  end
+
+  private
+
+  def dates_not_more_than_365_days_apart?(start_date, end_date)
+    end_date - 365.days <= start_date
+  end
+
+  def start_date_not_after_end_date?(start_date, end_date)
+    start_date <= end_date
+  end
+end
--- a/spec/validators/budget_date_validator_spec.rb
+++ b/spec/validators/budget_date_validator_spec.rb
@@ -1,0 +1,75 @@
+require "rails_helper"
+
+RSpec.describe BudgetDatesValidator do
+  subject { build(:budget) }
+
+  context "when the start date is the same as the end date" do
+    it "is valid" do
+      subject.period_start_date = Date.today
+      subject.period_end_date = Date.today
+
+      expect(subject).to be_valid
+    end
+  end
+
+  context "when the start date is before the end date" do
+    it "is valid" do
+      subject.period_start_date = Date.yesterday
+      subject.period_end_date = Date.today
+
+      expect(subject).to be_valid
+    end
+  end
+
+  context "when the start date is after the end date" do
+    it "is not valid" do
+      subject.period_start_date = Date.tomorrow
+      subject.period_end_date = Date.today
+
+      expect(subject).not_to be_valid
+    end
+
+    it "adds an error message to the :period_start_date" do
+      subject.period_start_date = Date.tomorrow
+      subject.period_end_date = Date.today
+
+      subject.valid?
+      expect(subject.errors.messages[:period_start_date]).to include I18n.t("activerecord.errors.models.budget.attributes.period_start_date.not_after_end_date")
+    end
+  end
+
+  context "when the dates are within 365 days of each other" do
+    it "is valid" do
+      subject.period_start_date = Date.today
+      subject.period_end_date = subject.period_start_date + 6.months
+
+      expect(subject).to be_valid
+    end
+  end
+
+  context "when the dates are exactly 365 days apart from each other" do
+    it "is valid" do
+      subject.period_start_date = Date.today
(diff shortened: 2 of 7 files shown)


spec/validators/budget_date_validator_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      subject.valid?
+      expect(subject.errors.messages[:period_start_date]).to include I18n.t("activerecord.errors.models.budget.attributes.period_start_date.not_after_end_date")
+    end
+  end
+
+  context "when the dates are within one year of each other" do
+    it "is valid" do
+      subject.period_start_date = Date.today
+      subject.period_end_date = subject.period_start_date + 6.months
```
tahb: We've got really good test coverage for the cases when:

- dates are both the same
- dates are 2 day apart
- dates are 6 months apart

To complete the set could we have a final test that check what happens if the period is exactly 365 days apart? Then I wonder about how we deal with leap years and if that's a problem too!

app/validators/budget_dates_validator.rb
```diff
@@ -0,0 +1,25 @@
+class BudgetDatesValidator < ActiveModel::Validator
+  def validate(record)
+    return unless record.respond_to?(:period_start_date) && record.respond_to?(:period_end_date)
```
tahb: These guards are diligant 👍 

I do think that the class is explicitly named for the object they should be used for so I think this line isn't obsviously required however I'm not sure of the harm of it. I did notice that there are no tests for either of these guard clauses.

mec: @tahb: I found it super hard to write the tests for the two guards on this which makes me question the need for them.

For the case where the object being validated does not respond to `period_start_date` or `period_end_date`: I've created a simple class that has Activerecord validations and uses the `BudgetDatesValidator`. I can then call `be_valid` on it and expect it to still be valid, this works on my assumption that an object like this is valid by default.

For the case where `period_start_date` and `period_end_date` are nil: I still needed a custom object here as a real `Budget` would use additional validation to check for the prescence of the two attributes which means I cannot test the `BudgetDatesValidator` in isolation.

I am not 100% happy with this, I feel like I should add the validation to the Budget model to remove the need for any of the guards.

Would you mind looking it over and letting me know what your thoughts are?

tahb: I wonder what would happen if we did not try to rescue the misuse of this validator, what would the impact be of failing loudly?

mec: @tahb I talked it over with @MariluzHerrera and realised that only a developer could misue the 'wrong object' issue so we got rid of it.

The guard against nil values for `period_start_date` and `period_end_date` is tricky. We are confident that we do not need the guard as we have `prescence_of` validation on the Budget model, but the Shoulda matcher test fails as they do not seem to be able to work along with the `validates_with`. We talked to @lozette as well and decided to keep the guard for nil values even though the test for it seems really complicated and difficult to parse.

You view appreciated!

mec: Cripes! I've had a chat with @erbridge as welll and I am still wondering if this just should go in the model as it is so coupled to Budget - I think that is what I want to do to feel good about this…

tahb: > we decided to keep the guard for nil values even though the test for it seems really complicated and difficult to parse

This is an important point. Having something that is difficult to test like this is a good sign that there's a problem in the design or somewhere in the responsiblities. 

I wonder how much of this problem comes down to the validator trying to do too much. A great place to be is for the validator to do one thing really well. We might achieve this by giving the responsibility of when to call it along with what data over to the Budget itself. We trust the Budget already that we will always be given 2 Date objects for comparison, rather than 2 Strings. If we try and make the validator do more it starts to feel painful.

Does passing that responsibility and removing the guard make you feel any better about this code?

```
validates_with CustomValidator, start: start, end: end, if: { |budget | budget.start.present? && budget.end.present? }
```

To be clear I'm happy for you to move it into budgets or keep it separate if that suggestion works out!

mec: @tahb Were you using Jedi mind tricks on me yesterday?! I had exatcly this realisation on the train home. Using conditional validation on the `validates_with` makes this much nicer and feels like the responsibility is (almost) in the right place.

I've added this with a lambda and tested it on the Budget following the reciepe I think shoulda-matchers uses:

- set invalid state
- validate
- check for expected errors

This feels a little odd to me, but I can't think of a better way.

I need to stop thinking about this PR now - it works and is much better than it was, I think it can go in and we can come back as the need for similar validation arises and split it out there?

tahb: I can confirm I am not a Jedi 😆 Let's merge!

lozette:
This should pass CI if you rebase against develop 🤞

PR #221, opened 2020-02-28
(288) Activity date validations

tahb:
## Changes in this PR

- Activity planned start and end dates are mandatory inline with IATI rules
- Activity actual start and end dates are clearly signed as optional fields
- If Activity actual start and end dates are provided, they must not be in the future

## Screenshots of UI changes

Screenshot 2020-02-28 at 12 27 30

Screenshot 2020-02-28 at 12 25 08

## Next steps


tahb:
Passes on the date front (fails due to another known bug we have a card for in the backlog) https://test-validator.iatistandard.org/view/dqf/files/16823091-9c22-4bfa-b3be-a6b2e5c75749?isTestfiles=true

PR #224, opened 2020-02-28
Chore/Use a server-side session store

tahb:
## Changes in this PR

* Before this change the Rails default is to use an encrypted cookie on each users browser: https://api.rubyonrails.org/classes/ActionDispatch/Session/CookieStore.html
* This is a problem because :
    > Session cookies do not invalidate themselves and can be maliciously reused. It may be a good idea to have your application invalidate old session cookies using a stored timestamp. - https://guides.rubyonrails.org/security.html#session-storage
* In this change we replace the session store with Redis
* This change fixes a common CookieOverflow error that occurs when too much data is stored in the session, also referenced here: https://api.rubyonrails.org/classes/ActionDispatch/Session/CookieStore.html
* This change does not encrypt the session data (although it is Marshalled)
https://redis.io/topics/security and instead relies on the network layers to firewall redis with ACL etc so that it is not open to the public.

## Next steps

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,14 @@ jobs:
     script: |
       docker network create test
       docker run -d --name pg --network test -p 5432:5432 postgres:11.6
+      docker run -d --name redis --network test -p 6379:6379 redis:4.0.14
       docker run -d \
         --name test-container \
         --network test \
         -e RAILS_ENV=test \
         -e DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true \
         -e DATABASE_URL=postgres://postgres@pg:5432/roda_test \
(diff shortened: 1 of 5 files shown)


tahb:
This runs locally without Docker but I cannot get it to work on CI with Docker. It seems to be unable to set session data in feature specs using either a real Redis container or an instance of MockRedis. 

The feedback loop is so big given we aren't using Docker in development that it is taking too long.

I plugged in `active_record_session` gem but found the same problem which points to a problem with `rack_session_access`. I'm still adverse to using this gem without rake tasks to auto clean expired session data from the database.

Closing for now. We'll have to trust in the Rails default of encrypted cookies for the moment.

PR #231, opened 2020-03-02
(231) Seed the database with delivery partner organisations & BEIS, and common user accounts

lozette:
## Changes in this PR

Trello: https://trello.com/c/pcdbxmEM/320-seed-the-database-with-known-delivery-partners-and-beis-itself

- Reorganise the seeds into individual files for organisation/users/activities
- Run the seeds in order (organisations first, as users & activities depend on them)
- Create the organisations from a YAML file
- Add a second generic delivery partner user (login details in the 1password vault as usual)

One deficiency here is that if we wish to update an organisation's IATI identifier in
the `organisations.yml` file we will have to drop all the organisations and re-seed them.
But I'm not sure of a better attribute to `find_or_create_by` than iati_reference?

## Next steps

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,22 +8,8 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
-# Generic development user
 if Rails.env.development?
-  organisation_params = FactoryBot.build(
-    :organisation, name: "Department for Business, Energy & Industrial Strategy"
-  ).attributes
-  organisation = Organisation.find_or_create_by(organisation_params)
-
-  user = User.find_or_initialize_by(
-    name: "Generic development user",
-    email: "roda@dxw.com",
-    identifier: "auth0|5dc53e4b85758e0e95b062f0",
-    role: :administrator
-  )
-  user.organisation = organisation
-  user.save
-
-  activity_params = FactoryBot.build(:activity, title: "GCRF", organisation: organisation).attributes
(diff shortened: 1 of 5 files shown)


db/seeds.rb
```diff
@@ -10,20 +10,7 @@
 
 # Generic development user
```
tahb: This line looks like it can be chopped our on our way through if we have time!

db/seeds/organisations.rb
```diff
@@ -0,0 +1,13 @@
...
+organisations = YAML.safe_load(File.read(File.join(Rails.root, "db", "seeds", "organisations.yml")))
+organisations.each do |organisation|
+  organisation_params = {
+    name: organisation["name"],
+    organisation_type: organisation["type"],
+    language_code: "en",
+    default_currency: "gbp",
+    service_owner: organisation["service_owner"],
+  }
+  Organisation.find_or_create_by(iati_reference: organisation["reference"]).update(organisation_params)
```
tahb: I like the way this is implemented as we should be able to add more records through our seeds idempotently.

db/seeds/organisations.yml
```diff
@@ -0,0 +1,68 @@
...
+  service_owner: false
+- name: British Academy
+  reference: GB-COH-RC000053
+  type: 22
+  service_owner: false
+- name: Royal Academy of Engineering
+  reference: GB-COH-RC000519
+  type: 22
+  service_owner: false
+- name: UK Research & Innovation
```
tahb: After chatting with Odeta in the Slack thread I think the UKRI reference is still unclear. As we know UKRI will be onboarded last I think we can remove them and their councils until later 👍

PR #233, opened 2020-03-03
Add terraform to deploy and set up the app

rjw1:
## Changes in this PR

Add the terraform which creates
- spaces
- services
- routes

and deploys the app and its worker

## Next steps

- [x] Do any environment variables need amending or adding to Travis?


terraform/variables.tf
```diff
@@ -0,0 +1,55 @@
...
+
+variable "rollbar_access_token" {
+  type        = string
+  description = "Rollbar access token"
+}
+
+variable "docker_image" {
+  type        = string
+  description = "docker image to use"
+  default     = "develop"
```
rjw1: I currently wondering if we want a default here and if we do is `develop` the right one

tahb: If this value determines what gets deployed if a variable isn't provided, I think it should default to the safest stable version which should be `master`.

There would be less impact in staging being set back to master, than pushing master forward unexpectedly to the latest changes (which might not be stable).

terraform/postgres.tf
```diff
@@ -0,0 +1,22 @@
...
+
+# Create a postgres database named with the environment.
+# Enable the pgcrypto and plpgsql extensions.
+# Increase the timeouts since the default of 15 minutes is too
+# short for postgres db creation
+
+resource "cloudfoundry_service_instance" "beis-roda-postgres" {
+  name         = "beis-roda-${var.environment}-postgres"
+  space        = cloudfoundry_space.space.id
+  service_plan = data.cloudfoundry_service.postgres.service_plans["small-ha-11"]
```
tahb: I wonder if the plan size should now be 100GB (with `medium-ha-11`) instead of 20GB from the off as we are likely to forget about this and have to deal with it quite quickly in live when we start backfilling years worth of transactions.

terraform/app.tf
```diff
@@ -0,0 +1,30 @@
+# Create the web app.
+
+resource "cloudfoundry_app" "beis-roda-app" {
```
tahb: Do we need to define what to use as a health check for the web server?

terraform/worker.tf
```diff
@@ -0,0 +1,29 @@
...
+  command           = "bundle exec sidekiq"
+  health_check_type = "none"
+  memory            = 1024
+  disk_quota        = 3072
+  timeout           = 120
+  docker_image      = "thedxw/beis-report-official-development-assistance:${var.docker_image}"
+  service_binding { service_instance = cloudfoundry_service_instance.beis-roda-redis.id }
+  service_binding { service_instance = cloudfoundry_service_instance.beis-roda-postgres.id }
+  service_binding { service_instance = cloudfoundry_user_provided_service.papertrail.id }
+  environment = {
```
tahb: These environment variables are the same when provided to both the web server and the worker. I presume there's no easy way to define them once and reuse them? We'd be less likely to forget to update the Sidekiq copy when adding new or changing old variables. Only if it's easy, not essential.

terraform/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Setup
+- install terraform (use `tfenv`)
+- install the latest [cloundfoundry provider](https://github.com/cloudfoundry-community/terraform-provider-cf/wiki#installations)
+
+### Deployment
+- use the correct branch
+  - develop for staging
+  - master for production
+- export environment variables for AWS credentials
```
tahb: How would we find the AWS creds needed to access a bucket stored in the GPaaS AWS account?

terraform/README.md
```diff
@@ -0,0 +1,37 @@
...
+- export environment variables for PaaS credentials
+- build and push a new docker image if needed
+  `cf app` can tell you what image is currently being used
```
tahb: Sweet!

```
name:              beis-roda-prod
requested state:   started
routes:            beis-roda-prod.london.cloudapps.digital
last uploaded:     Wed 04 Mar 13:27:46 GMT 2020
stack:
docker image:      thedxw/beis-report-official-development-assistance:1460-722421d8ea305a6f378d2befbd8448daa679ee4d
```

terraform/README.md
```diff
@@ -0,0 +1,37 @@
...
+- export environment variables of the form TF_VAR_variablename to match the variables in [varaibles.tf](variables.tf)
```
tahb: Type on variables

```suggestion
- export environment variables of the form TF_VAR_variablename to match the variables in [variables.tf](variables.tf)
```

doc/architecture/decisions/0016-use-terraform-to-set-up-and-deploy-to-the-paas.md
```diff
@@ -0,0 +1,24 @@
+# 16. Use Terraform to set up and deploy to the PaaS
```
tahb: Agree :+1:

.travis.yml
```diff
@@ -5,6 +5,16 @@ before_install:
...
   include:
+  - stage: Terraform
+    name: validate and lint the Terraform
+    script: |
+      git clone https://github.com/tfutils/tfenv.git ~/.tfenv
+      export PATH="$HOME/.tfenv/bin:$PATH"
+      tfenv install
+      bash -c "$(curl -fsSL https://raw.github.com/cloudfoundry-community/terraform-provider-cf/master/bin/install.sh)"
+      terraform init terraform
+      terraform fmt  -diff -check -recursive terraform
```
tahb: A good idea to check this. I wonder about the ordering of this new step considering the number of future builds that will include Terraform changes compared to the much larger amount of Rails changes. As I think it runs jobs top to bottom I think we could move this until after the application tests have been run so that the more common build process gets faster feedback?

terraform/variables.tf
rjw1: Indeed. I'll change it to master.

.travis.yml
rjw1: Its a parallel job. So they run at the same time assuming there is enough capacity. It could be a lot a lot quicker but it has to wait for the docker image to be built.

terraform/README.md
rjw1: They are in 1password. I'll add that.

terraform/worker.tf
rjw1: when changing they are shared. since they use terraform variables which are set from the deploy process. Adding new ones is the problem. There is a way with locals I think.  I think if we add another env var we should look at the refactor then.

terraform/app.tf
rjw1: we can do it. it defaults to /

terraform/postgres.tf
rjw1: sure.

terraform/variables.tf
rjw1: Thinking about this I think it makes sense for it to not have a default.

terraform/postgres.tf
rjw1: I think this should be another PR which gets rolled carefully when we are ready for it. While Adam is currently adding data is probably not the best time.

.travis.yml
rjw1: I've shifted this since it was easy enough

terraform/app.tf
rjw1: I've changed this to /health_check

terraform/postgres.tf
tahb: Agree. This ship has sailed :)

PR #234, opened 2020-03-03
(287) BEIS users can mark other users (i.e. delivery partner users) as active or inactive

lozette:
## Changes in this PR

Trello: https://trello.com/c/L6bIzPW7/287-create-ability-to-make-a-user-eg-delivery-partner-inactive-in-the-application

BEIS users can mark other users as active or inactive. The GOVUK design system prefers yes/no type options to be presented as radio buttons as opposed to a single checkbox, for clarity (see https://govuk-form-builder.netlify.com/form-elements/checkboxes/#generating-a-single-checkbox )

An inactive user will still have an Auth0 account, they just cannot access the RODA service.

## Screenshots of UI changes

Screenshot 2020-03-03 at 14 57 39

Screenshot 2020-03-03 at 14 58 04

Screenshot 2020-03-03 at 14 58 24

## Next steps

### DIFF
--- a/app/controllers/staff/users_controller.rb
+++ b/app/controllers/staff/users_controller.rb
@@ -17,6 +17,7 @@ class Staff::UsersController < Staff::BaseController
 
   def create
     @user = User.new(user_params)
+    @user.active = params[:user][:active]
     authorize @user
     @organisations = policy_scope(Organisation)
 
@@ -46,6 +47,7 @@ class Staff::UsersController < Staff::BaseController
     @organisations = policy_scope(Organisation)
 
     @user.assign_attributes(user_params)
+    @user.active = params[:user][:active]
 
     if @user.valid?
       result = UpdateUser.new(user: @user, organisation: organisation).call
--- a/spec/features/staff/users_can_sign_in_spec.rb
+++ b/spec/features/staff/users_can_sign_in_spec.rb
@@ -100,4 +100,23 @@ RSpec.feature "Users can sign in with Auth0" do
       expect(page).to have_content(I18n.t("page_content.errors.auth0.failed.prompt"))
     end
   end
+
+  context "when the user has been deactivated" do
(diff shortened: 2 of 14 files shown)


lozette:
2 questions:

1. We're not stopping user's from marking themselves as inactive, which means we could end up with a sole administrator locking themselves out. Do we need to mitigate this?
2. Thoughts on the content design - is active/inactive OK?

spec/features/staff/users_can_sign_in_spec.rb
```diff
@@ -100,4 +100,23 @@
       expect(page).to have_content(I18n.t("page_content.errors.auth0.failed.prompt"))
     end
   end
+
+  context "when the user has been deactivated" do
+    scenario "the user cannot log in and sees an informative message" do
```
tahb: Thorough test coverage 👍

app/controllers/staff/users_controller.rb
```diff
@@ -17,6 +17,7 @@ def new
 
   def create
     @user = User.new(user_params)
+    @user.active = params[:user][:active]
```
tahb: Wondering if we could make `user_params` deal with this as we do with the other params and we can continue to rely on strong parameters?

lozette: Unfortunately brakeman objects to `active` as a param name for mass assignment (https://dxw.slack.com/archives/CN0H2L066/p1583249732015700) - I can change the name or tell brakeman to ignore the warning?

lozette:
@tahb As we have a content designer next week we can maybe wait til then and see if they have suggestions (I'm sure we'll be changing a lot of the text as it is!)

lozette:
Screenshots of text changes from ec6bcd6

Screenshot 2020-03-04 at 12 02 54

Screenshot 2020-03-04 at 12 04 34

PR #242, opened 2020-03-04
Document deployment steps for production and staging

tahb:
Inspired by https://crown-commercial-service.github.io/ReportMI-service-manual/#/deployments?id=production which adds steps to the pull request model of creating releases and updating the change log.

Change logs are a a decision we've already made as dxw to track project progress https://github.com/dxw/tech-team-rfcs/blob/master/rfc-019-use-changelogs-to-track-changes.md.

This is the process used in release-1 as it was better to have some structure around the changelog. I'd like to use this pull request to check what we want to do and if there are any learnings and improvements we can make.

## Next steps

### DIFF
--- a/doc/deployment-process.md
+++ b/doc/deployment-process.md
@@ -1,0 +1,57 @@
+# Deployment process
+
+## Production
+
+As outlined in the [dxw development workflow guide](http://playbook.dxw.com/#/guides/development-workflow?id=deploying), production deploys are done by manually merging develop into master. To give us a slightly more formal process around what gets deployed and when and also to give us visibility into the things that have been deployed, we additionally follow these steps when releasing to production:
+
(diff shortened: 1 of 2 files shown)


doc/deployment-process.md
```diff
@@ -0,0 +1,57 @@
...
+### Steps
+
+1. Create a release branch and make a pull request
+  * Create a branch from develop for the release called release-X where X is the release number
+  * Update CHANGELOG.md to:
+    document the changes in this release in a bullet point form
+    add a link to the diff at the bottom of the file
+  * Document the changes in the commit message as well
+  * git push <your-branch-name>, e.g. git push release-45
+  * Create a tag for the release in the format release-X
```
mec: When does the tag get pushed?

tahb: Good spot.

I will swap them round and add `--follow-tags` to the push command!

PR #248, opened 2020-03-06
(286) Show actual Auth0 error when user creation fails

lozette:
## Changes in this PR

Trello: https://trello.com/c/falHEEs2/286-address-bug-when-email-address-has-already-been-used-to-create-a-user-in-auth0

Previously, we were showing a generic error message when user creation failed in
Auth0. Now, we will attempt to parse the error message received from Auth0 and
show it to the user. If the error message is unparseable, a generic error will
be shown.

## Screenshots of UI changes

Screenshot 2020-03-06 at 11 43 06

## Next steps

### DIFF
--- a/app/controllers/staff/users_controller.rb
+++ b/app/controllers/staff/users_controller.rb
@@ -27,7 +27,7 @@ class Staff::UsersController < Staff::BaseController
         flash.now[:notice] = I18n.t("form.user.create.success")
         redirect_to user_path(@user.reload.id)
       else
-        flash.now[:error] = I18n.t("form.user.create.failed")
+        flash.now[:error] = I18n.t("form.user.create.failed", error: result.error_message)
         render :new
       end
     else
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -1,4 +1,4 @@
-Result = Struct.new(:success, :object) {
+Result = Struct.new(:success, :object, :error_message) {
   def success?
     success == true
   end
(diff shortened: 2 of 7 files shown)


app/controllers/staff/users_controller.rb
```diff
@@ -27,7 +27,8 @@ def create
         flash.now[:notice] = I18n.t("form.user.create.success")
         redirect_to user_path(@user.reload.id)
       else
-        flash.now[:error] = I18n.t("form.user.create.failed")
+        message = extract_error_message(result)
```
tahb: I like the way we are leveraging the result object to help us bubble up the error message. 

The naming of this method gave me pause. It appears generic however I think it's quite specific to only being able to extract Auth0 failures. It doesn't for instance extract errors from the sending of welcome emails or a message for database rollback errors, which are reasons the result could be a failure.

I think we could push the logic of this error passing out of the controller and back down into the service with a little more refactoring.

Adding a new `:error_message` attribute to our `Result` class seems reasonable. When results fail there will generally be error messages to go with them! 

If we added this `Result = Struct.new(:success, :object, :error_message)` then we could push the extraction into the service object with something like:

```
# controller 
        flash.now[:error] = I18n.t("form.user.create.failed", error: result.error_message)

# service
      rescue Auth0::Exception => e
        result.success = false
        result.error_message = extract_auth0_error_message(e)
```

We could then add a unit test that has us worried about a `JSON::ParserError` occuring so we make sure we catch any regressions.

lozette: Amended! Thank you for the advice 🙌

PR #249, opened 2020-03-06
(303) Activities collect recipient country or recipient region

mec:
## Changes in this PR

Lots going on here, see trello card for more background:

https://trello.com/c/lWsmuCqr

This work first allows us to capture both a `recipient_region` and `recipient_country`. It goes on to add the concept of `geography` to let the user decide whether to provide a region or a country.

Geography was choosen instead of the planned 'providing_recipient_country' as it is a concept used in the IATI standard:

http://reference.iatistandard.org/203/activity-standard/overview/geography/

this means the question is not as per the design. However, for this stage of work the data structure follows the pattern we already have.  It also makes the implementation simpler than the true or false value that I planned to use, we could also add further geographies simply as IATI defines 'location'

Getting the validation and the branching of the stepped form was difficult. I settled on two decisions I would like to share:

- to change geography a user has to edit the geography step, go back to the activity page and then add the new attibute. This invloves more steps than is ideal, but I've choosen to prioritise delivering the end user goal over the best user experience. Managing the steps in the Wizard form was beyond me in this work and seem to require more work than I was willing to give this. I decided we should prioritise that work as a team against our other goals.

- validation on this is odd. I feel this work raised the issue of when is an Activity actually valid? I decided to only validate `recipient_region` and `recipient_country` when the form step is in use, preventing a user submitting empty values. The additional validation that is present when `wizard_status` is 'complete' was removed to allow for an Activity having one kind of geography over the other. I feel like validation contexts would make this simpler, but again felt that refactor was beyond the scope of what I am trying to do with this work.

The final piece here is to add the `recipient_region` or `recipient_country` to the xml and check it validates. Thankfully that went smoothly! 😄 

Activity with recipient_region geography:

https://stage.dataworkbench.io/validate/dcb321e1-eedf-4acd-aebd-79b60cbdaac9

I've added the explicit `vocabulary` attribute to this as well.

Activity with recipient_country geography:

https://stage.dataworkbench.io/validate/c2b98263-51e5-4b8e-a7b0-c563c71e7373

## Where is the rest?
The actual story for this work was to allow the user to search for a country rather than select from a drop down. I split the story into stages as I was concerned about the size of the PR.  This first PR should lay the ground for the actual coutnry search.

## Screenshots of UI changes

The new 'geography' step
Screenshot_2020-03-06 Will the benefitting recipient be a region or country — Report Official Development Assistance

The region and country steps
Screenshot_2020-03-06 Recipient country — Report Official Development Assistance
Screenshot_2020-03-06 Recipient region — Report Official Development Assistance

A completed activivity with a recipient_region
Screenshot_2020-03-06 Activity Programme 02 — Report Official Development Assistance(2)

A completed activity with a recipient_country
Screenshot_2020-03-06 Activity Programme 02 — Report Official Development Assistance

An incomplete activty showing add recipient_region
Screenshot_2020-03-06 Activity Programme 02 — Report Official Development Assistance(1)

An incomplete activity showing add recipient_country
Screenshot_2020-03-06 Activity Programme 02 — Report Official Development Assistance(3)

## Next steps


erbridge:
That hint text for region probably belongs on the radio, too. That's the point where I'm making the decision.

config/locales/codelists/2_03/iati.en.yml
```diff
@@ -91,6 +91,258 @@ en:
...
+      VU: Vanuatu
+      WF: Wallis and Futuna
+      WS: Samoa
+      XK: Kosovo
+      YE: Yemen
+      YT: Mayotte
+      ZA: South Africa
+      ZM: Zambia
+      ZW: Zimbabwe
+      'false': Norway
```
lozette: Eeek! YAML recognises "NO" as "false", so to have "NO" in this list you need to encapsulate it in quotes, i.e. `'NO'`

app/helpers/codelist_helper.rb
```diff
@@ -23,6 +23,11 @@ def region_select_options
     objects.unshift(OpenStruct.new(name: "Developing countries, unspecified", code: "998")).uniq
   end
 
+  def country_select_options
+    objects = yaml_to_objects(entity: "activity", type: "recipient_country", with_empty_item: false)
+    objects.unshift(OpenStruct.new(name: "Select a recipient country", code: "")).uniq
```
lozette: Can we have this string internationalised, please 🙏

config/locales/codelists/2_03/iati.en.yml
mec: Waaaaa?! Thank you! Lesson learnt!

app/helpers/codelist_helper.rb
mec: Yep, wasn't sure where it should go but settled on this.

app/models/activity.rb
```diff
@@ -49,8 +56,16 @@ class Activity < ApplicationRecord
...
+  def geography_step?
+    wizard_status == "geography" || wizard_complete?
+  end
+
+  def region_step?
+    wizard_status == "region"
+  end
+
+  def country_step?
+    wizard_status == "country"
```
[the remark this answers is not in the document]
mec: Yeah, I've tried to cover this in the PR message, but obviously failed! :)

`wizard_complete?` is true when the user has completed all of the steps in the wizard.

If the user supplies a region and not a country, at the point the wizard status is complete the activity is no longer valid as it has no recipient_country.

I decided to loosen the validation to run only on the step collecting the value for the two branched attributes, so whichever attribute they chose they must supply a value.

Does that make sense? I am struggling to explain this in writing!?

mec:
@erbridge 

> That hint text for region probably belongs on the radio, too. That's the point where I'm making the decision.

For this stage of this work I have purposely avoiding hint text (and copy generally)as the journey is set to change, happy to put something in though, if we think it is a good time to.

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -9,6 +9,8 @@ class Staff::ActivityFormsController < Staff::BaseController
     :sector,
     :status,
     :dates,
+    :geography,
```
erbridge: I think this name quite confusing. Perhaps calling it something like `geography_level` would make it clearer? Or even better would be to connect it to what it determines. I'd have (and did have) a hard time looking at this and following what `geography` represents.

app/helpers/codelist_helper.rb
```diff
@@ -34,6 +34,11 @@ def region_select_options
     objects.unshift(OpenStruct.new(name: "Developing countries, unspecified", code: "998")).uniq
   end
 
+  def country_select_options
+    objects = yaml_to_objects(entity: "activity", type: "recipient_country", with_empty_item: false)
+    objects.unshift(OpenStruct.new(name: t("page_content.activity.recipient_country.default_selection_value"), code: "")).uniq
```
erbridge: What was the motivation for using i18n for this value, when the values near it are inline?

app/models/activity.rb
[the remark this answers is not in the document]
erbridge: Ok. I think that makes sense. I think what confuses me is the name of the methods in here. Like why is `wizard_complete?` being checked in tests to see if we're on a step? Would it be clearer to refactor this to run validations on multiple conditions: `*_step?` and `wizard_complete?`?

app/views/staff/activity_forms/country.html.haml
```diff
@@ -1,2 +1,2 @@
 = render layout: "wrapper" do |f|
-  = f.govuk_collection_select :recipient_region, region_select_options, :code, :name, options: { selected: f.object.recipient_region || region_select_options.first.code }, label: { tag: 'h1', size: 'xl' }
+  = f.govuk_collection_select :recipient_country, country_select_options, :code, :name, options: { selected: f.object.recipient_country || country_select_options.first.code}, label: { size: "xl", tag: "h1" }
```
erbridge: I _think_ `code` is an empty string by default. Is that intended?

app/helpers/codelist_helper.rb
```diff
@@ -34,6 +34,11 @@ def region_select_options
     objects.unshift(OpenStruct.new(name: "Developing countries, unspecified", code: "998")).uniq
   end
 
+  def country_select_options
+    objects = yaml_to_objects(entity: "activity", type: "recipient_country", with_empty_item: false)
```
erbridge: Should this be sorted?

config/locales/codelists/2_03/iati.en.yml
```diff
@@ -91,6 +91,261 @@ en:
...
+      MY: Malaysia
+      MZ: Mozambique
+      NA: Namibia
+      NC: New Caledonia
+      NE: Niger (the)
+      NF: Norfolk Island
+      NG: Nigeria
+      NI: Nicaragua
+      NL: Netherlands (the)
+      'NO': Norway
```
erbridge: Love YAML :)

config/locales/en.yml
```diff
@@ -168,6 +168,8 @@ en:
         label: Finance
       flow:
         label: Flow
+      geography:
+        label: Benefitting recipient geography
```
erbridge: Is this label clear? Has it been tested? I'm not sure I'd know what this meant.

spec/factories/activity.rb
```diff
@@ -100,15 +104,28 @@
...
+    geography { nil }
     recipient_region { nil }
+    recipient_country { nil }
     flow { nil }
     finance { nil }
     aid_type { nil }
     tied_status { nil }
   end
 
-  trait :at_country_step do
```
erbridge: Do we need an `at_country_step`, too? Does this imply that we're not testing that case?

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
```diff
@@ -130,11 +130,18 @@
...
+
+        choose "Region"
+        click_button I18n.t("form.activity.submit")
         expect(page).to have_content I18n.t("page_title.activity_form.show.region")
 
-        # Region has a default and can't be set to blank so we skip
-        select "Developing countries, unspecified", from: "activity[recipient_region]"
+        # region has the default value already selected
         click_button I18n.t("form.activity.submit")
+
```
erbridge: Can we go back to geography and test country too?

spec/features/staff/users_can_edit_an_activity_spec.rb
```diff
@@ -222,10 +222,18 @@ def assert_all_edit_links_go_to_the_correct_form_step(activity:)
...
+      activity_step_path(activity, :geography)
+    )
+  end
+  click_on(I18n.t("generic.link.back"))
+
   within(".recipient_region") do
     click_on(I18n.t("generic.link.edit"))
     expect(page).to have_current_path(
-      activity_step_path(activity, :country)
+      activity_step_path(activity, :region)
```
erbridge: Shouldn't we have something like this for `country`, too?

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -7,6 +7,41 @@
...
+
+        it "contains the recipient region code and fixed vocabulary code of 1" do
+          visit organisation_activity_path(organisation, activity, format: :xml)
+
+          expect(xml.at("iati-activity/recipient-region/@code").text).to eq(activity.recipient_region)
+          expect(xml.at("iati-activity/recipient-region/@vocabulary").text).to eq("1")
+        end
+      end
+
+      context "whent he activity has recipient_country geography" do
```
erbridge:
```suggestion
      context "when the activity has recipient_country geography" do
```

spec/helpers/activity_helper_spec.rb
```diff
@@ -59,19 +60,20 @@
       end
     end
 
-    context "when the activity has passed the country step" do
+    context "when the activity has passed the region step" do
```
erbridge: Do we need the equivalent for `country`, too?

spec/helpers/codelist_helper_spec.rb
```diff
@@ -122,6 +122,13 @@
       end
     end
 
+    describe "#country_select_options" do
+      it "returns an array of country objects with '' as the first (default) option" do
```
erbridge: This doesn't test that there's an array with anything other than the default option in it.

spec/support/form_helpers.rb
```diff
@@ -17,7 +17,9 @@ def fill_in_activity_form(
     actual_end_date_day: "1",
     actual_end_date_month: "2",
     actual_end_date_year: "2020",
+    geography: "recipient_region",
     recipient_region: "Developing countries, unspecified",
+    recipient_country: "Chile",
```
erbridge: Why are we setting this when `geography` is set to be region, not country?

vendor/data/codelists/IATI/2_03/activity/recipient_country.yml
```diff
@@ -0,0 +1,765 @@
+---
```
erbridge: Is there some process that took this to create the values in i18n?

erbridge:
> For this stage of this work I have purposely avoiding hint text (and copy generally)as the journey is set to change, happy to put something in though, if we think it is a good time to.

I think there is value in putting it in, even if it's a quick pass. There's a _plan_ to change this, but maybe it won't?

app/helpers/codelist_helper.rb
lozette: The motivation was probably me asking for it to be internationalised 😆 

I hadn't realised the others weren't internationalised. Internationalise all the things!!

app/controllers/staff/activity_forms_controller.rb
mec: I really struggled with this. Is it the step name or the attribute name or both that were confusing? 

I settled on `geography` as that is the term IATI used to represent these concepts:

http://reference.iatistandard.org/203/activity-standard/overview/geography/

I went for what I felt was consistent with the rest of the model.

app/helpers/codelist_helper.rb
mec: @lozette  review: 

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/249#discussion_r389724181

app/views/staff/activity_forms/country.html.haml
mec: Yes, country is a select box so I put in a empty value as the default that fails the validation.

app/helpers/codelist_helper.rb
mec: Yes, I think so:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/ce530674dc8d497b0d1876852d8a3f7e711349f7/app/helpers/codelist_helper.rb#L11

config/locales/codelists/2_03/iati.en.yml
mec: I know! I had no idea!

config/locales/en.yml
mec: I think it is as clear as I could make it right now. It has not been tested no. We are only just getting real users to look at the app this week.

I feel that this says what the value represents right now - early feedback is looking like users do not undersatnd a large amount of the IATI language and terms so that may be a wider issue, would we then change every attribute or change just the content in the user facing part of the app – I think this is one of the many hard question we face.

spec/factories/activity.rb
mec: I don't think it implies that. This change is part of the first refactor I did. We originally stored `recpient_region` but asked for country – I wanted it to all be region to make way for also being able to ask for country.

I don't believe we explicitly test for how the steps behaviour for every step, it just happend that we wrote the test for an 'incomplete' wizard from using the country step, see:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/ce530674dc8d497b0d1876852d8a3f7e711349f7/spec/features/staff/users_can_edit_an_activity_spec.rb#L56

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
mec: I feel like I covered that here:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/feature/303-activities-collect-recipient-country/spec/features/staff/users_can_manage_activity_geography_spec.rb

spec/features/staff/users_can_edit_an_activity_spec.rb
mec: Same as previous, I feel like that is here:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/feature/303-activities-collect-recipient-country/spec/features/staff/users_can_manage_activity_geography_spec.rb

We have the most feature specs I have ever worked with and I've found it hard to navigate them, I feel we could end up with giant features that try to do too much and are (even more) difficult to follow.

spec/helpers/activity_helper_spec.rb
mec: I don't think we want to test every step here, just test that the step logic works - is this another place where country was refactored to region?

spec/helpers/codelist_helper_spec.rb
mec: Interesting - Looking at the spec my feeling was that `yaml_to_objects` is tested in that way and that this spec just looks for the default value as this is added later - perhaps the wording is way off?

app/models/activity.rb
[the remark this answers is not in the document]
mec: I am not sure it would be clearer, but I do agree the whole step/which step/complete/valid setup needs more work - I was not convinced this was the place to do it though?

vendor/data/codelists/IATI/2_03/activity/recipient_country.yml
mec: Me downloading the csv and converting it to yaml?

mec:
>I think there is value in putting it in, even if it's a quick pass. There's a plan to change this, but maybe it won't?

So hint text for the 'geography' question and the country question?

erbridge:
I think just put the region hint text into the region radio button.

vendor/data/codelists/IATI/2_03/activity/recipient_country.yml
erbridge: I mean to get from here to the locale file.

spec/helpers/codelist_helper_spec.rb
erbridge: Yeah, perhaps it's the wording. I haven't got a complete picture of what's covered in the test suite.

app/controllers/staff/activity_forms_controller.rb
erbridge: I think it's both that are confusing. I get that IATI have a certain value, but I would hope to be able to read the code without having to refer to IATI docs to understand the meaning of fields.

Perhaps: `geographic_coverage` is a good name? That's wording used in the IATI docs. Or `geographic_scale`?

app/helpers/codelist_helper.rb
erbridge: Ah ok. I think it feels a little weird because some default values in here are real values, while others are placeholders. It's hard to tell from the implementation which is which. I'd love for that to be clarified better. Not a thing for now, though, I suspect.

app/models/activity.rb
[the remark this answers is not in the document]
erbridge: I think you might be right that this isn't the right place to do it, but wouldn't `if: [:*_step?, :wizard_complete?]` (I think that's valid?) be clearer?

app/views/staff/activity_forms/country.html.haml
erbridge: Would it be clearer to set it to `nil`, since that's what you're really representing?

app/helpers/codelist_helper.rb
erbridge: Ah. Great.

config/locales/en.yml
erbridge: Understood. Happy to leave it for now and see what users think.

erbridge:
So I think you're right that country doesn't need to be tested every time region is, but I don't see much feature test coverage of the bahviour of country. In some senses, it's more interesting and liable to failure than region, due to not having a valid default value, so I would have liked to see more testing of it. You have a better idea of the test coverage though, so if you feel confident about this, don't let it block you.

vendor/data/codelists/IATI/2_03/activity/recipient_country.yml
tahb: Sounds like Meyric did this by hand. FYI we made a script for this that we can use locally https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/lib/tasks/yaml_to_locale.rake

spec/helpers/codelist_helper_spec.rb
tahb: It does look like bulk of the testing is defferred to `yaml_to_objects`. Ideally this and nearby methods that are similar would test the full range of their responses. The fact that another `yaml_to_objects` method is used in the implementation shouldn't change that. I think what's happened is that `yaml_to_objects` was added first and we built new methods around it.

`yaml_to_objects` is still used directly as a public method.

If we wanted to make things clearer here I'd suggest we move the testing for `yaml_to_objects` into a shared example and reused it in each test:

```
describe "#yaml_to_objects" do
  it_behaves_like "returns an array of options as objects"
describe "#country_select_options" do
  it_behaves_like "returns an array of options as objects"
  it "sets a blank default as the first option"
describe "#flow_select_options" do
  it_behaves_like "returns an array of options as objects"
  it "sets ODA as the default option"
```

Given the difficulties in this feature and in getting this pull request merged I think it's reasonable to accept that whist it does touch this code it doesn't introduce a problem. As we have test coverage I'm happy that we leave this refactoring opportunity to the next code that touches it.

app/views/staff/activity_forms/country.html.haml
tahb: I agree it would be slightly more intuitive if this was set to nil since we are unlikely to ever set a default country.

tahb: That said, I don't think it should block this work. We'll have 2 more iterations planned to this country form page.

PR #250, opened 2020-03-06
(297) Choosing a Sector for an Activity presents the user with duplicate options

lozette:
## Changes in this PR

Trello: https://trello.com/c/ozGnGd0z/297-choosing-a-sector-for-an-activity-presents-the-user-with-duplicate-options

When creating an activity, we noticed that some of the Sector options were duplicated.

It soon became apparent that some of the codelist items from IATI are marked as "status: withdrawn". IATI marks items as withdrawn to indicate they are deprecated: https://support.iatistandard.org/hc/en-us/articles/214390846-Add-a-withdrawn-flag-to-code-names-to-indicate-deprecation

First, we amended the CodelistHelper to only put codelist items into the dropdowns if they are not marked as withdrawn. 

However, we still had two duplicates - two entries for "Rural development" which were marked as "status: active". Following discussion with the service owner, we decided to amend one of the names on the codelist until IATI can advice us further. This means we no longer have any duplicates in the dropdown list for sector. 

## Next steps

### DIFF
--- a/spec/helpers/codelist_helper_spec.rb
+++ b/spec/helpers/codelist_helper_spec.rb
@@ -66,6 +66,17 @@ RSpec.describe CodelistHelper, type: :helper do
           expect(list.first.name).to eq("Active code")
         end
       end
+
+      it "does not add any duplicate-named items to the list" do
+        list = helper.yaml_to_objects(
+          entity: "activity",
+          type: "sector",
+          with_empty_item: false
+        )
+        grouped_by_name = list.group_by { |item| item["name"] }
+        duplicate_groups = grouped_by_name.values.select { |a| a.size > 1 }.flatten
+        expect(duplicate_groups.count).to eql(0)
+      end
     end
 
     describe "#currency_select_options" do
--- a/spec/helpers/codelist_helper_spec.rb
+++ b/spec/helpers/codelist_helper_spec.rb
@@ -48,6 +48,24 @@ RSpec.describe CodelistHelper, type: :helper do
           with_empty_item: false
         ).last).to eq(OpenStruct.new(name: "Zloty", code: "PLN"))
       end
+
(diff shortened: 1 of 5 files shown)


spec/helpers/codelist_helper_spec.rb
```diff
@@ -48,6 +48,20 @@
           with_empty_item: false
         ).last).to eq(OpenStruct.new(name: "Zloty", code: "PLN"))
       end
+
+      it "does not add any items to the list with 'status: withdrawn'" do
+        list = helper.yaml_to_objects(
```
mec: It's a shame we can't overide the yaml here to make it clearer what is going on - I feel like I have to go and look at the yaml file to properly understand this. If we did that would it count as dependancy injection?

lozette: I was wondering this too. When I originally wrote these tests way back when, I had a dummy YAML file for testing in the fixtures, but I was advised that using the "real" data was safer. I feel like using a dummy YAML for this instance might be clearer though?

mec: It would defo be clearer for me!

If `yaml_to_objects` took a `data` argument, I feel like we could pass in a nice clear piece of yaml in the test - not sure if it's worth the effort here though?

lozette: Yeah I'm going to add a small fixture with some dummy data, I think it's best (and that's two votes for!)

mec: This is defintately better, I don't have to parse and understand that giant sectors file!

The one simple thing that would help me is to change the context to read:

"when there is only one active and two withdrawn items" it "filters out the withdrawn items"

or something like that, that explains the fixture without me needing to go look. then I would add an  expectation for the active item, somthing like:

`expect(list[1].status).to eq "active"`

that would make me super happy.

PR #254, opened 2020-03-06
Stop running the test script when a command errors

rjw1:
## Changes in this PR

Using `set -e` at the top of the script ensures that if one of the
following commands errors the script will stop and return an exit code
of not zero which will let CI know that something has failed.

Otherwise a command will fail during the script but Travis thinks
everything is green.

Also set `u` so if a variable is unbound it will also error out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ jobs:
     env:
     - RAILS_ENV=test
     script: |
+      set -eu
       docker network create test
       docker run -d --name pg --network test -p 5432:5432 postgres:11.6
       docker run -d \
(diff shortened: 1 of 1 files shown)


tahb:
It seems to be failing after a rebase on the install of `terraform-provider-cloudfoundry`. I notice that the terraform version being installed does not match our `.terraform-version`.

We have `0.12.21`
CI is logging `Installation of terraform v0.12.23 successful`

tahb:
Had a chat with bob. We think that whilst the Terraform version is mismatch is an issue it's not related to this CI failure and so will be addressed in another PR.

A restart of this job fixed the problem. We think it is due to a fragility in the GitHub API which is caclled by travis with the following:

```
bash -c "$(curl -fsSL https://raw.github.com/cloudfoundry-community/terraform-provider-cf/master/bin/install.sh)"
```

We can look at changing the install method to a different more reliable one. Again this fault was not introduced during this work so I'm going to hit merge :)

PR #258, opened 2020-03-10
(296) Date fields accept 0

lozette:
## Changes in this PR

Trello: https://trello.com/c/c9DkwHY9/296-bug-dates-accept-0

We noticed Papertrail recording `ActiveRecord::MultiparameterAssignmentErrors` when a user input `0` into any of the three inputs on any date field (generated via `govuk_date_field`)

Rails will attempt to coerce any values from a `date_field` view helper into a date. If any of those values are 0, the coercion will fail with `ActiveRecord::MultiparameterAssignmentErrors`

Unfortunately, because this coercion happens *before* the values are assigned to the model, we cannot catch them with a validator on the model and return errors on the attributes in question.

Taking a cue from https://thoughtbot.com/blog/ruby-on-fails we are catching the 
`ActiveRecord::MultiparameterAssignmentErrors` in `ActivityFormsController` and adding a generic flash error message to let the user know one or more of the entered dates was not valid. This is less than ideal but better than what we had before. 

## Screenshots of UI changes

Screenshot 2020-03-10 at 10 25 47

## Next steps

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -30,9 +30,14 @@ class Staff::ActivityFormsController < Staff::BaseController
     @activity = Activity.find(params[:activity_id])
     authorize @activity
 
-    @activity.assign_attributes(activity_params)
-    update_wizard_status
+    begin
+      @activity.assign_attributes(activity_params)
+    rescue ActiveRecord::MultiparameterAssignmentErrors
+      flash[:error] = I18n.t("activerecord.errors.models.activity.attributes.generic_date.invalid")
+      return render_wizard
+    end
 
+    update_wizard_status
     render_wizard @activity
   end
 
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -1,6 +1,5 @@
 class Staff::ActivityFormsController < Staff::BaseController
   include Wicked::Wizard
-  include DateHelper
   include ActivityHelper
 
   FORM_STEPS = [
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -1,6 +1,7 @@
 class Staff::ActivityFormsController < Staff::BaseController
   include Wicked::Wizard
   include ActivityHelper
+  include DateHelper
 
   FORM_STEPS = [
     :identifier,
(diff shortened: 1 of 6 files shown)


lozette:
Commit 7643e0cf976ef470b58169494f89a464f571aa07 is an idea but also a horrible hack so it can be dropped if necessary!

I have attempted to assign error messages to the correct date fields after `ActiveRecord::MultiparameterAssignmentErrors`

Because the params are named things like `planned_start_date(3i)` I have had to slice the key names up to get the attribute name, then assign an error message to that key name. It is *extremely* hacky, but gets the desired result:

Screenshot 2020-03-10 at 11 19 42

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -31,12 +30,28 @@ def update
...
+      assign_error_messages_to_record
+      return render_wizard
+    end
 
+    update_wizard_status
     render_wizard @activity
   end
 
+  def assign_error_messages_to_record
+    filtered_params = activity_params.select { |param| activity_params[param] == "0" }
```
erbridge: Is this going to also catch e.g. a code of `0`? Would feel safer if we explicitly checked the date params, perhaps before doing the assignment. However, doesn't this make more sense to be on the `Activity`? Validate date fields to not contain `0` date components?

config/locales/models_and_forms.en.yml
```diff
@@ -60,13 +60,17 @@ en:
             actual_end_date:
               between: Date must be between %{min} years ago and %{max} years in the future
               not_in_future: Actual end date must not be in the future
+              zero: Date may not contain 0
```
erbridge:
```suggestion
              zero: Date component may not be 0
```

I would expect `07` to be valid.

erbridge: Or even "{Day,Month,Year}" may not be 0"

app/controllers/staff/activity_forms_controller.rb
erbridge: Oh. I read your description again. I see why it's not on the model. Does the error occur on assignment or on save?

lozette: On assignment, full trace here:

```
1) Users can create a fund level activity when the user belongs to BEIS validations validation errors work as expected
     Failure/Error: @activity.assign_attributes(activity_params)
     
     ActiveRecord::MultiparameterAssignmentErrors:
       2 error(s) on assignment of multiparameter attributes [error on assignment [1, 0, 2010] to planned_start_date (mon out of range),error on assignment [0, 12, 2010] to planned_end_date (mday out of range)]
     # ./app/controllers/staff/activity_forms_controller.rb:34:in `update'
     # ./spec/features/staff/users_can_create_a_fund_level_activity_spec.rb:134:in `block (4 levels) in <main>'
     # ./spec/rails_helper.rb:73:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:72:in `block (2 levels) in <top (required)>'
```

lozette: In *theory* entering 0 into other params should be OK, as they won't cause `ActiveRecord::MultiparameterAssignmentErrors` and we won't get here. But I agree, it's fragile as hell and I think this commit is a bad idea. But I wanted to see if i could try to get the error messages on the right fields, and not just one generic error message.

config/locales/models_and_forms.en.yml
lozette: Agree!

app/controllers/staff/activity_forms_controller.rb
erbridge: Perhaps selecting only fields that have a `_date` suffix and reraising otherwise would make it more robust. I think it would also help clarify what's happening here.

erbridge:
Not sure input validations are reliable enough across browsers, but otherwise, what Tom said. Can we run the date fields through our own validator before assigning it?

lozette:
I had considered adding ranges to the input fields, at the moment the view is just
```
  = f.govuk_date_field :planned_start_date
  = f.govuk_date_field :planned_end_date
  = f.govuk_date_field :actual_start_date
  = f.govuk_date_field :actual_end_date
```
so I assume it's OK to replace these with something more HTML-flavoured?

erbridge:
Can you add extra properties (`min`, `max`) via a second argument?

lozette:
@erbridge Unfortunately I don't think so!

lozette:
I'm going to close this PR as my most recent attempt to resolve the issue overrides all the other commits in this branch.

PR #259, opened 2020-03-10
Purpose of activity is now dynamic

MariluzHerrera:
## Changes in this PR

Trello: https://trello.com/c/OWw5dOUE

Addressing the feedback from user research, during the creation of an activity the form step for 'Purpose' needed to be dynamically populated with the level of the activity. That would work as a reminder for users that what they are creating is either a fund, a programme or a project.

In order to achieve this, I have added that activity level to the yaml file so it could be displayed in the view for the purpose step. After doing that and checking all the tests were passing and the view on the browser was actually what I wanted to be, I realised that the page title on the browser tab was displaying
`Purpose of %{level}` (see image attached at the bottom of this section) instead of being dynamically populated. I believe this page title comes from a meta tag that it is common to the entire application. I think changing this meta tag would be too disruptive and complicated for this small tweak, so I decided that the best option would be to create another entry on the yaml file for `page_title.activity_form.show.purpose_level` where I could specify the level of the activity and pass that to the view and the tests. This allows me to leave `page_title.activity_form.show.purpose` as only 'Purpose' so the browser tab will not display a literal interpolation anymore. I am aware this is only _a_ way of doing it but not necessarily the right way, so suggestions on this will be appreciated.

Screenshot 2020-03-09 at 17 42 38

### Before

Screenshot 2020-03-10 at 10 54 14

### After

Screenshot 2020-03-10 at 10 16 39

## Next steps

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -267,7 +267,8 @@ en:
         finance: Finance
         flow: Flow
         identifier: Your unique identifier
-        purpose: Purpose of activity
+        purpose: Purpose
+        purpose_level: Purpose of %{level}
         region: Recipient region
         sector: Sector
         status: Status
--- a/spec/support/form_helpers.rb
+++ b/spec/support/form_helpers.rb
@@ -21,7 +21,8 @@ module FormHelpers
     flow: "ODA",
     finance: "Standard grant",
     aid_type: "General budget support",
-    tied_status: "Untied"
+    tied_status: "Untied",
+    level: "fund"
   )
 
     expect(page).to have_content I18n.t("activerecord.attributes.activity.identifier")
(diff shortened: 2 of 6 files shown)


config/locales/en.yml
```diff
@@ -267,7 +267,8 @@ en:
         finance: Finance
         flow: Flow
         identifier: Your unique identifier
-        purpose: Purpose of activity
+        purpose: Purpose
+        purpose_level: Purpose of %{level}
```
lozette: I think this is fine!

spec/support/form_helpers.rb
```diff
@@ -21,15 +21,16 @@ def fill_in_activity_form(
     flow: "ODA",
     finance: "Standard grant",
     aid_type: "General budget support",
-    tied_status: "Untied"
+    tied_status: "Untied",
+    level: nil
```
lozette: I think it would be nice to have a default for the level. If one calls `fill_in_activity_form` without passing in a level param, the result of line 33 is `"Purpose of "`. The test still passes but it's a bit odd. As all the other params have a default value, I think it's fine to add one to level too.

MariluzHerrera: It is a good point. I added the level param in every `fill_in_activity_form` from the specs files to make sure they all pass a parameter (fund, programme or project) to the form_helpers. But I don't see any harm in adding a default value here. Should we use the generic `activity` as default value?

lozette: I think use `fund` as the default as it's the top level activity type, but it doesn't matter particularly! I don't think `activity` is a level though.

MariluzHerrera: True. Activity is not a level. I'll use `fund`. Thank you Laura!

lozette:
Rebase it and ship it! 🚀

PR #260, opened 2020-03-10
(382) Upgrade redis from 3.x to 4.x

tahb:
## Changes in this PR

GPaaS now supports Redis services of version 4 and 5. This is a recent
change and one we were expecting from our recent move to the PaaS from
Heroku.

New Redis service plans are now available
https://docs.cloud.service.gov.uk/deploying_services/redis/#redis

We are switching to version 4 as that's what we were on before the
migration. Though version 5 is available there may be other unexpected
consequences of jumping 2 major versions.

As part of this work we keep the same size redis instance (tiny) and the
same high availability setting (HA).

I've tested this Terraform change in a plan against the staging
environment and will include screenshots of the outcome in the pull
request.

I am not sure what the impact of suffixing the version with '_x' is.
Options for minor options are not listed on a `cf martketplace -s redis`.

NB. Whilst the documentation specify dots in their syntax, Terraform
requires them to be translated to underscores. `tiny-ha-4.x` becomes
`tiny-ha-4_x`.

## Screenshots of changes

Screenshot 2020-03-10 at 11 28 46

## Next steps


tahb:
I deployed one manually and deleted it on staging. Something spun up when using `4_x` so let's see!

PR #261, opened 2020-03-10
Fix/ci stops failing to download cloudfoundary plugin

tahb:
## Changes in this PR

- stop deploying to Heroku as we no longer have Heroku environments and their output in Travis was misleading the debugging
- install CF Terraform plugin manually instead of automatically, following the CF documentation https://github.com/cloudfoundry-community/terraform-provider-cf/wiki#manually

This change is not fully testable on a local environment since Travis will be the one to download this package and subsequent Terraform commands on real environments. We could spin up a new environment to test this if we're concerned enough to take the time.

## Next steps

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+os: linux
 language: minimal
 services:
 - docker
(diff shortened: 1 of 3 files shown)


tahb:
Superceded by bob's version  https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/265

PR #266, opened 2020-03-12
(296) Date fields accept 0

lozette:
## Changes in this PR

Trello: https://trello.com/c/c9DkwHY9/296-bug-dates-accept-0

We noticed Papertrail recording ActiveRecord::MultiparameterAssignmentErrors when a user input 0 into any of the three inputs on any date field (generated via govuk_date_field)

Rails will attempt to coerce any values from a date_field view helper into a date. If any of those values are 0, the coercion will fail with ActiveRecord::MultiparameterAssignmentErrors

Unfortunately, because this coercion happens before the values are assigned to the model, we cannot catch them with a validator on the model and return errors on the attributes in question.

After trying several solutions, we have extended DateHelper (which was being included in several controllers but not actually being used!) to catch and reject any dates which contain a zero in any of the 3 date parameters. 

In the controller, we update all the attributes via mass assignment, *except* date parameters. The date parameters are then passed to DateHelper and checked for correctness (all fields present, no only-zeros in any of the fields).

 Dates which are invalid are then validated on the UI as if they had been entered as blank.

## Screenshots of UI changes

Screenshot 2020-03-12 at 11 42 06

## Next steps

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -40,13 +40,39 @@ class Staff::ActivityFormsController < Staff::BaseController
     @activity = Activity.find(params[:activity_id])
     authorize @activity
 
-    @activity.assign_attributes(activity_params)
+    update_activity_dates
+    update_activity_attributes_except_dates
+
     update_wizard_status
 
     render_wizard @activity
   end
 
   private
+
+  def date_field_params_regex
+    # This regex will match the three date params from `date_field`;
+    # e.g. actual_start_date(2i) actual_start_date(1i) actual_start_date(3i)
+    /^(planned|actual)_(start|end)_date\([1-3]i\)$/
+  end
+
+  def update_activity_attributes_except_dates
+    activity_params_except_dates = activity_params.reject { |param| param.match(date_field_params_regex) }
+    @activity.assign_attributes(activity_params_except_dates)
+  end
+
+  def update_activity_dates
+    activity_date_params = activity_params.select { |param| param.match(date_field_params_regex) }
+    return if activity_date_params.empty?
+    @activity.update(planned_start_date: format_date(date_params("planned_start_date")))
+    @activity.update(planned_end_date: format_date(date_params("planned_end_date")))
+    @activity.update(actual_start_date: format_date(date_params("actual_start_date")))
+    @activity.update(actual_end_date: format_date(date_params("actual_end_date")))
+  end
+
+  def date_params(date_type_string)
+    {day: activity_params["#{date_type_string}(3i)"], month: activity_params["#{date_type_string}(2i)"], year: activity_params["#{date_type_string}(1i)"]}
+  end
 
   def activity_params
     params.require(:activity).permit(:identifier, :sector, :title, :description, :status,
--- a/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
+++ b/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
@@ -120,6 +120,17 @@ RSpec.feature "Users can create a fund level activity" do
         expect(page).to have_content "Planned start date can't be blank"
         expect(page).to have_content "Planned end date can't be blank"
 
+        # Dates cannot contain only a zero
+        fill_in "activity[planned_start_date(3i)]", with: 1
+        fill_in "activity[planned_start_date(2i)]", with: 0
+        fill_in "activity[planned_start_date(1i)]", with: 2010
+        fill_in "activity[planned_end_date(3i)]", with: 0
+        fill_in "activity[planned_end_date(2i)]", with: 12
+        fill_in "activity[planned_end_date(1i)]", with: 2010
+        click_button I18n.t("form.activity.submit")
+        expect(page).to have_content "Planned start date can't be blank"
+        expect(page).to have_content "Planned end date can't be blank"
+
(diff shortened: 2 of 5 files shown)


app/controllers/staff/activity_forms_controller.rb
```diff
@@ -31,14 +31,42 @@ def update
...
+
     update_wizard_status
 
     render_wizard @activity
   end
 
   private
 
+  def update_activity_dates
+    return if activity_params.select { |param| param.match(/(.*)date(.*)/) }.empty?
```
erbridge: Actually, would it be better to drop the regex and guard each update with the presence of the matching param?

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -31,14 +31,42 @@ def update
...
+    @activity.update(planned_start_date: format_date(planned_start_date_params))
+    @activity.update(planned_end_date: format_date(planned_end_date_params))
+    @activity.update(actual_start_date: format_date(actual_start_date_params))
+    @activity.update(actual_end_date: format_date(actual_end_date_params))
+  end
+
+  def planned_start_date_params
+    {day: activity_params["planned_start_date(3i)"], month: activity_params["planned_start_date(2i)"], year: activity_params["planned_start_date(1i)"]}
+  end
```
erbridge: This could be more terse by making a `date_params(param_name)` method that does this in the general case using string concatenation, then use the regular expression above to find all of the matching parameter names and do this as an iterator. That way you also catch future date fields.

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -31,14 +31,42 @@ def update
     @activity = Activity.find(params[:activity_id])
     authorize @activity
 
-    @activity.assign_attributes(activity_params)
+    # Update activity dates, after checking they are valid via format_date
+    update_activity_dates
+    # Update all attributes except dates
+    @activity.assign_attributes(activity_params.select { |param| param.match(/^((?!date).)*$/) })
```
erbridge: I suggest flipping this `select` to a `reject` and using the same regex you're using in `update_activity_dates`.

app/controllers/staff/activity_forms_controller.rb
lozette: If I do that it's going to be something like 

```
@activity.update(planned_start_date: format_date(planned_start_date_params)) if [activity_params["planned_start_date(1i)"],activity_params["planned_start_date(2i)"],activity_params["planned_start_date(3i)"]].any?
```
because there's some Rails magic going on with the date params

lozette:
I've done a little refactoring, it's ready for another look!

spec/helpers/date_helper_spec.rb
```diff
@@ -21,5 +21,10 @@
       params = {day: "40", month: "13", year: "2020"}
       expect(helper.format_date(params)).to eq(nil)
     end
+
+    it "returns nil when given a zero parameter" do
```
mec: The commit message confused me for a while. I was expecting some code in DateHelper but there is only this extra test?

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -16,6 +16,10 @@ class Staff::ActivityFormsController < Staff::BaseController
     :tied_status,
   ]
 
+  # This regex will match the three date params from `date_field`;
+  # e.g. actual_start_date(2i) actual_start_date(1i) actual_start_date(3i)
+  DATE_PARAM_REGEX = /.*_date\(.*/
```
mec: This is clever! When I see a comment I wonder if a nice, well named method would help slow-pokes like me figure out what this does?

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
```diff
@@ -123,6 +123,17 @@
...
 
+        # Dates cannot contain only a zero
+        fill_in "activity[planned_start_date(3i)]", with: 1
+        fill_in "activity[planned_start_date(2i)]", with: 0
+        fill_in "activity[planned_start_date(1i)]", with: 2010
+        fill_in "activity[planned_end_date(3i)]", with: 0
+        fill_in "activity[planned_end_date(2i)]", with: 12
+        fill_in "activity[planned_end_date(1i)]", with: 2010
+        click_button I18n.t("form.activity.submit")
+        expect(page).to have_content "Planned start date can't be blank"
```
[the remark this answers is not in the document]
lozette: Have a look at my closed PR to see the hoops I had to go through to get a better error message 😆  https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/258

spec/helpers/date_helper_spec.rb
lozette: Good point, I'll amend the commit message

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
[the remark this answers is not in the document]
mec: I thought you may have! Yoinks!

app/controllers/staff/activity_forms_controller.rb
lozette: Done 👍

lozette:
Good call on the regex, I am def a bit too greedy with mine

PR #271, opened 2020-03-13
(353) Switch Tied Status from a select box to a set of radio buttons with h…

lozette:
…ints

## Changes in this PR

Trello: https://trello.com/c/mibiNC05/353-create-activity-tied-status-change-to-radio-btn-with-hints

As part of this, I have renamed CodelistHelper#yaml_to_status_objects to
CodelistHelper#yaml_to_objects_with_description to make it more generic.

Unfortunately I don't think govuk_collection_radio_buttons takes a default selected
option so we have lost the "Untied" default selection.

## Screenshots of UI changes

Screenshot 2020-03-13 at 11 22 46

## Next steps

### DIFF
--- a/app/helpers/codelist_helper.rb
+++ b/app/helpers/codelist_helper.rb
@@ -17,7 +17,7 @@ module CodelistHelper
     objects
   end
 
-  def yaml_to_status_objects(entity:, type:)
+  def yaml_to_objects_with_description(entity:, type:)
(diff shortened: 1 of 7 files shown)


lozette:
Commit split!

I wonder if content design would recommend we mark the previously-default option with something like "Most common" or "Recommended default" in the label?

PR #273, opened 2020-03-16
(457) Fix cookie overflow error by switching to Redis for cookie storage

tahb:
By default Rails stores session data in an encrypted cookie in the user's browser. This is prone to a CookieOverflow error when the data stored exceeds 4K.

> Sessions typically contain at most a user_id and flash message; both fit within the 4K cookie size limit. A CookieOverflow exception is raised if you attempt to store more than 4K of data.

https://api.rubyonrails.org/classes/ActionDispatch/Session/CookieStore.html

## Changes in this PR

- switch session storage from client-side to server-side with using the [Active Record session store gem](https://github.com/rails/activerecord-session_store) to remove the limit on what amount of data can be stored in the session
- session stubbing for feature tests has changed to now Mock OAuth behaviour at the boundary of the service. This adds more overhead to each feature test that requires logging in but increases the coverage and fixes a problem where all feature tests fail when the session store is set to something other than cookie/Rack based. The RackSessionAccess gem can now be removed with some additional changes to feature tests in order to move tests for signed in users, within a context block.

## Next steps


spec/services/create_project_activity_spec.rb
```diff
@@ -1,49 +1,96 @@
...
-  let(:user) { create(:administrator, organisation: delivery_partner_organisation) }
-  let(:programme) { create(:programme_activity, organisation: beis) }
-
   describe "#call" do
-    let(:result) {
-      described_class.new(user: user, organisation_id: user.organisation.id, programme_id: programme.id).call
-    }
-
     it "sets the Organisation to that of users organisation" do
+      beis = create(:beis_organisation)
```
mec: I find this super interesting! This kind of repetition is what would usually have us reaching for `let` and here we think it is playing a part in making this fragile? 

Whilst I think I am fan of clear and simple over clever yet obfuscated, this is even making me feel odd!

I wonder two things:

- do we know that removing the mystery guests has fixed the issue?
- if it has, how do we prevent future us refactoring this test back to `lets` and returning this to it's fragile state?

I feel like there must be some way to setup a `create_project_activity` object with all the same attributes and then place it under test?

tahb: You are quite right for flagging this. As this was an intermittent failing test that I noticed whilst trying to fix another problem I am going to defer the problem to a seperate task in order to get the cookie error fixed asap.

.travis.yml
```diff
@@ -13,12 +13,14 @@ jobs:
...
       docker network create test
       docker run -d --name pg --network test -p 5432:5432 postgres:11.6
+      docker run -d --name redis --network test -p 6379:6379 redis:4.0.14
       docker run -d \
         --name test-container \
         --network test \
         -e RAILS_ENV=test \
         -e DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true \
         -e DATABASE_URL=postgres://postgres@pg:5432/roda_test \
+        -e REDIS_URL=postgres://redis:6379 \
```
mec: Shouldn't this be `redis://`?

config/initializers/session_store.rb
```diff
@@ -0,0 +1,18 @@
+redis_url = ENV.fetch("REDIS_URL", nil)
+
+if redis_url.nil? || Rails.env.test?
```
mec: Now we have redis in CI can we remove the check for `Rails.env.test?`?

doc/architecture/decisions/0018-use-a-server-side-session-store.md
```diff
@@ -0,0 +1,34 @@
...
+
+Rails acknowledges [5] that the default cookie storage is fast but prone to this error.
+
+## Decision
+
+Use Redis [1] as a server-side session store.
+
+## Consequences
+
+This gem is maintained by Redis and is a popular choice for session management.
```
mec: The ADR improved loads!

config/initializers/session_store.rb
```diff
@@ -0,0 +1,18 @@
...
+    servers: {
+      host: redis_uri.host,
+      port: redis_uri.port,
+      db: 1,
+      namespace: "roda:session",
+    },
+    expire_after: 12.hours,
+    threadsafe: true,
+    signed: true,
+    secure: true
```
mec: Do we also need `key` here?

tahb:
I got an `InvalidAuthenticityToken` error as well. It seems to be due to a combination of configuration: 

1. `secure` should not be true in developmenr since we don't require HTTPS
2.  `signed` doesn't work at all. I've dug through the config and it looks to be [recently available to help us encrypt cookie data](https://github.com/redis-store/redis-actionpack/releases/tag/v5.2.0) but I can't figure out why it's a problem. I suspect there's a conflicting option elsewhere. I am removing the option and allowing it to choose the default.

mec:
Ah! I just got the `secure` - did you see my later question about the `key`?

mec:
My guess on `signed` would be it requires https?

config/initializers/session_store.rb
tahb: From testing `namespace: "roda:session"` has an affect on the key name but the `key` argument doesn't. This might be because we have the `redis-namespace` gem in the mix too.

In this screenshot of my local session keys you can see the first few entries where only `key` is used and the last where the `namespace` is used.

Screenshot 2020-03-18 at 13 00 30

I figure that we don't but I don't really like the magic going on as it does say `key` in `redis-rails` 🤷‍♂

mec: Odd, my take on `key` is it the name of the cookie used locally - mine blows up without it!?

mec: here is mine set to `_roda_session` in FF:

Screenshot 2020-03-18 at 13 08 49

tahb:
I left a reply on the `key` question 👍 

From [the redis-rails documentation](https://github.com/redis-store/redis-rails/blob/bc2a75fc530d2318053f1f7bcf6efdb26d724cde/README.md#session-storage) my read is that `secure` forces cookie exchanges over HTTPS so they are encrypted in transit. 

On reading `signed` again I _think_ it's not relevant to us:

> signed uses signed/encrypted cookies to store the local session on a client machine, preventing a malicious user from tampering with its contents.

A benefit of storing the data in Redis would be that session contents were **not** stored on the client. If session data is not on the client's machine within a cookie then this option sounds irrelevant.

mec:
> A benefit of storing the data in Redis would be that session contents were **not** stored on the client. If session data is not on the client's machine within a cookie then this option sounds irrelevant.

Yeah agreed, don't really understand, in my head you do need a cookie to tie up the user with the session and I think this is the cookie in question? from the redis-rails docs:

> we present the session data to the user as a signed/encrypted cookie, but we persist the data in Redis.

I didn't find the docs super clear, especially the part about Rails 5.2

config/initializers/session_store.rb
tahb: Doh 🤦‍♂ yes of course there is a redis key **and** a client key. Updated!

tahb:
@mec ready for another look 🤞

config/initializers/session_store.rb
```diff
@@ -0,0 +1,18 @@
+redis_url = ENV.fetch("REDIS_URL", nil)
+return unless redis_url
```
mec: If we return here, what `session_store` will Rails use? I've tested it and it works - but I don't see how!?

mec: Okay, looking at the docs it defaults to `:cookie_store`, makes sense now.

Gemfile.lock
```diff
@@ -296,8 +296,24 @@ GEM
     rb-inotify (0.10.0)
       ffi (~> 1.0)
     redis (4.1.3)
+    redis-actionpack (5.2.0)
```
mec: Reading the docs it does kind of imply we only need `redis-actionpack` but I don't think it is an issue that should block this - just interested to hear your take.

config/initializers/session_store.rb
tahb: I was being explicit about this in an earlier version. Whilst I knew it would be the cookie store after dealing with it it's important to note that the next person to read it is likely to wonder the same so I'll add it back in!

Gemfile.lock
tahb: I wondered this but when looking at redis-actionpack they link back to redis-rails if it's being used in a Rails context:

> If you are using redis-store with Rails, head on over to the redis-rails README to learn how to integrate this gem into your Rails application.

https://github.com/redis-store/redis-actionpack#usage

Now it works, I will test if only using actionpack works and report back as I agree less dependencies the better!

config/initializers/session_store.rb
tahb: Done 👍

Gemfile.lock
tahb: All works! Done 👍

config/initializers/session_store.rb
tahb: Another advantage of being explicit is that we can set the same session names and expiration dates to provide some consistency.

PR #281, opened 2020-03-19
(466) Enable BEIS users to see all Budgets and Transactions that belong to an Activity

lozette:
## Changes in this PR
Trello: https://trello.com/c/Xnef9CW2/466-bug-beis-users-do-not-see-transations-or-budgets-that-are-not-associated-with-their-organisation

Regardless of which organisation the Budgets or Transactions belong to, BEIS users
should be able to see them.

This ticket was raised because BEIS users could not see Budgets and Transactions
in the downloaded IATI XML, because the Budgets/Transactions did not belong to the
user's organisation.

Now, in the Activity XML BEIS users can see all budgets and transactions that belong to
the Activity.

## Next steps

### DIFF
--- a/spec/policies/budget_policy_spec.rb
+++ b/spec/policies/budget_policy_spec.rb
@@ -3,18 +3,95 @@ require "rails_helper"
 RSpec.describe BudgetPolicy do
   let(:user) { create(:administrator) }
   let(:activity) { create(:activity, organisation: user.organisation) }
+  let(:other_activity) { create(:activity, organisation: create(:organisation)) }
   let(:budget) { create(:budget, activity: activity) }
+  let!(:other_budget) { create(:budget, activity: other_activity) }
 
   subject { described_class.new(user, budget) }
 
-  it { is_expected.to permit_action(:index) }
-  it { is_expected.to permit_action(:show) }
-  it { is_expected.to permit_new_and_create_actions }
-  it { is_expected.to permit_edit_and_update_actions }
-  it { is_expected.to permit_action(:destroy) }
+  context "as a BEIS user" do
+    let(:user) { create(:beis_user) }
 
-  it "includes budget in resolved scope" do
-    resolved_scope = described_class::Scope.new(user, Budget.all).resolve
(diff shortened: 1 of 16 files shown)


spec/policies/budget_policy_spec.rb
```diff
@@ -13,8 +16,21 @@
   it { is_expected.to permit_edit_and_update_actions }
   it { is_expected.to permit_action(:destroy) }
 
-  it "includes budget in resolved scope" do
-    resolved_scope = described_class::Scope.new(user, Budget.all).resolve
-    expect(resolved_scope).to include(budget)
+  context "as a BEIS user" do
+    let(:user) { create(:beis_user) }
+
+    it "includes all budgets in the resolved scope" do
```
tahb: Does this still ensure that only budgets that belong to the current activity are viewed, as opposed to budgets from all activities?

lozette: Yeeees it might help if I implemented that 🤦‍♂ 😅

lozette:
Ah good point, the initial request was for the XML but yes seeing them in the HTML would be good (I missed that bit)

lozette:
Done, but https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/281/commits/6533590aea3b7f9331375c85cb2f2b955ca955e5 is messy again as we can't do it just with Policies. 

Specifically, in `transactions/_table.html.haml` and `budgets/_table.html.haml` we need to call `policy(:project).create?` and know if we're calling it against a Budget/Transaction which has a Project as its activity. Maybe there is a better way I'm not seeing?

tahb:
As mentioned on Slack I've added 2 commits as a suggestion for an alternative approach for your review https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commits/fix/466-bug-beis-users-do-not-see-transations-or-budgets-that-are-not-associated-with-their-organisation-tom-spike

lozette:
Looks great, thank you for your help! 👍

PR #282, opened 2020-03-19
Sector step renamed to Focus Area for clarity and hint text added to …

tahb:
…the form

## Changes in this PR

The form step changes dynamically to include the level of activity being created. Once this contextual guidance has been used to create the activity, the static "Focus area" is used instead of "Sector" through the rest of the service (apart from IATI where sector is used).

"Focus area in {level}" is only used within the form when that concept is introduced to users. Once introduced the theory is that users can link this back when they see the content within the service and the dynamic inclusion becomes redundant. For example, viewing a fund level activity and seeing "Focus area for your fund" in the list of contents

Co-authored-by: Agz <agz@dxw.com>

### Before

Screenshot 2020-03-19 at 17 02 31

### After

Screenshot 2020-03-19 at 17 01 01
Screenshot 2020-03-19 at 17 01 48
Screenshot 2020-03-19 at 17 02 01

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,5 +83,6 @@
   for delivery partners to report on the programme
 - fix cookie error by switching session storage to Redis
 - Activity aid type is now selected by radio button, not a dropdown select box  
+- Iterate the form content for the sector field by renaming it to "focus area"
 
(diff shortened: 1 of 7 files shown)


deberny:
This is great! 🚀

PR #283, opened 2020-03-19
(471) Unknown Auth0 error messages are not displayed

mec:
This work came out of the Pentest which was performed on the service in
March 2020.

An attacker could inject their own error message into a fake
Auth0 response and get the user to inteact with a bogus service.

By checking the error message is something we know and understand and
then displaying our own error message instead we prevent this attack.

If we see a unknown error message, we log to Rollbar so we can learn
what errors we are seeing and display a generic error message.

As ever, I struggled to find a good home for the locales that display
the errors, I went for a location that already had 'errors' and 'auth0'
over creating more separation.

I reworded some of the specs to better reflect what is under test.

### DIFF
--- a/app/controllers/auth0_controller.rb
+++ b/app/controllers/auth0_controller.rb
@@ -15,7 +15,12 @@ class Auth0Controller < ApplicationController
   end
 
   def failure
-    # show a failure page or redirect to an error page
-    @error_message = request.params["message"]
+    message = request.params["message"]
+    @error_message = t message.to_sym,
+      scope: "page_content.errors.auth0.error_messages",
+      raise: true
+  rescue I18n::MissingTranslationData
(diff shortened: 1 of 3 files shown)


mec:
[Authentication error messages cannot be modified [ID:WEB-M1]](https://trello.com/c/Kn6bhkDb/471-authentication-error-messages-cannot-be-modified-idweb-m1)

app/controllers/auth0_controller.rb
```diff
@@ -15,7 +15,12 @@ def callback
   end
 
   def failure
-    # show a failure page or redirect to an error page
-    @error_message = request.params["message"]
+    message = request.params["message"]
+    @error_message = t message.to_sym,
```
lozette: I had no idea you can do this, v clever!

PR #286, opened 2020-03-20
Enable Rails 6 hosts whitelisting to counteract host header poisoning

lozette:
## Changes in this PR
Trello: https://trello.com/c/fKK1MwEE/473-http-host-header-poisoning-id-web-l3

During a recent pen test, our application was found to be vulnerable to a Host
header poisoning attack, in which a "bad" hostname can be injected into the
request headers and passed back to the client, where it may be used to redirect
users (or in any scenario where `url_for` is used or the `location` is passed
back to the client)

Using a built-in feature in Rails 6 https://github.com/rails/rails/pull/33145
we are able to only allow known hostnames to be passed on from a request header
into the application.

## Next steps

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -118,4 +118,10 @@ Rails.application.configure do
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
 
   config.require_master_key = false
+
+  # See https://github.com/rails/rails/issues/29893
+  # This whitelists the hosts the application can trust when using `url_for` and related helpers
+  config.hosts = [
+    ENV["DOMAIN"],
+  ]
 end
--- a/spec/requests/host_urls_are_whitelisted_spec.rb
+++ b/spec/requests/host_urls_are_whitelisted_spec.rb
@@ -1,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe "host urls are checked against a whitelist", type: :request do
+  scenario "a whitelisted host is accepted and and OK response is received" do
+    get root_path, headers: {"Host" => "test.local"}
+    expect(response).to have_http_status("200")
+  end
+
+  scenario "a bad host is rejected and the request is forbidden" do
+    get root_path, headers: {"Host" => "baddomain.com"}
+    expect(response).to have_http_status("403")
+  end
+end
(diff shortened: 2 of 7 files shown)


mec:
@lozette I tried this locally and I don't think the spec works! :( 

I think it's passing in CI because the spec filename does not have '_spec' in it so it does not run unless you call it explicitilly, I think rspec looks for "*_spec.rb", although I am not 100% on that.

When I run the file, it fails on the "a bad host is rejected and the request is forbidden" spec.

lozette:
@mec Oh no! Well spotted and you're right!

lozette:
@mec ready for another look!

config/environments/production.rb
```diff
@@ -118,4 +118,12 @@
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
 
   config.require_master_key = false
+
+  # See https://github.com/rails/rails/issues/29893
+  # This whitelists the hosts the application can trust when using `url_for` and related helpers
+  config.hosts = [
+    "beis-roda-prod.london.cloudapps.digital",
```
mec: Do we need these? The ENV["DOMAIN"] is set in both environments and will likely change.

spec/requests/users_can_sign_out_spec.rb
```diff
@@ -5,10 +5,11 @@
     allow(ENV).to receive(:[]).with("BULLET_DEBUG").and_return("false")
     allow(ENV).to receive(:[]).with("AUTH0_CLIENT_ID").and_return("123456")
     allow(ENV).to receive(:[]).with("AUTH0_DOMAIN").and_return("test.auth0")
+    host! "test.local"
```
mec: If we have the global `config.before` do we still need this?

spec/requests/host_urls_are_whitelisted_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe "host urls are checked against a whitelist", type: :request do
+  scenario "a whitelisted host is accepted and and OK response is received" do
+    get root_path, headers: {"Host" => "test.local"}
```
mec: I like the explicit `test.local` here as it helps me understand what the test does. 👍

config/environments/production.rb
lozette: Good point, don't think so!

spec/requests/users_can_sign_out_spec.rb
lozette: I put this in so that the change on line 13 of `www.example.com` to `test.local` has some context.

mec: That is fair enough!

PR #287, opened 2020-03-20
Anonymize user's IP address when logging to Rollbar

lozette:
## Changes in this PR
Trello: https://trello.com/c/WCwPC52H/452-anonymise-ip-addresses-we-log-and-send-onwards

Much of this is lifted from https://github.com/DFE-Digital/teacher-vacancy-service/

We are currently logging users' IP addresses in full to Rollbar, which is a
privacy concern. This PR uses an anonymizing method to strip the last octet from
IP addresses so that they are still recognisable for tracking issues, but cannot
be used to trace the exact location/information of a user.

## Next steps


app/controllers/concerns/ip.rb
```diff
@@ -0,0 +1,34 @@
+module Ip
+  require "ipaddr"
+
+  extend ActiveSupport::Concern
+
+  def request_ip
+    anonymize_ip(request.ip)
+  end
+
```
mec: Should all this be private?

spec/controller/application_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+RSpec.describe ApplicationController, type: :controller do
+  describe "#request_ip" do
+    it "returns the anonymized IP with the last octet zero padded" do
+      expect(controller.request_ip).to eql("0.0.0.0")
```
mec: I'm struggling with this, I can't work out the setup? Does an RSpec controller test (contraversial! :P) have a preset IP? Has this done anything to it and how would I know?

spec/controller/application_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    context "when the IP is at the max range" do
+      it "returns the anonymized IP with the last octet zero padded" do
+        allow_any_instance_of(ActionController::TestRequest)
```
mec: This makes a bit more sense, I can see the setup and how the code has transformed the IP.

spec/controller/application_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+RSpec.describe ApplicationController, type: :controller do
```
mec: Should we also test the case when it is an IPv6 address?

config/initializers/rollbar.rb
```diff
@@ -13,6 +13,9 @@
     config.enabled = false
   end
 
+  # Anonymize user IP addresses when logging, for privacy
+  config.anonymize_user_ip = true
```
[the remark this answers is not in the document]
lozette: Good question!!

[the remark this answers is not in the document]
lozette: As discussed in standup today, this is because we want to anonymize the IP address anywhere it might be sent other than Rollbar (e.g. Papertrail or future applications) 👍

app/controllers/concerns/ip.rb
lozette: `request_ip` can't be but everything else can be 👍

spec/controller/application_controller_spec.rb
lozette: I've made this clearer by explicitly passing a v4 IP address

spec/controller/application_controller_spec.rb
lozette: Added!

PR #289, opened 2020-03-24
Recipient country list reduced

MariluzHerrera:
Trello: https://trello.com/c/R1sH0nhV

## Changes in this PR

When creating an activity, in the recipient country step there was a list including all the countries on the ISO list.

Following advice from users regarding this list being too extensive and including countries that ODA do not consider recipients, I have reduced the list of countries leaving only the ones that ODA report as recipients.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,10 +82,11 @@
 - Make it clearer that Programme should have an extending organisation in order
   for delivery partners to report on the programme
 - fix cookie error by switching session storage to Redis
-- Activity aid type is now selected by radio button, not a dropdown select box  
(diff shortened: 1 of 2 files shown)


mec:
@MariluzHerrera it looks like something has gone a bit wonky on your rebase? Shall we have a look and see if we can fix it together? Give me a shout if you would like to?

MariluzHerrera:
> @MariluzHerrera it looks like something has gone a bit wonky on your rebase? Shall we have a look and see if we can fix it together? Give me a shout if you would like to?

I think it is fixed now following your advice. Looks a lot cleaner. Thanks Meyric!!!

PR #291, opened 2020-03-24
IATI reporting-org should be BEIS regardless of the user's organisation

lozette:
## Changes in this PR
Trello: https://trello.com/c/CfLUXjCa/459-iati-reporting-org-should-be-beis-regardless-of-the-users-organisation

Funds and programmes should have their reporting org set as BEIS

Projects created by UKSA (our current only delivery partner) should have their
reporting org set as BEIS

Other future DPs will have themselves set as their reporting org (for now).

Note that these changes are only in the XML representation of the activity, not
the database record.

## Next steps

### DIFF
--- a/app/views/staff/shared/xml/_activity.xml.haml
+++ b/app/views/staff/shared/xml/_activity.xml.haml
@@ -1,7 +1,12 @@
 %iati-activity{"default-currency" => activity.default_currency, "xml:lang" => activity.organisation.language_code }
   %iati-identifier= activity.iati_identifier
-  %reporting-org{"type" => activity.organisation.organisation_type, "ref" => activity.organisation.iati_reference }
-    %narrative= activity.organisation.name
+  - service_owner = Organisation.find_by(service_owner: true)
+  - if activity.organisation.is_government_organisation?
+    %reporting-org{"type" => service_owner.organisation_type, "ref" => service_owner.iati_reference }
+      %narrative= service_owner.name
+  - else
+    %reporting-org{"type" => activity.organisation.organisation_type, "ref" => activity.organisation.iati_reference }
+      %narrative= activity.organisation.name
   %title
     %narrative= activity.title
   %description{"type" => "1"}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,5 +88,6 @@
 - Anonymize user's IP addresses before logging them outside the application, by removing the last octet of the address. Also use Rollbar's built-in IP address anonymizer.
 - BEIS users can view Transactions & Budgets on a project, but not create or edit them
 - Country list for recipient countries when creating an activity has been reduced to only those ODA uses as recipients.
+- Reporting org in the IATI XML is always BEIS for funds, programmes and projects created by UKSA, despite what the database record says
 
 [release-2]: https://github.com/dxw/DataSubmissionService/compare/release-2...release-1
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -7,4 +7,8 @@ class Organisation < ApplicationRecord
 
   scope :sorted_by_name, -> { order(name: :asc) }
   scope :delivery_partners, -> { sorted_by_name.where(service_owner: false) }
+
+  def is_government_organisation?
+    %w[10 11].include?(organisation_type)
+  end
 end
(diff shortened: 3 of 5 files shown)


app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -1,7 +1,15 @@
 %iati-activity{"default-currency" => activity.default_currency, "xml:lang" => activity.organisation.language_code }
   %iati-identifier= activity.iati_identifier
-  %reporting-org{"type" => activity.organisation.organisation_type, "ref" => activity.organisation.iati_reference }
-    %narrative= activity.organisation.name
+  - service_owner = Organisation.find_by(service_owner: true)
+  - if activity.fund? || activity.programme?
```
mec: Do we need this first if? Funds and Programmes can only be created by BEIS?

app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -1,7 +1,15 @@
...
+    %reporting-org{"type" => service_owner.organisation_type, "ref" => service_owner.iati_reference }
+      %narrative= service_owner.name
+  - elsif activity.project? && activity.organisation.name == "UKSA"
```
mec: What do you think about changing this to:

`elsif activity.project? && activity.organisation.is_goverment_organisation?`

And implementing that on Organisation? It feels like a small change that works for all current 'goverment' DPs if we do something like:

```
def is_goverment_organisation?
    ["Goverment", "Local Goverment"].include? organisation_type
end
````

Or something?

lozette:
@mec I don't know about the card! I took it at face value tbh, but now I'm not so sure

app/views/staff/shared/xml/_activity.xml.haml
lozette: God this is so much more elegant 👌

lozette:
@mec I have changed the template and it's much neater with `is_government_organisation?`

I don't know about adding another db column. I did it in the template because the XML is the only place where this is going to be expressed, but if we're going to put it in the UI as well, a db column would be good. I dunno, I'm torn! 

@tahb @MariluzHerrera what do you think - a new db column for something only expressed in the XML or not?

MariluzHerrera:
> @mec I have changed the template and it's much neater with `is_government_organisation?`
> 
> I don't know about adding another db column. I did it in the template because the XML is the only place where this is going to be expressed, but if we're going to put it in the UI as well, a db column would be good. I dunno, I'm torn!
> 
> @tahb @MariluzHerrera what do you think - a new db column for something only expressed in the XML or not?

That's something that did not occur to me. It is a good question. I assumed that a column on the table would be necessary since we are handling that type of information (regardless of being only in the XML). Actually, on two of my cards I deleted steps from the form and made their values default (Finance and Tied Status) and it never occur to me to delete the respective columns from the DB. Now I don't know what the best approach is!

lozette:
@MariluzHerrera I think with your cards it would be best to leave them in, just in case we decide to make the values selectable in future.

mec:
@lozette @tahb @MariluzHerrera 

I guess for me this comes down to my view of object oriented design, I always prefer a model object to hold all the 'knowledge' and views to be a dumb as possible – I would prefer a view to ask the model who the reporting org is than have one view contain that knowledge – it may be that my in-experience is showing (or my super old school OO knowlege is showing its age!).

It just feels 'right' to me to look at the model to understand what the attributes are rather than having to go out to a view.

Hoes does it feel if we do not have a column, but instead have a method on Activity that returns the correct reporting org, something like:

````ruby
def reporting_organisation
    if organisation.is_goverment_organisation?
        organisation
    else
        Organisation.where(service_owner: true)
    end
end
````

This way the view is even simpler and all the logic is on the model?

@MariluzHerrera I kind of feel the same way about your issue that we spoke about, having a view be the only place where I go to understand an attributes seems odd to me, when we wrote the test to see how it looked in the model I ❤️ how self documenting it was.

lozette:
Just spoke with Tom and we're going to refactor it make a new association `reporting_org` (I have never studied OO design, can you tell 😆 )

mec:
@lozette no worries - me either - I think this is just how it comes together in my head, I think I get the ideas from Objective-C so I always worry they are not applicable on Rails! 😺

lozette:
I'm going to close this bad boy as the new implementation is a bit different and this PR has got messy!

PR #294, opened 2020-03-25
Finance value defaulted and step removed from Create Activity journey

MariluzHerrera:
## Changes in this PR

Trello: https://trello.com/c/yfr1eiz2

Feedback from user research when creating an activity showed that the Finance step could possibly be omitted as the users always document the activities' finance under the "Standard grant" category. After discussion with them and knowing is highly likely they will keep using the same category in the future, we decided to default the finance category to "Standard grant" for every activity in every level and omit this step from the form the user needs to fill in when they create activities. This would reduce the time users take to complete this task too.

I have deleted the Finance step from the controller and the views. I have deleted the references to this step on the Spec files and the yaml files that are not needed anymore. I have also removed the Finance column from the activities table on the DB.
On the XML that is created for IATI, Finance is still there with the default value of code "110" ("Standard grant"). This value comes from a constant on the Activity model. 
I have checked some XMLs on the IATI public validator and they were valid.  

### Before

Screenshot 2020-03-25 at 09 07 20

### After

Screenshot 2020-03-24 at 15 52 56

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,7 +93,8 @@
 
 ## [unreleased]
 
+- When creating an activity the Finance step has been defaulted to `Standard grant` and omitted from the user journey
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-3...HEAD
 [release-3]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-2...release-3
 [release-2]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-1...release-2
-
--- a/app/presenters/activity_presenter.rb
+++ b/app/presenters/activity_presenter.rb
@@ -56,11 +56,6 @@ class ActivityPresenter < SimpleDelegator
     I18n.t("activity.flow.#{super}")
   end
 
-  def finance
-    return if super.blank?
-    I18n.t("activity.finance.#{super}")
-  end
-
   def tied_status
     return if super.blank?
     I18n.t("activity.tied_status.#{super}")
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -159,11 +159,6 @@ RSpec.describe Activity, type: :model do
       it { should validate_presence_of(:flow) }
     end
 
-    context "when finance is blank" do
-      subject { build(:activity, finance: nil, wizard_status: :finance) }
-      it { should validate_presence_of(:finance) }
-    end
-
     context "when tied_status is blank" do
       subject { build(:activity, tied_status: nil, wizard_status: :tied_status) }
       it { should validate_presence_of(:tied_status) }
@@ -181,7 +176,6 @@ RSpec.describe Activity, type: :model do
       it { should_not validate_presence_of(:actual_end_date) }
       it { should validate_presence_of(:geography) }
       it { should validate_presence_of(:flow) }
-      it { should validate_presence_of(:finance) }
       it { should validate_presence_of(:tied_status) }
     end
 
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -1,6 +1,13 @@
 require "rails_helper"
 
 RSpec.describe Activity, type: :model do
+  describe "#finance" do
+    it "always returns Standard Grant, code '110'" do
+      activity = Activity.new
+      expect(activity.finance).to eq "110"
+    end
+  end
+
   describe "scopes" do
     describe ".funds" do
       it "only returns fund level activities" do
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -13,7 +13,6 @@ class Staff::ActivityFormsController < Staff::BaseController
     :region,
     :country,
     :flow,
-    :finance,
     :aid_type,
     :tied_status,
   ]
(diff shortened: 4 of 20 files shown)


mec:
Nice work @MariluzHerrera, concise and small commits that I could follow easily - thank you!

I do wonder if there are still places that the concept of 'finance' still exists?

It would be great if the changelog went with the commit that actually drives the change - perhaps when we switch the value in 70b9c5d8fc876237d371181de9405032c6e82c22?

When I think about the flow of this work, I would imagine a test that checks `finance` is always `code=110`, that would document this behaviour of the data and drive my work - what do you think about a test, something like 'valid activity XML always has a a finance code of 110 standard grant'? 

This test might also shed light on some places where 'finance' still exists, I would need an activity with finance not set to `110` for example, how would I do that?

MariluzHerrera:
> Nice work @MariluzHerrera, concise and small commits that I could follow easily - thank you!
> 
> I do wonder if there are still places that the concept of 'finance' still exists?
> 
> It would be great if the changelog went with the commit that actually drives the change - perhaps when we switch the value in [70b9c5d](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/70b9c5d8fc876237d371181de9405032c6e82c22)?
> 
> When I think about the flow of this work, I would imagine a test that checks `finance` is always `code=110`, that would document this behaviour of the data and drive my work - what do you think about a test, something like 'valid activity XML always has a a finance code of 110 standard grant'?
> 
> This test might also shed light on some places where 'finance' still exists, I would need an activity with finance not set to `110` for example, how would I do that?

Thanks for your review. I'll have a look at the changes you requested. I'll include the changelog on that commit you mentioned (I did not know where to include the changelog so I did it pretty much at the end). 

I thought about tests since all I did was deleting tests. Did not realise making a test for the default value would be a great idea, so I'll do that.

`Finance` is still around in some places. It is still needed for the tests that check validations of the XMLs files, right?

vendor/data/codelists/IATI/2_03/activity/finance.yml
```diff
@@ -1,241 +0,0 @@
----
```
erbridge: I think we probably want to keep this file. It's useful for reference, even if we only use one code.

MariluzHerrera: I was trying to avoid having unnecessary files. There is another [yaml file](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/config/locales/codelists/2_03/iati.en.yml) in the locales for all IATI codes, including the finance ones. We could use those ones for reference if needed.

MariluzHerrera: After a conversation with @mec he explained to me the differences between vendor files and locales in terms of development, which I didn't know. Now I think it makes more sense to leave the vendor file on the code and delete the finance references from the locales. I'll do it that way

MariluzHerrera:
> One small comment, otherwise, I think this looks good!
> 
> I do wonder about the fragility of relying on a const for published activities. If we set every activity to have the code and stored it in the database, that would mean activities keep their codes even if the code is changed for future activities.

I considered the need of this column on the DB, and I discussed the change with other developers in the team. I deleted it because I think it is redundant since the value is given by a constant, and can cause confusion to future developers. If the specification changes in the future, then we can revisit this change.

CHANGELOG.md
```diff
@@ -89,5 +89,7 @@
 - BEIS users can view Transactions & Budgets on a project, but not create or edit them
 - Country list for recipient countries when creating an activity has been reduced to only those ODA uses as recipients.
 - Add feedback form link to phase banner
+- When creating an activity the Finance step has been defaulted to "Standard grant" and omitted from the user journey
+- Finance column has been deleted from activities table on the schema
```
mec, line 93: I am not sure we need this in the Changelog at all, but it is not in this commit eitherway?

I like to think of the changelog as features a user might understand, I guess there is an exception for security fixes so maybe 'something the product owner understands' is a better description?

app/presenters/activity_presenter.rb
```diff
@@ -56,11 +56,6 @@ def flow
     I18n.t("activity.flow.#{super}")
   end
 
-  def finance
```
mec, line 59: I'm not sure this counts as 'deleting from spec files' so it might have snuck into this commit by accident?

spec/models/activity_spec.rb
```diff
@@ -1,6 +1,13 @@
 require "rails_helper"
 
 RSpec.describe Activity, type: :model do
+  describe "#finance" do
```
mec, line 4: This spec is added here and I think is better placed along with the implementation?

MariluzHerrera: Yes, this one slipped in here by accident. I'll put it in the right commit. Thanks

app/presenters/activity_presenter.rb
MariluzHerrera: I'll move it. Thanks

CHANGELOG.md
MariluzHerrera: Let's talk about that later. I might have not understood the purpose of a Changelog.

MariluzHerrera: I've amended this file. Now it's showing the right information

app/presenters/activity_presenter.rb
MariluzHerrera: I've moved this to the right commit. Thanks for pointing that out.

spec/models/activity_spec.rb
MariluzHerrera: This test now lives in the right commit. Thank you @mec

PR #295, opened 2020-03-25
Feature/478 users no longer select tied status

MariluzHerrera:
## Changes in this PR

Trello: https://trello.com/c/qFZMNyvN

After discussion with service users, they informed us that the only category they use for `Tied status` is `Untied`, hence there is no need for a form step with different options when they create activities. This is unlikely to change in the future so we are confident that removing this step will only be beneficial for the users as they will be able to complete the journey quicker without affecting the quality of the documentation they create.

To achieve this, I introduced a default value of "5" ("Untied") for `Tied status` on activity model, as a constant, and removed the form step from the controller and views. The Spec files no longer test for the presence of this form step or a value been given by the user. I have also made tests to assert the default value will be always added to the XML file.

I have deleted the `tied_status` column from the DB as it serves no purpose anymore.

In order to make sure documentation will not be altered by this change, I have tested several XML files I created on different activity levels and check them on the IATI XML validator. All of them passed the test validator.

### Before

Screenshot 2020-03-25 at 09 07 20

### After

Screenshot 2020-03-25 at 15 17 39

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,11 +89,12 @@
 - BEIS users can view Transactions & Budgets on a project, but not create or edit them
 - Country list for recipient countries when creating an activity has been reduced to only those ODA uses as recipients.
 - Add feedback form link to phase banner
-- Activity to Activity association renamed `child_activities` (from `activities`) to avoid an association name clash with the `public_activity` gem
 
 ## [unreleased]
 
+- Activity to Activity association renamed `child_activities` (from `activities`) to avoid an association name clash with the `public_activity` gem
 - When creating an activity the Finance step has been defaulted to `Standard grant` and omitted from the user journey
+- When creating an activity, the `Tied status` step has been removed from the user journey and it has now a default value of `Untied`, code "5"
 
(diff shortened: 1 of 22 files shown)


mec:
I think my comments on #294 stand for this one as well - what do you think?

MariluzHerrera:
> I think my comments on #294 stand for this one as well - what do you think?

Yes, very similar. I'll move the changelog.md file to a different commit and I'll do a test for the default value. 
I might need a hand deciding where `Tied status` is needed in the code and where is not. I'll message you later. Thanks Meyric!

CHANGELOG.md
```diff
@@ -93,6 +93,7 @@
 
 ## [unreleased]
 
+- When creating an activity, the `Tied status` step has been removed from the user journey and it has now a default value of `Untied`, code "5"
```
mec, line 96: I think, by convention, we always add the new change to the bottom of the list.

MariluzHerrera: Yes. That was my mistake when rebasing. Thanks for pointing that out. I've fixed it now.

PR #296, opened 2020-03-26
(507) Setup JavaScript testing environment

mec:
## Changes in this PR

The goal of this work was to introduce the ability for the team to test JavaScript features in the test suite.

The approach is a standard one, we use [Capybara](https://github.com/teamcapybara/capybara#drivers) and pass `js: true`  to any specs that need to test JS behaviour. This relies on there being a web broweser and selenium driver available for use.

I've opted to use Firefox and Geckodriver (instead of Chrome and chrome driver) for no other reason that it is the default, if we feel strongly, switching the browser and driver should be relatively straight forward.

In local development you will need to have both Firefox and Geckodriver installed, a simple way to achieve this is in the readme as part of this work. A way to bypass the JS test is also documented for developers who may want to avoid the slow specs during day-to-day work, although the entire suite should always be run before pushing changes.

In CI a new `test` build stage was introduced to the Docker image that installs Firefox and Geckodriver, having a seperate stage means we can ignore that part of the build, which should give us a small speed improvement when deploying the app.

Whilst Firefox can be installed via the apt package manager, Geckodriver has to be installed manually, whilst this process is straight forward, it would to give any security implications someone elses viewpoint.

The version of Geckodriver can be overidden at build time with the `gecko_driver_version` argument that has a default of 0.26.0 (the latest current release).

In order to test the testing (!?) the pattern from GOVUK frontend has been followed of running a small, inline JS script that adds a class to the body element, by testing if this is present or not we can confirm that the testing environement is working both with and without JS enabled.

A stretch goal of tidying the Dockerfile and removing some superfluous code has been achieved. The main goal was the break the build into stages with the base requirements to run the app first, followed by the dependencies, then the application code itself and lastly the testing.

A summary of the chagnes is in the commit message, if we feel these changes are too much or unecessary, we can drop the related commit.

## Next steps

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: minimal
 services:
 - docker
 before_install:
-- docker build . --build-arg RAILS_ENV=test -t app/test
+- docker build . --target test --build-arg RAILS_ENV=test -t app/test
 jobs:
   include:
   - stage: Test
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ jobs:
         -e CI=true \
         --env-file .env.test \
         app/test /bin/bash -c "tail -f /dev/null"
-      docker exec test-container rake
+      docker exec test-container bundle exec rake
       docker exec test-container bundle exec brakeman
       echo "==== testing terrafrom ===="
       cd terraform
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,59 +1,80 @@
 # ------------------------------------------------------------------------------
-# dependency stage
+# base
 # ------------------------------------------------------------------------------
-FROM mkenney/npm:latest AS dependencies
-ENV INSTALL_PATH /deps
-RUN mkdir -p $INSTALL_PATH
-WORKDIR $INSTALL_PATH
-COPY package.json ./package.json
-COPY package-lock.json ./package-lock.json
-RUN npm set progress=false && npm config set depth 0
-RUN npm install --save govuk-frontend
-
-# ------------------------------------------------------------------------------
-# release stage
-# ------------------------------------------------------------------------------
-
-FROM ruby:2.6.3 as release
+FROM ruby:2.6.3 AS base
 MAINTAINER dxw <rails@dxw.com>
+
 RUN apt-get update && apt-get install -qq -y \
   build-essential \
   libpq-dev \
   --fix-missing --no-install-recommends
+
+ENV APP_HOME /app
+ENV DEPS_HOME /deps
+
+ARG RAILS_ENV
+ENV RAILS_ENV ${RAILS_ENV:-production}
+ENV NODE_ENV ${RAILS_ENV:-production}
+
+# ------------------------------------------------------------------------------
+# dependencies
+# ------------------------------------------------------------------------------
+FROM base AS dependencies
+
+# Set up install environment
+RUN mkdir -p ${DEPS_HOME}
+WORKDIR ${DEPS_HOME}
+# End
+
 RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - \
         && apt-get install -y nodejs
 
-ENV INSTALL_PATH /srv/beis-roda
-RUN mkdir -p $INSTALL_PATH
-
-# This must be ordered before rake assets:precompile
-COPY --from=dependencies ./deps/node_modules $INSTALL_PATH/node_modules
-COPY --from=dependencies ./deps/node_modules/govuk-frontend/govuk/assets $INSTALL_PATH/app/assets
-
-WORKDIR $INSTALL_PATH
-
-# set rails environment
-ARG RAILS_ENV
-ENV RAILS_ENV=${RAILS_ENV:-production}
-ENV RACK_ENV=${RAILS_ENV:-production}
-
-COPY Gemfile $INSTALL_PATH/Gemfile
-COPY Gemfile.lock $INSTALL_PATH/Gemfile.lock
+# Install Ruby dependencies
+COPY Gemfile ${DEPS_HOME}/Gemfile
+COPY Gemfile.lock ${DEPS_HOME}/Gemfile.lock
 
 RUN gem update --system 3.0.3
 RUN gem update rake 13.0.1
 RUN gem install bundler
 
-# bundle ruby gems based on the current environment, default to production
-RUN echo $RAILS_ENV
-RUN \
-  if [ "$RAILS_ENV" = "production" ]; then \
-    bundle install --jobs 4 --without development test --retry 10; \
+RUN bundle config set frozen 'true'
+
+# Configure bundler for the environment
+RUN if [ ${RAILS_ENV} = "production" ]; then \
+  bundle config set without 'development test'; \
+  elif [ ${RAILS_ENV} = "test" ]; then \
+  bundle config set without 'development'; \
   else \
-    bundle install --jobs 4 --retry 10; \
+  bundle config set without 'test'; \
   fi
 
-COPY . $INSTALL_PATH
+RUN bundle config
+RUN bundle install --retry 3 --jobs 4
+# end
+
+# Install JavaScript dependencies
+COPY package.json ${DEPS_HOME}/package.json
+COPY package-lock.json ${DEPS_HOME}/package-lock.json
+
+RUN npm set progress=false && npm config set depth 0
+RUN npm install
+#end
+
+# ------------------------------------------------------------------------------
+# web
+# ------------------------------------------------------------------------------
+FROM dependencies AS web
+
+# Set up install environment
+RUN mkdir -p ${APP_HOME}
+WORKDIR ${APP_HOME}
+# end
+
+COPY . ${APP_HOME}
+
+# This must be ordered before rake assets:precompile
+RUN cp -R $DEPS_HOME/node_modules $APP_HOME/node_modules
+RUN cp -R $DEPS_HOME/node_modules/govuk-frontend/govuk/assets $APP_HOME/app/assets
 
 RUN RAILS_ENV=$RAILS_ENV SECRET_KEY_BASE="super secret" bundle exec rake assets:precompile --quiet
 
@@ -70,12 +91,11 @@ EXPOSE 3000
 CMD ["bundle", "exec", "puma"]
 
 # ------------------------------------------------------------------------------
-# test stage
+# test
 # ------------------------------------------------------------------------------
+FROM web as test
 
-FROM release as test
-
-RUN apt-get install -yy firefox-esr
+RUN apt-get install -qq -y firefox-esr
 
 ARG gecko_driver_version=0.26.0
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
-# Install dependencies into a seperate and isolated Docker stage
-# that is thrown away apart from any subsequent COPY commands
+# ------------------------------------------------------------------------------
+# dependency stage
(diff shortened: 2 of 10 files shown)


Dockerfile
```diff
@@ -63,3 +68,21 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
...
+# test stage
+# ------------------------------------------------------------------------------
+
+FROM release as test
+
+RUN apt-get install -yy firefox-esr
+
+ARG gecko_driver_version=0.26.0
+
+RUN wget https://github.com/mozilla/geckodriver/releases/download/v$gecko_driver_version/geckodriver-v$gecko_driver_version-linux64.tar.gz
```
tahb: On your question around security: I think this is okay. The package is downloaded from a reputable place and over HTTPS. Should it be compromised in any way in the future the impact would be limited to test runs of the image given the use of docker stages. Code installed here cannot make it into the same image used to run production code.

docker-compose.test.yml
```diff
@@ -0,0 +1,24 @@
+version: "3.7"
+services:
+  test:
```
tahb: Having this around for testing purposes is something I would have found useful when debugging CI so thank you 👍 

Usually for this type of work I need to give docker compose access to my local files so changes I make are made available without having to rebuild the image everytime. [I'd usually do something like this](https://github.com/LBHackney-IT/report-a-defect/blob/develop/docker-compose.test.yml#L15) in a combination with running the container with Spring to autoload the changes.

I think this change could be merged in without this suggestion as the only downside is that rebuilding takes more time (if indeed this is an issue).

Dockerfile
```diff
@@ -1,59 +1,84 @@
...
-# This must be ordered before rake assets:precompile
-COPY --from=dependencies ./deps/node_modules $INSTALL_PATH/node_modules
-COPY --from=dependencies ./deps/node_modules/govuk-frontend/govuk/assets $INSTALL_PATH/app/assets
+ARG RAILS_ENV
+ENV RAILS_ENV ${RAILS_ENV:-production}
+ENV NODE_ENV ${RAILS_ENV:-production}
 
-WORKDIR $INSTALL_PATH
+# ------------------------------------------------------------------------------
+# dependancies
```
tahb:
```suggestion
# dependencies
```

Dockerfile
```diff
@@ -70,16 +95,15 @@ EXPOSE 3000
...
 
-FROM release as test
-
-RUN apt-get install -yy firefox-esr
+RUN apt-get install -qq -y firefox-esr
 
 ARG gecko_driver_version=0.26.0
 
-RUN wget https://github.com/mozilla/geckodriver/releases/download/v$gecko_driver_version/geckodriver-v$gecko_driver_version-linux64.tar.gz
+RUN wget --quiet  https://github.com/mozilla/geckodriver/releases/download/v$gecko_driver_version/geckodriver-v$gecko_driver_version-linux64.tar.gz
```
tahb: Is there a reason you chose to make the output as quiet as possible? There can be lots of output but the trade off is that with no output you can't see what progress the build command is making, which can be frustrating when it takes a while to complete.

Dockerfile
```diff
@@ -1,59 +1,84 @@
...
+
+# Set up install environment
+RUN mkdir -p ${APP_HOME}
+WORKDIR ${APP_HOME}
+# end
+
+# This must be ordered before rake assets:precompile
+COPY --from=dependencies ${DEPS_HOME}/Gemfile ${APP_HOME}/Gemfile
+COPY --from=dependencies ${DEPS_HOME}/Gemfile.lock ${APP_HOME}/Gemfile.lock
+COPY --from=dependencies ${GEM_HOME} ${GEM_HOME}
```
tahb: I can't see that `GEM_HOME` is defined anywhere, it seems to build without it though so maybe we can remove this line?

Dockerfile
```diff
@@ -1,59 +1,84 @@
...
 RUN gem update --system 3.0.3
 RUN gem update rake 13.0.1
 RUN gem install bundler
 
-# bundle ruby gems based on the current environment, default to production
-RUN echo $RAILS_ENV
-RUN \
-  if [ "$RAILS_ENV" = "production" ]; then \
-    bundle install --jobs 4 --without development test --retry 10; \
+RUN bundle config set frozen 'true'
```
tahb: What does this do?

.travis.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         -e CI=true \
         --env-file .env.example \
         app/test /bin/bash -c "tail -f /dev/null"
-      docker exec test-container rake
+      docker exec test-container bundle exec rake
```
tahb: In theory since we are only installing the dependencies we need in a completely clean environment we should be able to remove all instance of bundle exec. 

If we are prompted to use it to run the command then we have probably installed 2 different versions of the same package somewhere so it would be useful to know that.

I propose that this commit flips direction to remove all cases for consistency (though it was probably me that added it for brakeman, sorry!)

travis-docker-push.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 TAG="${TRAVIS_BUILD_NUMBER}-${TRAVIS_COMMIT}"
 echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
-docker build -t "thedxw/beis-report-official-development-assistance:$TAG" .
+docker build --target release -t "thedxw/beis-report-official-development-assistance:$TAG" .
```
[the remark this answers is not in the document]
mec: Yep, that is my understanding. I've been explicit in `.travis.yml` and used `--target test` but we could leave that out and it seems to go through the stages in order.

Dockerfile
mec: Thanks for your view!

docker-compose.test.yml
mec: Ah! Good shout, I will change that.

Dockerfile
mec: wget is very verbose, that was the reason, but I think your right that more information at this point is better.

.travis.yml
mec: What about different version of dependancies? Will it handle that case - does that case even exsist? I'd like to understand this more.

Dockerfile
mec: This is section is lifted out of Claims Dockerfile, my goal was to get rid of the `--without` as it's deprecated, however that involves the new style bundle config so I kept this, my understanding is that `frozen` tells bunder to not change the Gemfile and that would mean if the Gemfile and .lock don't match we get an error - the other case is the Gemfile getting re-written based off the .lock without us knowing – I will admit my knowledge is weak here!

Looking at the docs, it is a little confusing to me:

https://bundler.io/v2.0/man/bundle-config.1.html

`frozen` seems to imply `deployment` but perhaps not `--deployment` - I'd like to understand this more, this explains things a little better:

https://bundler.io/guides/deploying.html#deploying-your-application

This makes me think that we should drop the `frozen` and instead use `--deployment` on the bundle install (which includes `frozen`) - but then I think we may have to use `bundle exec` everywhere? (see your later comment)

.travis.yml
mec: Also see https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/296#discussion_r400844590

Dockerfile
mec: Interesting! I assumed GEM_HOME is set by ruby itself so the contianer has it already, I also assumed this is where the system gems are installed and that is location the bundler would use!

I am going to have to go and poke the container to see what I think about this one, it may also relate to using (or not using) the `--deployment` options (see below)

docker-compose.test.yml
mec: Spring is something I have little understanding of, it would be good to use this to learn more. I'll definately add the volumes though.

travis-docker-push.sh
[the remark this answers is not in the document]
erbridge: My suggestion is to be explicit. You don't want adjustments to the Dockerfile to have unexpected consequences for which image you're building.

Dockerfile
mec: Looking at it now, I am kind of thinking the whole thing is a bit odd!? 

GEM_HOME does seem to be set, so our container has it as `/usr/local/bundle` if that is the case is the third copy doing a copy from the dependency stage to the current stage which is pointless as it's the same location?

There also seems little point copying the Gemfile and Gemfile.lock as they come along later in line 77 anyway?

.travis.yml
mec: My non-understanding grows wider with this bundler business! 

I can call `bundle` and `rake` as they are both in the path at `/usr/local/bundle/bin/` so any call to `rake` can be sans `bundle exec` - is this sound as the container only has one single version of rake and one 'global' bundle?

I also have a binstub for rake in `/app/bin` but that is part of the code so I don't use it.

`rails` on the other hand is a gem, so must be run via `bundle exec`?

mec: ☝️ I think this is why I run `rails` everywhere!

mec:
One more comment for this today! Regarding the failing spec, I've found this:

https://github.com/omniauth/omniauth/wiki/Integration-Testing#mocking-failure

This seems to be the issue, I can set the `redirect_to_failure` and get a consistent passing of your seed for the testing environment. I am not 100% on this one yet as I don't feel like I fully understand what and why - but I thought I would post the finding anyway.

mec:
@tahb I've pushed a new commit that fixes the failing spec on seed 38144, could you try it on your environment? I am left strugging to understand why this works and how the specs ever pased so I woud like to confirm it works and try to understand what is going on!

I would then appreciate if we could spend some time going over the rest of this together and making some decisions.

travis-docker-push.sh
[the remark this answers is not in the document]
mec: After much discussion with myself and @tahb, we've decided that although we would like the web stage to be the last and so 'default' stage for the image, for those that may want to use it outisde of deployments, the test stage must come after the web stage to work as expected.

We also noticed that although our Docker file is in stages we do not throw any stage away and so some of our code is less useful.

We can however still stop at the web stage and not build the test stage, saving use a small amount of time.

mec:
@tahb lines 74 and 75 in the Dockerfile have to stay - it still feels like the strangest way to add assets to the pipeline to me but, if we don't copy the assets like this the precompile will fail – we could fix this, but I don't think it should be in this work.

This threw up another flickering spec with seed 10466 which I have attempted to address in #307 

I hope I have covered everything else we've dicuessed and this is ready to look at again!

Dockerfile
```diff
@@ -1,54 +1,80 @@
...
+# ------------------------------------------------------------------------------
+FROM dependencies AS web
+
+# Set up install environment
+RUN mkdir -p ${APP_HOME}
+WORKDIR ${APP_HOME}
+# end
+
+# This must be ordered before rake assets:precompile
+COPY --from=dependencies ./deps/node_modules $APP_HOME/node_modules
```
tahb, line 74: Though this works I think this existing bit of code (specifically `--from=depedencies`) risks misleading the reader into thinking this is being copied from another build stage, when this is not the case.

I think a refactor to the following copy might be more honest:

```
RUN cp -R ./deps/node_modules $APP_HOME/node_modules
RUN cp -R ./deps/node_modules/govuk-frontend/govuk/assets $APP_HOME/app/assets
```

mec: Yep, that is a nice middle ground and is more intention revealing - thanks for the suggestion!

PR #297, opened 2020-03-26
Chore/Document database backup and restore

tahb:
## Changes in this PR

As part of preparing the pentest environment for the accessibility audit we have loaded it with real data from production for the first time.

This pull request includes documentation of the process I went through to do it.

I needed an easy way to seed the staging users to I've refactored the existing user seed file into something clearly named. They are both identical apart from the auth0 identifiers, which is not clear to see! 

The main point of interest is around the need to run the import twice. I can't figure out why it always gets stuck on extending organisations not existing, however it's fine creating other records which need dependent data i.e transactions and activities, or organisations and users. Any idea what we could try here?

## Next steps


tahb:
@mec I think that's right there is a foreign key constraint on extending_organisation! https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/db/schema.rb#L123

I looks like we are being inconsistent with our use of constraints. I read over [the documentation on constraints](https://guides.rubyonrails.org/active_record_migrations.html#active-record-and-referential-integrity) and it is phrased in such a way that doing database contraints is not the common "Rails way". It appears to advocate that we rely on the Model to store logic but also acknowledges that this "cannot guarantee referencial integrity" which is the problem we have run into!

I presume that the more constraints we add for different records the more times we would have to run the import command. The instruction to run it twice may not always remain true.

I think we have 2 choices:

1. remove all foreign key constraints at the database level and rely on our Rails models to ensure associations are always present on create
2. change approach and start adding in foreign key constraints in for all associations which would guarantee at the database level. I'm not sure what this would mean for our database import. Perhaps there is a cleverer way to import whilst ignoring constraints?

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/models/activity.rb#L19

tahb:
We might be able to do add a step to modify the sql file in order to reliably skip the database triggers that are causing these (and any future) errors using `session_replication_role`: https://www.postgresqlco.nf/en/doc/param/session_replication_role/

> Since foreign keys are implemented as triggers, setting this parameter to replica also disables all foreign key checks, which can leave data in an inconsistent state if improperly used. 

I'd like to see if this works on a real environment!

tahb:
I've succeeded in importing the prod database into my local environment without this hack using:

```
$ psql 
  > DROP DATABASE "roda-development";
  > CREATE DATABASE "roda-development";
  > CREATE ROLE "rdsbroker_de78448a_1869_4e6c_9ffc_1256420e96f3_manager" WITH SUPERUSER CREATEDB CREATEROLE;

$ psql -d roda-development < DATA_FILE_NAME.sql

$ rake db:prepare
$ rake db:seed
```

I need to add this into this proposal properly and test it on a live environment to see if it works there too.

tahb:
@mec I have successfully tested this on both the pentest environment and locally! I've changed approach slightly for production->staging to remove the repeat `psql < dump.sql` step without having to change the way the service was using constracints (though this may still be a seperate issue). 

I've added local steps as I have already found it useful to verify the production data to find the data integirty problem of not all activities having a value for `geography` 🙌  Please do try it out and let me know how you find it!

doc/database-backup-and-restore.md
```diff
@@ -0,0 +1,106 @@
...
+      > CREATE ROLE "rdsbroker_de78448a_1869_4e6c_9ffc_1256420e96f3_manager" WITH SUPERUSER CREATEDB CREATEROLE;
+    ```
+1. Add production data to the clean local database
+    ```
+    psql -d roda-development < PROD_DATA_FILE_NAME.sql
+    ```
+1. Prepare and seed the database
+    ```
+    rake db:prepare
+    rake db:seed
```
mec, line 102: This will add all our seed data ontop of the prod data right?

doc/database-backup-and-restore.md
```diff
@@ -0,0 +1,106 @@
...
+    cf services # to get the SERVICE_NAME of postgres
+    cf conduit SERVICE_NAME -- pg_dump --file PROD_DATA_FILE_NAME.sql
+    ```
+1. Find the Database role name from the database dump. It should be of the form: `rdsbroker_<uuid>_manager`
+    ```
+    grep -n rdsbroker PROD_DATA_FILE_NAME.sql
+    ```
+1. Use the Postgres binary to wipe and prepare the local database (if you've done this process before you'll get a warning that this role already exists)
+    ```
+    psql
```
mec, line 90: When devs have installed postgres with Homebrew this will often need to be `psql postgres` - don't think we need to do anything, just something I've seen.

doc/database-backup-and-restore.md
tahb: Yeah that's the idea. As the seed will do a find or create it should avoid creating more records than it needs to. The primary reason for this is so that we can sign in without having to create a User record with the right auth0_identifier etc.

tahb:
@mec I added a note to the commit, does this answer your question in enough detail if I was to include that in the PR description too?

> Since the data from the dump was originally owned by a role called rdsbroker_<uuid>_manager we need to set that role in the target database so that it is present before the data that links to it is added. Without this the restore command fails with many role errors.

PR #298, opened 2020-03-26
Ensure that organisation iati_references are unique at the database level

tahb:
## Changes in this PR

During the production database restore to the pentest environment I noticed that if you skip activerecord you can have as many organisations with GB-GOV-13 as you like. This is a pretty important field to be unique for publishing to IATI purposes to lets validate at the database level as well as ActiveRecord.

## Next steps

### DIFF
--- a/db/migrate/20200326173005_add_uniqueness_to_organisation_iati_reference.rb
+++ b/db/migrate/20200326173005_add_uniqueness_to_organisation_iati_reference.rb
@@ -1,0 +1,5 @@
+class AddUniquenessToOrganisationIatiReference < ActiveRecord::Migration[6.0]
+  def change
+    add_index :organisations, :iati_reference, unique: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_03_160846) do
+ActiveRecord::Schema.define(version: 2020_03_26_173005) do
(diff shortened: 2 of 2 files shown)


mec:
@tahb it kind of looks like the build process doesn't run migrations? The CI tests run on this but when it moves on to the build, I cannot see where mirgrations would be triggered?

Have we had any migrations since moving to GPaas? Do we need to add `docker exec rake db:prepare` on line 32 in `.travis.yml`?

tahb:
@mec From what I remember the [Dockerfile is instructed to use a docker-entrypoint.sh that runs before hooks](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/Dockerfile#L58) which are run before the every `CMD` statement. 

The only hook we currently have is to set up the database with a`rake db:prepare`: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/docker-entrypoint.sh.

That's how I'd expect it to be run on CI. CI builds the image, and as part of the `docker run` the database is set up in the background. I'd expect the schema to be loaded rather than any migrations being run as the database is clean.

tahb:
I think this is failing because I forgot to commit the new schema.rb!

tahb:
@lozette good question. We have tests for our Rails uniquness of this field https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/models/organisation.rb#L6. What we don't have is validation when something tries to modify this outside of the Rails context with a direct connection. I'm not sure how we'd test this without doing the same.

PR #299, opened 2020-03-26
Validate that iati references are prefixed with the correct structure

tahb:
## Changes in this PR

I noticed this bug when restoring production data to the pentest environment. UKSA have an invalid identifier of 1234. This makes publishing their data to IATI invalid.

Following the IATI guidance on organisation identifiers: http://reference.iatistandard.org/203/organisation-identifiers/#two-components-of-iati-organisation-identifiers. This change validates that the "namespace" code exists as a prefix and is separated by a hyphen.

Whilst IATI do say that each component of the namespace code MAY also include a hyphen the real world examples we have seen for BEIS do not. It feels like we can layer on this logic later if required.

## Screenshots of UI changes

Screenshot 2020-03-26 at 18 15 49

## Next steps

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -3,7 +3,10 @@ class Organisation < ApplicationRecord
   has_many :funds
 
   validates_presence_of :name, :organisation_type, :language_code, :default_currency
-  validates :iati_reference, uniqueness: {case_sensitive: false}, presence: true
+  validates :iati_reference,
+    uniqueness: {case_sensitive: false},
+    presence: true,
+    format: {with: /\A[a-zA-Z]{2,}-[a-zA-Z]{3}-.+\z/, message: I18n.t("activerecord.errors.models.organisation.attributes.iati_reference.format")}
 
   scope :sorted_by_name, -> { order(name: :asc) }
   scope :delivery_partners, -> { sorted_by_name.where(service_owner: false) }
--- a/config/locales/models_and_forms.en.yml
+++ b/config/locales/models_and_forms.en.yml
@@ -80,6 +80,10 @@ en:
             period_start_date:
               between: Date must be between %{min} years ago and %{max} years in the future
               not_after_end_date: The period start date cannot be after the period end date
+        organisation:
+          attributes:
+            iati_reference:
+              format: Identifiers must start with a country code and company type separated by a dash, eg. GB-GOV-13
         transaction:
           attributes:
             date:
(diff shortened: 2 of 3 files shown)


mec:
You've got a brakeman validation warning on this - it might be causing the CI to fail, it just seems to stop at the point of the warning? I am not sure if a brakeman warning should be a fail or not?

lozette:
I think a brakeman warning should make CI fail. @tahb do you want me to address this, as you're p busy? 

```
Confidence: High
Category: Format Validation
Check: ValidationRegex
Message: Insufficient validation for `iati_reference` using `/\A[a-zA-Z]{2,}-[a-zA-Z]{3}-.+/`. Use `\A` and `\z` as anchors
File: app/models/organisation.rb
Line: 9
```

tahb:
@mec I think a Brakeman warnings should fail a CI run so we can review it?

I did try to add Brakeman to the default `rake` initially but found there were some problems with it and Brakeman advise that we run outside of rake. https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/223/commits/70826badf8e0f6ad06fd5d4798fa6d993cc2bd70

@lozette thank for you for the offer of help! Sadly I only noticed it once I came to looking at this PR again as a hole. I've added in the closing anchor and all looks to be well 🤞

app/models/organisation.rb
```diff
@@ -3,7 +3,10 @@ class Organisation < ApplicationRecord
   has_many :funds
 
   validates_presence_of :name, :organisation_type, :language_code, :default_currency
-  validates :iati_reference, uniqueness: {case_sensitive: false}, presence: true
+  validates :iati_reference,
+    uniqueness: {case_sensitive: false},
+    presence: true,
+    format: {with: /\A[a-zA-Z]{2,}-[a-zA-Z]{3}-.+\z/, message: "Identifiers must start with a country code and company type separated by a dash, eg. GB-GOV-13"}
```
lozette, line 9: You're gonna hate me but..... should this be internationalised?

tahb: Ha, yeah it should be! Thanks for calling me out. I'm on it 🏃

mec:
@tahb yes for sure brakeman should run! Becasue we don't run brakeman in the default task, I thought you may have missed it locally before pushing this! 😄

tahb:
@mec I definitely did miss the brakeman feedback before pushing! A problem I have is that the builds are taking just a little too long now so my attention moves onto the next thing before it completes and provides feedback. Ideally I'd wait and confirm it was all green before expecting anyone else to review it.

PR #303, opened 2020-03-31
(459) IATI reporting org should be BEIS regardless of the user's organisation

lozette:
Trello: https://trello.com/c/CfLUXjCa/459-iati-reporting-org-should-be-beis-regardless-of-the-users-organisation

Funds and programmes should have their reporting org set as BEIS

Projects created by UKSA (our current only delivery partner, and a government organisation) should have their reporting org set as BEIS

Other future DPs will have themselves set as their reporting org (for now).

**NOTE** This PR includes a rake task (can be added as a data migration) to be run on deployment to production.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@
   Javascript is available
 - Add privacy policy to site
 - Empty optional dates for `actual start date` and `actual end date` are not included on the activity XML
+- Reporting org in the IATI XML is always BEIS for funds, programmes and projects created by governmental organisations, and the activity's organisation if it is a non-governmental organisation
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-3...HEAD
 [release-3]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-2...release-3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,7 +92,6 @@
 
 ## [unreleased]
 
-- Activity to Activity association renamed `child_activities` (from `activities`) to avoid an association name clash with the `public_activity` gem
 - When creating an activity the Finance step has been defaulted to `Standard grant` and omitted from the user journey
 - When creating an activity, the `Tied status` step has been removed from the user journey and it has now a default value of `Untied`, code "5"
 - Progressively enhance the country select element into a combo box when
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -23,6 +23,7 @@ class Activity < ApplicationRecord
   belongs_to :organisation
   belongs_to :extending_organisation, foreign_key: "extending_organisation_id", class_name: "Organisation", optional: true
   has_many :implementing_organisations
+  belongs_to :reporting_organisation, foreign_key: "reporting_organisation_id", class_name: "Organisation"
 
   enum level: {
     fund: "fund",
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -10,4 +10,8 @@ class Organisation < ApplicationRecord
 
   scope :sorted_by_name, -> { order(name: :asc) }
   scope :delivery_partners, -> { sorted_by_name.where(service_owner: false) }
+
+  def is_government?
+    %w[10 11].include?(organisation_type)
+  end
 end
--- a/app/services/create_fund_activity.rb
+++ b/app/services/create_fund_activity.rb
@@ -9,6 +9,7 @@ class CreateFundActivity
     activity = Activity.new
     activity.organisation = Organisation.find(organisation_id)
     activity.reporting_organisation = activity.organisation
+    activity.extending_organisation = service_owner
 
     activity.wizard_status = "blank"
     activity.level = :fund
@@ -21,9 +22,13 @@ class CreateFundActivity
     activity.accountable_organisation_reference = "GB-GOV-13"
     activity.accountable_organisation_type = "10"
 
-    activity.extending_organisation = Organisation.find_by!(service_owner: true)
-
     activity.save(validate: false)
     activity
+  end
+
+  private
+
+  def service_owner
+    Organisation.find_by_service_owner(true)
   end
 end
--- a/app/services/create_fund_activity.rb
+++ b/app/services/create_fund_activity.rb
@@ -8,7 +8,7 @@ class CreateFundActivity
   def call
     activity = Activity.new
     activity.organisation = Organisation.find(organisation_id)
-    activity.reporting_organisation_reference = activity.organisation.iati_reference
+    activity.reporting_organisation = activity.organisation
 
     activity.wizard_status = "blank"
     activity.level = :fund
--- a/app/services/create_project_activity.rb
+++ b/app/services/create_project_activity.rb
@@ -13,7 +13,7 @@ class CreateProjectActivity
 
     activity = Activity.new
     activity.organisation = reporting_organisation
-    activity.reporting_organisation_reference = reporting_organisation.iati_reference
+    activity.reporting_organisation = reporting_organisation
 
     programme = Activity.find(programme_id)
     programme.child_activities << activity
--- a/app/services/create_project_activity.rb
+++ b/app/services/create_project_activity.rb
@@ -8,18 +8,10 @@ class CreateProjectActivity
   end
 
   def call
-    service_owner = Organisation.find_by(service_owner: true)
-    reporting_organisation = Organisation.find(organisation_id)
-
     activity = Activity.new
-    activity.organisation = reporting_organisation
-
-    if reporting_organisation.is_government?
-      service_owner = Organisation.find_by(service_owner: true)
-      activity.reporting_organisation = service_owner
-    else
-      activity.reporting_organisation = reporting_organisation
-    end
+    activity.organisation = creating_organisation
+    activity.reporting_organisation = reporting_organisation
+    activity.extending_organisation = creating_organisation
 
     programme = Activity.find(programme_id)
     programme.child_activities << activity
@@ -35,8 +27,21 @@ class CreateProjectActivity
     activity.accountable_organisation_reference = service_owner.iati_reference
     activity.accountable_organisation_type = service_owner.organisation_type
 
-    activity.extending_organisation = reporting_organisation
     activity.save(validate: false)
     activity
+  end
+
+  private
+
+  def service_owner
+    Organisation.find_by_service_owner(true)
+  end
+
+  def creating_organisation
+    Organisation.find(organisation_id)
+  end
+
+  def reporting_organisation
+    creating_organisation.is_government? ? service_owner : creating_organisation
   end
 end
--- a/app/services/create_project_activity.rb
+++ b/app/services/create_project_activity.rb
@@ -13,7 +13,13 @@ class CreateProjectActivity
 
     activity = Activity.new
     activity.organisation = reporting_organisation
-    activity.reporting_organisation = reporting_organisation
+
+    if reporting_organisation.is_government?
+      service_owner = Organisation.find_by(service_owner: true)
+      activity.reporting_organisation = service_owner
+    else
+      activity.reporting_organisation = reporting_organisation
+    end
 
     programme = Activity.find(programme_id)
     programme.child_activities << activity
--- a/db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
+++ b/db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
@@ -12,7 +12,7 @@ class AddReportingOrgAssociationToActivity < ActiveRecord::Migration[6.0]
         else
           organisation
         end
-        activity.save(validate: false)
+        activity.save!
       end
     end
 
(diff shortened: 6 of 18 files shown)


spec/factories/activity.rb
```diff
@@ -21,9 +21,7 @@
     wizard_status { "complete" } # wizard is complete
 
     association :organisation, factory: :organisation
-    before(:create) do |activity|
-      activity.reporting_organisation_reference = activity.organisation.iati_reference
-    end
+    association :reporting_organisation, factory: :beis_organisation
```
tahb, line 24: This feels much better than the old version 🙌

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
```diff
@@ -0,0 +1,6 @@
+class AddReportingOrgAssociationToActivity < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :activities, :reporting_organisation_reference, :string
```
tahb, line 3: I'm nervous that the column us removed as a separate task to the one that preserves the existing data. I'm not sure of the strategy that would allow us to use them both to switch over to the new method?

I think there are at least 2 options to help us. The easier one is probably to use migrations for this to change both schema and data in one:

1. add new column
1. start a database transaction
1. migrate data into new column 
1. end database transaction
1. delete old column

Our down operation can be coded to take the iati_reference from the record and save the string too. 

An advantage of this is that we will not have to remember to run the rask task on each environment in the correct sequencing of migrations. When migrations are run the data is switched over as part of it automatically.

[Here's an example I dug up from MyTVH](https://github.com/dxw/mytvh-web/blob/develop/db/migrate/20140619111622_change_force_changed_passphrase_for_user.rb) which illustrates what I'm thinking of.

app/models/organisation.rb
```diff
@@ -7,4 +7,8 @@ class Organisation < ApplicationRecord
 
   scope :sorted_by_name, -> { order(name: :asc) }
   scope :delivery_partners, -> { sorted_by_name.where(service_owner: false) }
+
+  def is_government_organisation?
```
tahb, line 11: Given the object is of the class `Organisation` and the method is an instance method it will always be called _on_ an organisation. Given this I'd consider chopping the name down to `organisation.is_government?`. Not something that should block this work though!

app/services/create_project_activity.rb
```diff
@@ -13,7 +13,13 @@ def call
 
     activity = Activity.new
     activity.organisation = reporting_organisation
-    activity.reporting_organisation = reporting_organisation
+
+    if reporting_organisation.is_government_organisation?
```
tahb, line 17: This looks like it does the trick 👍 

I noticed that there's some related code that uses `reporting_organisation` for setting the exetending organsiation association too. It feels like we could make the pair a little clearer?

I wondered if the use of private methods could help break this down to make it more readable, what do you think?

```
private 
def creating_organisation
  Organisation.find(organisation_id)

def service_owner_organisation
  Organisation.find_by(service_owner: true)

def extending_organisation
  creating_organisation

def reporting_organisation
  if managing_organisation.is_government_organisation?
    service_owner_organisation
  else 
    managing_organisation
```

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
lozette: Good call 👍

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    remove_column :activities, :reporting_organisation_reference, :string
+  end
+
+  def down
+    add_column :activities, :reporting_organisation_reference, :string
+
+    ActiveRecord::Base.transaction do
+      activities = Activity.where(reporting_organisation_reference: nil)
```
tahb, line 27: As we've just added that column I'm not sure we need this `where` query. Perhaps instead we do want to only run this code on activities that have an associated `reporting_organisation` present?

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
```diff
@@ -0,0 +1,37 @@
...
+      service_owner = Organisation.find_by_service_owner(true)
+
+      activities.each do |activity|
+        organisation = activity.organisation
+        activity.reporting_organisation = if organisation.is_government?
+          service_owner
+        else
+          organisation
+        end
+        activity.save(validate: false)
```
tahb, line 16: I _think_ that with `save` (without a bang) and `validate: false` being used here the value a transaction block provides is reduced. It might only be helpful now if there are connection problems with the database?

It would be great if we could assume that when working with our records that they are always in a valid state and change this to: `activity.save` 

This would help ensure that the records stay valid and help us prevent introduced new problems unwittingly by using `validate: false`. Is there a case or database state you've seen that has made you cautious? Maybe we can restore that integrity first separately so that this can run with confidence. If we went to each environment and asked all activities if they were `valid?` would that give us that confidence?

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
```diff
@@ -0,0 +1,37 @@
+class AddReportingOrgAssociationToActivity < ActiveRecord::Migration[6.0]
+  def up
+    add_reference :activities, :reporting_organisation, type: :uuid, foreign_key: {to_table: :organisations}
+
+    ActiveRecord::Base.transaction do
+      activities = Activity.where(reporting_organisation: nil)
```
tahb, line 6: Given the association is created a few lines about I expect this will always be nil. I presume this is being used as a guard in case we run the migration more than once, which I'm not sure we will be doing in real environments.

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
lozette: I used `validate: false` because we save Activities to the database which are incomplete. So if, for example, we were updating an Activity which hadn't get been completed (e.g. if the user hadn't yet added a geography or whatever) then the Activity reporting organisation would still be saved even if it's missing a geography. 

If you don't think there will be any incomplete activities in prod then I suppose we can leave it out, but I worry if there are any, we will end up with activities without reporting orgs (because there is no way to add a reporting org via the UI)

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
lozette: Would `activities = Activity.where.not(reporting_organisation_id: nil)` work here?

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
tahb: Ah I see what you mean. Hopefully we'll still be able to use the `wizard_status` in the conditional validations to allow us to update those activities?

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
tahb: Yes 🙌

db/migrate/20200331111947_add_reporting_org_association_to_activity.rb
lozette: I'm not sure... I added the `validate: false` because I had this problem running the migration locally! Some Activities wouldn't update because they were missing attributes.

tahb: I've replicated data from production to check what this is. You're right it is missing `geography`, we must have forgot to migrate the data when adding that field. I will create a migration that if merged first will allow us to use `save!` here.

There was a further problem with a hole in our conditional validation. The identifier field is required and must be unique, these are both being run with the first wizard_status: :identifier. I think we can fix this conditional validation as a separate problem by adding a new "blank" step. This will allow us to call `save!` here!

tahb: 1. https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/310

tahb: 1. https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/311

lozette:
Dependent on #310 #311

spec/services/create_project_activity_spec.rb
```diff
@@ -2,9 +2,10 @@
 
 RSpec.describe CreateProjectActivity do
   let(:beis) { create(:beis_organisation) }
-  let(:delivery_partner_organisation) { create(:delivery_partner_organisation) }
+  let(:delivery_partner_organisation) { create(:delivery_partner_organisation, organisation_type: organisation_type) }
   let(:user) { create(:administrator, organisation: delivery_partner_organisation) }
   let(:programme) { create(:programme_activity, organisation: beis) }
+  let(:organisation_type) { 10 }
```
mec, line 8: Do we need to `let` this? I can't see it used again? Could it be set explicitily? There are already loads of let here…

````
create(:delivery_partner_organisation, organisation_type:  10)
````

File: app/services/create_fund_activity.rb [hunk 9]
```diff
@@ -21,9 +22,13 @@ def call
...
-    activity.extending_organisation = Organisation.find_by!(service_owner: true)
-
     activity.save(validate: false)
     activity
   end
+
+  private
+
+  def service_owner
+    Organisation.find_by_service_owner(true)
```
mec, line 32: @tahb @MariluzHerrera @lozette : Thinking out load/question: Could `service_owner` be a class method on `Organisation` - it's like a constant and I keep looking for a nicer way to express it, my undertanding is that then we could do things like:

````
activity.extending_organisation = Organisation.service_owner
````
which is quite nice in my mind…

File: app/services/create_fund_activity.rb [hunk 10]
```diff
@@ -21,9 +22,13 @@ def call
     activity.accountable_organisation_reference = "GB-GOV-13"
     activity.accountable_organisation_type = "10"
 
-    activity.extending_organisation = Organisation.find_by!(service_owner: true)
```
mec, line 24: @lozette I would ❤️ to see why we wanted this in the private method in the commit message - I could learn so much about style.

CHANGELOG.md
```diff
@@ -92,7 +92,6 @@
 
 ## [unreleased]
 
-- Activity to Activity association renamed `child_activities` (from `activities`) to avoid an association name clash with the `public_activity` gem
```
mec, line 95: Oop! I think I've already seen this back in 815c816cb3cfa21d013769d9a6ad2c92389ebc51

lozette: Oops, merge mistake!

app/models/activity.rb
```diff
@@ -23,6 +23,7 @@ class Activity < ApplicationRecord
   belongs_to :organisation
   belongs_to :extending_organisation, foreign_key: "extending_organisation_id", class_name: "Organisation", optional: true
   has_many :implementing_organisations
+  belongs_to :reporting_organisation, foreign_key: "reporting_organisation_id", class_name: "Organisation"
```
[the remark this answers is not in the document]
lozette: I think mine are even worse tbh, I think I added this because it was a cut & paste from elsewhere.

spec/services/create_project_activity_spec.rb
lozette: No I think this is left over from a refactor!

app/services/create_fund_activity.rb
lozette: @tahb requested similar for projects here https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/303#discussion_r401512422 and I decided to go ahead and do it for funds as well 😆  Not sure if it was the right thing to do.

app/models/activity.rb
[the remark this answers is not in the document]
mec: @tahb I think this is the issue you refer to in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/297#issuecomment-610472101 - before we introduce another db constraint shall we talk it over - now feels like the time to sort this if we switch to a solution in the app code?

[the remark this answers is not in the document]
tahb: I definitely agree. Now that we're about to add our second constraint is a great time to review what our approach is. I think the right thing to do is to add constraints in rather than take them away. I will review and start a conversation with you both as soon as I can.

app/services/create_fund_activity.rb
tahb: I think it would be nice to be able to say that however whilst it is technically possible to have more than one organisations marked as a `service_owner` then I think this approach would risk implying there is more security than there is and lead to unexpected problems in the future.

Getting to a state where the system can only have a single service owner would be great. It seems like we need a way to represent a new type of resource that quacks in almost entirely the same way as an organisation - `ServiceOwner`? Perhaps single table inheritance _would be_ appropriate as a way to help with this one?

app/models/activity.rb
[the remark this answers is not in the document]
tahb: I've opened a pull request and started a conversation here https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/321. I propose that this work is blocked pending the outcome of that convo?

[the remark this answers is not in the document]
lozette: Yeah it will be. This PR is going to haunt me forever!!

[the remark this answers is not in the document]
tahb: Mec isn't working today and it'd be great to unblock this. Suggesting we merge this as 2 of us feel adding constraints is the better way to go. The cost to reverting that decision to go the other way is low since removing constraints is much easier than adding them!

PR #309, opened 2020-04-02
(434) User actions are auditable

lozette:
## Changes in this PR

Trello: https://trello.com/c/jEaQCbZp/434-user-actions-are-auditable

Part of the Service Standard stipulates the service should be logging what changes were made by which user. 

We have previously used the `public_activity` gem to track changes in the Teaching Vacancies service for the DfE. 

This work has been somewhat tricky as `public_activity` reserves the association `activities` on models to log CUD events. We have configured the gem to use a database table called `auditable_events` instead of `activities`. Unfortunately this does not rename the model association. 

On the models we have refactored the Activity association to be called `parent_activity`, so a Budget now has a `parent_activity` which is an Activity ("our" Activity), and an association called `activities` which is a list of PublicActivity::Activities

## Next steps


db/schema.rb
```diff
@@ -10,7 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_31_085103) do
+ActiveRecord::Schema.define(version: 2020_03_30_130958) do
```
erbridge: This looks like it's going earlier rather than later. Is that correct?

app/models/activity.rb
```diff
@@ -120,8 +123,8 @@ def has_implementing_organisations?
   end
 
   def parent_activities
-    return [activity] if programme?
-    return [activity.activity, activity] if project?
+    return [parent_activity] if programme?
```
erbridge, line 126: It's annoying we don't have control over these names. Can we maybe wrap the model mixin in one of our own that aliases `activity` to `event` instead, and disables public access to `activity`? I think the rename to `parent_activity` is a good one regardless, because it describes the relationship more clearly.

db/schema.rb
lozette: That'll be my mashed brain doing a bad rebase 😆  This PR has not been kind to me

spec/rails_helper.rb
```diff
@@ -10,6 +10,10 @@
 # Add additional requires below this line. Rails is not loaded until this point!
 require "mock_redis"
 
+# Testing for public_activity gem
+require "public_activity/testing"
+PublicActivity.enabled = false
```
tahb, line 15: Nice idea to turn if off until we need it for testing. It will be creating a lot of records otherwise!

db/migrate/20200407140854_rename_budget_activity_id_to_parent_activity_id.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    ActiveRecord::Base.transaction do
+      Budget.all.each do |budget|
+        budget.parent_activity = budget.activity
+        budget.save!
+      end
+    end
+
+    remove_column :budgets, :activity_id
+  end
```
tahb, line 14: I wonder how this is going to play now that the database contraints have gone in ahead of this. Do we need to add_foreign_key to the new column?

spec/features/staff/users_can_create_a_budget_spec.rb
```diff
@@ -17,6 +17,24 @@
...
+          click_on(fund_activity.title)
+
+          click_on(I18n.t("page_content.budgets.button.create"))
+
+          fill_in_and_submit_budget_form
+
+          budget = Budget.last
+          auditable_event = PublicActivity::Activity.find_by(trackable_id: budget.id)
+          expect(auditable_event.key).to eq "budget.create"
+          expect(auditable_event.owner_id).to eq user.id
```
tahb, line 35: Is it possible to `expect` to see that the event connects to the budget record? I can see it goes in as a param but I think the important thing is we can get access to it down the road.

db/migrate/20200409133152_rename_transaction_activity_id_to_parent_activity_id.rb
```diff
@@ -0,0 +1,29 @@
...
+    add_index :transactions, :parent_activity_id
+
+    ActiveRecord::Base.transaction do
+      Transaction.all.each do |transaction|
+        transaction.parent_activity = transaction.activity
+        transaction.save!
+      end
+    end
+
+    remove_column :transactions, :activity_id
```
tahb, line 13: Similar point to Budgets around the foreign key constraints.

db/migrate/20200407140854_rename_budget_activity_id_to_parent_activity_id.rb
lozette: I think I've amended this correctly but it'd be good if you can scrutinise it for me to make sure I've got it right

tahb: It looks like the foreign key constraint for the new field is ok. I wonder if we need to remove the old foreign key constraint? 

An idea could be to checkout develop along with setting the correct database state. Then checkout your branch and apply this migration. If that works up, and then on a rollback we're likely ok 👍

db/migrate/20200408173214_add_foreign_key_constraints.rb
```diff
@@ -12,7 +12,8 @@ def change
     # add_foreign_key :activities, :organisations, column: "reporting_organisation_id"
 
     add_foreign_key :transactions, :activities, on_delete: :cascade
-    add_foreign_key :budgets, :activities, on_delete: :cascade
+    # Covered in 20200407140854
+    # add_foreign_key :budgets, :activities, on_delete: :cascade
```
tahb, line 16: This has already been applied on all environments so we probably shouldn't retrospectively edit it as this migration will never be rerun. We should assume it's already there and in our new migration approach it as a change.

db/schema.rb
```diff
@@ -121,9 +140,8 @@
 
   add_foreign_key "activities", "activities"
   add_foreign_key "activities", "organisations", column: "extending_organisation_id"
-  add_foreign_key "activities", "organisations", column: "reporting_organisation_id"
```
tahb, line 124: Odd that this unrelated bit was removed, oddness with database state during a rebase?

db/migrate/20200408173214_add_foreign_key_constraints.rb
lozette: Yes I wasn't sure about this! I will remove it

app/models/activity.rb
lozette: I'd like to do it but I'm not entirely sure how either 😬

PR #314, opened 2020-04-06
(461) Add privacy policy

mec:
## Changes in this PR

The content has been through reviews with BEIS colleagues.

Uses the `<address>` element for better page semantics, which meant adding
a stylesheet.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address

## Screenshots of UI changes
Screenshot_2020-04-06 Privacy policy — Report Official Development Assistance

## Next steps

### DIFF
--- a/app/views/public/privacy_policy/index.html.haml
+++ b/app/views/public/privacy_policy/index.html.haml
@@ -1,0 +1,225 @@
+=content_for :page_title_prefix, t("page_title.privacy_policy")
+
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds.app-visitor-welcome
+      %h1.govuk-heading-xl
+        = t("page_title.privacy_policy")
+
+      %p.govuk-body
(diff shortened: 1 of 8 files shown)


mec:
[Privacy policy - Users can see how their data is stored and used by the service](https://trello.com/c/eJZGOuWu/461-privacy-policy-users-can-see-how-their-data-is-stored-and-used-by-the-service)

app/views/public/privacy_policy/index.html.haml
```diff
@@ -0,0 +1,225 @@
...
+        %br
+        United Kingdom
+
+        %p.govuk-body
+          Email:
+          = link_to "enquiries@beis.gov.uk", "mailto:enquiries@beis.gov.uk", class: "govuk-link"
+
+        %p.govuk-body
+          Telephone:
+          = link_to "20 7215 5000", "tel:+442072155000", class: "govuk-link"
```
tahb, line 210: `20 7215 5000` is not recognised when I ring it. I think we need to include an additional zero here: 020 7215 5000

mec: Great spot! Thanks!

PR #315, opened 2020-04-06
(463) Country uses accessible autocomplete

mec:
## Changes in this PR
Progressively enhnace the Country select element into a combo box.

A user can type the name of a country and see a list of matching countries or click the element and select a country from the list available.

When JavaScript is not available, the select element is shown.

We felt this was a good trade off in terms of making selecting a country simpler for a user vs how many users may not be able to run javascript.

As mentioned in the commit message two desicions where made to keep this work simple:

1. use the ugliyfied css and js so no build step is needed
2. load the script inline on the view that uses it, as there is only one

It's worth noting that the country field is still not validated correctly under certain conditions, but this was not part of this work.

## Screenshots of UI changes

autocomplete

## Next steps


mec:
[Use accessible autocomplete for country selection and reduce the country list](https://trello.com/c/1CKR66RS/463-use-accessible-autocomplete-for-country-selection-and-reduce-the-country-list)

spec/features/public/visitors/javascript_detection_spec.rb
```diff
@@ -1,15 +0,0 @@
-RSpec.feature "JavaScript detection" do
```
tahb, line 1: If we're removing the test do we need to have the `js-enabled`  class remain in the layout, or is that used as a conventional way to communicate with other tooling such as screen readers?

app/assets/stylesheets/partials/_autocomplete.scss
```diff
@@ -0,0 +1,12 @@
+// the autocomplete does not set a font, this only effects the autocomplete
+// items, so we set it here
+li.autocomplete__option {
+  @include govuk-font($size: 19);
```
tahb, line 4: I think a problem remains with the fonts. I believe the option shown in the list below is correct but the 'selected' option is using something different. It's a close match but the bottom of the "l" character is different.

Screenshot 2020-04-08 at 11 05 45

When I inspect the 'selected' element Firefox says that it's something other than GDS transport.

Screenshot 2020-04-08 at 11 17 00

I'm surprised by this type of problem in a GDS package. It looks like their [demo page](https://alphagov.github.io/accessible-autocomplete/examples/) has the same issue though so I'm not sure it would be a quick fix.
Screenshot 2020-04-08 at 11 19 40

spec/features/public/visitors/javascript_detection_spec.rb
mec: My view is that it is part of the GOVUK frontend template so we should keep it, but happy either way:

https://design-system.service.gov.uk/styles/page-template/

app/assets/stylesheets/partials/_autocomplete.scss
mec: Wow! Great spot! I'll take a look - yeah not sure why it is like this.

mec:
There seems to be an issue with the Accessible autocomplete dropdown arrow and the GOVUK Frontend, see:

https://github.com/alphagov/accessible-autocomplete/issues/351

We've used the work-around while we look into a long-term fix.

PR #318, opened 2020-04-07
Empty optional dates are not included in XML

MariluzHerrera:
## Changes in this PR

Trello: https://trello.com/c/2LJYjzhh

This PR fixes a bug we detected while validating XMLs of activities on the IATI validator. When a user creates an activity, they're prompted to input start dates and end dates as part of the journey. The `planned dates` are mandatory, but the `actual dates` are optional. 

Before this change, the XML file would include the fields for the optional dates (`actual_start_date` and `actual_end_date`) even when these dates have been left blank by the user. This situation would case the IATI validator to throw an error on validation of this XML file, as it would encounter two fields with not data on them. Now, the optional dates fields that are left blank by the user will not be included on the XML and the validation will not fail as a result. 

I added tests to assert this functionality and I have checked several XML files agains IATI validator. They have all passed validation.

## Next steps

### DIFF
--- a/spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
+++ b/spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
@@ -42,6 +42,26 @@ RSpec.feature "Users can view an activity as XML" do
         end
       end
 
+      context "when the activity does not have actual dates (optional dates)" do
+        let(:activity) {
+          create(:fund_activity,
+            organisation: organisation,
+            identifier: "IND-ENT-IFIER",
+            actual_start_date: nil,
+            actual_end_date: nil)
+        }
+        let(:xml) { Nokogiri::XML::Document.parse(page.body) }
+
+        it "does not include empty optional dates" do
+          visit organisation_activity_path(organisation, activity, format: :xml)
+          optional_start_date = xml.at("iati-activity/activity-date[@type = '2']")
+          optional_end_date = xml.at("iati-activity/activity-date[@type = '4']")
+
+          expect(optional_start_date).to be_nil
+          expect(optional_end_date).to be_nil
+        end
+      end
+
       context "when the activity is a fund activity" do
         let(:activity) { create(:fund_activity, organisation: organisation, identifier: "IND-ENT-IFIER") }
         let(:activity_presenter) { ActivityXmlPresenter.new(activity) }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,7 @@
 - Progressively enhance the country select element into a combo box when
   Javascript is available
 - Add privacy policy to site
+- Empty optional dates for `actual start date` and `actual end date` are not included on the activity XML
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-3...HEAD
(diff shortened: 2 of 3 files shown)


spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -42,6 +42,27 @@
...
 
+      context "when the activity does not have actual dates (optional dates)" do
+        let(:activity) {
+          create(:fund_activity,
+            organisation: organisation,
+            identifier: "IND-ENT-IFIER",
+            actual_start_date: nil,
+            actual_end_date: nil)
+        }
+        let(:activity_presenter) { ActivityXmlPresenter.new(activity) }
```
mec, line 53: Do we need this? I can't see where we use it?

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -42,6 +42,27 @@
...
+        let(:xml) { Nokogiri::XML::Document.parse(page.body) }
+
+        it "does not include empty optional dates" do
+          visit organisation_activity_path(organisation, activity, format: :xml)
+          optional_start_date = xml.at("iati-activity/activity-date[@type = '2']")
+          optional_end_date = xml.at("iati-activity/activity-date[@type = '4']")
+
+          expect(optional_start_date.present?).to eq(false)
```
mec, line 61: Given your `xml.at` returns nil to pass, what do you think about loosing the `present?` and being explicit? Something like:

```` ruby
expect(optional_start_date).to be_nil
````

MariluzHerrera: That's a really good point! This was my first full-on TDD experience on my own and I struggled to create the test. I followed other examples and then I added what I needed, but definitely I could've done my life easier as you suggest. I didn't consider that nil will be the same than "not present". Happy to make this change as I think it's easier to read

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
MariluzHerrera: I wasn't sure whether I need the activity xml presenter. I thought since we are checking if the info is displayed in the xml we would need it, but that's due to my lack of understanding with presenters (I think I mentioned this to you in the past). Thanks for the suggestion

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
mec: Well, I think you are right `nil.present?`does equal false, I just wonder if it is needed. 

This xml stuff is hard - you've done a great job on your own! 👍

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
mec: No worries, as a general rule-of-thumb we only want 'just enough' to get the test setup. 

As this is a feature test it is excercising the whole stack, we are making a request, returning an entire page of xml and then looking for a specific element, that process will take care of any presenter needed so you shouldn't have to.

MariluzHerrera:
> This looks great, exactly how I would have done it (if that's a good endorsement 🤣 )

It definitely is @lozette!!! Thank you! :-)

PR #319, opened 2020-04-08
(WIP) CI uses docker-compose

mec:
## Changes in this PR
This is an attempt to use `docker-compose` in CI to build and run the tests. The motivation for this is seeing the test suite fail in CI becasue Redis is not available.

The way we were using `docker run` could potentially cause a race condition where the tests run before Redis has started.

Whilst switching to docker-comopse [cannot remove the risk](https://docs.docker.com/compose/startup-order/) it would be interesting to see if this makes any difference. 

The complete fix is to check for redis before moving on to start the application, [wait-for-it](https://github.com/vishnubob/wait-for-it) looks like a zero dependancy way to do that and I may take a look if we think it is worthwhile?

I think we should be able to build the image, run the tests with docker-compose and then push the same image to docker hub for deployment - this would save us building the image twice, but to do that we would have to change our multistage build so `web` is the last stage and we would have to install the testing dependancies on the deployed image, @tahb this is something we have spoken about in the past.

I'd be interested to hear folks views on this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,25 @@
 language: minimal
-services:
-- docker
 before_install:
-- docker build . --target test --build-arg RAILS_ENV=test -t app/test
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+  - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+  - sudo apt-get update
+  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
+  - docker -v
 jobs:
   include:
   - stage: Test
-    name: Build and test the current version of the application using Docker
-    env:
-    - RAILS_ENV=test
+    name: Build and test the current version of the application using Docker compose
     script: |
-      set -eu
-      docker network create test
-      docker run -d --name pg --network test -p 5432:5432 postgres:11.6
-      docker run -d --name redis --network test -p 6379:6379 redis:4.0.14
-      docker run -d \
-        --name test-container \
-        --network test \
-        -e RAILS_ENV=test \
-        -e DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true \
-        -e DATABASE_URL=postgres://postgres@pg:5432/roda_test \
-        -e REDIS_URL=redis://redis:6379 \
-        -e TRAVIS="$TRAVIS" \
-        -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" \
-        -e TRAVIS_BRANCH="$TRAVIS_BRANCH" \
-        -e TRAVIS_PULL_REQUEST="$TRAVIS_PULL_REQUEST" \
-        -e COVERALLS_SERVICE_NAME=travis-ci \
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,37 +1,20 @@
...
-        --env-file .env.test \
-        app/test /bin/bash -c "tail -f /dev/null"
-      docker exec test-container bundle exec rake
-      docker exec test-container bundle exec brakeman
+      echo "==== build image ===="
+      TAG="${TRAVIS_BUILD_NUMBER}-${TRAVIS_COMMIT}"
+      docker build --target test -t "thedxw/beis-report-official-development-assistance:$TAG" .
+      echo "==== testing app ====" 
+      docker-compose -f docker-compose.ci.yml up --detach
+      docker exec test-container /bin/bash -c "./wait-for-it.sh redis:6379 --strict -- bundle exec rails db:prepare"
```
tahb, line 14: Having to add a new shell script to wait for Redis to be started feels like quite heavy handed. I can't imagine we're the only ones with this issue, I've used docker compose to manage live services like our dashboards and have never had a need for it - I'm concerned there is another issue here and this is a plaster.

.travis.yml
```diff
@@ -1,37 +1,20 @@
...
+      docker exec test-container /bin/bash -c "bundle exec standardrb"
+      docker exec test-container /bin/bash -c "bundle exec rspec"
```
tahb, line 16: I think we should continue to use `rake` as our single command that runs our test commands by default. A single place means we are less likely to forget to come and update this script whenever we change this definition. 

I understand that Brakeman is an odd case that can't follow our existing model, which is a shame.

Dockerfile
```diff
@@ -95,7 +95,7 @@ CMD ["bundle", "exec", "puma"]
 # ------------------------------------------------------------------------------
 FROM web as test
 
-RUN apt-get install -qq -y firefox-esr
+RUN apt-get update && apt-get install --fix-missing -qq -y firefox-esr
```
tahb, line 98: What did adding `--fix-missing` fix?

wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
tahb, line 1: Adding 182 lines of bash makes me feel uneasy as there is a lot of stuff here for us to check and for us to look after as direct attributions to our code base.

.travis.yml
tahb, line 14: `db:prepare` should be being run within the docker-entrypoint.sh when every container starts so I don't think we need this

mec: Yeah, it's a fair point. I don't think we have to, it was one of the recommended ways from the Docker docs. I am still interested to see if simply using compose will help on it's own.

mec: The entrypoint is overridden in the compse file so we do have to run this.

Dockerfile
mec: From memory (it's been a whilte since I looked at this!) the firefox install stopped working, I can only assume it was after an update to the package. The fix may have been running apt-get update first and the fix-missing was just a hangover from trying to diagnose the issue.

mec:
This branch was an experiment so I will close it now.

PR #329, opened 2020-04-14
(549) Activities can have a previous identifier

mec:
When an activity has already been published to IATI the identifier must
not change. As we are now standardising the identifier we may have to
use two identifiers for some activities.

As existing activities will be imported, do not expose the previous
identifier attribute to users.

Where an activity has a previous identifier use it as the
`iati-identifier` in the xml and output the new identifier using the
`other-identifier` attribute, see:

http://reference.iatistandard.org/203/activity-standard/iati-activities/iati-activity/other-identifier/

I did wonder if we should include the `owner-org` for the `other-identifier` and set it to BEIS, but I won't let that block this work.

http://reference.iatistandard.org/203/activity-standard/iati-activities/iati-activity/other-identifier/owner-org/

## XML validation
https://stage.dataworkbench.io/validate/e47224d5-0eca-4fd0-8199-6e98032c99aa

## Next steps

### DIFF
--- a/spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
+++ b/spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
@@ -7,6 +7,30 @@ RSpec.feature "Users can view an activity as XML" do
     context "when the user belongs to BEIS" do
       before { authenticate!(user: user) }
 
+      context "when the activity has a previous activity identifier" do
+        let(:activity) {
+          create(:fund_activity,
+            organisation: organisation,
+            identifier: "IND-ENT-IFIER",
+            previous_identifier: "PREV-IND-ENT-IFIER")
+        }
+        let(:xml) { Nokogiri::XML::Document.parse(page.body) }
+
+        it "shows the previous identifier as the actvitiy identifier" do
+          visit organisation_activity_path(organisation, activity, format: :xml)
+
+          expect(xml.at("iati-activity/iati-identifier").text).to eq(activity.previous_identifier)
+        end
+
+        it "shows the activity identifier as the other identifier" do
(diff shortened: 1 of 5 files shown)


mec:
[Add a field to activities to support tracking previous versions of an activity on IATI](https://trello.com/c/BjNMFPJj/549-add-a-field-to-activities-to-support-tracking-previous-versions-of-an-activity-on-iati)

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -7,6 +7,30 @@
...
+
+        it "shows the previous identifier as the actvitiy identifier" do
+          visit organisation_activity_path(organisation, activity, format: :xml)
+
+          expect(xml.at("iati-activity/iati-identifier").text).to eq(activity.previous_identifier)
+        end
+
+        it "shows the activity identifier as the other identifier" do
+          iati_identifier = ActivityXmlPresenter.new(activity).iati_identifier
+
```
MariluzHerrera, line 27: I think I'm understanding more and more the uses of the presenters thanks to examples like this one :-)

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -7,6 +7,30 @@
     context "when the user belongs to BEIS" do
       before { authenticate!(user: user) }
 
+      context "when the activity has a previous activity identifier" do
+        let(:activity) {
+          create(:fund_activity,
+            organisation: organisation,
+            identifier: "IND-ENT-IFIER",
```
tahb, line 14: For ingested records these fields are the same thing. What was the intent behind how we could populate the `identifier` field in this case?

PR #332, opened 2020-04-16
(520) BEIS users only see an organisation's activities on that organisation's show page

lozette:
## Changes in this PR

Trello: https://trello.com/c/XhZ45G6F/520-beis-users-only-see-an-organisations-activities-on-the-organisation-show-page

BEIS users have the right to view all activities, no matter which organisation they belong to.
The can also view any organisation's show page.

However, because of a bug, BEIS user's would see *all* activities on any organisation's
show page, because the list of activities is not filtered by organisation.

Now, when a BEIS user views an organisation's show page, they will only see activities
which belong to that organisation. That is, fund and projects which belong to the organisation,
and programmes which have the organisation as its extending organisation.

When a BEIS user views the BEIS organisation, they will see *all* activities belonging
to all organisations.

Non-BEIS users will see the same activities on their organisation show page as they
did before.

## Next steps

### DIFF
--- a/app/controllers/staff/organisations_controller.rb
+++ b/app/controllers/staff/organisations_controller.rb
@@ -12,15 +12,9 @@ class Staff::OrganisationsController < Staff::BaseController
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
 
-    if organisation.service_owner
-      fund_activities = policy_scope(Activity.funds, policy_scope_class: FundPolicy::Scope).includes(:organisation).order("created_at ASC")
-      programme_activities = policy_scope(Activity.programme, policy_scope_class: ProgrammePolicy::Scope).includes(:organisation).order("created_at ASC")
-      project_activities = policy_scope(Activity.project, policy_scope_class: ProjectPolicy::Scope).includes(:organisation).order("created_at ASC")
-    else
-      fund_activities = policy_scope(Activity.funds, policy_scope_class: FundPolicy::Scope).includes(:organisation).where(organisation: organisation).order("created_at ASC")
-      programme_activities = policy_scope(Activity.programme, policy_scope_class: ProgrammePolicy::Scope).includes(:organisation).where(extending_organisation: organisation).order("created_at ASC")
-      project_activities = policy_scope(Activity.project, policy_scope_class: ProjectPolicy::Scope).includes(:organisation).where(organisation: organisation).order("created_at ASC")
-    end
+    fund_activities = FindFundActivities.new(organisation: organisation, current_user: current_user).call
+    programme_activities = FindProgrammeActivities.new(organisation: organisation, current_user: current_user).call
+    project_activities = FindProjectActivities.new(organisation: organisation, current_user: current_user).call
 
     @fund_activities = fund_activities.map { |activity| ActivityPresenter.new(activity) }
     @programme_activities = programme_activities.map { |activity| ActivityPresenter.new(activity) }
--- a/app/controllers/staff/organisations_controller.rb
+++ b/app/controllers/staff/organisations_controller.rb
@@ -12,14 +12,19 @@ class Staff::OrganisationsController < Staff::BaseController
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
 
-    fund_activities = policy_scope(Activity.funds).includes(:organisation).where(organisation: organisation).order("created_at ASC")
+    if organisation.service_owner
+      fund_activities = policy_scope(Activity.funds, policy_scope_class: FundPolicy::Scope).includes(:organisation).order("created_at ASC")
+      programme_activities = policy_scope(Activity.programme, policy_scope_class: ProgrammePolicy::Scope).includes(:organisation).order("created_at ASC")
+      project_activities = policy_scope(Activity.project, policy_scope_class: ProjectPolicy::Scope).includes(:organisation).order("created_at ASC")
+    else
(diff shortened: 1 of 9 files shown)


app/controllers/staff/organisations_controller.rb
```diff
@@ -12,14 +12,19 @@ def show
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
 
-    fund_activities = policy_scope(Activity.funds).includes(:organisation).where(organisation: organisation).order("created_at ASC")
-    @fund_activities = fund_activities.map { |activity| ActivityPresenter.new(activity) }
+    if organisation.service_owner
+      fund_activities = policy_scope(Activity.funds).includes(:organisation).order("created_at ASC")
+      programme_activities = policy_scope(Activity.programme, policy_scope_class: ProgrammePolicy::Scope).includes(:organisation).order("created_at ASC")
```
MariluzHerrera, line 17: Hi @lozette! I was having a look at your PR (I'm practising code review today :-) ) and I have a question about the line above. Why do funds not have their own class `FundPolicy::Scope` ? Is it because they are the main "wrappers" for programmes and projects?

lozette: Uhhh do you know what I have absolutely no idea! The change was from @mec 's commit here https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/c3684969dc312580bd7be6ce47594f7f9e30da28#diff-432213d9aea29cede7ca815df9804248 - @mec can you shed some light?

mec: Well, funds do have a policy, and probably should use their own policy scope here - otherwise they are using the base `ActivityPolicy` which may not be the best. 

Looking at `ActivityPolicy` we have a lot of ugly logic in there that using a class would remove, it also looks like the scope includes all funds for administrator users (everyone) which may not be correct whereas the `FundPolicy` includes funds only for BEIS users, which seems more resaonable.

I haven't looked at this PR apart from your question so I don't know if this line introduces this issue or if we are just following what has come berfore, but either way it's a great spot and a great question! :D

lozette: I was following what came before but I'll amend the PR to use the Fund `policy_scope_class` and see if anything breaks!!

app/controllers/staff/organisations_controller.rb
```diff
@@ -12,14 +12,19 @@ def show
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
 
-    fund_activities = policy_scope(Activity.funds).includes(:organisation).where(organisation: organisation).order("created_at ASC")
-    @fund_activities = fund_activities.map { |activity| ActivityPresenter.new(activity) }
+    if organisation.service_owner
```
tahb, line 15: Before we added this `if` condition there was already quite a lot of logic involved in finding funds, programmes and projects: scoping, filtering and ordering. It looked like a good candidate for some refactoring.

Duplicating the majority of this logic and omitting the part that makes it behave differently is one way to solve this problem but I do worry about the ease of working with it again in future. I think what would make this code easier to read and work with would be to defer these responsibilities out of the controller and into a class. Classes would help us reuse the sorting behaviour so that we would only have to change it in one place, etc.


Some ideas:

```
fund_activities = FindActivities.new(organisation: organisation).call(level: :fund)

# or

fund_activities = FindFunds.new(organisation: organisation).call
programme_activities = FindProgrammes.new(organisation: organisation).call
project_activities = FindProjects.new(organisation: organisation).call
```

---

There is a larger peice of debt we have here. The way we treat the BEIS organisation as an `Organisation` that's only distinguishable by `service_owner` is still a problem. It's forcing us to add logic like this to make it behave differently throughout. I think creating a new `ServiceOwner` resource is the right way forward but I don't know when we will get the chance to address it. Though, in that world we'd have a `ServiceOwnerController#show` and we could reuse these finder objects to make it an easier change.

lozette: How do I always manage to pick up "quick tickets" that turn out to be refactoring monsters 😆

app/services/find_fund_activities.rb
```diff
@@ -0,0 +1,20 @@
...
+    @organisation = organisation
+    @current_user = current_user
+  end
+
+  def call
+    funds = policy_scope(Activity.funds, policy_scope_class: FundPolicy::Scope)
+      .includes(:organisation)
+      .order("created_at ASC")
+    unless organisation.service_owner
+      funds = funds.select { |fund| fund.organisation == organisation }
```
tahb, line 16: It would be more efficient to have this as part of the database query rather than the 2 parter of loading the query and then enumerating it. We might be able to keep building the query before loading it early with:

```
    funds = policy_scope(Activity.funds, policy_scope_class: FundPolicy::Scope)
      .includes(:organisation)
      .order("created_at ASC")

    funds = if organisation.service_owner
      funds.all
    else
      funds.where(organisation_id: organisation.id)
    end

    funds
```

It's also a touch more verbose with the `funds.all` when `service_owner` which I think improves the readability a fraction more.

lozette: I totally got out of the habit of `variable = if ...` syntax bc rubocop is usually against it but it def works here 👍

app/controllers/staff/organisations_controller.rb
tahb: I know this was said in jest :) If it'd be helpful to talk more about this and to share my perspective I'm very happy to try!

lozette: No it's fine!! It makes sense to refactor it, the controller was getting too chunky.

PR #336, opened 2020-04-17
public_activity update events on Activities are more granular

lozette:
## Changes in this PR

Trello: https://trello.com/c/ZQlLq4nF/551-record-when-users-go-through-each-step-of-the-activity-form

Users update an Activity via a multi-step form, a step for each attribute. The
previous implementation of public_activity meant we had an `activity.update`
event for each step. This change changes the event keys to be more obvious
*what* is being created or updated.

We decide if an activity attribute is being created or updated by the `wizard_complete?`
method on Activities - if the wizard is complete, we can assume an attribute is being
updated; if not, we can assume the attribute is being created.

So the event keys will now be `activity.action.attribute`, e.g. activity.create
for the initial Activity creation; activity.create.identifier for the identifier
creaton, and activity.update.identifier for when the identifier is updated.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,13 +99,14 @@
 - Add privacy policy to site
 - Empty optional dates for `actual start date` and `actual end date` are not included on the activity XML
 - Reporting org in the IATI XML is always BEIS for funds, programmes and projects created by governmental organisations, and the activity's organisation if it is a non-governmental organisation
-- User actions are tracked on Activities, Budgets, Transactions and Users.
 
 ## [unreleased]
 
+- User actions are tracked on Activities, Budgets, Transactions and Users.
 - Activities that have an identifier already can use it as the iati-identifier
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -99,10 +99,12 @@
 - Add privacy policy to site
 - Empty optional dates for `actual start date` and `actual end date` are not included on the activity XML
 - Reporting org in the IATI XML is always BEIS for funds, programmes and projects created by governmental organisations, and the activity's organisation if it is a non-governmental organisation
-- User actions are tracked on Activities, Budgets, Transactions and Users.
 
 ## [unreleased]
 
+- User actions are tracked on Activities, Budgets, Transactions and Users.
```
lozette, line 105: I don't know how this one ended up in `released` after the rebase with develop in #434, but it has definitely not been released!

PR #357, opened 2020-04-27
Remove Transaction reference

lozette:
## Changes in this PR

Trello: https://trello.com/c/Azapqja1/554-remove-reference-from-create-a-transaction

In user testing, the Transaction reference was found to be confusing for users.
It is also not required by IATI.

Remove the reference entirely.

## Next steps

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_04_14_101829) do
+ActiveRecord::Schema.define(version: 2020_04_27_130408) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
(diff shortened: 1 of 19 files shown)


lozette:
I can't check the XML output at the moment as the validator is erroring https://github.com/data4development/IATI-data-validator/issues/80

db/schema.rb
```diff
@@ -141,7 +140,6 @@
 
   add_foreign_key "activities", "activities"
   add_foreign_key "activities", "organisations", column: "extending_organisation_id"
-  add_foreign_key "activities", "organisations", column: "reporting_organisation_id"
```
mec, line 144: Is this supposed to be here?

spec/factories/implementing_organisation.rb
```diff
@@ -1,4 +1,6 @@
 FactoryBot.define do
+  sequence(:reference) { |n| "organisation-#{n}" }
```
mec, line 2: I am assuming these are declared globally inside FactoryBot so we were relying on the one from `transaction` before it got deleted?

db/schema.rb
lozette: No! I think I need to clean up my db, this keeps reappearing 😬

spec/factories/implementing_organisation.rb
lozette: Yeah we were... I moved it in here because this is where it's used, I was confused at first because I couldn't figure out where it was declared!

PR #358, opened 2020-04-28
(462) sector step by category

mec:
## Changes in this PR
The goal of this work is to split the sector quesiton in two. It turns out we had all the sectors lumped together in one, very long, list when in-fact they each fall into a categories.

Research showed that users liked choosing a 'sector category' first and then seeing a more refined list of sectors to choose from.

The category list is still quite long - but this is defined by IATI.

I chose to approach this as simply as possible so I did things in this order:

- make the sector question use radio buttons
- make the list of sector able to accept a catgory and show only sectors in that cateogry
- add the sector category as a seperate question included in the flow
- make the flow dynamic based on the answer to the sector category question
- remove the sector category question as it is not really a data poing the users cares about, it's just a means to an end
- backfill exisiting activities with the correct sector category

Editing the dependent answer woks as expected:

A user decides to change their response to sector is first asked which category, the category they have previously selected is shown. They can keep the cateogry of change it before moving on to choose the sector from a refined list. Once they choose the sector they return to the summary view.

## Decisions along the way

I chose not to follow the existing pattern of having both vendor IATI data and locale data for the category - I find this duplication un-necessary until a point we need to retain the vendor data and map different language over it, which I don't belive we are at yet.

I chose to put a `sector_category_name` helper into the Activity model rather than the activity presenter. Although it felt like a good place for this to map the category code back to a name, we do not consisitently use the presenter in controllers which meant I would have to refactor there which would have made this PR too large for my taste!

I found getting the dependent answers to work really complicated! I came to the decision that the shoulda spec that tests validation when the wizard status is complete, was not a representativie test of the reality inside the applicaiton.

At no point do we do anything with an activity that has wizard status complete that would require validaiton. I can see a point not far off that we will have some concept of submitting the activitiy and it will need to be validated at that point, but that time is not now. By removing this requirement I was able to deliver the expected behaviour inside the forms controller.

The solution I have lets us pass the step that we want to end the 'editing' on and return the user to the summary view.

## Screenshots of UI changes

Activity show view, we do not show the sector category as it is not a real questions, just the means to aid the user to answer the sector question:
Screenshot_2020-04-29 Activity Global Challenges Research Fund (GCRF) — Report your official development assistance

Sector category question, list is still quite long:
Screenshot_2020-04-29 What is the focus area category for this fund — Report your official development assistance

Sector question with reduce responses:
Screenshot_2020-04-29 Which Unallocated Unspecified sector is the focus area for this fund — Report your official developme

## Next steps

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -21,8 +21,8 @@ class Staff::ActivityFormsController < Staff::BaseController
   steps(*FORM_STEPS)
 
   def show
-    @page_title = t("page_title.activity_form.show.#{step}")
     @activity = Activity.find(params[:activity_id])
+    @page_title = t("page_title.activity_form.show.#{step}", sector_category: @activity.sector_category_name, level: @activity.level)
     authorize @activity
 
     case step
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -7,6 +7,7 @@ class Staff::ActivityFormsController < Staff::BaseController
     :blank,
     :identifier,
     :purpose,
+    :sector_category,
     :sector,
     :status,
     :dates,
@@ -82,7 +83,7 @@ class Staff::ActivityFormsController < Staff::BaseController
   end
 
   def activity_params
-    params.require(:activity).permit(:identifier, :sector, :title, :description, :status,
+    params.require(:activity).permit(:identifier, :sector_category, :sector, :title, :description, :status,
       :planned_start_date, :planned_end_date, :actual_start_date, :actual_end_date,
       :geography, :recipient_region, :recipient_country, :flow,
       :aid_type)
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -34,6 +34,7 @@ class Staff::ActivityFormsController < Staff::BaseController
       skip_step if @activity.recipient_region?
     end
 
+    @activity.update(wizard_status: "complete") if params["editing_until"] == step.to_s
     render_wizard
   end
 
@@ -45,6 +46,10 @@ class Staff::ActivityFormsController < Staff::BaseController
     update_activity_dates
     update_activity_attributes_except_dates
     record_auditable_activity
+
+    if @activity.wizard_complete?
+      reset_sector_answer if step == :sector_category
+    end
 
     update_wizard_status
 
@@ -104,5 +109,10 @@ class Staff::ActivityFormsController < Staff::BaseController
     else
       @activity.wizard_status = step
     end
+  end
+
+  def reset_sector_answer
+    @activity.update(sector: nil, wizard_status: "editing")
+    jump_to(:sector, editing_until: :sector)
   end
 end
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -1,5 +1,6 @@
 class Activity < ApplicationRecord
   include PublicActivity::Common
+  include CodelistHelper
 
   STANDARD_GRANT_FINANCE_CODE = "110"
   UNTIED_TIED_STATUS_CODE = "5"
@@ -7,6 +8,7 @@ class Activity < ApplicationRecord
   validates :identifier, presence: true, if: :identifier_step?
   validates_uniqueness_of :identifier, if: :identifier_step?
   validates :title, :description, presence: true, if: :purpose_step?
+  validates :sector_category, presence: true, if: :sector_category_step?
   validates :sector, presence: true, if: :sector_step?
   validates :status, presence: true, if: :status_step?
   validates :geography, presence: true, if: :geography_step?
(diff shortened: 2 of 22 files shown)


mec:
[Implement radio buttons for sector step](https://trello.com/c/rc5CArj7/462-implement-radio-buttons-for-sector-step)

app/models/activity.rb
```diff
@@ -49,6 +51,14 @@ def tied_status
     UNTIED_TIED_STATUS_CODE
   end
 
+  def sector_category_name
```
tahb, line 54: As you point out in the commit description, the right place for this presentiional logic is in the helper. I'm concerned that if the propoect of refactoring the activity controller is daunting now, it's going to make leak complexity elsewhere in the app with models and presenters sharing the responsibility of view logic. 

I would like to see if we can make this better and get it out of the model.

config/locales/en.yml
```diff
@@ -191,6 +191,8 @@ en:
         label: Recipient region
       sector:
         label: Focus area
```
tahb, line 193: Now we have 2 types of "focus area", I wonder if we can give a better name to the old value more generally throughout this work to give the reader more clarity.

lozette:
I absolutely agree with your concerns that the Wizard is becoming a nightmare to maintain. 

Unfortunately I've found a bug. If you have a completed Activity, and click on the Focus area edit, choose a sector category then hit Back, the new Sector Category is not saved. I have no idea how to handle this, though - it's correct that the sector category is not saved (because you haven't chosen a sector) but it's also a bad experience for the user. 

This plus all this wrangling with the `editing` Wizard status makes me really nervous - the Wizard is just going to turn into more of a monster as we go along. @tahb is there time/scope to think about improving the Wizard or are we just going to have to live with it because there's so much to do?

config/locales/en.yml
lozette: Agree, one for content design I think?

mec: Yes, I did cover this with @deberny and we felt it best to use something and the circle back in content review?

app/models/activity.rb
mec: There is a similar issue on the third party project PR - I wondered there is we use locales to pass the correct level strong - this might work here as well.

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/361#pullrequestreview-403360691

mec:
@lozette I share your concerns - from my point of view we are trying make Wizard do things I am not convinced it is designed for.

We might be able to resolve the UX by not setting `sector` to nil and instead leave it as the previous value?

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -105,4 +110,9 @@ def update_wizard_status
       @activity.wizard_status = step
     end
   end
+
+  def reset_sector_answer
+    @activity.update(sector: nil, wizard_status: "editing")
```
tahb, line 115: I'm cautious about adding a new wizard_status for editing. I think it introduces a new scenario where a user marks an activity as editing but then fails to complete their transaction (for whatever reason) which marks it back as complete. This could cause odd behaviour. It opens questions like: what happens if 2 people edit the same document at the same time?

Could we rely on this editing state to live only in a query parameter for the request? We're already headed that way with `editing_until`. Perhaps `editing=true` and `editing_step=sector_category`?

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -46,6 +47,10 @@ def update
     update_activity_attributes_except_dates
     record_auditable_activity
 
+    if @activity.wizard_complete?
+      reset_sector_answer if step == :sector_category
```
tahb, line 51: I can see how we can make an easy extension to this for more steps, like geography 👍

app/controllers/staff/activity_forms_controller.rb
tahb: Another idea: A new field that tracks the state of when an activity has been fully filled out. If that's marked as true, we can infer that all future updates are edits. We'd still need to pass the step around in a query param.

mec:
@tahb @lozette I am not convinced yet that the bad UX is bad, I am still looking at it. That said I have found a different problem which is to do with validation - so I'll have another look at this and try and address some of your concerns.

db/data/20200428152559_add_sector_category_to_activities.rb
```diff
@@ -0,0 +1,17 @@
+class AddSectorCategoryToActivities < ActiveRecord::Migration[6.0]
+  include CodelistHelper
+
+  def up
+    sectors = sector_radio_options
+    activities_with_sectors = Activity.where.not(sector: [nil, ""])
+
+    activities_with_sectors.each do |activity|
+      sector = sectors.find { |sector| sector.code == activity.sector }
```
tahb, line 9: Is it valid for an activity to exist without a `sector_category`? What should happen if one cannot be found? If `activity.update(sector_category: nil)` happens, will it blow up to to validation?

config/locales/en.yml
tahb: Happy with that 👍

mec:
Closing this in favour of the refactor work on #368 and then we'll re-introduce the sector categories.

PR #361, opened 2020-04-29
(586) Add third party projects (level D activities)

lozette:
## Changes in this PR

Trello: https://trello.com/c/ZVwI3ta9/586-add-third-party-projects

Add third-party projects (sub-projects), or level D activities.

These are identical to level C activities (projects) except they sit under level C activities, with the level C activity as their parent activity. They are creatable and editable only by delivery partners. They can be downloaded as XML by BEIS users. 

## Next steps

### DIFF
--- a/app/presenters/activity_presenter.rb
+++ b/app/presenters/activity_presenter.rb
@@ -64,4 +64,9 @@ class ActivityPresenter < SimpleDelegator
     return "Untitled (#{id})" if title.nil?
     title
   end
+
+  def level
+    return if super.blank?
+    I18n.t("page_content.activity.level.#{super}")
+  end
 end
--- a/app/services/create_third_party_project_activity.rb
+++ b/app/services/create_third_party_project_activity.rb
@@ -1,0 +1,47 @@
+class CreateThirdPartyProjectActivity
+  attr_accessor :user, :organisation_id, :project_id
+
+  def initialize(user:, organisation_id:, project_id:)
+    self.organisation_id = organisation_id
+    self.project_id = project_id
+    self.user = user
+  end
+
+  def call
+    activity = Activity.new
+    activity.organisation = creating_organisation
+    activity.reporting_organisation = reporting_organisation
+    activity.extending_organisation = creating_organisation
+
+    project = Activity.find(project_id)
+    project.child_activities << activity
+
+    activity.wizard_status = "blank"
+    activity.level = :third_party_project
+
+    activity.funding_organisation_name = service_owner.name
+    activity.funding_organisation_reference = service_owner.iati_reference
+    activity.funding_organisation_type = service_owner.organisation_type
+
+    activity.accountable_organisation_name = service_owner.name
+    activity.accountable_organisation_reference = service_owner.iati_reference
+    activity.accountable_organisation_type = service_owner.organisation_type
+
+    activity.save(validate: false)
+    activity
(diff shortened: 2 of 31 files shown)


spec/features/staff/users_can_create_a_third_party_project_spec.rb
```diff
@@ -0,0 +1,49 @@
+RSpec.feature "Users can create a project" do
+  let(:beis) { create(:delivery_partner_organisation) }
```
mec, line 2: This probably shouldn't be called 'beis'? I don't think we actually even use this?

app/presenters/activity_presenter.rb
```diff
@@ -64,4 +64,9 @@ def display_title
     return "Untitled (#{id})" if title.nil?
     title
   end
+
+  def level
+    return if super.blank?
+    super.tr("_", " ")
```
mec, line 70: What do you think about `super.humanize(capitalize: false)` here, I love all the Rails active support stuff?

We could even use the old 'locale trick' and get the hyphen in? Something like:

````
t("activity_level.#{super}")

activity_level:
  fund: "Fund"
  …
  third_party_project: "Third-party project"
````

File: spec/features/staff/organisation_show_page_spec.rb [hunk 3]
```diff
@@ -122,10 +131,19 @@
       expect(page.find("table.projects  tbody tr:first-child")[:id]).to have_content(yet_another_project.id)
       expect(page.find("table.projects  tbody tr:last-child")[:id]).to have_content(project.id)
     end
+
     scenario "they do not see projects that they are not the reporting organisation of" do
       expect(page).not_to have_content another_project.identifier
     end
 
+    scenario "they see a list of all their third-party projects" do
```
mec, line 139: Do you think we should test they do not see third-party projects that are not theirs (like we do for projects) or are we happy that is covered elsewhere?

app/services/create_third_party_project_activity.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    activity.accountable_organisation_name = service_owner.name
+    activity.accountable_organisation_reference = service_owner.iati_reference
+    activity.accountable_organisation_type = service_owner.organisation_type
+
+    activity.save(validate: false)
+    activity
+  end
+
+  private
```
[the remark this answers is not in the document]
lozette: I prefer a `private` section as opposed to declaring individual methods as private, but I tend to follow what's in a class already if I'm adding to it. If I'm making a new class I'll add a `private` section

app/presenters/activity_presenter.rb
lozette: Oooooh I forgot that existed!

spec/features/staff/users_can_create_a_third_party_project_spec.rb
lozette: Aaah, the old cut & paste 😳

spec/features/staff/organisation_show_page_spec.rb
lozette: Can't hurt to add another test 👍

lozette:
@mec good shout on the `third_party_project` level string appearing as-is! I have done some interpolation-within-interpolation (yes you can do it!) to address this.

PR #362, opened 2020-05-01
Proof of concept IATI ingest tool for UKSA

tahb:
## Changes in this PR

This is the code that was created to test the ingesting of existing IATI data for UKSA into RODA. The general approach looks for fixture files that have been taken from the IATI data registry and attempts to create valid activity and transaction records within RODA.

It's not in a rake task as it's not ready to be used in anger on a real environment until these problems have been addressed. You can run this locally (after seeding your database) using the test fixtures using the following command:

```
IngestIatiActivities.new(delivery_partner: Organisation.find_by(name: "UK Space Agency"), file_io: File.read("#{Rails.root}/spec/fixtures/activities/uksa/real_and_complete_legacy_file.xml")).call
```

The proposed code is verbose and more complex than it could be given more time. It has been built out with test coverage which gives me confidence that it holds enough value to be documented within the repo. Whether it is thrown away later and only serves as documentation of what is possible or becomes the foundation of further work. Some basic refactoring has been done of various responsibilities into private methods. The goal of this spike was not to invest too heavily in the quality of the code though prior refactoring should be made before further extension.

The playback of this exercise along with screenshots and the problems we found are included in this deck which can serve to document the issues: https://docs.google.com/presentation/d/16BV9uxr1A9YNp0EAmb1njn7LC6z1LeZS2Xtw2IBLIvQ/edit#slide=id.g7ff9a40c52_2_187. Some of these include:

- hardcoding the disbursement channels for transactions
- using the previous activity identifier to try and create new programmes intelligently. It created a lot more programmes than was required
- it includes a manual change to a real transaction value of 0.57 to 1.57 to pass a validation problem


spec/fixtures/activities/uksa/with_country.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tahb, line 1: There are multiple seed files as ingested the full file with 36 activities to test one specific condition was extra overhead. I made new files which are subsets of the original.

app/services/ingest_iati_activities.rb
```diff
@@ -0,0 +1,175 @@
...
+      new_activity.accountable_organisation_type = accountable_organisation.organisation_type
+      new_activity.accountable_organisation_reference = accountable_organisation.iati_reference
+    end
+
+    # There is no DP reference in the extending element. We must ask the user for it.
+    new_activity.extending_organisation = delivery_partner
+
+    implementing_org_elements = legacy_activity.elements.select { |element| element.name.eql?("participating-org") && element.attributes["role"].value.eql?("4") }
+    implementing_org_elements.each do |org_element|
+      new_activity.implementing_organisations << ImplementingOrganisation.create!(
```
tahb, line 94: Create action are always called with a bang so the user is quickly notified of problems. This could be better built out with the ability to dry run this task and receive a digest of errors to fix all in one go.

app/services/ingest_iati_activities.rb
```diff
@@ -0,0 +1,175 @@
...
+        "Unknown description"
+      end
+
+      providing_organisation_name = transaction_element.children.detect { |child| child.name.eql?("provider-org") }.children.detect { |child| child.name.eql?("narrative") }.text
+      providing_organisation_reference = transaction_element.children.detect { |child| child.name.eql?("provider-org") }.attributes["ref"].value
+      receiving_organisation_name = transaction_element.children.detect { |child| child.name.eql?("receiver-org") }.children.detect { |child| child.name.eql?("narrative") }.text
+      receiving_organisation_reference = transaction_element.children.detect { |child| child.name.eql?("receiver-org") }.attributes["ref"].try(value, nil)
+
+      transaction = Transaction.new(
+        reference: SecureRandom.hex,
```
tahb, line 122: Reference is being removed during other ongoing work so this sits here now as something we don't have to think about too hard.

app/services/ingest_iati_activities.rb
```diff
@@ -0,0 +1,175 @@
...
+        new_activity.flow = legacy_activity.elements.detect { |element| element.name.eql?("default-flow-type") }.attributes["code"].value
+        new_activity.aid_type = legacy_activity.elements.detect { |element| element.name.eql?("default-aid-type") }.attributes["code"].value
+
+        add_dates(legacy_activity: legacy_activity, new_activity: new_activity)
+        add_geography(legacy_activity: legacy_activity, new_activity: new_activity)
+        add_transactions(legacy_activity: legacy_activity, new_activity: new_activity)
+
+        new_activity.ingested = true
+        # Set the status to invoke validations
+        new_activity.wizard_status = :complete
```
tahb, line 39: The initial approach for saving and validating is to only create valid records. All mandatory fields must be required before creation in RODA and where data is missing we add default values that a user can identify through the front end and correct before publishing to IATI. Rather than creating records with `save(:validate_false)` which could work but introduces the chances of records being in unexpected states, leading to errors. If we can create valid records by default, we should.

mec:
@lozette @tahb worth looking at #368 we add validation contexts and we may be able to leverage them to change the validation during ingest rather than skipping it entirely.

lozette:
I'd like to suggest that this is ready to be reviewed & merged, with the caveat that a better fix for https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/362/commits/4d461fabd57f3241254428cf4cd14c88040cab10 should be spun out into a different card (https://trello.com/c/Nbh9Taqp/615-how-do-we-address-the-fact-that-dates-in-budgets-published-to-iati-are-not-valid)

As @mec suggests above, we can introduce the concept of a draft budget using validation contexts. But I think that's separate work to this branch. So the `skip_validation` fix could stay for now and we can add validation contexts to Budgets ASAP to address it.

app/services/ingest_iati_activities.rb
```diff
@@ -159,7 +159,7 @@ def call
         parent_activity: new_activity
       )
 
-      budget.save!
+      budget.save!(validate: false)
```
mec, line 164: Can we flag budgets (or any entity we ingest) as ingested (or something like `source: ingested`) like we propose on Activity and then a simple validation context would work well here, something like:

`validates :period_start_date, :period_end_date, date_within_boundaries: true, unless: :ingested?`

mec: Bah! That's actually a conditional vlidation - soz, but I still think the approach is sound!

lozette:
If/when https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/382 gets merged then I will need to refactor 24d15d6202acb2a7223f340fa634f5513b81fc6e to only include the ingest part.

spec/services/ingest_iati_activities_spec.rb
```diff
@@ -176,5 +175,19 @@
...
+      it "normalizes the text" do
+        _beis = create(:beis_organisation)
+        uksa = create(:organisation, name: "UKSA", iati_reference: "GB-GOV-EA31")
+        legacy_activities = File.read("#{Rails.root}/spec/fixtures/activities/uksa/with_escaped_characters.xml")
+
+        described_class.new(delivery_partner: uksa, file_io: legacy_activities).call
+
+        activity = Activity.find_by(previous_identifier: "GB-GOV-13-GCRF-UKSA_TZ_UKSA-021")
+
+        expect(activity.description).to eql("Both Ethiopia and Kenya are flood & drought prone with significant mortality & economic losses attributed to these events in each country.")
```
tahb, line 190: I like the fact you're testing for the exact string you hope to see as a result of the private method used. This gives me more confidence than seeing the implementation or method repeated as so:

```
expect(activity.description).to eql(CGI.unescape("Both Ethiopia and Kenya are flood & drought prone with significant mortality &amp; economic losses attributed to these events in each country.").strip)
```

spec/validators/budget_date_validator_spec.rb
```diff
@@ -72,4 +72,16 @@
...
   end
+
+  context "when the budget is ingested" do
+    context "when the dates are more than 365 days apart" do
+      it "is valid (skips validation)" do
+        subject.ingested = true
+        subject.period_start_date = Date.today
+        subject.period_end_date = subject.period_start_date + 366.days
+
+        expect(subject).to be_valid
```
tahb, line 83: Nice 👍

tahb:
@lozette LGTM 👍 Please can you approve and merge on my behalf!

PR #368, opened 2020-05-05
(598) Refactor activity validations

mec:
## Changes in this PR
The goal of this work is to to simplify the stepped form and related Activity creation:

- stop skipping validation when creating a new activity
- use validations contexts to run validation during the form steps
- when all form steps are completed by the user, run all the validations

With the refactor in-place the form steps work as expect when one question depends on the answer to another, by resetting the value to nil the activity is no longer valid so the reset step is shown as the user expects - you can see this in 3ec5514 where the `recipient_country` or `recipient_region` is rest and the spec needed to be updated to reflect the working behaviour. This relates to https://trello.com/c/GwGnh3N1

The use of the word 'wizard' has been minimised to only methods from the gem to try and make the intention of things clearer to developers.

### DIFF
--- a/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
+++ b/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
@@ -73,6 +73,34 @@ RSpec.feature "Users can create a fund level activity" do
       expect(activity.extending_organisation).to eql(user.organisation)
     end
 
+    context "when there is an existing activity with a nil identifier" do
+      scenario "successfully create a activity" do
+        visit organisation_path(user.organisation)
+        click_on(I18n.t("page_content.organisation.button.create_fund"))
+
+        visit organisation_path(user.organisation)
+        click_on(I18n.t("page_content.organisation.button.create_fund"))
+
+        fill_in_activity_form(level: "fund")
+
+        expect(page).to have_content I18n.t("form.fund.create.success")
+      end
+    end
+
+    context "when there is an existing activity with the same identifier" do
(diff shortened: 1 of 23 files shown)


mec:
[Refactor the wizard steps for validation](https://trello.com/c/9VfW5d0T/598-refactor-the-wizard-steps-for-validation)

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
```diff
@@ -73,6 +73,32 @@
...
+        visit organisation_path(user.organisation)
+        click_on(I18n.t("page_content.organisation.button.create_fund"))
+
+        fill_in_activity_form(level: "fund")
+
+        expect(page).to have_content I18n.t("form.fund.create.success")
+      end
+    end
+
+    context "when there is an exsiting activity with the same identifier" do
```
lozette, line 90: Is this correct? From the commit message it sounds like you shouldn't be able to create an activity with the same (non-nil) identifier?

lozette: Also you have a typo in "existing" :)

mec: Oh, man!! I had this on my list and forgot! :( look down there, that "A-Unique_Identifier" has an underscore so it's passing!

Thank you for spotting!

lozette: Phew it wasn't me going mad then 🤣

PR #378, opened 2020-05-07
(618) Validate values down to 1 pence

mec:
## Changes in this PR
The value attribute for both budgets and transactions accepted a
minimum value of 1 (one pound in sterling), looking at existing data
revealed transactions with a value of less that 1 (0.57).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,8 @@
 
 ## [unreleased]
 
+- Transaction and budget values are validated down to 0.01
+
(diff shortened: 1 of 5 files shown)


mec:
[Reduce the minimum validation for a transaction value (and probably budget and disbursement) down to a penny.](https://trello.com/c/fhlSdTXw/618-reduce-the-minimum-validation-for-a-transaction-value-and-probably-budget-and-disbursement-down-to-a-penny)

PR #380, opened 2020-05-07
(462) Sector step uses categories

mec:
## Changes in this PR
The goal of this work is to split the sector quesiton in two. It turns out we had all the sectors lumped together in one, very long, list when in-fact they each fall into a categories.

Research showed that users liked choosing a 'sector category' first and then seeing a more refined list of sectors to choose from.

The category list is still quite long - but this is defined by IATI.

- make the sector question use radio buttons
- make the list of sector able to accept a catgory and show only sectors in that cateogry
- make the flow dynamic based on the answer to the sector category question
- backfill exisiting activities with the correct sector category

## Screenshots of UI changes

Activity show view, we do not show the sector category as it is not a real questions, just the means to aid the user to answer the sector question:
Screenshot_2020-04-29 Activity Global Challenges Research Fund (GCRF) — Report your official development assistance

Sector category question, list is still quite long:
Screenshot_2020-04-29 What is the focus area category for this fund — Report your official development assistance

Sector question with reduce responses:
Screenshot_2020-04-29 Which Unallocated Unspecified sector is the focus area for this fund — Report your official developme

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,7 @@
 - Add Skylight application performance monitoring
 - No longer show the geography response on the activity summary and when changing country or region the flow includes the geography question
 - Add concept of `ingested` to budgets, and skip validation on ingested budgets. This is in preparation for ingesting legacy data from IATI.
+- Sector questions asks for a category and uses radio buttons
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-5...HEAD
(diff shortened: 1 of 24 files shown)


mec:
[Implement radio buttons for sector step](https://trello.com/c/rc5CArj7/462-implement-radio-buttons-for-sector-step)

mec:
@lozette I've had to add to this once it was rebased onto the ingest work - could you have another look?

PR #387, opened 2020-05-13
Transaction disbursement channel is no longer mandatory

lozette:
## Changes in this PR

https://trello.com/c/iDDYOe9j/632-transaction-disbursement-channel-field-to-become-an-optional-field

While working on UKSA data exported from IATI we discovered that many
Transactions do not have a disbursement channel set. This field is not mandatory
in IATI, so it should not be mandatory in our application.

## Next steps

### DIFF
--- a/app/services/ingest_iati_activities.rb
+++ b/app/services/ingest_iati_activities.rb
@@ -114,6 +114,9 @@ class IngestIatiActivities
       date = transaction_element.children.detect { |child| child.name.eql?("value") }.attributes["value-date"].value
       value = transaction_element.children.detect { |child| child.name.eql?("value") }.children.text
       transaction_type = transaction_element.children.detect { |child| child.name.eql?("transaction-type") }.attributes["code"].value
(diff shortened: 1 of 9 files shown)


app/services/ingest_iati_activities.rb
```diff
@@ -114,6 +114,9 @@ def call
       date = transaction_element.children.detect { |child| child.name.eql?("value") }.attributes["value-date"].value
       value = transaction_element.children.detect { |child| child.name.eql?("value") }.children.text
       transaction_type = transaction_element.children.detect { |child| child.name.eql?("transaction-type") }.attributes["code"].value
+      disbursement_channel = if transaction_element.children.detect { |child| child.name.eql?("disbursement-channel") }.present?
```
tahb, line 117: I think it's good that we are trying to save this field if it's found. It would be good to have an assertion for this but I can't see one in any of our existing fixtures. Could we find a real activity on IATI with a `disbursement-channel` or fake one for testing purposes?

PR #389, opened 2020-05-13
(636) A BEIS transparency user can download all activities for each delivery partner in a single XML file

lozette:
## Changes in this PR

Trello: https://trello.com/c/XpXfmNXM/636-a-beis-transparency-user-can-download-the-xml-all-activities-for-each-delivery-partner-in-a-single-file

A BEIS user needs to be able to download all the (project C or third-party project D, whichever is lowest) activities for an organisation as a single file. Currently the BEIS transparency user has to go to each Activity show page to download each Activity individually. 

In this PR I have added `has_many` association to Activities linking to Budgets and Transactions. This is after a lot of thought about the simplest way to get all budgets/transactions for each Activity. 

In Organisation#show we are getting all the project & third-party activities for that organisation. Without these associations, we would have to iterate over each activity in the Organisation controller and get that activity's budgets/transactions. Adding this association seemed like the simplest way to address this.

In future these files can potentially become extremely large, as each Organisation has to retrieve more and more activities to display. We should think about ways we can ensure the application remains performant, possibly via caching of activities and/or their budgets/transactions?

### After

Screenshot 2020-05-19 at 14 15 07

## Next steps

### DIFF
--- a/app/controllers/staff/organisations_controller.rb
+++ b/app/controllers/staff/organisations_controller.rb
@@ -17,10 +17,22 @@ class Staff::OrganisationsController < Staff::BaseController
     project_activities = FindProjectActivities.new(organisation: organisation, current_user: current_user).call
     third_party_project_activities = FindThirdPartyProjectActivities.new(organisation: organisation, current_user: current_user).call
 
-    @fund_activities = fund_activities.map { |activity| ActivityPresenter.new(activity) }
-    @programme_activities = programme_activities.map { |activity| ActivityPresenter.new(activity) }
-    @project_activities = project_activities.map { |activity| ActivityPresenter.new(activity) }
-    @third_party_project_activities = third_party_project_activities.map { |activity| ActivityPresenter.new(activity) }
+    respond_to do |format|
+      format.html do
+        @fund_activities = fund_activities.map { |activity| ActivityPresenter.new(activity) }
+        @programme_activities = programme_activities.map { |activity| ActivityPresenter.new(activity) }
+        @project_activities = project_activities.map { |activity| ActivityPresenter.new(activity) }
+        @third_party_project_activities = third_party_project_activities.map { |activity| ActivityPresenter.new(activity) }
+      end
+      format.xml do
+        @project_activities = if third_party_project_activities.present?
+          third_party_project_activities.map { |activity| ActivityXmlPresenter.new(activity) }
+        else
+          project_activities.map { |activity| ActivityXmlPresenter.new(activity) }
+        end
+        response.headers["Content-Disposition"] = "attachment; filename=\"#{organisation.iati_reference}.xml\""
+      end
+    end
   end
 
   def new
--- a/app/controllers/staff/organisations_controller.rb
+++ b/app/controllers/staff/organisations_controller.rb
@@ -25,10 +25,13 @@ class Staff::OrganisationsController < Staff::BaseController
         @third_party_project_activities = third_party_project_activities.map { |activity| ActivityPresenter.new(activity) }
       end
       format.xml do
-        @project_activities = if third_party_project_activities.present?
+        @activities = case level
+        when "project"
+          project_activities.map { |activity| ActivityXmlPresenter.new(activity) }
+        when "third_party_project"
           third_party_project_activities.map { |activity| ActivityXmlPresenter.new(activity) }
         else
-          project_activities.map { |activity| ActivityXmlPresenter.new(activity) }
+          []
         end
         response.headers["Content-Disposition"] = "attachment; filename=\"#{organisation.iati_reference}.xml\""
       end
@@ -83,5 +86,9 @@ class Staff::OrganisationsController < Staff::BaseController
 
   def organisation_params
     params.require(:organisation).permit(:name, :organisation_type, :default_currency, :language_code, :iati_reference)
+  end
+
+  def level
+    params[:level]
   end
 end
--- a/app/views/staff/organisations/show.html.haml
+++ b/app/views/staff/organisations/show.html.haml
@@ -73,3 +73,8 @@
 
       - if policy(:organisation).edit?
         = link_to t("page_content.organisation.button.edit_details"), edit_organisation_path(@organisation_presenter), class: "govuk-button"
+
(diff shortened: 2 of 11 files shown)


app/views/staff/organisations/show.xml.haml
```diff
@@ -0,0 +1,11 @@
+!!! XML
+%iati-activities{"version" => "#{IATI_VERSION.gsub('_', '.')}",
+"generated-datetime" => "#{Time.now.strftime("%Y-%m-%dT%H:%M:%S")}"}
+  - @fund_activities.each do |activity|
+    = render partial: "staff/shared/xml/activity", locals: { activity: activity, transactions: activity.transactions, budgets: activity.budgets }
```
lozette, line 5: This is where I'm using those new associations on `Activity`

app/views/staff/organisations/show.xml.haml
```diff
@@ -0,0 +1,7 @@
+!!! XML
+%iati-activities{"version" => "#{IATI_VERSION.gsub('_', '.')}",
+"generated-datetime" => "#{Time.now.strftime("%Y-%m-%dT%H:%M:%S")}"}
+  - @project_activities.each do |activity|
+    = render partial: "staff/shared/xml/activity", locals: { activity: activity, transactions: activity.transactions, budgets: activity.budgets }
+  - @third_party_project_activities.each do |activity|
```
tahb, line 6: When this trello card was made I forgot that level D activities had been shipped. I think there might be a bit more logic involved here now based on the type of organisation. We should check this with Dan, I think it'll be manageable to deal with here 🤞 

I _think_ that:

- for UKSA only projects are published to IATI
- for Met Office and other NGO organisations _only_ third party projects are published to IATI
- there is no case when both levels should be published to IATI in the same file

I'll raise this with Dan on Slack and cc you in.

app/views/staff/organisations/show.html.haml
```diff
@@ -74,7 +74,11 @@
       - if policy(:organisation).edit?
         = link_to t("page_content.organisation.button.edit_details"), edit_organisation_path(@organisation_presenter), class: "govuk-button"
 
-  .govuk-grid-row
-    .govuk-grid-column-two-thirds
-      - if policy(:organisation).download?
+  - if policy(:organisation).download?
```
tahb, line 77: I agree with making Pundit do the heavy lifting 👍

tahb:
Also, I think it was a good idea to add those associations to make it easier to get transactions and budgets from activities. Thank you for chipping away at this and taking the time to make it better for all 👍

app/views/staff/organisations/show.xml.haml
tahb: https://dxw.slack.com/archives/CN0H2L066/p1589540222086000

config/locales/en.yml
```diff
@@ -245,6 +245,9 @@ en:
       default_currency:
         label: Default currency
       details: Organisation details
+      download:
+        explanation: Download all project activities for this organisation
```
lozette, line 249: Probably need some content design on this!

app/views/staff/organisations/show.xml.haml
tahb: To close the loop on this for the record: Dan and Odeta confirmed this to be the shared understanding of how it should work. Resolving thread.

app/controllers/staff/organisations_controller.rb
```diff
@@ -17,10 +17,22 @@ def show
...
-    @third_party_project_activities = third_party_project_activities.map { |activity| ActivityPresenter.new(activity) }
+    respond_to do |format|
+      format.html do
+        @fund_activities = fund_activities.map { |activity| ActivityPresenter.new(activity) }
+        @programme_activities = programme_activities.map { |activity| ActivityPresenter.new(activity) }
+        @project_activities = project_activities.map { |activity| ActivityPresenter.new(activity) }
+        @third_party_project_activities = third_party_project_activities.map { |activity| ActivityPresenter.new(activity) }
+      end
+      format.xml do
+        @project_activities = if third_party_project_activities.present?
```
tahb, line 28: I wonder if we could make a more reliable check by looking at the `organisation.organisation_type` field which should include one of the values from this list: http://reference.iatistandard.org/203/codelists/OrganisationType/. I'd worry that in the future if we changed out minds and allowed the users to add more information, even though we chose not to publish it, we'd inadvertently wipe out all project information from IATI.

I'd be interested to know if we could say `if @organisation.is_non_government?`which checked the values against a whilist for a code of 21, 22, 23 or 24. I'd have to check with Dan if that assertion or similar could be made.

lozette: At the moment we have `is_government?` which checks for codes 10 & 11. I'm *assuming*. that non-government is the opposite?

lozette: After discussion with Odeta https://dxw.slack.com/archives/CN0H2L066/p1589798935095100 it looks like the govermental/governmental switch might be on whether or not we export *level B* activities, not level C/D

lozette: After more discussion I'm going to add a "Download" button to Level C & Level D activities (with appropriate content around the buttons) so that the transparency user can choose the most appropriate level to download

lozette:
I've re-jigged the download activity action again, this time the transparency user can choose to download either level c or level d if they want to (assuming those activities exist)

Screenshot 2020-05-19 at 14 15 07

app/controllers/staff/organisations_controller.rb
```diff
@@ -25,10 +25,13 @@ def show
         @third_party_project_activities = third_party_project_activities.map { |activity| ActivityPresenter.new(activity) }
       end
       format.xml do
-        @project_activities = if third_party_project_activities.present?
+        @activities = case level
+        when "c"
```
tahb, line 29: I think it would be preferable to keep using consistent language internally for `projects` and `third-party projects` rather than adding a mix of both for simplicity sake. I suppose it would make the query param longer but I think that's okay?

lozette: Good call!

PR #390, opened 2020-05-13
Debugging to check why userinfo isn't always set in the session

tahb:
## Changes in this PR

Only replicatable on staging so we have to push to develop.

On our initial debugging 048c0fd2ffde499a483ed78f7d08ac1b4f6afa43 we learnt that the user does have a session, keys are present but not the userinfo key. Let's verify if the callback endpoint recieves this data when auth0 sends a callback.

## Next steps

### DIFF
--- a/app/controllers/auth0_controller.rb
+++ b/app/controllers/auth0_controller.rb
@@ -4,6 +4,13 @@ class Auth0Controller < ApplicationController
     # and the IdP
     session[:userinfo] = request.env["omniauth.auth"]
 
+    # DEBUG: Is the session set correctly on the auth0 callback? Is it ready to
+    # read straight away or is there a delay?
+    if Rails.env.production?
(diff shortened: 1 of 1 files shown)


app/controllers/auth0_controller.rb
```diff
@@ -4,6 +4,13 @@ def callback
     # and the IdP
     session[:userinfo] = request.env["omniauth.auth"]
 
+    # DEBUG: Is the session set correctly on the auth0 callback? Is it ready to
+    # read straight away or is there a delay?
+    if Rails.env.production?
+      Rails.logger.info("* Auth0 callback contained the following id: #{request.env.dig("omniauth.auth", "uid")}")
```
tahb, line 10: I'm trying to use dig here because as this code is going to develop it should be safe to run in production before I can revert it.

PR #406, opened 2020-05-20
(594) Additional planned disbursement features

mec:
## Changes in this PR
This work introduces the addiitonal features to Planned disbursments.

## Export to XML
Valid IATI XML is produced that includes planned disbursements.

https://test-validator.iatistandard.org/view/dqf/files/292a5e17-6e8d-4bb7-86af-7d0464954145?isTestfiles=true

## Ingest
Although the ingest work is not yet finalised it was a good oportunity for me to look at the ingest code. There is a great deal of complexity around ingested legacy data that is not well-formed in RODA's world so this is just the basics in-line with the rest of the ingest code.

When we ingest and there is no IATI reference for the receiving organisation, we don't export the XML attributes, this keeps the IATI XML valid.

## Pre-filll providing organisation details on new form
This was picked up in user research so I stretched to complete this.  The providing organisation can be derived from the Activity organisation type and level:

when the acitivty is a project (level C)
and the organisation is government type
the providing organisation is BEIS

when the acitivty is a third-party project (level D)
and the organisation is government type
the providing organisation is BEIS

when the activity is a project (level C)
and the organisation is non-government type
the providing organisation is the BEIS

when the acitivty is a third-party project (level D)
and the organisation is non-government type
the providing organisationis the delivery partner

See discussion here:

https://dxw.slack.com/archives/CN0H2L066/p1589901111126800

I decided to keep the UI rather than just set the attributes, this allows users to change the values when circustances arise that we didn't anticiapate.

We may want to update the content on the form to reflect this, but we should also look at the existing planned disbursement content as a whole anyway and this can be factored in there.

## Screenshots of UI changes

Screenshot_2020-05-20 Add a planned disbursement — Report your official development assistance

Screenshot_2020-05-20 Add a planned disbursement — Report your official development assistance(1)

## Next steps

### DIFF
--- a/spec/presenters/planned_disbursement_xml_presenter_spec.rb
+++ b/spec/presenters/planned_disbursement_xml_presenter_spec.rb
@@ -1,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe PlannedDisbursementXmlPresenter do
+  let(:planned_disbursement) { build_stubbed(:planned_disbursement) }
+
+  describe "#period_start_date" do
+    it "returns a date formatted for IATI XML" do
+      planned_disbursement.period_start_date = "25 June 2020"
+      expect(described_class.new(planned_disbursement).period_start_date).to eq("2020-06-25")
+    end
+  end
+
+  describe "#period_end_date" do
+    it "returns a human readable date" do
+      planned_disbursement.period_end_date = "October 20, 2020"
+      expect(described_class.new(planned_disbursement).period_end_date).to eq("2020-10-20")
+    end
+  end
+
+  describe "#value" do
+    it "returns the value to two decimal places formatted for IATI XML" do
+      expect(described_class.new(planned_disbursement).value).to eq("100000.00")
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,7 @@
 ## [unreleased]
 
 - If an activity has a recipient_country set, the recipient_region is inferred from the recipient_country
+- Planned disbursements are exported in the IATI XML
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-6...HEAD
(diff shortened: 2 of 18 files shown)


mec:
[Planned disbursements can be ingested and exported as xml](https://trello.com/c/EGcWqoPA/639-planned-disbursements-can-be-ingested-and-exported-as-xml)

spec/presenters/planned_disbursement_xml_presenter_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe PlannedDisbursementXmlPresenter do
+  let(:planned_disbursement) { build_stubbed(:planned_disbursement) }
+
+  describe "#period_start_date" do
+    it "returns a human readable date" do
+      planned_disbursement.period_start_date = "2020-06-25"
```
lozette, line 8: These tests are simply returning the date as it was inputted, which is fine, but it would be useful to have a test to show it actually reformatting a date, e.g. 25-06-2020 -> 2020-06-25

spec/presenters/planned_disbursement_xml_presenter_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  describe "#period_end_date" do
+    it "returns a human readable date" do
+      planned_disbursement.period_end_date = "2020-10-20"
+      expect(described_class.new(planned_disbursement).period_end_date).to eq("2020-10-20")
+    end
+  end
+
+  describe "#value" do
```
mec, line 20: @lozette I also don't like this one - it doesn't actually do what it says - I stole it from another presenter spec, but I am concerned that if I do anything to the value I am introducing rounding errors to the monetary value? Do you have any thoughts on what we can do to output this as `x.00` - then we could apply it elsewhere?

lozette: How about `ActionController::Base.helpers.number_to_currency(super, unit: "", delimiter: "")` - similar to the non-XML presenter but without the unit and the comma delimiter.

PR #407, opened 2020-05-20
Set the Activity#recipient_region when the #recipient_country is set

lozette:
## Changes in this PR
Trello: https://trello.com/c/b60vhOYB/631-when-an-actvity-has-the-country-set-the-region-is-set-automatically

A user can choose a recipient region or recipient country when creating or editing
an activity. If they choose a recipient country, we want to infer from that country
code which recipient_region to set.

This `country => region` mapping was supplied by BEIS and generated from a
spreadsheet.

**Question:** is the format used in `/vendor/data/country_to_region_mapping.rb` acceptable, or should I switch it for a `.yml` file in keeping with the other codelists? I feel the hash constant I've used it probably far faster than a YAML file. Are we happy with it being placed in this location?

## Next steps


app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -31,8 +31,7 @@
   %activity-date{"iso-date" => activity.planned_end_date, type: "3"}/
   - if activity.actual_end_date?
     %activity-date{"iso-date" => activity.actual_end_date, type: "4"}/
-  - if activity.recipient_region?
-    %recipient-region{"code" => activity.recipient_region, vocabulary: "1"}/
+  %recipient-region{"code" => activity.recipient_region, vocabulary: "1"}/
```
lozette, line 34: `recipient_region` will always be present now so no need for a check

vendor/data/country_to_region_mapping.rb
```diff
@@ -0,0 +1,253 @@
+﻿REGION_CODES = {
+"AF" => "679",
+"AX" => "89",
+"AL" => "89",
+"DZ" => "189",
+"AS" => "889",
+"AD" => "89",
+"AO" => "289",
+"AI" => "380",
+"AQ" => "",
```
mec, line 10: Do we know that these missing countries are no in the country list?

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -1,3 +1,5 @@
+require "#{Rails.root}/vendor/data/country_to_region_mapping.rb"
```
mec, line 1: I guess yaml is consistent with the rest of the vendor data, but if you are saying there is a performance benefit I wonder why we chose yaml to begin with?

vendor/data/country_to_region_mapping.rb
```diff
@@ -0,0 +1,253 @@
+﻿REGION_CODES = {
```
mec, line 1: I wonder if putting this in a 'beis' folder makes the distinction about 'who' did the mapping simpler to access?

app/controllers/staff/activity_forms_controller.rb
lozette: The YAML files we use have multiple things in each data "chunk" (e.g. code, description, name) so a one-to-one mapping with a hash would make less sense. I think!?

vendor/data/country_to_region_mapping.rb
lozette: I've moved it into a BEIS folder 👍  

I still am not sure if this is a very Rails-y way of doing it, maybe a module with a constant in it is better? 🤔

mec: When I think about how all the other big lists of data come into the applicaiton, I would look for something like this  not a module - I would probably look for yaml and a code list helper - as that seems to be the pattern we already have.

That's not to say this doesn't work and is more elegant, it's more about how aproachable the code?

vendor/data/country_to_region_mapping.rb
lozette: Good point... Maybe I should add a default to "Developing countries, unspecified" for the ones which are nil, given this is the default in the application 🤔

vendor/data/country_to_region_mapping.rb
lozette: Yeah I think you're right... I will switch this to YAML for consistency!

vendor/data/country_to_region_mapping.rb
lozette: I've checked all the "" countries and none of them appear on the DAC country list, so I think we're OK here!

PR #408, opened 2020-05-20
Create meaningful project identifiers from IATI data

tahb:
## Changes in this PR

- Infer project identifiers from IATI identifiers to make them more meaningful when users are navigating RODA. Rather than seeing a SecureRandom.uuid they will see UKSA_AF_UKSA-19 instead which users are more likely to relate to

### Before

Screenshot 2020-05-20 at 15 49 32

### After

Screenshot 2020-05-20 at 15 37 28

## Next steps


spec/services/ingest_iati_activities_spec.rb
```diff
@@ -3,14 +3,25 @@
 
 RSpec.describe IngestIatiActivities do
   describe "#call" do
-    it "creates 36 new projects" do
+    it "creates 35 new projects" do
```
mec, line 6: I am struggling to see why this has changed when the fixture has not?

app/services/ingest_iati_activities.rb
```diff
@@ -210,8 +210,15 @@ def call
   end
 
   private def add_identifiers(legacy_activity:, new_activity:)
-    new_activity.identifier = SecureRandom.uuid
     new_activity.previous_identifier = legacy_activity.elements.detect { |element| element.name.eql?("iati-identifier") }.children.text
+    new_activity.identifier = infer_internal_identifier(old_identifier: new_activity.previous_identifier.dup)
```
mec, line 214: Could you help me to understand why the `dup` is required here?

app/services/ingest_iati_activities.rb
```diff
@@ -210,8 +210,15 @@ def call
   end
 
   private def add_identifiers(legacy_activity:, new_activity:)
```
mec, line 212: I'd be interested to hear if you have a preference for inline `private` like this or using a private block - our code base is incosistent in it's use - I keep meaning to see if this is something StandardRB will make a call on so we don't have to?

spec/services/ingest_iati_activities_spec.rb
tahb: This is a good point. I haven't explained why this happened! I have subsequently changed this back to 36 in the [more recent work to change how the tool deals with programmes](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/411). Once that's merged I'll fix up this pull request so that it should disappear 👍

app/services/ingest_iati_activities.rb
tahb: I believe I needed to `dup` because when I passed this value in: `new_activity.previous_identifier` the method then made changes to it which worked but changed the original object too which was not what I wanted. The value of `new_activity.previous_identifier` was mutated based on the behaviour of the infer_internal_identifier method.

I think I could have written it a different way to assign it to a new piece of memory but it's slightly more verbose so I skipped it:

```
previous_identifier = new_activity.previous_identifier
infer_internal_identifier(previous_identifier)
```

app/services/ingest_iati_activities.rb
tahb: I'm of course happy to follow a linting rule, that would make things more consistent. My preference is for inline ones since I find it a simple and clear way to work out the scope of a method. I do not have to trawl around a class (especially a big one) to make sure the method exists in the right place in the vertical hierarchy.

```
def
private def
protected def
```

Using this syntax methods can appear anywhere in a class and in any order without changing their effective scope:

```
private def
protected def
def
```

For big files I found enough occasions where the pain of having to figuring out a file such as the following was too much needless pain:

```
def 
def
def
def
def
private 
  def
  def
  def (is this in a private scope? I have to scroll up)
def (is this within a scope? I have to scroll up. It's actually meant to be public but has been accidentally reordered)
protected
  def
  def

tahb:
Just to repeat my comment here to bring it to the top: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/408#discussion_r428682192 I

tahb:
@mec this pull request is still based off of an [unmerged PR](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/412) however an updated version of the code required for this job can be found in a single commit: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/408/commits/ab390caa0c91e329a14e89b401f72e5240dd2de6

tahb:
I restarted the stuck Travis build and the coveralls isn't updating. Hopefully after a rebase the different sha will kick it into gear.

PR #409, opened 2020-05-20
Switch our version of the form builder gem to the latest

LornaHarwood:
## Changes in this PR

- During a separate piece of work we have been trying to take advantage
of a new feature of this gem.
- Since we were using a dxw fork, we were unable to do this, so had to
either switch away from it or update our fork. Given that it looked like
the gem was now providing the feature that caused us to originally fork,
we think that it is safe to switch back.
- On updating the gem we found that there were able to remove the monkey
patching for Internationalisation and replace it with gem configuration
https://govuk-form-builder.netlify.app/building-blocks/localisation/#customising-locale-structure
- We found that we needed to continue to provide an override so that we
could continue to use a single set of local values rather than have it
duplicated. This is consist with the initial approach our application
was using.
- Whilst by default the gem looks for locale values under helpers.legend
we were already using helpers.fieldset, so we made another configuration
change to allow us to continue to use helpers.fieldset
- We have changed html_safe? to html_safe without a question mark after
a test failure made us aware of the fact that this hint text was never
being used. Since the update to the gem we found that we do need to have
this line for the hint text appear.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,6 +144,7 @@
 - Ingest tool will try to set more meaningful identifiers for projects
 - The providing organisation is pre-filled in for new transactions
 - No longer lint the automatic schema changes made by the data_migrate gem
+- Switched to the latest form builder gem version from our fork
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-6...HEAD
 [release-6]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-5...release-6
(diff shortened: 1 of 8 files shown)


LornaHarwood:
[Delivery partners can see codes when selecting aid type](https://trello.com/c/2xrjDcXk/477-delivery-partners-can-see-codes-when-selecting-aid-type)

LornaHarwood:
[Switch from forked version of the GOVUKFormBuilder](https://trello.com/c/3ugctxMK/140-switch-from-forked-version-of-the-govukformbuilder)

lozette:
This looks really great, well done 👍 

The build is failing because of `standard`, if you run `rake standard:fix` it will fix up the issues automagically

PR #410, opened 2020-05-21
Cookie statement

lozette:
## Changes in this PR

Trello: https://trello.com/c/mysPfWry/524-create-and-publish-a-cookie-policy

Copy approved by BEIS legal team.

Co-Authored-By: meyric@dxw.com

### After

Screenshot 2020-05-21 at 12 12 32

## Next steps


lozette:
I've cherry-picked the commit from https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/413 into this PR as it makes more sense to be with this work.

config/locales/en.yml
```diff
@@ -353,6 +354,9 @@ en:
...
... cookies do not store your personal data but are required for the service to operate.</p> <table class="govuk-table"> <thead class="govuk-table__head"> <tr class="govuk-table__row"> <th scope="col" class="govuk-table__header">Name</th> <th scope="col" class="govuk-table__header">Purpose</th> <th scope="col" class="govuk-table__header">Expires</th> </tr> </thead> <tbody class="govuk-table__body"> <tr class="govuk-table__row"> <td class="govuk-table__cell">_roda_session</td> <td class="govuk-table__cell">Identifies your session in RODA</td> <td class="govuk-table__cell">When you close your browser or 24 hours</td> </tr> </tbody> </table> <h2 class="govuk-heading-m">Gathering your feedback</h2> <p class="govuk-body">We use Google Forms software to collect information about what you thought of the RODA service. We do this to help make sure the site is meeting the needs of its users and to help us make improvements.</p> <p class="govuk-body">For more information on Google‘s cookies, please see the <a href="https://policies.google.com/technologies/cookies?hl=en-UK" class="govuk-link">Google Privacy and Terms</a> site.</p> <p class="govuk-body published-dates">Last updated 21 May 2020</p>
```
tahb, line 358: I'm not sure how a HTML blob is easier to maintain than a structured view to be honest. I would assume that it is easier to edit HTML when looking at it with its structure and indentation? I realise this file would be changing less frequently but we work with the en.yml file a lot but would be less likely to open a cookies.html. 

As a similar static page our privacy policy content has already been set up as conventional view file rather than being managed as a blob so this new approach introduces an inconsistency. 

Have I missed a benefit here?

app/controllers/public/cookie_statement_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Public::CookieStatementController < ApplicationController
+  include Pundit
```
tahb, line 2: I see we have a test that is checking we can see the cookie policy when both signed in and signed out which is great. It seems a bit odd to have authorisation code loaded into a controller with the name "public".

I think I've answered my own question: It looks to be because the `_navigation` partial that's rendered has logic that uses Pundit methods. This makes sense as we want the same view to be rendered for staff and the public without needing to duplicate views. 

Do you think it is worth including these in each public or API controller as we need or should we include Pundit in all controllers as a default?

config/locales/en.yml
lozette: I've only just realised that the privacy policy isn't internationalised 😱 

You haven't missed a benefit, I will make the cookie statement like the other public pages (only internationalised)

app/controllers/public/cookie_statement_controller.rb
lozette: Assuming all pages will need the navigation maybe we should add it everywhere.

PR #411, opened 2020-05-21
(634) Prevent ingest tool from creating new programmes

tahb:
## Changes in this PR

- Ingest tool will now check each new project against a predefined list of programmes to projects to check which RODA programme to associate with
- Start a refactoring to the ingest script by defering logic through a new class called `LegacyActivity`. I felt that the service in its previous form was too complex to add anymore logic into so I started to break it out. I was tempted to refactor the whole service to move all XML query commands into it but I felt that that was a rabbit hole and that this would be enough for now. It does set a direction for future changes so that would be useful to review
- For UKSA all projects should link to the single programme of international partnerships however other delivery partners will have many more so using a CSV seemed like an appropriately flexibile option

### Before

Screenshot_2020-04-27_at_15 49 29

### After

Screenshot 2020-05-21 at 14 48 55

## Next steps


app/services/ingest_iati_activities.rb
```diff
@@ -13,15 +14,15 @@ def initialize(delivery_partner:, file_io:)
   def call
     doc = Nokogiri::XML(file_io, nil, "UTF-8")
 
-    legacy_activities = doc.xpath("//iati-activity")
+    legacy_activity_nodes = doc.xpath("//iati-activity")
+    legacy_activity_nodes.each do |legacy_activity_node|
+      legacy_activity = LegacyActivity.new(activity_node_set: legacy_activity_node)
 
-    legacy_activities.each do |legacy_activity|
```
tahb, line 18: I thought that `legacy_activity` was too generic a name, I've tried to be more exact by including `node` in the names of these variables to communicate the type of variable.

app/services/ingest_iati_activities.rb
```diff
@@ -13,15 +14,15 @@ def initialize(delivery_partner:, file_io:)
...
       ActiveRecord::Base.transaction do
-        new_activity = Activity.new(level: :project, organisation: delivery_partner)
+        new_activity = Activity.find_or_create_by(identifier: legacy_activity.identifier, level: :project, organisation: delivery_partner)
```
tahb, line 22: `new_activity` doesn't strictly make sense in the case when a record is "found" though naming it to activity here and throughout the rest of the code here where `legacy_activity` is used felt more confusing.

lib/legacy_activity.rb
```diff
@@ -0,0 +1,42 @@
+require "csv"
```
tahb, line 1: As a class that was representing data not sourced by this service I felt that a home within lib/ would be more appropriate that models/. A models folder we use regularly and if it included both `activity` and `legacy_activity` that would create a lot of unnecessary confusion.

spec/lib/legacy_activity_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+          <?xml version="1.0" encoding="UTF-8"?>
+          <iati-activities generated-datetime="2019-09-25T22:01:44.470000+00:00" version="2.03">
+            <!--Data generated by IATI CoVE. Built by Open Data Services Co-operative: http://iati.cove.opendataservices.coop/-->
+            <iati-activity default-currency="GBP" hierarchy="2" last-updated-datetime="2019-09-25T22:01:44.470000+00:00">
+              <iati-identifier>GB-GOV-13-GCRF-UKSA_NS_UKSA-019</iati-identifier>
+            </iati-activity>
+          </iati-activities>
+        XML
+
+        activity_node_set = Nokogiri::XML(activity_xml, nil, "UTF-8").xpath("//iati-activity").first
```
tahb, line 44: As I said in my commit. I really didn't want to do this conversion each time. I'd love to give the class a `Nokogiri::XML::Element` object but I could not figure out how to create one!

lib/legacy_activity.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def elements
+    activity_node_set.elements
+  end
+
+  def to_xml
+    activity_node_set.to_xml
+  end
+  end
```
lozette, line 41: You have an extra `end` here but it's removed in a later commit!

spec/lib/legacy_activity_spec.rb
lozette: `Nokogiri::XML(activity_xml, nil, "UTF-8").xpath("/iati-activities/iati-activity")` will return an `Nokogiri::XML::Element`, and you can then pass that into LegacyActivity.

You'd have to adjust `lib/legacy_activity.rb:21` to be `identifier_element = activity_node_set.children.detect { |element| element.name.eql?("iati-identifier") }` to get this test passing.

Not sure if that's what you were after!

tahb: That's the only way I can think to do it but it requires us to have a`activity_xml` and do all the nokogiri parsing stuff.

I'd like to avoid that and say something shorter like:

```
legacy_nodes = Nokogiri::XML::Element.new(…………)
result = service(legacy_nodes).do_thing
```

lozette: Yeah I thought I was being too simplistic there 😁 I will have a think, but I agree maybe it's worth merging this as-is so we can move on to AMS, and circle back later to refactor this, maybe together?

tahb: Yeah happy to come back to it if there's no obvious improvement I was missing 👍

lib/legacy_activity.rb
tahb: Fixed 👍

PR #412, opened 2020-05-21
(665) Ingesting updates existing projects before creating new ones

tahb:
## Changes in this PR

- Update existing UKSA production data with the associated IATI identifiers to provide a reliable way to update existing records
- Never ingest the same activity twice to avoid double entry
- If the ingest finds an existing activity that matches the IATI activity, it will do a smaller set of actions. Namely creating transactions, budgets and planned descriptions, as well as saving the XML onto the record for future updates
- if the ingest cannot find a matching record (as is the case for 11 activities) then it will seek to create new records

### After

Screenshot 2020-05-21 at 17 59 10

Screenshot 2020-05-22 at 10 47 40
Screenshot 2020-05-22 at 10 47 47

Screenshot 2020-05-21 at 18 01 00

## Next steps

### DIFF
--- a/app/services/ingest_iati_activities.rb
+++ b/app/services/ingest_iati_activities.rb
@@ -17,33 +17,37 @@ class IngestIatiActivities
     legacy_activity_nodes = doc.xpath("//iati-activity")
     legacy_activity_nodes.each do |legacy_activity_node|
       legacy_activity = LegacyActivity.new(activity_node_set: legacy_activity_node, delivery_partner: delivery_partner)
+      existing_activity = Activity.find_by(previous_identifier: legacy_activity.identifier)
 
       ActiveRecord::Base.transaction do
-        new_activity = Activity.find_or_create_by(identifier: legacy_activity.identifier, level: :project, organisation: delivery_partner)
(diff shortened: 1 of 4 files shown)


tahb:
Requires https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/411 to be merged first.

app/services/ingest_iati_activities.rb
```diff
@@ -154,14 +158,14 @@ def call
       providing_organisation_reference = transaction_element.children.detect { |child| child.name.eql?("provider-org") }.attributes["ref"].value
       receiving_organisation_name = transaction_element.children.detect { |child| child.name.eql?("receiver-org") }.children.detect { |child| child.name.eql?("narrative") }.text
       receiving_organisation_reference = transaction_element.children.detect { |child| child.name.eql?("receiver-org") }.attributes["ref"].try(value, nil)
-
+# binding.pry
```
lozette: 😱

tahb: 🤭

tahb: @lozette I've removed the pry, thanks for spotting.

PR #413, opened 2020-05-22
Move privacy policy feature test into matching public folder

tahb:
## Changes in this PR

The controller and views for this are nested under "public"  so this change moves the test to be 
consistent.

## Next steps


mec:
@tahb #410 changes the name of the spec for this - it might be easier to address this there - it needs a review anyway?

lozette:
Agree with @mec, I will make the change in #410

PR #416, opened 2020-05-26
Change to titles for the "create fund" journey

duffylouise:
Changed 7 titles from statements to questions based on the external content review

## Next steps

### DIFF
--- a/spec/features/staff/users_can_choose_recipient_country_spec.rb
+++ b/spec/features/staff/users_can_choose_recipient_country_spec.rb
@@ -14,7 +14,7 @@ RSpec.feature "Users can choose a recipient country" do
 
     context "with JavaScript disabled" do
       scenario "countries are choosen from a select box" do
-        expect(page).to have_select(I18n.t("page_title.activity_form.show.country"))
+        expect(page).to have_select(I18n.t("activerecord.attributes.activity.recipient_country"))
       end
 
       scenario "choosing a recipient country sets a recipient region associated to that country" do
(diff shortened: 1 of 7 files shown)


lozette:
I am going to hold off fixing the tests on this one until #409 is merged as that changes some of the labelling/hint texts

duffylouise:
No problem. There might be a couple more PRs coming your way this afternoon.

spec/features/staff/users_can_choose_recipient_country_spec.rb
```diff
@@ -14,7 +14,7 @@
 
     context "with JavaScript disabled" do
       scenario "countries are choosen from a select box" do
-        expect(page).to have_select(I18n.t("page_title.activity_form.show.country"))
```
tahb, line 17: I wasn't expecting page_title to be affected by any changes we made to the form builder gem. The `@page_title` is set in the `ActivityFormController` and then passed to the `_wrapper` on each form step. This allows us to keep our form titles different from individual field names.

Having a quick look I think we've accidentally broken page titles for country prior to this work?

Our en.yml says:

```
  page_title:
    activity_form:
      show:
        country: What country will benefit from this activity?
```

But the form view on staging just has "country" and I think we want the former which would make these switches to activerecord.etc no longer needed!
Screenshot 2020-05-29 at 10 58 06

Page titles seem to be working for other fields like identifier and sector.

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
```diff
@@ -141,7 +141,7 @@
 
         choose "Primary education"
         click_button I18n.t("form.activity.submit")
-        expect(page).to have_content I18n.t("page_title.activity_form.show.status")
+        expect(page).to have_content I18n.t("activerecord.attributes.activity.status")
```
tahb, line 144: If the page title for status is staying as "Status" then I'd be looking to see why our pape_title code in the wrapper stopped working. If we stop using page title we lose the ability to change this value to something other than the active record attribute name, which removes the previous flexibility.

lozette:
I'm going to close this PR and re-start it fresh

duffylouise:
Good thinking - thanks.

PR #418, opened 2020-05-27
(672) Provider organisation pre filled for transactions

mec:
## Changes in this PR
The providing organisation of a transaction can be derived from the activity level and the delivery partner organisation type.

- For government orgs all providing organisation at all levels is BEIS
- For non-governement orgs all proviging organisation at all level is BEIS, excpet Level D (Third-party projects) where it is the delivery partner.

This work builds on previous work done on planned disbursements which follows the same set of rules.

We could remove the providing org altogether and not ask the user at all, but it was felt we can build our confidence that this is a good call be testing with users before we remove it entirely, there may be edge cases we are not aware of and keeping the form means a user can override the values if needed.

## Screenshots of UI changes
Screenshot_2020-05-27 Add a planned disbursement — Report your official development assistance(1)

Screenshot_2020-05-27 Add a planned disbursement — Report your official development assistance

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,6 +142,7 @@
 - Planned disbursement dates are validated within boundaries
 - Ingest tool creates or updates existing projects if they match on the IATI identifier
 - Ingest tool will try to set more meaningful identifiers for projects
+- The providing organisation is pre-filled in for new transactions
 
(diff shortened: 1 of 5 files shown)


mec:
[Provider organisation is pre-filled for transactions](https://trello.com/c/XG2VeDUs/672-provider-organisation-is-pre-filled-for-transactions)

PR #422, opened 2020-05-27
Script the ingesting of UKSA IATI data onto production environments

tahb:
## Changes in this PR

- ~~this is blocked until the following data migration has been run https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/db/data/20200521165254_add_iati_ids_to_existing_uksa_data.rb through a production deploy~~ this has now happened https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/434
- run the task to automatically ingest historic data from UKSA
- use of the fixture as the source - IATI data has not been changed or added to by BEIS in the last 6 months. The fixture we've used for testing was also used to set up the pentest environment which has had it's data verified by Dan https://dxw.slack.com/archives/CN0H2L066/p1590769465039200
- this task will fail if there is no organisation with the `iati-reference` for UKSA: `GB-GOV-EA31`
- a previous data migration has been run on production that enables this task to perform the necessary _update or create_ https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/db/data/20200521165254_add_iati_ids_to_existing_uksa_data.rb#L2

Even though we have automatic database restore available through GPaaS I have taken a manual database restore just in case something unexpected happens. Should we need to restore our data schema will be reverted too which will allow this task to rerun into a broken state. We could ensure that the next deployment to production included a fix or removal of this script to put the service back to a safe starting position.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -160,6 +160,8 @@
   element is missing
 - Ingest tool fails loudly if any activity fails to be created
 
+- Ingest UKSA data from IATI
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-7...HEAD
 [release-7]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-6...release-7
 [release-6]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-5...release-6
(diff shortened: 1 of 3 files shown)


tahb:
Odeta has flagged a last minute potential problem with the ingest tool that we might be better off fixing before this goes in https://dxw.slack.com/archives/CN0H2L066/p1591022224022000

PR #425, opened 2020-05-28
(579) Fix mobile navigation

mec:
## Changes in this PR
Picked this bug up from the backlog as I had nothing else that looked suitable.

I've fixed the mobile navigaiton whilst keeping the exsiting header design, but as it does not follow the GOVUK pattern I also updated the whole header in the last two commits – if this is not something we think is right we can drop those.

This does mean updating a few specs and that:

- the sign in link is shown as the green button on the start page only
- the user name is not shown in the header

I am not sure the rational for these departures from the pattern (@deberny maybe you know), I did look at through the git history to try to find out.

Docs for the header component are here:

https://design-system.service.gov.uk/components/header/

### Before
Screenshot_2020-05-28 Organisation Department for Business, Energy and Industrial Strategy — Report your official developme (3)

Screenshot_2020-05-28 Organisation Department for Business, Energy and Industrial Strategy — Report your official developme (2)

### After
Screenshot_2020-05-28 Organisation Department for Business, Energy and Industrial Strategy — Report your official developme

Screenshot_2020-05-28 Organisation Department for Business, Energy and Industrial Strategy — Report your official developme (1)

## Next steps


mec:
[Sub navigation doesn't work for smaller browser sizes](https://trello.com/c/pgLBsmHg/579-sub-navigation-doesnt-work-for-smaller-browser-sizes)

mec:
> The mobile menu now works fine as expected. Was it the lack of #navigation on
> 
>     that made it broken?

Yeah, and declaring the `aria` twice in the haml, we declared it and then declared it again, essentially overwriting the previous value - I think both are required for it to work as expected?

I interntionally kept the fix in the first commit 0e90e1c519b095c70f54bd82680e480d54a06ab6 so you can see the difference.

mec:
@deberny could you take one more look at this, I forgot to delete the now redundant styles in 5a1a7dfc8f7ddfd8b5e6f4583f91903ab28d26b4! Sorry!

PR #426, opened 2020-05-28
(693) Ingest AMS Newton fund data into RODA from IATI dataset 

lozette:
## Changes in this PR

Trello: https://trello.com/c/d4R5Aqlb/693-get-data-into-roda-from-iati-for-ams-newton-dataset

This PR:
* Adds the legacy IATI data sets for AMS Newton 
* Adds the project -> programme mapping for AMS Newton fund data
* Runs a data migration to ingest the AMS Newton fund data

We have run a script on `staging` to create the necessary programmes (listed in the mapping file) so that the data migration will run on staging deploy. The programmes will need to be created on `production` before a production deploy.

## Next steps


spec/fixtures/activities/ams/gcrf/gcrf_real_and_complete_legacy_file.xml
```diff
@@ -0,0 +1,1460 @@
+
```
tahb, line 1: If this file isn't used as part of the Newton ingest I wonder if it should wait to be included in the work that makes the GCRF ingest possible, keeping the related changes together.

spec/fixtures/activities/ams/newt/newt_real_and_complete_legacy_file.xml
```diff
@@ -0,0 +1,9687 @@
+
```
tahb, line 1: I'm slightly regretting my past decision to use fixtures for the real import. It technically works because we trust that the IATI files aren't changing but it feels misleading to the reader. Do you think it's an issue and if so worth doign something about? Should we have these files timestamped in vendor maybe?

spec/fixtures/activities/ams/gcrf/gcrf_real_and_complete_legacy_file.xml
lozette: Gone!

spec/fixtures/activities/ams/newt/newt_real_and_complete_legacy_file.xml
lozette: Yeah fixtures doesn't feel right for the "real" data tbh. I think moving them to vendor with the IATI codelists is a good shout. 

Should that be done here or in a separate PR moving the UKSA ones as well?

tahb: Okay let's do it then! The UKSA ship has sailed, the data migration documents what it used at the time so it might be best to leave that as an imperfection and focus on making it better for AMS in this PR.

PR #428, opened 2020-05-29
(fix) Tidy up old Activities which may not have a reporting_organisation

lozette:
## Changes in this PR

Trello: https://trello.com/c/VAjGDwZs/671-practice-ingesting-uksa-iati-data-ontop-of-production-data#comment-5ecfc215edfc381404373ecd

While working on the pentest environment we discovered some Activities with a nil
reporting_organisation. These records are old and we are no longer able to create
activities without a reporting organisation.

This data migration ensures there are no old records without a reporting organisation.

It follows the same rules as those in the Create[Level]Activity services.

It is irreversible because Activities are not valid without a reporting organisation.

## Next steps

### DIFF
--- a/db/data/20200529093419_add_reporting_org_to_activities_where_missing.rb
+++ b/db/data/20200529093419_add_reporting_org_to_activities_where_missing.rb
@@ -1,0 +1,25 @@
+class AddReportingOrgToActivitiesWhereMissing < ActiveRecord::Migration[6.0]
+  def up
+    activities = Activity.where(reporting_organisation: nil)
+    service_owner = Organisation.find_by_service_owner(true)
+    activities.each do |activity|
+      creating_organisation = activity.organisation
+      reporting_organisation = case activity.level
+      when :fund
+        creating_organisation
+      when :programme
+        creating_organisation
+      when :project
+        service_owner
+      when :third_party_project
+        service_owner
+      end
+      activity.reporting_organisation = reporting_organisation
+      activity.save!
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/db/data_schema.rb
+++ b/db/data_schema.rb
@@ -1,1 +1,2 @@
-DataMigrate::Data.define(version: 20200507142459)
+# encoding: UTF-8
+DataMigrate::Data.define(version: 20200529093419)


db/data/20200529093419_add_reporting_org_to_activities_where_missing.rb
```diff
@@ -0,0 +1,25 @@
+class AddReportingOrgToActivitiesWhereMissing < ActiveRecord::Migration[6.0]
+  def up
+    activities = Activity.where(reporting_organisation: nil)
+    service_owner = Organisation.find_by_service_owner(true)
+    activities.each do |activity|
+      creating_organisation = activity.organisation
+      reporting_organisation = case activity.level
+      when :fund
+        creating_organisation
```
tahb, line 9: These always scramble my brain. I _think_ fund and programme always needs to have BEIS as the reporter, however project and sub-project can always be the organisation that created it? If so these should be flipped around.

tahb: Looking at Meyrics mapping table, the reporter is always BEIS! https://miro.com/app/board/o9J_kv27qEM=/?moveToWidget=3074457347355940764&cot=13 What do we set for UKSA when they create a project on our latest version?

lozette: I took the logic from the Create[Level]Service classes, in which case are they wrong? https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/services/create_fund_activity.rb#L11

lozette: For projects it varies according to https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/services/create_project_activity.rb#L44

lozette: This is to tidy up two very old activities which are otherwise entirely nil

tahb: Understood 👍 As we know funds and programmes can only be created by BEIS, the BEIS org will only ever be the `creating_organisation` so it becomes equivalant to `service_owner`. I think this is fine as it'll only be run once 👍

PR #429, opened 2020-05-29
(fix) Transaction date ingest element

mec:
## Changes in this PR
The value-date element is different to the transaction-date element.
Whilst they are likely to be the same, we export the transaction-date as
the value-date ourselves, we should ingest from the correct xml element.
Transaction reference docs:

http://reference.iatistandard.org/203/activity-standard/iati-activities/iati-activity/transaction/

### DIFF
--- a/app/services/ingest_iati_activities.rb
+++ b/app/services/ingest_iati_activities.rb
@@ -121,7 +121,7 @@ class IngestIatiActivities
     transaction_elements = legacy_activity.elements.select { |element| element.name.eql?("transaction") }
     transaction_elements.each do |transaction_element|
       currency = transaction_element.children.detect { |child| child.name.eql?("value") }.attributes["currency"].value
-      date = transaction_element.children.detect { |child| child.name.eql?("value") }.attributes["value-date"].value
+      date = transaction_element.children.detect { |child| child.name.eql?("transaction-date") }.attributes["iso-date"].value
       value = transaction_element.children.detect { |child| child.name.eql?("value") }.children.text
       transaction_type = transaction_element.children.detect { |child| child.name.eql?("transaction-type") }.attributes["code"].value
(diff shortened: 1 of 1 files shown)


app/services/ingest_iati_activities.rb
```diff
@@ -121,7 +121,7 @@ def call
     transaction_elements = legacy_activity.elements.select { |element| element.name.eql?("transaction") }
     transaction_elements.each do |transaction_element|
       currency = transaction_element.children.detect { |child| child.name.eql?("value") }.attributes["currency"].value
-      date = transaction_element.children.detect { |child| child.name.eql?("value") }.attributes["value-date"].value
+      date = transaction_element.children.detect { |child| child.name.eql?("transaction-date") }.attributes["iso-date"].value
```
tahb, line 124: I'm surprised a test didn't need to change to support this change in some way. Perhaps there was no existing assertion for transaction dates? Did dates not work at all? If not, should we add one?

mec: I don't think there is - we don't check the right element was used, only that the value is the expected one. In our fixtures all the values match up so `transaction-date == value-date` - this is what lead me to it - I changed the transaction-date in the fixture and expected failure but got none.

If we think we should ingest `value-date` as it's own value inside the application, we will have to add a new column to store it as we currently use the value of transaction-date to fill in value-date, see the xml view here:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/3643bdf810550a7def86023033e54397c7993674/app/views/staff/shared/xml/_transaction.xml.haml#L4

Worth noting that value-date is described as

> The date to be used for determining the exchange rate for currency conversions.

and we have no way to handle currency changes so I am not 100% how much value there is?

Perhaps a question for BEIS to decide the risk vs the effort?

PR #430, opened 2020-05-29
(682) Fix transaction date boundaries

mec:
## Changes in this PR
We have seen instances of users inputting unlikely, but valid years,
0202 for example.

We have a date_within_boundaries custom validator already that will help
here.

The validator will stop future dates being more than 25 years in the
future from the current date, but this feels reasonable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -147,6 +147,8 @@
 - Switched to the latest form builder gem version from our fork
 - Planned disbursement create and update actions are recorded
 - User role hint text is shown
+- Transaction dates are validated to be no more than 10 years ago and 25 years
+  in the future
 
 ## [unreleased]
 - The IATI identifier on an activity, transaction, planned disbursement,
--- a/app/models/transaction.rb
+++ b/app/models/transaction.rb
@@ -14,5 +14,5 @@ class Transaction < ApplicationRecord
     :receiving_organisation_name,
     :receiving_organisation_type
   validates :value, inclusion: 0.01..99_999_999_999.00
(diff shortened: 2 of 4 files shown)


mec:
[Transaction dates should not be valid so far in the past](https://trello.com/c/9MnJrf9N/682-transaction-dates-should-not-be-valid-so-far-in-the-past)

mec:
> This looks to fix the problem concisely 👍 Have you thought about what we can do for our production data that is now in an invalid state?

Bah! Good thinking! :)

Okay, we haven't ingested the UKSA legacy data to production yet so my first response is that as UKSA have agreed to verify that data, maybe they could pass over all their data fixing this as they go?

I mean we assume the `0202` was supposed to be `2020` but is it up to us to make that decision?

tahb:
If memory serves this transaction is actually one manually created by UKSA and not one that came in during the ingest. I don't know if it's limited to only 1 transactions. If it is, perhaps a manual correction with the help of Dan to production might be enough?

mec:
I am going to connect to production and run a sandbox console to try and see how many records are in this state, then we can make a better informed decision.

mec:
I've checked on production and only 2 Transactions have dates before 1900, I've got the IDs so can give the URLs to Dan to make the fix.

mec:
If we get this work into production before we run the ingest, it should stop the next ingest from allowing unlikely dates like `0202`

mec:
Dan has fixed the two records in production that UKSA put in, I will merge this now. I am working on the assumption we are manually managing when we run the ingest on production.

PR #435, opened 2020-06-01
(581) strip whitespace from iati identifier

mec:
## Changes in this PR
We have seen an instance where leading whitespace in the IATI
Identifier value caused the XML to fail the IATI validator.

The IATI spec states 'The iati-identifier should have no leading or
trailing whitespace.' therefore we should strip the whitespace before we
save the value.

http://reference.iatistandard.org/203/activity-standard/overview/iati-identifier/

The strip_attributes gem was chosen:

https://github.com/rmm5t/strip_attributes

This was after much discussion around the various approaches could
take. We felt the gem was simple enough and gave us the functionality we
wanted without a large overhead. Under the hood, it uses the before_save
callback, which was one of the preferred approaches, but the gem has a
nice interface to allow targeting the attributes to act upon.

Configure RSpec to use the strip_attribute matchers so we can use them
in our specs.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,8 +149,9 @@
 - User role hint text is shown
 
 ## [unreleased]
-- The IATI identifier on an activity is stripped of leading and trailing
-  whitespace
+- The IATI identifier on an activity, transaction, planned disbursement,
+  organisation and implementing organisation  is stripped of leading and
+  trailing whitespace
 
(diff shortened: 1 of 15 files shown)


mec:
[Organisation IATI references can break the IATI validator error due to blank spaces before or after](https://trello.com/c/T9V76aBC/581-organisation-iati-references-can-break-the-iati-validator-error-due-to-blank-spaces-before-or-after)

PR #436, opened 2020-06-01
Improve "create activity" content

lozette:
## Changes in this PR

Trello: https://trello.com/c/4bIfITjU/669-improve-create-fund-content-dates-recipient-region-and-country

Improve the labelling for the Create Activity form steps.

This ticket has involved changing many of the strings used in the Activity
create form. Previously the page title strings were being used as labels for
the fields/fieldsets in the create activity form, but this is no longer
feasible as the content for page titles & fieldset labels need to be different.

## Next steps

### DIFF
--- a/config/locales/models_and_forms.en.yml
+++ b/config/locales/models_and_forms.en.yml
@@ -118,10 +118,17 @@ en:
       activity:
         actual_end_date: Actual end date (optional)
         actual_start_date: Actual start date (optional)
+        dates: What are the start and end dates for this %{level}?
+        flow: What is the flow type?
+        identifier: Enter a unique identifier
         planned_end_date: Planned end date
         planned_start_date: Planned start date
+        purpose: What is the purpose of the %{level}?
+        recipient_country: What country will benefit from this activity?
+        recipient_region: What region will benefit from this activity?
         sector_category:
           html: What area of the economy or society is your %{level} helping? For example, research, education or small to medium-sized enterprise (SME) development. Choose one of the <a href='https://www.oecd.org/dac/stats/documentupload/2015%20CRS%20purpose%20codes%20EN_updated%20April%202016.pdf' target='_blank' class='govuk-link'>CRS purpose codes</a>
+        status: What is the status of the %{level}?
       user:
         organisation_ids: Organisations
         role: Role
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,6 +150,8 @@
 
 ## [unreleased]
 
+- Improve content on "Create Activity" form steps
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-7...HEAD
 [release-7]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-6...release-7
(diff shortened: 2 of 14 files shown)


config/locales/models_and_forms.en.yml
```diff
@@ -118,10 +118,17 @@ en:
       activity:
         actual_end_date: Actual end date (optional)
         actual_start_date: Actual start date (optional)
+        dates: What are the start and end dates for this %{level}?
```
tahb, line 121: I think the `helpers.fieldset.x` key is used by the govuk form builder gem to automatically match onto the HTML element `fieldset` without us needing to specify it in each view. I also think `fieldset` is only used for radio form fields like status, as well as dates, so these might be different elements.

I reckon we could add locale values which are then automatically picked up by the gem. If we place them in `helper.label` or `helper.fieldset` following [the gem documentation](https://govuk-form-builder.netlify.app/building-blocks/localisation/#customising-locale-structure) we could remove all the logic we're adding to the views here.

```
helpers:
  label:
    person:
      role: What role do you play?
```

tahb: This list now looks similar to the `page_title` copy [over in en.yml](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/config/locales/en.yml#L347). Do we need to remove or change what's in that file?

What is the approach to our content now? Is it

- `en.page_title` is used only for the browser page title
- `helpers.x is used for form labels/titles
- `activerecord.attributes` are used for error messages

mec: I don't think we can use the formbuilder gem 'magic' when we are intopolating the string - but I agree that it would be good to settle on some kind of structure for all the keys.

tahb: Ah, something else that is relevant: we recently make a change to the gem config to point it to our form labels in `activerecord.attributes` instead following Leekys earliy lead. https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/config/initializers/form_builder.rb#L10. This was intended so that the form labels and their error message would always match up https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/config/locales/models_and_forms.en.yml#L9.

lozette: Yes I tried editing the `activerecord.attributes` labels but that f*cked up the error messages, so you'd get something like "*What are the start and end dates for this project? can't be blank*" on error. Hence adding the fieldset legends instead.

lozette: > This list now looks similar to the page_title copy over in en.yml. Do we need to remove or change what's in that file?

Well no because Louise did that in the first version of this PR (which I closed) and it messed up the page titles 🤦

lozette:
@mec "I'm still not loving internationalize all the things™,"

I agree as it's getting really complex but we're going to be in trouble if they want a Welsh version and we don't have it 🤷

mec:
> @mec "I'm still not loving internationalize all the things™,"
> 
> I agree as it's getting really complex but we're going to be in trouble if they want a Welsh version and we don't have it 🤷

croesi pontydd pan gyrhaeddwn nhw rhag ofn na wnawn ni hynny? 😄

mec:
Replace by #456

PR #438, opened 2020-06-01
(668) improve create fund content - status

lozette:
## Changes in this PR

To be merged *after* https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/436

Trello: https://trello.com/c/rR0rbCSQ/668-improve-create-fund-content-economic-societal-area-and-status

This text has been lifted from IATI's official codelists, so if we ever update
the codelists from 2.03, this change might be undone.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,6 +150,8 @@
 
 ## [unreleased]
 
+- Improve content on "Create Activity" form steps
+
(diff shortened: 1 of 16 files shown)


mec:
These all sit on top of one and other so it's been a bit of a trek to get here, but reading the card I was expecting 'SME' to be 'Small Medium Enterprise' (or whatever it stands for) at this point?

mec:
Replaced by #456

PR #442, opened 2020-06-03
(hotfix) (pentest only) Run db:schema:load before running migrations

lozette:
2020-06-03 a failed manual terraform deploy means the database on pentest
needs to have its schema reloaded before migrations are run.


lozette:
(Don't merge)

PR #443, opened 2020-06-03
Fix application unable to start on live environment that has a freshly imported database

tahb:
## Changes in this PR

After our recent problems resyncing the pentest environment this pull requests seeks to include a fix and document the steps we've taken in order to make it easier next time.

- include seed for pentest environment users as we are going to be wiping this environment a lot of the coming sprints
- all `data:migrate` (not to be confused with the more common `rake db:migrate`) to be toggled off to fix a problem with apps crashing when starting on a fresh database
- document the latest paas data migration steps including a different nuclear destroy the old postgres instance


docker-entrypoint.sh
```diff
@@ -12,9 +12,16 @@ setup_database()
 
 run_data_migrations()
 {
-  echo "Running data migrations…"
-  bundle exec rake data:migrate
-  echo "Finished running data migrations."
+  if [ "$DATA_MIGRATE" == "false" ];
+  then
+    echo "Running data schema load…"
+    bundle exec rake data:schema:load
```
tahb, line 18: Only if it's explicitly set to `false` do we do the unusual route of only loading the schema. All other times the default behaviour should be the previous `data:migrate` as we trust the state of the schema to manage what migrations need to be run.

doc/database-backup-and-restore.md
```diff
@@ -39,34 +39,31 @@ You will need to be a user of the services GPaaS account. You're OK if you can s
...
+    cf unbind-service beis-roda-pentest-worker beis-roda-pentest-postgres
+    cf delete-service beis-roda-pentest-postgres
     ```
-1. Connect to the target database with a PostgreSQL prompt
+1. Use Terraform to recreate it (this may take 10-20 minutes)
     ```
-    cf target -s staging
-    cf services # to get the SERVICE_NAME for the destination postgres
-    cf conduit SERVICE_NAME -- psql
+    terraform apply --var-file=pentest.tfvars --var 'data_migrate=false'
```
tahb, line 50: Instead of documenting steps to update the tfvars correctly I've included a deliberate `var` flag on the end of this command to ensure for any deploys after a data wipe it doesn't migrate the data.

lozette:
Thank you for this, it all makes a lot of sense (and I would never have figured this out on my own).

PR #446, opened 2020-06-04
(705) Refactor locale files – part 1

mec:
## Changes in this PR
This is the first part of the work to refactor our locales into better shape to help the team manage them.

The work has been discussed with the team and thsoe thoughts are here:

https://miro.com/app/board/o9J_kv27qEM=/?moveToWidget=3074457348258073280&cot=13

Part 1 delivers:

- start to bring the new structure to the locales on disk
- extract any pages into their own files in a views directory to contextualise them
- configure Rails (8191d7b)
- stop enforcing I18n-task normalization (2165b2a)
- move the old 'generic' strings into the new default domain
- start to be clear about html page titles and content page titles as we think the may be different

## Next
- Work to extract each model to it's own file
- configure the GOVUK form builder to use the new model files
- contrinue to extract strings from `en.yml` to the new structure
- generally make the locales more approachable (remains to be seen! 😄 )

After this we can get back to @duffylouise cards and get them sorted.

### DIFF
--- a/app/views/public/visitors/index.html.haml
+++ b/app/views/public/visitors/index.html.haml
@@ -1,19 +1,19 @@
-=content_for :page_title_prefix, t("page_title.home")
+=content_for :page_title_prefix, t("start_page.document_title")
 
 %main.govuk-main-wrapper#main-content{ role: "main" }
   .govuk-grid-row
     .govuk-grid-column-two-thirds.app-visitor-welcome
       %h1.govuk-heading-xl
-        = t("page_title.welcome")
+        = t("start_page.title")
 
       %p.govuk-body
-        = t("page_content.start_page.introduction")
+        = t("start_page.introduction")
 
       %p.govuk-body
-        = t("page_content.start_page.use_this_page_to")
+        = t("start_page.use_this_page_to")
 
       %ul.govuk-list.govuk-list--bullet
-        - t("page_content.start_page.uses").each do |use|
+        - t("start_page.uses").each do |use|
           %li
             = use
 
--- a/app/views/public/visitors/index.html.haml
+++ b/app/views/public/visitors/index.html.haml
@@ -18,6 +18,6 @@
             = use
 
       = button_to '/auth/auth0', class: 'govuk-button', role: 'button', draggable: 'false' do
-        = t("generic.link.sign_in")
+        = t("header.link.sign_in")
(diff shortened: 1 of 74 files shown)


app/views/public/visitors/index.html.haml
```diff
@@ -1,23 +1,23 @@
-=content_for :page_title_prefix, t("page_title.home")
+=content_for :page_title_prefix, t("start_page.document_title")
```
tahb, line 1: You mentioned we could look at making these "title" locales more explicit to help distinguish between browser windows and those on the page. Is that what we mean here with "document"?

config/locales/default.en.yml
```diff
@@ -0,0 +1,23 @@
+---
```
tahb, line 1: This file feels much better. A file for default/shared/global/generic pieces of content that do not require context of any particular model or view.

app/views/public/visitors/index.html.haml
mec: Yes, I tried `html_document_title` but it felt too techincal. I referenced this on MDN to try and see what lanaguage was used there: 

https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML

If you look in the locale file `config/locales/views/start.en.yml` you'll see how the difference would play out - there is no difference in the strings so there is repetition but right now we think that is okay.

config/locales/views/cookie_statement.en.yml
```diff
@@ -0,0 +1,27 @@
+---
```
tahb, line 1: Again isolating all of these locals that are specific to a single static view is a good step foward. There are quite a lot keys here so grouping them all under a page would help me feel confident in making a change without having to hunt through whatever else in the big en.yml might be related 👍

app/views/public/visitors/index.html.haml
tahb: Roger! Do you think it'd be a bridge too far to refactor/rename the `page_title_prefix` into `document_title_prefix` to follow this lead? It might also make sense to come back later as there's also `@page_title` being created and passed around.

mec: no, it's all to play for (as ever) - I need to keep things small and managable, but I'll add that to the list.

mec:

PR #447, opened 2020-06-05
(705) Refactor locale files part two

mec:
## Changes in this PR
Hot on the heels of #446 comes this epic tale of refcatoring the locales! 😄 

The goal of this work is to get as many translation string out of the existing `en.yml` and into the new locale file structure.

I want to stop here as this is already 17 (small and nice) commits so this PR is only about moving the keys without re-organising them (that will follow!)

As our Acitivy models represents four variations we could have changed the `models` directory to something else (`entities` perhaps?) but I decided to keep it – I think most deveopers will get the intention behind this decision.

The `en.yml` now only contains the `page_titles` that are hopefully the browser window tites (and may be used as other content) so the next step will include moving those to the model files and renaming to `document_title` in-line with the plan, which is documented here:

https://miro.com/app/board/o9J_kv27qEM=/?moveToWidget=3074457348258073280&cot=13

and in the last PR is this series (#446)

So next up:

- browser page titles moved
- changing the key layout where beneficial
- re-organising `models_and_forms.yml` to the new world view
- configure the GOVUK form builder to use the new model files
- further optimisations


mec:

PR #449, opened 2020-06-05
(691) Address ingested activities with missing mandatory dates

lozette:
## Changes in this PR

Trello: https://trello.com/c/bHOqDiH7/691-address-activities-with-missing-mandatory-dates

This PR is in two parts, and only the second commit is strictly necessary to address the issue in the trello card.

When ingesting AMS Newton fund activities locally, we found that 17 of the activities in the IATI XML file had only `activity-date` elements of type 1 (planned start date) and 3 (planned end date). According to the ingest script, types 1 & 2 are expected, so these activities were not ingested successfully.

This change is addressed in 44aaff9652e72401ffcd4f05276ccfab706b0abb

However while fixing this, we noticed that according to the IATI standard, activity dates 1 and 2 are the mandatory ones - and *either* 1 (planned start) *or* 2 (actual start) are expected https://andylolz.github.io/iatistandard-multilingual/en/203/codelists/ActivityDateType/

So our ActiveRecord validations, which expect both 1 (planned start) and 3 (planned end) to be present, do not line up with the IATI standard.

This change is addressed in 10bbd9c6891b85f79840908f3dd4632a5e4ba757 but it is not necessary to allow successful ingestion of the AMS Newton fund activities, as they all have dates 1 & 3 present. 

We can safely drop commit 10bbd9c6891b85f79840908f3dd4632a5e4ba757 and ingest AMS Newton fund activities, but our ActiveRecord validations will not match the mandatory fields in IATI. This means in future we may encounter problems with delivery partners who want to input activities without a planned end date (which in IATI they are permitted to do) but are unable to in the RODA interface. 

## Next steps

### DIFF
--- a/app/services/ingest_iati_activities.rb
+++ b/app/services/ingest_iati_activities.rb
@@ -260,13 +260,13 @@ class IngestIatiActivities
 
   private def add_dates(legacy_activity:, new_activity:)
     planned_start_date_element = legacy_activity.elements.detect { |element| element.name.eql?("activity-date") && element.attributes["type"].value.eql?("1") }
-    new_activity.planned_start_date = planned_start_date_element.attributes["iso-date"].value || nil
+    new_activity.planned_start_date = planned_start_date_element ? planned_start_date_element.attributes["iso-date"].value : nil
     actual_start_date_element = legacy_activity.elements.detect { |element| element.name.eql?("activity-date") && element.attributes["type"].value.eql?("2") }
-    new_activity.actual_start_date = actual_start_date_element.attributes["iso-date"].value || nil
+    new_activity.actual_start_date = actual_start_date_element ? actual_start_date_element.attributes["iso-date"].value : nil
     planned_end_date_element = legacy_activity.elements.detect { |element| element.name.eql?("activity-date") && element.attributes["type"].value.eql?("3") }
-    new_activity.planned_end_date = planned_end_date_element.attributes["iso-date"].value || nil if planned_end_date_element
+    new_activity.planned_end_date = planned_end_date_element ? planned_end_date_element.attributes["iso-date"].value : nil
     actual_end_date_element = legacy_activity.elements.detect { |element| element.name.eql?("activity-date") && element.attributes["type"].value.eql?("4") }
-    new_activity.actual_end_date = actual_end_date_element.attributes["iso-date"].value || nil if actual_end_date_element
+    new_activity.actual_end_date = actual_end_date_element ? actual_end_date_element.attributes["iso-date"].value : nil
   end
 
   private def add_identifiers(legacy_activity:, new_activity:)
--- a/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
+++ b/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
@@ -153,8 +153,7 @@ RSpec.feature "Users can create a fund level activity" do
         expect(page).to have_content I18n.t("page_title.activity_form.show.dates")
 
         click_button I18n.t("form.activity.submit")
-        expect(page).to have_content "Planned start date can't be blank"
-        expect(page).to have_content "Planned end date can't be blank"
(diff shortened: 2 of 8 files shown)


spec/services/ingest_iati_activities_spec.rb
```diff
@@ -6,366 +6,387 @@
...
 
-        expect { service.call }.to raise_error(ActiveRecord::RecordInvalid)
+  context "AMS activities" do
+    context "Newton fund activities" do
+      it "handles activities without an `actual_start_date` (`activity-date` type 2)" do
+        ams = create(:organisation, name: "Academy of Medical Sciences", iati_reference: "GB-COH-03520281")
+        legacy_activities = File.read("#{Rails.root}/spec/fixtures/activities/ams/newt/fake_activity_with_dates.xml")
+        programme = create(:programme_activity, organisation: ams)
+        # Temporary fix until we have the actual programme-project mappings
+        allow_any_instance_of(LegacyActivity).to receive(:find_parent_programme).and_return(programme)
```
tahb, line 382: Nice stub! I think we should have stubbed the inital use of the mappings file in the same way for UKSA to decouple the 2. I reckon we'll have to use this more rather than add files and use it less.

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
```diff
@@ -154,7 +154,6 @@
 
         click_button I18n.t("form.activity.submit")
         expect(page).to have_content "Planned start date can't be blank"
```
tahb, line 156: Perhaps we should replace this error message with a new expectation that in this case an error appears around the `actual_start_date`?

Screenshot 2020-06-08 at 14 29 06

spec/services/ingest_iati_activities_spec.rb
```diff
@@ -6,366 +6,387 @@
   let!(:existing_programme) { create(:programme_activity, identifier: "GCRF-INTPART", organisation: beis) }
 
   describe "#call" do
-    it "creates 36 new projects" do
-      uksa = create(:organisation, name: "UKSA", iati_reference: "GB-GOV-EA31")
-      legacy_activities = File.read("#{Rails.root}/spec/fixtures/activities/uksa/real_and_complete_legacy_file.xml")
-
-      service_object = described_class.new(delivery_partner: uksa, file_io: legacy_activities)
+    context "UKSA activities" do
```
tahb, line 9: Tricky to follow the diff in this file to view the move versus the new test. I presume it is only the last test that was added and everything else is an indent, despite GitHub guessing that lots of lines were added and removed?

app/services/ingest_iati_activities.rb
```diff
@@ -260,13 +260,13 @@ def call
 
   private def add_dates(legacy_activity:, new_activity:)
     planned_start_date_element = legacy_activity.elements.detect { |element| element.name.eql?("activity-date") && element.attributes["type"].value.eql?("1") }
-    new_activity.planned_start_date = planned_start_date_element.attributes["iso-date"].value || nil
```
tahb, line 263: Thanks for cleaning up my bad use of `||` 💩

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
tahb: Now I look at it, I think there's 2 additional things to consider with the new view:

1. should planned_end_date now have "(optional)" in the label?
2. how do we label the `planned_start_date` and `actual_start_date` to inform the user that it requires one or the other?

spec/services/ingest_iati_activities_spec.rb
lozette: Yeah I think adding this new context has made the diff a nightmare to read, I will rejig it.

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
lozette: Added the `(optional)` tag to planned end date and added a better error message (with input from Louise)

PR #451, opened 2020-06-05
(705) Refactor locale files part three

mec:
## Changes in this PR
It's a trilogy! 😄 This is the draft of part 3 that includes:

- the final move of translations into their new model homes
- a refactoring of the Rails flash messages (5e244bd)

## There's more?!?! 😱 
Yes, I'm afraid so… Next up we'll (finally):

- add the planned new structure for forms, tables, document_titles, etc
- refactor any related helpers
- refactor the govuk form builder
- remove any resulting redundant translations

See you in part four…


config/locales/models/budget.en.yml
```diff
@@ -34,3 +34,34 @@ en:
     budget:
       edit: Edit budget
       new: Create budget
+  activerecord:
```
tahb, line 39: I've probably said this already but I'd like to say it again: I think having a file per resource like this is excellent. It is contextual and you can see how language about the same resource is used in different places without distraction 👌

config/locales/models_and_forms.en.yml
```diff
@@ -1,6 +0,0 @@
----
```
tahb, line 1: 🙌 Bye bye!

app/controllers/staff/extending_organisations_controller.rb
```diff
@@ -14,7 +14,7 @@ def update
 
     if @activity.valid?(:update_extending_organisation)
       @activity.save
-      flash[:notice] = I18n.t("form.activity.update.success")
+      flash[:notice] = I18n.t("action.#{@activity.level}.update.success")
```
tahb, line 17: Good catch 👍

mec:

PR #453, opened 2020-06-09
(705) Refactor locale files part four

mec:
## Changes in this PR
Part four (! 😪 ) of the refactor of the locales:

- configures the GOVUK formbilder to use a nice clear strututre of `label`, `legened` and `hint`
- updates every(!) form to use the new pattern
- updates each model to use the pattern as well as `table` and 'summary` for those components

## Still to do:
Gah!!
- introduce the concept of document title (the browser tab) and page title (the content on the page)
- come up with a good way to organise what is left


mec:

mec:
2. is a quick fix.
1. I'll have a look - I know we do something with these errors...

mec:
@tahb I think it is good to use the `activerecord.attributes` key to override these attribute names - I've done it here for focus area and focus area category - I still need to go back over and see if I missed any others and try and understand the code that enables this behaviour, I think you and @LornaHarwood added it?

PR #459, opened 2020-06-11
(643) Flag incomplete activities in the activity table views

lozette:
## Changes in this PR

Trello: https://trello.com/c/2hbGT0GG/643-user-can-see-the-reporting-status-of-an-activity-in-all-tables

To ensure a user can quickly see which Activities are missing required attributes,
we have added a complete/incomplete flag to the Activity tables on organisation & activity
views.

I have used the Gov.uk tag pattern here https://design-system.service.gov.uk/components/tag/
but have had to add CSS overrides t get the colours working correctly (after pairing with @deberny )

### After

Screenshot 2020-06-11 at 16 58 15

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,8 +168,10 @@
 - Amend Activity date validations - either `planned_start_date` *OR* `actual_start_date`
   must be present, in line with the IATI `activity-date` XML standard
 - Amend ingest service to successfully ingest Activities without an `activity-date` type
+  2 (`actual_start_date`) 
 - XML download does not contain any incomplete activities    
-- Ingest AMS Newton fund data from IATI  
+- Ingest AMS Newton fund data from IATI   
+- Flag incomplete activities in the activity table views
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-8...HEAD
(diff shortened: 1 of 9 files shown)


deberny:
@lozette I've noticed there is no title for the column in status. I know we haven't probably found the right language for it yet. Can @duffylouise help? 

The warning text pattern work really well when it breaks out large blocks of text, to stand out important information that may have legal consequences. 

For the table, could you try and use the 'Tag' pattern https://design-system.service.gov.uk/components/tag/ , it might work better?


Screenshot 2020-06-11 at 13 01 54

On the side: Do we have a ticket anywhere to change the generat 'Title' table column heading to 'Programme title' as per content audit?
Screenshot 2020-06-11 at 12 56 24

duffylouise:
I've got the 'title' thing as a Trello card so should be able to update that soon. And happy to help with the other stuff

lozette:
@duffylouise Would you be able to have a think as to what the title of the Complete/Incomplete column would be?

It's indicating if all the required information for an activity has been added. We could use "Status" but an activity already has a status (which is what state the activity is in, in IATI terms). This is the "status" of the Activity in RODA (as in do we have all the required information or not). 

Do you think it's safe to use "Status" here as well? Maybe something else?

duffylouise:
I think we should probably avoid using "status" as it's used elsewhere so what about "stage" instead? Does that work?

lozette:
Thanks both 👍

PR #462, opened 2020-06-12
(725) Activity region & country names are exposed in a `narrative` element in the XML

lozette:
## Changes in this PR

Trello: https://trello.com/c/c2twcAzs/725-expose-the-country-region-name-as-a-narrative-element-in-the-xml

The `recipient-region` and `recipient-country` elements in the Activity XML
should contain the region/country name in a nested `narrative` element like so:

```
<recipient-region code="589" vocabulary="1">
        <narrative>Middle East, regional<narrative>
</recipient-region>
```

## Next steps

### DIFF
--- a/spec/helpers/codelist_helper_spec.rb
+++ b/spec/helpers/codelist_helper_spec.rb
@@ -122,10 +122,22 @@ RSpec.describe CodelistHelper, type: :helper do
       end
     end
 
+    describe "#region_name_from_code" do
+      it "returns the name of the region from a code number" do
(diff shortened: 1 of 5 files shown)


spec/helpers/codelist_helper_spec.rb
```diff
@@ -122,13 +122,25 @@
       end
     end
 
+    describe "#region_name_from_code" do
+      it "returns the name of the region from a code number" do
+        expect(helper.region_name_from_code("998")).to eq("Developing countries, unspecified")
```
tahb, line 127: Nice 👍

PR #464, opened 2020-06-15
A programme's implementing organisation is the same as its extending organisation

lozette:
## Changes in this PR

Trello: https://trello.com/c/usZXtaXu/501-programmes-have-an-implementing-organisation

When a programme has its extending organisation set, the same organisation is set
as the implementing organisation.

Previously we were not requiring an implementing organisation for programmes, but
if we are going to publish programmes to IATI we will need them to have an
implementing organisation set. This will always be the same as their extending
organisation.

The second commit contains a data migration to add the implementing organisations to existing programmes on staging/production.

## Next steps

### DIFF
--- a/app/controllers/staff/extending_organisations_controller.rb
+++ b/app/controllers/staff/extending_organisations_controller.rb
@@ -13,6 +13,7 @@ class Staff::ExtendingOrganisationsController < Staff::BaseController
     @activity.assign_attributes(extending_organisation_id: extending_organisation_id)
 
     if @activity.valid?(:update_extending_organisation)
+      set_implementing_organisation
       @activity.save
       flash[:notice] = I18n.t("action.#{@activity.level}.update.success")
       redirect_to organisation_activity_path(@activity.organisation, @activity)
@@ -22,6 +23,15 @@ class Staff::ExtendingOrganisationsController < Staff::BaseController
   end
 
   private
+
+  def set_implementing_organisation
+    extending_organisation = Organisation.find(extending_organisation_id)
+    implementing_organisation = ImplementingOrganisation.new(name: extending_organisation.name,
+                                                             organisation_type: extending_organisation.organisation_type,
+                                                             reference: extending_organisation.iati_reference,
+                                                             activity_id: activity_id)
+    implementing_organisation.save!
+  end
 
   def activity_id
     params[:activity_id]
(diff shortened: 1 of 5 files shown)


app/controllers/staff/extending_organisations_controller.rb
```diff
@@ -23,6 +23,19 @@ def update
...
+  def set_implementing_organisation
+    extending_organisation = Organisation.find(extending_organisation_id)
+    implementing_organisation = ImplementingOrganisation.new(name: extending_organisation.name,
+                                                             organisation_type: extending_organisation.organisation_type,
+                                                             reference: extending_organisation.iati_reference,
+                                                             activity_id: activity_id)
+    if implementing_organisation.valid?
+      implementing_organisation.save
+    else
+      flash[:notice] = I18n.t("activerecord.errors.models.activity.general")
```
lozette, line 35: I haven't added a test for this unhappy path as I'm not sure we "should" ever get here, as the extending organisation should exist and have all its attributes. If we do need a test, I'm not sure how to spec it out 🤔

mec: I think the test for this being hard is pointing out that we perhaps don't need the code we cannot test? As setting the implementing organisation is not a user facing action, I am not sure we need the if/else and message, maybe we just call `save!` if something is wrong we get an exception and in this case that feels right, if one of the extending_orgs is not valid something has gone wrong somewhere?

What do you think?

db/data/20200615144002_add_implementing_organisations_to_programmes.rb
```diff
@@ -0,0 +1,23 @@
...
+    programmes = Activity.programme
+    programmes.each do |programme|
+      next if programme.implementing_organisations.present?
+      next unless programme.extending_organisation_id
+      extending_organisation = Organisation.find(programme.extending_organisation_id)
+      implementing_organisation = ImplementingOrganisation.new(name: extending_organisation.name,
+                                                               organisation_type: extending_organisation.organisation_type,
+                                                               reference: extending_organisation.iati_reference,
+                                                               activity_id: programme.id)
+      if implementing_organisation.valid?
```
mec, line 12: It feels the same here, with no implementing org the record will not pass IATI validation, so perhaps `save!` ?

app/controllers/staff/extending_organisations_controller.rb
lozette: I think you're right and I'm being *too* cautious!

db/data/20200615144002_add_implementing_organisations_to_programmes.rb
mec: @lozette what is the thinking around not raising an exception during the ingest?

lozette: No you're right, we should raise an exception!

PR #472, opened 2020-06-18
Users can see codes when selecting aid type

lozette:
## Changes in this PR

Trello: https://trello.com/c/2xrjDcXk/477-delivery-partners-can-see-codes-when-selecting-aid-type

When selecting an aid type for an Activity, the aid type code is displayed
alongside the Aid Type name for clarity.

### After

Screenshot 2020-06-18 at 14 44 58

## Next steps

### DIFF
--- a/app/helpers/codelist_helper.rb
+++ b/app/helpers/codelist_helper.rb
@@ -17,11 +17,16 @@ module CodelistHelper
     objects
   end
 
-  def yaml_to_objects_with_description(entity:, type:)
+  def yaml_to_objects_with_description(entity:, type:, code_displayed_in_name: false)
     data = load_yaml(entity: entity, type: type)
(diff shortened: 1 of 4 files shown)


app/helpers/codelist_helper.rb
```diff
@@ -17,11 +17,16 @@ def yaml_to_objects(entity:, type:, with_empty_item: true)
     objects
   end
 
-  def yaml_to_objects_with_description(entity:, type:)
+  def yaml_to_objects_with_description(entity:, type:, code_displayed_in_name: false)
```
tahb, line 20: Given that all(?) the IATI yaml files have codes this seems relevant as a generic change to this method. I notice it's taking on a fair few responsibilities though so one for us to keep an eye on.

PR #480, opened 2020-06-22
Changes to purpose of programme heading, title, and hint text

duffylouise:
## Changes in this PR

Changed "Purpose of programme" to "What is the purpose of the programme?"
Changed "Title" to "Programme name"
Changed hint text "A short title that explains the fund's activity" to "A short title that explains the purpose of the programme"

## Next steps

### DIFF
--- a/app/views/staff/activity_forms/purpose.html.haml
+++ b/app/views/staff/activity_forms/purpose.html.haml
@@ -1,4 +1,7 @@
 = render layout: "wrapper" do |f|
   = f.govuk_fieldset legend: { text: t("form.legend.activity.purpose", level: t("page_content.activity.level.#{f.object.level}")), tag: :h1, size: "xl" } do
-    = f.govuk_text_field :title, label: { text: t("form.label.activity.title", level: t("page_content.activity.level.#{f.object.level}")).humanize }
+    = f.govuk_text_field :title,
+      label: { text: t("form.label.activity.title", level: t("page_content.activity.level.#{f.object.level}")).capitalize },
+      hint_text: t("form.hint.activity.title", level: t("page_content.activity.level.#{f.object.level}"))
+
     = f.govuk_text_area :description, rows: 5
--- a/app/views/staff/shared/activities/_activity.html.haml
+++ b/app/views/staff/shared/activities/_activity.html.haml
@@ -17,7 +17,7 @@
 
   .govuk-summary-list__row.title
     %dt.govuk-summary-list__key
-      = t("summary.label.activity.title")
+      = t("summary.label.activity.title", level: activity_presenter.level).capitalize
     %dd.govuk-summary-list__value
       = activity_presenter.title
     %dd.govuk-summary-list__actions
--- a/config/locales/models/activity.en.yml
+++ b/config/locales/models/activity.en.yml
@@ -44,7 +44,7 @@ en:
         planned_end_date: For example, 28 11 2020
         planned_start_date: For example, 27 3 2020
         sector_category: What area of the economy or society is your %{level} helping? For example, research, education or small to medium-sized enterprise (SME) development. Choose one of the <a href='https://www.oecd.org/dac/stats/documentupload/2015%20CRS%20purpose%20codes%20EN_updated%20April%202016.pdf' target='_blank' class='govuk-link'>CRS purpose codes</a>
-        title: A short title that explains the fund's activity
+        title: A short title that explains the purpose of the %{level}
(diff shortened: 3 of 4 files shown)


duffylouise:
[Improve "create programme" content - purpose of programme and value](https://trello.com/c/rI0cKhqL/674-improve-create-programme-content-purpose-of-programme-and-value)

config/locales/models/activity.en.yml
```diff
@@ -46,7 +46,7 @@ en:
         planned_end_date: For example, 28 11 2020
         planned_start_date: For example, 27 3 2020
         sector_category: What area of the economy or society is your %{level} helping? For example, research, education or small to medium-sized enterprise (SME) development. Choose one of the <a href='https://www.oecd.org/dac/stats/documentupload/2015%20CRS%20purpose%20codes%20EN_updated%20April%202016.pdf' target='_blank' class='govuk-link'>CRS purpose codes</a>
-        title: A short title that explains the fund's activity
+        title: A short title that explains the purpose of the programme
```
lozette, line 49: The activity level here has been changed from fund to programme. I actually think it should be `%{level}` so the activity type name can change with the level.

config/locales/models/activity.en.yml
```diff
@@ -77,7 +77,7 @@ en:
         stage: Stage
         status: Status
         third_party_projects: Third-party projects
-        title: Title
+        title: Programme name
```
lozette, line 80: Again this should be `%{level} name` because it might not always be a programme.

duffylouise: Should I do these again in another pull request?

duffylouise: If you have time, that would be super helpful.

mec:
@tahb @leeky I've resolved this for @duffylouise, could one of you take a look in @lozette abscence?

app/views/staff/activity_forms/purpose.html.haml
```diff
@@ -1,4 +1,7 @@
 = render layout: "wrapper" do |f|
   = f.govuk_fieldset legend: { text: t("form.legend.activity.purpose", level: t("page_content.activity.level.#{f.object.level}")), tag: :h1, size: "xl" } do
-    = f.govuk_text_field :title, label: { text: t("form.label.activity.title", level: t("page_content.activity.level.#{f.object.level}")).humanize }
+    = f.govuk_text_field :title,
+      label: { text: t("form.label.activity.title", level: t("page_content.activity.level.#{f.object.level}")).humanize },
+      hint_text: t("form.hint.activity.title", level: t("page_content.activity.level.#{f.object.level}"))
```
leeky, line 5: Should the `hint_text` also be `humanize`d ?

app/views/staff/shared/activities/_activity.html.haml
```diff
@@ -17,7 +17,7 @@
 
   .govuk-summary-list__row.title
     %dt.govuk-summary-list__key
-      = t("summary.label.activity.title")
+      = t("summary.label.activity.title", level: activity_presenter.level.humanize)
```
leeky, line 20: Not sure if it matters, but https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/480/files#diff-f0cc68c5ea5e4239efd88b1881df8dddR4 calls `humanize` on the whole I18n string rather than the level part, i.e.

`= t("summary.label.activity.title", level: activity_present.level).humanize`

app/views/staff/activity_forms/purpose.html.haml
mec: I think we are using humanize to simply capitalize the string, so here title needs capitalistion e.g. "Fund name" and the hint does not e.g. "A short title that explains the purpose of the fund".

I guess it should be `capitalize` in that case, but I am not sure it matters too much - depends on how much it offends you I guess?

app/views/staff/shared/activities/_activity.html.haml
mec: Again, I think both work and it could probably use `capitalize`.

mec:
@lozette go back to support! 😄 I think I just over-wrote yours! Soz!

@leeky I've changed this to use capitalize as that is the intent.

app/views/staff/activity_forms/purpose.html.haml
leeky: Fair enough, missed the difference between the hint and the label here 😟

PR #491, opened 2020-06-25
Increase application layout width

mec:
## Changes in this PR

The application uses a lot of horizontal space so it makes sense to
increase the width of the layout from the standard 960px to a larger
value.

Once we enter a public beta phase we can always tweak the value based on
user analytics.

## Next steps


mec:
[Increase application layout width](https://trello.com/c/SN37LlRx/750-increase-application-layout-width)

PR #494, opened 2020-06-26
Chore/add a tree structure to activity and test performance

tahb:
## Changes in this PR

- refactor the parent relationship of activities so that instead of `activity.activity_id` we have `activity.parent_id`. This follows a found convention for Rails gems that deal trees and enables the installation of the `acts_as_tree` with minimal impact
- refactor `activity.parent_activities` by instead telling the `acts_as_tree` gem. This gem allows us find out about ancestors and children for any given node without us having to have checks on the level of each activitiy to infer how many parents might exist - quite a fragile check. It also makes it easy to extend our hierarchy by adding or removing levels in the future.
- preload activity parents when querying collections of activities (unless it's a fund as there's no current parent)

As mentioned in the commit that added [acts_as_tree](https://github.com/amerine/acts_as_tree). I investigated Ancestry and Acts_as_nested as two alternative options first. Ready the commit for more info on why I ended up picking acts_as_tree.

## Screenshots of UI changes

I created a temporary view to demonstrate what the impact of maintaining a tree structure. For 1945 activities of all levels it took the browser 2.02 seconds in total to load - this reduced to 1.3 seconds after replacing the lost parent_id database index.

The code looked like this:

```
%table.govuk-table
  %thead.govuk-table__head
    %tr.govuk-table__row
      %th.govuk-table__header
        Title
      %th.govuk-table__header
        Fund
      %th.govuk-table__header
        Programme
      %th.govuk-table__header
        Project

  %tbody.govuk-table__body
    - @activities.each do |activity|
      %tr.govuk-table__row
        %td.govuk-table__cell= activity.title
        %td.govuk-table__cell= activity.parent_activities.first&.title
        %td.govuk-table__cell= activity.parent_activities.second&.title
        %td.govuk-table__cell= activity.parent_activities.third&.title

```

I think the case can be made that a page with 100+ activities all rendered at once is not usable, and that pagination or downloading to a spreadsheet would be more appropriate. I thought it was a useful exercise to see how performant the view would be in an extreme case and I think 2 seconds is a lot better than I feared.

Screenshot 2020-06-26 at 15 17 52

Rendered the same amount of content on `develop` without these changes the time is 5.3 seconds, up from 2 seconds. 

Screenshot 2020-06-26 at 15 22 23

Screenshot_2020-06-26 — Report your official development assistance

(there were sub-projects in this page but the screenshot cropped it :) )

### 

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -186,13 +186,14 @@
 - Remove 3 unwanted activities from production
 - Content changes to fields for transaction value and activity identifier
 - Increase the width of the application layout
-- Activity show content is show in tabs for financials and details
 
 ## [unreleased]
 
 - Ingest RS Newton fund data from IATI
 - Allow BEIS users to redact activities from the IATI XML file, and to
   easily see on the Organisation show page which Activities are redacted
+- Activity show content is show in tabs for financials and details
+- Refactor how we can ask activities for their parents
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-10...HEAD
 [release-10]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-9...release-10
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -59,6 +59,8 @@ Rails.application.configure do
     Bullet.enable = true
     Bullet.bullet_logger = true
     Bullet.raise = true # raise an error if n+1 query occurs
+    Bullet.add_whitelist type: :unused_eager_loading, class_name: "User", association: :organisation
+    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :organisation
     Bullet.add_whitelist type: :unused_eager_loading, class_name: "Transaction", association: :provider
     Bullet.add_whitelist type: :unused_eager_loading, class_name: "Transaction", association: :receiver
   end
--- a/db/migrate/20200626100128_change_activity_parent_relationship_name.rb
+++ b/db/migrate/20200626100128_change_activity_parent_relationship_name.rb
@@ -1,0 +1,27 @@
+class ChangeActivityParentRelationshipName < ActiveRecord::Migration[6.0]
+  def up
+    add_column :activities, :parent_id, :uuid
+    Activity.transaction do
+      Activity.all.each do |activity|
+        activity.parent_id = activity.activity_id
+        activity.save!(validate: false)
+      end
+    end
+    add_index :activities, :parent_id
+    add_foreign_key :activities, :activities, column: "parent_id", on_delete: :restrict
+    remove_column :activities, :activity_id
+  end
+
+  def down
+    add_column :activities, :activity_id, :uuid
+    Activity.transaction do
+      Activity.all.each do |activity|
+        activity.activity_id = activity.parent_id
+        activity.save!(validate: false)
+      end
+    end
+    add_index :activities, :activity_id
+    add_foreign_key :activities, :activities, column: "activity_id", on_delete: :restrict
+    remove_column :activities, :parent_id
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_06_23_132105) do
+ActiveRecord::Schema.define(version: 2020_06_26_100128) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
@@ -34,7 +34,6 @@ ActiveRecord::Schema.define(version: 2020_06_23_132105) do
     t.string "aid_type"
     t.string "form_state"
     t.string "level"
-    t.uuid "activity_id"
     t.string "funding_organisation_name"
     t.string "funding_organisation_reference"
     t.string "funding_organisation_type"
@@ -50,10 +49,11 @@ ActiveRecord::Schema.define(version: 2020_06_23_132105) do
     t.boolean "ingested", default: false
     t.string "legacy_iati_xml"
     t.boolean "publish_to_iati", default: true
-    t.index ["activity_id"], name: "index_activities_on_activity_id"
+    t.uuid "parent_id"
     t.index ["extending_organisation_id"], name: "index_activities_on_extending_organisation_id"
     t.index ["level"], name: "index_activities_on_level"
     t.index ["organisation_id"], name: "index_activities_on_organisation_id"
+    t.index ["parent_id"], name: "index_activities_on_parent_id"
     t.index ["reporting_organisation_id"], name: "index_activities_on_reporting_organisation_id"
   end
 
@@ -163,7 +163,7 @@ ActiveRecord::Schema.define(version: 2020_06_23_132105) do
     t.index ["role"], name: "index_users_on_role"
   end
 
-  add_foreign_key "activities", "activities"
+  add_foreign_key "activities", "activities", column: "parent_id", on_delete: :restrict
   add_foreign_key "activities", "organisations", column: "extending_organisation_id"
   add_foreign_key "activities", "organisations", column: "reporting_organisation_id"
   add_foreign_key "activities", "organisations", on_delete: :restrict
--- a/spec/services/find_programme_activities_spec.rb
+++ b/spec/services/find_programme_activities_spec.rb
@@ -8,7 +8,24 @@ RSpec.describe FindProgrammeActivities do
   let!(:extending_organisation_programme) { create(:programme_activity, extending_organisation: other_organisation) }
   let!(:other_programme) { create(:programme_activity) }
 
+  before(:all) do
+    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :parent
+  end
+
+  after(:all) do
+    Bullet.delete_whitelist(type: :unused_eager_loading, class_name: "Activity", association: :parent)
+  end
+
   describe "#call" do
+    it "eager loads the organisation and parent activity" do
+      expect_any_instance_of(ActiveRecord::Relation)
+        .to receive(:includes)
+        .with(:organisation, :parent)
+        .and_call_original
+
+      described_class.new(organisation: service_owner, current_user: user).call
+    end
+
     context "when the organisation is the service owner" do
       it "returns all programme activities" do
         result = described_class.new(organisation: service_owner, current_user: user).call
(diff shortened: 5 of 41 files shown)


db/schema.rb
```diff
@@ -49,7 +48,7 @@
     t.string "sector_category"
     t.boolean "ingested", default: false
     t.string "legacy_iati_xml"
-    t.index ["activity_id"], name: "index_activities_on_activity_id"
```
tahb, line 52: Do I need to re add this back?

tahb: Adding this back in for `parent_id` reduced the page load time from 2 seconds to 1.3 seconds.

spec/support/bullet.rb
```diff
@@ -3,6 +3,7 @@
     config.before(:each) do
       Bullet.add_whitelist type: :unused_eager_loading, class_name: "User", association: :organisation
       Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :organisation
+      Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :parent
```
mec, line 6: This makes me a bit nervous - mostly because I don't fully understand bullet!

Can we disable bullet around the tests we know it is a false positive rather than for the whole `Activity` class?

tahb: It makes me nervous too because it might hide alerts about other places we are eager loading but not using it, which may cause things to slow down unnecessarily. 

I'll take another look and see what an alternative might look like and report back here 👍

mec: I `think` we could diasble bullet around the specs: https://github.com/flyerhzm/bullet#run-in-tests

db/migrate/20200626100128_change_activity_parent_relationship_name.rb
```diff
@@ -0,0 +1,23 @@
+class ChangeActivityParentRelationshipName < ActiveRecord::Migration[6.0]
+  def up
+    add_column :activities, :parent_id, :string
+    Activity.transaction do
+      Activity.all.each do |activity|
+        activity.update!(parent_id: activity.activity_id)
```
mec, line 6: We shold handle the case where an activity is incomplete - as discussed setting `validate: false` is a nice clear wat to do that.

tahb:
@mec I've now:

1. updated the migration with `save!(validate: false)`
2. configured Bullet to only whitelist `unused_eager_loading` in the 3 service classes for `Activity.parent`. It should flag for anything else, i.e `Activity.includes(:children)`, now or in the future. I've tested that this config overwride isn't persisted to the rest of the test suite
3. I've added some more tests into each service class for the addition of `includes` for good measure
4. I noticed a second place Bullet whitelists existed and settled for merging the 2 lists together for now

Hopefully this is ready for another look!

db/schema.rb
```diff
@@ -162,7 +162,6 @@
     t.index ["role"], name: "index_users_on_role"
   end
 
-  add_foreign_key "activities", "activities"
```
mec, line 165: I am trying to understand why there are no foreign keys here and what the relationship is we want, I think it's about making sure the parent of a child is not deleted, orphaning the child  but I cannot see that reflected in the schema?

spec/services/find_programme_activities_spec.rb
```diff
@@ -8,7 +8,20 @@
   let!(:extending_organisation_programme) { create(:programme_activity, extending_organisation: other_organisation) }
   let!(:other_programme) { create(:programme_activity) }
 
+  before(:each) do
+    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :parent
```
mec, line 12: I've said before I am not super clear on bullet but I feel like this is just being set globally, how do we check it is removed after `each`? When I look at `Bullet.whitelist` 'after(:each)` and `after(:all)` I still see `:parent` which I think is this config?

mec:
@tahb I would love to understand the two comments I've left.

config/environments/test.rb
```diff
@@ -59,8 +59,6 @@
     Bullet.enable = true
     Bullet.bullet_logger = true
     Bullet.raise = true # raise an error if n+1 query occurs
-    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Transaction", association: :provider
```
mec, line 62: I am wondering if this is actaully the 'correct' place to add whitelists as putting them in before and after hooks mean we keep adding the same thing over and over?

spec/services/find_programme_activities_spec.rb
tahb: Sorry, I should have made myself more clear as I had left some new thoughts on this in the commit message:

> I tested that setting the whitelist in this way does not overwrite the global Bullet configuration. I ran 2 tests, the former with the whitelist change, and the latter without. The first file passes and the second fails. This confirms that the whitelist did not persist to other subsequant tests.

I'm surprised that the `Bullet.whitelist` is different in other tests.

config/environments/test.rb
tahb: I take your point. I suspect the 2 points may be connected. If the config is reapplied after each test, that could be why I'm able to overwride it in my own before block that's run later. I'll check how much overhead this move removes from to the test suite run time.

tahb: Yes, with my version it looks like the Bullet whitelist hash is reset after every test. 

Moving these lines to test.rb does not appear to save any noticeable time to the running of the test suite, both at 1 minute 54 seconds ish on my machine. I guess this makes sense as it is building a small hash that doesn't do any further calculations.

Their documentation does not set out a convention for us to follow on where whitelists should be configured nor trade offs to consider. Moving whitelists that do not change to initialisation rather than `before(:each)` does seem reasonable. I think the trade off is we lose the ability to control these whitelists on a per test basis - something we need for our service classes in this PR.

If we move the whitelists to test.rb I think the only alternative is to turn Bullet off for the entire service class test before, and re-enable it afterwards. This also feels like a worse choice since we would have no monitoring at all.

I'd be in favour of keeping as much bullet monitoring as we can by leaving this in bullet.rb unless there's a way I've not seen to avoid doing `Before … bullet.enable = false`, `After … Bullet.enable = true`?

db/schema.rb
tahb: I missed this removal and am adding it back in. I believe it is about making sure the parent of a child is not deleted.

spec/services/find_programme_activities_spec.rb
tahb: I think what we do here is tied to where whitelist config here which is the subject of the other comment thread here https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/494#pullrequestreview-441483206

leeky:
@tahb For consistency, wouldn't it better to rename `activity_id` to `parent_activity_id`, as this name is in use in `Budget`, `Transaction` and `PlannedDisembursement`? 

Then you can drop the `Activity#parent_activity` method altogether, and just use that association as-is?

tahb:
@leeky `activity_id` is being renamed to`parent_id` as it appears to be a conventional column name that all the tree gems use. It feels right to keep with the convention and `parent` and `children` makes sense to me when associating objects of the same type up and down a tree.

I did notice the `parent_activity_id` column on Budgets etc. I felt that the association between Activity->Budget is not tree shaped as Activity->Activity is, and naming the columns the same may cause confusion. Arguable Budget could be renamed to have an `activity_id`, the fact it has to also include `parent` is a plaster because of a naming clash with the PublicActivity gem.

CHANGELOG.md
```diff
@@ -186,13 +186,14 @@
 - Remove 3 unwanted activities from production
 - Content changes to fields for transaction value and activity identifier
 - Increase the width of the application layout
-- Activity show content is show in tabs for financials and details
```
tahb, line 189: I noticed this was in the wrong place when I rebased. Hoping it's agreeable to fix this as I go on the basis that if another PR beats this change to it, I'll have to address a conflict before this can be merged in.

config/environments/test.rb
tahb: @leeky I'm concious that @mec may be heads down in other work. Do you have an opinion on our best course of action here?

mec: @tahb here is a quick thing you can do to demonstate what I am seeing that I would lke to understand:

In `spec/support/bullet.rb` add a line after 13:

`pp Bullet.whitelist`

So we print out the current whitelist after each spec runs, now run the suite and watch that whitelist grow! 

Perhaps it doesn't matter that the whitelist is growing exponentially or I am misunderstanding, if so I would like to understand.

We are using the before hook for adding the `:parent` association on Activity in three spec files so I believe the hooks are scoped to the specs inside that file. That means we see three entries in the whitelist for `:parent` three times (three for each file). Also notice once the three specs have run the three `:parent` entries remain in the whitelist for the rest of the suite, which again points to the global nature of using `add_to_whitelist`.

Rpsec itself is configured to add four entries to the whitelist for every spec in the suite, so there a lot of entries in the whitelist. This is what makes me want to put the existing `Bullet.add_whitelist` in one place rather than in every spec.

Ultimately, this is how bullet has been setup so I am not trying to block this - I would just like to learn and understand, if we want to just move this forward, I think it is fine in a wider context!

tahb: I can now see the whitelist growing linearly, though I'm not sure of the immediate impact it clearly isn't a healthy trait. 

I had wondered which set of problems we'd prefer: sensible whitelist configuration or making sure Bullet is monitoring for future performance issues in our classes. 

I've found another way which should give us both with `Bullet.delete_whitelist`. Unfortunantley I didn't spot this sooner!

```
  before(:all) do
    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :parent
  end

  after(:all) do
    Bullet.delete_whitelist(type: :unused_eager_loading, class_name: "Activity", association: :parent)
  end
```

config/environments/test.rb
```diff
@@ -59,6 +59,8 @@
     Bullet.enable = true
     Bullet.bullet_logger = true
     Bullet.raise = true # raise an error if n+1 query occurs
+    Bullet.add_whitelist type: :unused_eager_loading, class_name: "User", association: :organisation
```
tahb, line 62: Existing whitelists now all live in test.rb.

spec/services/find_third_party_project_activities_spec.rb
```diff
@@ -8,7 +8,24 @@
   let!(:organisation_project) { create(:third_party_project_activity, organisation: other_organisation) }
   let!(:other_project) { create(:third_party_project_activity) }
 
+  before(:all) do
+    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :parent
+  end
+
+  after(:all) do
```
tahb, line 15: We target the specific whitelist we added once we're done with all tests in the file, and remove it.

tahb:
@mec thank you for prodding to make this better 👍

PR #498, opened 2020-06-30
(751) Split the activity show view into tabs

mec:
## Changes in this PR
This is the first piece of work to change the delivery partner views to focus on their activties and allow them to find the information they need quickly and simply regardless whether they are looking for financial or activity details data.

By splitting these two groups of information accross two tabs we can give equal weight to both needs.

Future works will add more tabs to this view but this was not part of this work.

It became clear through this work that the back link will need some thought/improvements but this work does not address this issue.

I decided to keep the tab view simple and not split the tabs into a partial whilst there are only two, this meant no logic is needed to work out which is the current tab.

I also decided to keep the tab views in the `activities` view directory rather than the convention of placing them in direcotries with the controller names - I was not sure if you can 'nest' controllers in this way?

### Before
Screenshot_2020-06-30 Activity Airbus Flood and Drought — Report your official development assistance

### After
Screenshot_2020-06-30 Activity Airbus Flood and Drought — Financials — Report your official development assistance

Screenshot_2020-06-30 Activity Airbus Flood and Drought — Details — Report your official development assistance

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -186,6 +186,7 @@
 - Remove 3 unwanted activities from production
 - Content changes to fields for transaction value and activity identifier
 - Increase the width of the application layout
+- Activity show content is show in tabs for financials and details
 
 ## [unreleased]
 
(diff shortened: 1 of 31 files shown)


mec:
[Split the activity show page for level C and D into two seperate tabbed views](https://trello.com/c/uNBQwXaP/751-split-the-activity-show-page-for-level-c-and-d-into-two-seperate-tabbed-views)

PR #499, opened 2020-06-30
(760) Activity identifiers are only unique between sibling activities

leeky:
## Changes in this PR

Activities can now have the same identifier, but only when they belong to different parent activities.

We are loosening up the uniqueness validation on activity identifiers because we believe it is likely that the project and third-party project identifiers currently used by delivery partners have the possibility of being the same.

When using this reference to publish to IATI we think this approach would continue to work. The fund, programme and project identifiers are all combined which should guarantee uniqueness across all activities.

*NB: We originally added this because we found in practice that programme identifiers were being reused by BEIS. However, because they would still share a common fund activity, this change wouldn't actual have fixed this issue. In that specific case, we chose to change the identifier to suffix the delivery partner's initials.*

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -198,6 +198,8 @@
 
 ## [unreleased]
 
+- Activity identifiers are unique among siblings
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-11...HEAD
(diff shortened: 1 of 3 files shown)


leeky:
@tahb Since we paired on this, could you approve this change and also cast an eye over the commit message as I've rewritten to reflect that we decided there's value in this change at the project/third-party project level, rather than at the programme level when we first created it?

PR #501, opened 2020-07-01
(715) Redact selected activities from the XML file

lozette:
## Changes in this PR

Trello: https://trello.com/c/e8GbGaHP/715-redact-selected-activities-from-the-xml-file

In order for BEIS users to redact certain activities from the IATI XML file, we have introduced a boolean field onto the Activity table which enables complete activities to be left out of the XML download. Activities are set to be published to IATI by default, the option has to be manually turned off. 

This option is only available to BEIS users.

Additionally, the list view of Activities on the Organisation show page now has a label to clearly show which activities are redacted or not. 

## Screenshots of UI changes

Screenshot 2020-07-01 at 14 30 02

Screenshot 2020-07-01 at 17 24 41

Screenshot 2020-07-01 at 17 12 15

Screenshot 2020-07-01 at 17 12 23

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -63,7 +63,7 @@ class Activity < ApplicationRecord
 
   scope :funds, -> { where(level: :fund) }
   scope :programmes, -> { where(level: :programme) }
-  scope :publishable_to_iati?, -> { where(form_state: :complete, publish_to_iati: true) }
+  scope :publishable_to_iati, -> { where(form_state: :complete, publish_to_iati: true) }
 
   def valid?(context = nil)
     context = VALIDATION_STEPS if context.nil? && form_steps_completed?
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -63,7 +63,7 @@ class Activity < ApplicationRecord
 
   scope :funds, -> { where(level: :fund) }
   scope :programmes, -> { where(level: :programme) }
-  scope :publishable_to_iati?, -> { where(form_state: :complete) }
+  scope :publishable_to_iati?, -> { where(form_state: :complete, publish_to_iati: true) }
 
   def valid?(context = nil)
     context = VALIDATION_STEPS if context.nil? && form_steps_completed?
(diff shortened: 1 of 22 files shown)


app/controllers/staff/activities_controller.rb
```diff
@@ -31,6 +31,19 @@ def show
     end
   end
 
+  def edit
+    @activity = Activity.find(id)
```
lozette, line 35: I'm wondering if we want a custom action name here instead of edit/update. And if we do, how to fit it into our routes file, which is very unusual at the moment!!

leeky: How about using a singular resource?

```ruby
resources :activities, only: [], concerns: [:transactionable, :budgetable, :disbursement_plannable] do
      resource :redaction, only: [:edit, :update], controller: :activity_redactions
      .. etc ..
end
```

Then you can move these actions into `ActivityRedactionsController` and paths like `/activities/:activity_id/redaction/edit` ?

app/models/activity.rb
```diff
@@ -63,7 +63,7 @@ class Activity < ApplicationRecord
 
   scope :funds, -> { where(level: :fund) }
   scope :programmes, -> { where(level: :programme) }
-  scope :publishable_to_iati?, -> { where(form_state: :complete) }
+  scope :publishable_to_iati?, -> { where(form_state: :complete).where(publish_to_iati: true) }
```
leeky, line 66: How about combining the two `where`s?

```ruby
where(publish_to_iati: true, form_state: :complete)
```

spec/features/staff/organisation_show_page_spec.rb
```diff
@@ -86,6 +106,16 @@
         end
       end
 
+      scenario "they see a Publish to Iati column & status against projects" do
+        within(".projects") do
+          expect(page).to have_content I18n.t("summary.label.activity.publish_to_iati.label")
+        end
+
+        within("##{project.id}") do
```
leeky, line 114: I believe this also works: `within(id: project.id) do`

app/models/activity.rb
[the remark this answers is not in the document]
lozette: No we will also be publishing to Publish What You Fund among others so this is IATI specific.

lozette: Noice!

PR #502, opened 2020-07-02
Redacted activities redact their children

lozette:
## Changes in this PR

A missed part of https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/501

Trello: https://trello.com/c/e8GbGaHP/715-redact-selected-activities-from-the-xml-file

When redacting an activity, all its children activities are redacted as well.

In theory this will only happen for level C activities at the moment, as only
level C & D can be redacted, and only level C can have children. But if we
begin publishing level B to IATI, this will be useful for redacting any children
that level B activities have.

## Next steps

### DIFF
--- a/app/controllers/staff/activity_redactions_controller.rb
+++ b/app/controllers/staff/activity_redactions_controller.rb
@@ -11,6 +11,8 @@ class Staff::ActivityRedactionsController < Staff::BaseController
     authorize @activity, :redact_from_iati?
 
     @activity.update(publish_to_iati: activity_params["publish_to_iati"])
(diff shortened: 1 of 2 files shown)


app/controllers/staff/activity_redactions_controller.rb
```diff
@@ -11,6 +11,13 @@ def update
     authorize @activity, :redact_from_iati?
 
     @activity.update(publish_to_iati: activity_params["publish_to_iati"])
+
+    if @activity.child_activities.any?
```
leeky, line 15: I think this would work too:

```ruby
@activity.child_activities.update(publish_to_iati: activity_params["publish_to_iati"])
```

PR #506, opened 2020-07-06
(752) Add basic all activities page

mec:
## Changes in this PR

Create the navigation element and page itself.

The action and route were already in-place from a previous feature that
was later removed.

Shows a complete list of all activities at level C and D.

## Screenshots of UI changes
Screenshot_2020-07-06 Actvities — Report your official development assistance

## Next steps

### DIFF
--- a/app/controllers/staff/activities_controller.rb
+++ b/app/controllers/staff/activities_controller.rb
@@ -4,7 +4,9 @@ class Staff::ActivitiesController < Staff::BaseController
   include Secured
 
   def index
-    @activities = policy_scope(Activity).includes(:parent)
+    @activities = policy_scope(Activity.project_activities)
+    authorize @activities
+    @activity_presenters = @activities.includes([:organisation, :parent]).map { |activity| ActivityPresenter.new(activity) }
   end
 
   def show
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -66,6 +66,7 @@ class Activity < ApplicationRecord
   scope :funds, -> { where(level: :fund) }
   scope :programmes, -> { where(level: :programme) }
   scope :publishable_to_iati, -> { where(form_state: :complete, publish_to_iati: true) }
+  scope :project_activities, -> { where(level: :project).or(where(level: :third_party_project)) }
 
   def valid?(context = nil)
     context = VALIDATION_STEPS if context.nil? && form_steps_completed?
--- a/app/views/staff/shared/activities/_table.html.haml
+++ b/app/views/staff/shared/activities/_table.html.haml
@@ -1,0 +1,29 @@
+- unless activities.empty?
+  %table.govuk-table.activities
+    %thead.govuk-table__head
+      %tr.govuk-table__row
+        %th.govuk-table__header
+          = t("table.header.activity.identifier")
+        %th.govuk-table__header
+          = t("table.header.activity.title")
+        %th.govuk-table__header
+          = t("table.header.activity.level_a")
+        %th.govuk-table__header
+          = t("table.header.activity.level_b")
+        %th.govuk-table__header
+          = t("table.header.activity.level_c")
+        %th.govuk-table__header
+
+    %tbody.govuk-table__body
+      - activities.each do |activity|
+        %tr.govuk-table__row{ id: activity.id }
+          %td.govuk-table__cell= activity.identifier
+          %td.govuk-table__cell= activity.title
+          %td.govuk-table__cell= activity.parent_activities.first&.title
+          %td.govuk-table__cell= activity.parent_activities.second&.title
+          %td.govuk-table__cell= activity.parent_activities.third&.title
+          %td.govuk-table__cell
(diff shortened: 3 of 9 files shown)


mec:
[Refactor activities view for delivery partners](https://trello.com/c/BBolZObG/752-refactor-activities-view-for-delivery-partners)

app/views/staff/shared/activities/_table.html.haml
```diff
@@ -0,0 +1,29 @@
...
+        %th.govuk-table__header
+
+    %tbody.govuk-table__body
+      - activities.each do |activity|
+        %tr.govuk-table__row{ id: activity.id }
+          %td.govuk-table__cell= activity.identifier
+          %td.govuk-table__cell= activity.title
+          %td.govuk-table__cell= activity.parent_activities.first&.title
+          %td.govuk-table__cell= activity.parent_activities.second&.title
+          %td.govuk-table__cell= activity.parent_activities.third&.title
```
leeky, line 24: This seems odd to me, when the activity the row is about is a Level C activity, the "Level C"  column is blank. I suppose the presence of a value in this column is the only indicator that the row contains a "Level D" activity..?

app/controllers/staff/activities_controller.rb
```diff
@@ -4,7 +4,9 @@ class Staff::ActivitiesController < Staff::BaseController
   include Secured
 
   def index
-    @activities = policy_scope(Activity).includes(:parent)
+    @activities = policy_scope(Activity.project_activities)
+    authorize @activities
+    @activity_presenters = @activities.includes([:organisation, :parent]).map { |activity| ActivityPresenter.new(activity) }
```
leeky, line 9: Since the `includes` relates the to the ActiveRecord query rather than the presenter, how about rewriting as:

```ruby
@activities = policy_scope(Activity.project_activities).includes([:organisation, :parent])
authorize @activities
@activity_presenters = @activities.map { |activity| ActivityPresenter.new(activity) }
```

app/models/activity.rb
```diff
@@ -66,6 +66,7 @@ class Activity < ApplicationRecord
   scope :funds, -> { where(level: :fund) }
   scope :programmes, -> { where(level: :programme) }
   scope :publishable_to_iati, -> { where(form_state: :complete, publish_to_iati: true) }
+  scope :project_activities, -> { where(level: :project).or(where(level: :third_party_project)) }
```
leeky, line 69: There are scopes call `funds` and `programmes` already, and this seems to be breaking from the pattern, although I accept that this includes `projects` *AND* `third_party_projects` together.

spec/features/staff/users_can_view_an_activities_spec.rb
```diff
@@ -25,19 +25,50 @@
...
+
+    scenario "all acitvities can be viewed" do
+      activities = create_list(:project_activity, 5, organisation: user.organisation)
+
+      visit organisation_activities_path(user.organisation)
+
+      expect(page).to have_content I18n.t("page_title.activity.index")
+
+      first_activity = activities.first
+      last_actvitiy = activities.last
```
leeky, line 47: typo: `last_activity`

app/controllers/staff/activities_controller.rb
```diff
@@ -4,7 +4,9 @@ class Staff::ActivitiesController < Staff::BaseController
   include Secured
 
   def index
```
leeky, line 6: I'm wondering whether this is the right controller for this to go, as it's a list of only projects/third-party projects for the current user's organisation, but occupies a spot where I'd expect to find the code for viewing a list of *all* activities.

In the case of users, they can only ever belong to one organisation, so you could do have a top-level route:

`resources :projects, only: [:index, :show]`

And then have use the `current_user.organisation` in order to do the scoping?

spec/models/activity_spec.rb
```diff
@@ -44,6 +44,21 @@
...
     end
+
+    describe ".project_activities" do
+      it "only returns acitivites at level C and D i.e. Projects and Third party projects" do
+        fund = create(:fund_activity)
+        programme = create(:programme_activity, parent: fund)
+        project = create(:project_activity, parent: programme)
+        third_party_project = create(:third_party_project_activity, parent: project)
+
+        expect(Activity.project_activities).to include project
```
leeky, line 55: :+1:

spec/features/staff/users_can_view_an_activities_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.feature "Users can view an activity" do
```
leeky, line 1: Maybe rename the spec to `users_can_view_activities_spec.rb` ?

app/controllers/staff/activities_controller.rb
mec: We talked about this one and why the includes live where it is.

mec:
We've revisited this work and would like to simplyfy it taking @leeky comments into account. I will close it now and reopen once we've re-worked it.

PR #508, opened 2020-07-06
(content) Update the label of the identifier step

mec:
## Changes in this PR
A small content change that fell through the gaps.


mec:
[Improve "create fund" content - title and hint text](https://trello.com/c/WBp6gKFk/667-improve-create-fund-content-title-and-hint-text)

leeky:
It does beg the question why this label exists in two places.

PR #509, opened 2020-07-08
(fix) Move budgets on the progrmame financials tab

mec:
To be consistent with other activity levels.

### DIFF
--- a/app/views/staff/activities/financials.html.haml
+++ b/app/views/staff/activities/financials.html.haml
@@ -35,14 +35,15 @@
             = t("page_title.activity.financials")
 
           - if @activity.fund? && policy(:fund).create?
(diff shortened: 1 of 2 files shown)


mec:
[Make the ordering of the financials tab the same for all activity levels](https://trello.com/c/rVnRsgEz/775-make-the-ordering-of-the-financials-tab-the-same-for-all-activity-levels)

app/views/staff/activities/financials.html.haml
```diff
@@ -35,14 +35,15 @@
             = t("page_title.activity.financials")
 
           - if @activity.fund? && policy(:fund).create?
+            = render partial: "staff/shared/activities/budgets", locals: { activity: @activity, budget_presenters: @budget_presenters }
```
leeky, line 38: Took me a while to understand what had changed, because of the way GitHub has displayed the diff here!

PR #511, opened 2020-07-08
(780) XML contains ONLY recipient country OR region

leeky:
## Changes in this PR

The IATI activity standard specifies that `<recipient-region/>` and `<recipient-country/>` should be mutually exclusive, unless we are providing the percentage splits amongst them.

For now, we'll only include the recipient region when the country is not available.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,7 @@
 - Activity identifiers are unique among siblings
 - add accessibility statement
 - Move budgets to the top of the programme activity financials tab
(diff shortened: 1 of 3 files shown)


mec:
@leeky sorry I just rebased this and didn't mean to! 😞 I messed up the changelog on the accessibilty statement and now the fix is here! I think it's okay?

leeky:
@mec No worries, will rebase again once you've merged your changes into develop and all will be fine!

PR #514, opened 2020-07-09
(752) Add activities index page

mec:
## Changes in this PR

This is a simple view of an organisations activities.

The view takes a organisation_id and displays all the activities that
are associated with that organisation through the `organisation`
association.

BEIS users can supply a delivery partners organisation id and see those
activities. This functionality is not exposed in the UI as yet, but is
supported by the view, we expect later work to add this UI.

Delivery partners cannot use another organisation id.

The view can be augmented later with additional filters as we learn what
users want to see.

When an unexpected organisation_id is encountered we try to show the
current users own organisation activities instead, those instances might
be:

- no organisation id parameter
- an unknown organisation id
- an empty organisation id
- another organisations id when the user is not a BEIS user

We chose to handle these cases and display the current users activities
as we shouldn't encounter these situations under normal conditions.

There is a lot of change in `spec/features/staff/users_can_view_activities_spec.rb` as I had to reconfigure this spec and rename it.

## Screenshots of UI changes
BEIS user view BEIS actvities:
Screenshot_2020-07-09 Activities — Report your official development assistance

BEIS user viewing another organisations activties:
beis - uksa

Delivery partner:
Screenshot_2020-07-09 Activities — Report your official development assistance

## Next steps

- [X] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak wich has no impact outside the development team.
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -203,6 +203,7 @@
 - Move budgets to the top of the programme activity financials tab
 - Activity XML includes only recipient-country OR recipient-region, not both
 - Activity creation journey changed to ask for a level and parent
+- Add an activities page and navigation
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-11...HEAD
 [release-11]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-10...release-11
--- a/app/controllers/staff/activities_controller.rb
+++ b/app/controllers/staff/activities_controller.rb
@@ -51,17 +51,12 @@ class Staff::ActivitiesController < Staff::BaseController
 
     return current_user.organisation.id unless Organisation.exists?(organisation_id)
     return current_user.organisation.id if organisation_id.blank?
-    return current_user.organisation.id if requested_organisation_is_not_current_users?(organisation_id) && current_user_is_not_service_owner?
+    return current_user.organisation.id if requested_organisation_is_not_current_users?(organisation_id) && current_user.delivery_partner?
     organisation_id
   end
 
   def requested_organisation_is_not_current_users?(requested_organisation_id)
     requested_organisation_id != current_user.organisation.id
-  end
-
-  def current_user_is_not_service_owner?
-    return false if current_user.organisation.service_owner?
-    true
   end
 
   def fund_id
--- a/app/controllers/staff/activities_controller.rb
+++ b/app/controllers/staff/activities_controller.rb
@@ -4,7 +4,9 @@ class Staff::ActivitiesController < Staff::BaseController
   include Secured
 
   def index
-    @activities = policy_scope(Activity).includes(:parent)
+    @activities = policy_scope(Activity.where(organisation: organisation_id))
+    authorize @activities
+    @activity_presenters = @activities.includes(:organisation).map { |activity| ActivityPresenter.new(activity) }
   end
 
   def show
@@ -45,7 +47,21 @@ class Staff::ActivitiesController < Staff::BaseController
   end
 
   def organisation_id
-    params[:organisation_id]
+    organisation_id = params.fetch(:organisation_id, current_user.organisation).to_s
+
+    return current_user.organisation.id unless Organisation.exists?(organisation_id)
+    return current_user.organisation.id if organisation_id.blank?
+    return current_user.organisation.id if requested_organisation_is_not_current_users?(organisation_id) && current_user_is_not_service_owner?
+    organisation_id
+  end
+
+  def requested_organisation_is_not_current_users?(requested_organisation_id)
+    requested_organisation_id != current_user.organisation.id
+  end
+
+  def current_user_is_not_service_owner?
+    return false if current_user.organisation.service_owner?
+    true
   end
 
   def fund_id
--- a/spec/features/staff/users_can_view_activities_spec.rb
+++ b/spec/features/staff/users_can_view_activities_spec.rb
@@ -1,0 +1,185 @@
+RSpec.feature "Users can view activities" do
+  context "when the user is not logged in" do
+    it "redirects the user to the root path" do
+      activity = create(:activity)
+      visit organisation_activity_path(activity.organisation, activity)
+      expect(current_path).to eq(root_path)
+    end
+  end
+
+  context "when the user is signed in as a BEIS user" do
+    let(:user) { create(:beis_user) }
+    before { authenticate!(user: user) }
+
+    scenario "the activities index shows activities associated to the BEIS organisation" do
+      activities = create_list(:programme_activity, 2, organisation: user.organisation)
+      another_activity = create(:project_activity)
+
+      visit activities_path(organisation_id: user.organisation)
+      expect(page).to have_content I18n.t("page_title.activity.index")
+
+      first_activity = activities.first
+      last_activity = activities.last
+
+      expect(page).to have_content first_activity.identifier
+      expect(page).to have_content last_activity.identifier
+      expect(page).not_to have_content another_activity.identifier
+    end
+
+    scenario "they can view another organisations activities" do
+      delivery_partner = create(:delivery_partner_organisation)
+      activity = create(:project_activity, organisation: delivery_partner)
+
+      visit activities_path(organisation_id: activity.organisation)
+
+      expect(page).to have_content activity.identifier
+    end
+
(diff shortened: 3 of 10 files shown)


mec:
[Refactor activities view for delivery partners](https://trello.com/c/BBolZObG/752-refactor-activities-view-for-delivery-partners)

app/controllers/staff/activities_controller.rb
```diff
@@ -39,7 +41,25 @@ def id
...
+    requested_organisation_id != current_user.organisation.id
+  end
+
+  def current_user_is_not_service_owner?
+    return false if current_user.organisation.service_owner?
+    true
+  end
+
+  def known_organisation_id?(organisation_id)
+    Organisation.all.pluck(:id).include?(organisation_id)
```
tahb, line 62: The `exists?` method might remove some logic here and perhaps prevent the need to construct a new hash of ids with `pluck`.

```
Organisation.exists?(organisation_id)
```

Arguably it's now succint enough not to need it it's own method?

spec/features/staff/users_can_view_activities_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      expect(current_path).to eq(root_path)
+    end
+  end
+
+  context "when the user is signed in as a BEIS user" do
+    let(:user) { create(:beis_user) }
+    before { authenticate!(user: user) }
+
+    scenario "the activities index shows activities associated to the BEIS organisation" do
+      activities = create_list(:programme_activity, 5, organisation: user.organisation)
```
tahb, line 15: If we're only checking for 2 records, we might be able to save a little time by creating fewer than 5 new records?

spec/features/staff/users_can_view_activities_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+        activity = create(:programme_activity, organisation: user.organisation)
+
+        visit activities_path(organisation_id: "")
+
+        expect(page).to have_content activity.identifier
+      end
+    end
+
+    context "when the organisation id query parameter is not the BEIS organisation id" do
+      scenario "it shows the supplied organiastion activities" do
```
tahb, line 49: Caught a small typo on organisation

CHANGELOG.md
```diff
@@ -195,6 +195,7 @@
 - Activity show content is show in tabs for financials and details
 - Refactor how we can ask activities for their parents
 - Ingest BA Newton fund data from IATI
+- Add an activities page and navigation
```
tahb, line 198: I think this line should a few lines down under 'unreleased'.

app/controllers/staff/activities_controller.rb
```diff
@@ -39,7 +41,25 @@ def id
   end
 
   def organisation_id
-    params[:organisation_id]
+    organisation_id = params.fetch(:organisation_id, current_user.organisation)
+
+    return current_user.organisation.id unless known_organisation_id?(organisation_id)
+    return current_user.organisation.id if organisation_id.blank?
```
leeky, line 47: I don't think this line is needed, because the `params.fetch`  is doing the same thing.

Having said that, if we're setting the `organisation_id` parameter to one value, but the user doesn't have access to that information, shouldn't we be returning a 404 *or* redirect them to the page for their organisation?

app/controllers/staff/activities_controller.rb
```diff
@@ -39,7 +41,25 @@ def id
...
+    return current_user.organisation.id if requested_organisation_is_not_current_users?(organisation_id) && current_user_is_not_service_owner?
+    organisation_id
+  end
+
+  def requested_organisation_is_not_current_users?(requested_organisation_id)
+    requested_organisation_id != current_user.organisation.id
+  end
+
+  def current_user_is_not_service_owner?
+    return false if current_user.organisation.service_owner?
```
leeky, line 57: You could replace this with `!current_user.organisation.service_owner?`

app/controllers/staff/activities_controller.rb
mec: Lovely! Thank you.

app/controllers/staff/activities_controller.rb
mec: I thought I had checked that, with the `fetch` I am trying to catch the case when there is no param, so `/activities` and defaulting to the current user org. The `blank?` is a catch for when the param is present, but there is no value so `/acticities?organisation_id` - I thought a lot about all these case and if we should try to handle them, which leads nicely to…

I considered both when we might 404, 403, redirect, etc and ended up discussing with @tahb as I couldn't make the decision on which was best. 

I feel like the 'correct' thing to do it return the error page, but from a users point of view I am not sure that is helpful, we felt missing or changing the query param is not something a user would do intentionally and so returning an error may not be the most helpful and given a delivery partner can only see their own organisations activities, showing that is a safe bet.

Let me know what you think?

app/controllers/staff/activities_controller.rb
mec: 😄 It funny raising this as I debated this with myself a great deal! 

I find the bang infront of something a little hard to approach, it's so easily missed. Ideally I would have a `delivery_partner?` method as I think that allows the check to be for a postitive, which feels nice and seems to read clearer in Ruby, but that seemed like a push too far.

Happy to change this to the bang if you would like it, would you keep the method or just drop it inline?

app/controllers/staff/activities_controller.rb
leeky: I'd argue that we should be always trying to do the correct thing - if the user isn't going to fiddle with the query parameter, then surely there's no harm in showing an error message when they do? But I'm happy for this to go in because it sounds like you've thought about all of these situations! :shipit:

CHANGELOG.md
```diff
@@ -213,4 +214,4 @@
 [release-5]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-4...release-5
 [release-4]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-3...release-4
 [release-3]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-2...release-3
-[release-2]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-1...release-2
+release-2]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-1...release-2
```
leeky, line 217: `[` is missing from this line

app/controllers/staff/activities_controller.rb
mec: @tahb I am getting a brakeman sql injection warning with `exists?` any thoughts? Put an ignore in rethink? I would have thought Rails would make the actual SQL safe?

app/controllers/staff/activities_controller.rb
leeky: I think there are plenty of places where `delivery_partner?` would be useful. Perhaps a method belongs inside `app/models/organisation.rb`? Then you could use it inline rather than having a separate method in here.

mec: @leeky I've gone for the User model as this is where the existing `service_owner?` method lives - let me know what you think?

leeky: @mec Perfect 🎉

app/controllers/staff/activities_controller.rb
mec: Looking into this, it seems the accepted answer is to make sure the id is a string: 

https://github.com/presidentbeef/brakeman/issues/1045

Which seems to make brakeman happy.

mec: @tahb if you are happy with ☝️ that - I can merge this in – I want to make sure you have at least seen this as it was a potential security issue.

tahb: Thanks for flagging.

I wondered even with `.to_s` you could craft a malicious string to escape out and inject more into the query. https://rails-sqli.org/rails4#exists does say that `.to_s` is a way to make it safe though which is good enough for me.

> To be be safe, convert user input to a string or integer if using it as the primary key in exists?.

The risk is that we accidentally remove the `.to_s` later without realising however Brakeman is seemingly ready and waiting to let us know 👍

PR #515, opened 2020-07-09
(753) Create new activity journey now includes level and parent

tahb:
This pull request runs into many unexpected bumps in the road. There are 10 commits that are seeking to isolate various issues or add supporting methods before the 11th commit arrives. The hope was to make the last commit much smaller than it is but I couldn't logically pull it apart anymore than it is. I'd suggest only doing a commit-by-commit review of this one.

We had originally thought about a new form that could be much simpler by using a few query params instead of the existing multi-step form thar has grown large. Given these 2 pieces of information are always required and that the flows will always be connected, it felt natural to extend the existing journey instead. Creating a new controller called something like `ActivityFormPrerequisitesController` with views, new validations and managing the redirection between them and the `ActivityFormController` all felt rather forced when the existing flow could support it well after a short spike into it. 

[Wireframes for this can be found on Miro](https://miro.com/app/board/o9J_kqNtuVk=/?moveToWidget=3074457348587713807&cot=6)

## Changes in this PR

- All existing create fund, programme, project and third-party project buttons replaced with a new generic 'create activity'
- All users must now choose the level of which the new activity should exist, A or B for BEIS and C or D for delivery partners
- All users must now choose the parent activity for which the new activity should belong from their available parent activities. Funds are not applicable and this step is skipped
- Activity policies refactored to using the authoring organisation for the majority of checks, decoupling the heavy dependency on 'level' which is no longer guaranteed
- Level and Parent information is now shown in the activity details page to allow the user to continue if they stopped at either point. Note that they can only add. Editing worked but has been deliberatley removed due to the consequences that changing the hierarchy has on the IATI identifier that's used to publish new activities.
- ActivityFormController#Update is in dire need of a refactor. I considered against it given the amount of other additions and refactorings going on in this proposal

### BEIS user

Screenshot 2020-07-09 at 17 44 52
Screenshot 2020-07-09 at 17 44 58
Screenshot 2020-07-09 at 17 45 03
Screenshot 2020-07-09 at 17 45 09
Screenshot 2020-07-09 at 17 45 12
Screenshot 2020-07-09 at 18 14 44

---

### Delivery partner user

Screenshot 2020-07-09 at 18 03 01
Screenshot 2020-07-09 at 18 03 07

## Next steps


app/helpers/activity_helper.rb
```diff
@@ -12,7 +12,7 @@ def step_is_complete_or_next?(activity:, step:)
   end
 
   def activity_back_path(current_user:, activity:)
-    if activity.programme? && current_user.service_owner?
+    if activity.parent.present? && activity.programme? && current_user.service_owner?
```
tahb, line 15: We have agreed to lose these back links so this change is a very quick plaster to make it just work for now.

app/presenters/activity_presenter.rb
```diff
@@ -67,8 +67,18 @@ def display_title
...
+  end
+
   def level
     return if super.blank?
-    I18n.t("page_content.activity.level.#{super}")
+    case super
+    when "fund" then "Fund"
+    when "programme" then "Programme"
+    when "project" then "Project"
+    when "third_party_project" then "Third-party project"
```
tahb, line 81: The hyphen in 'third-party' here is why this method doesn't simply use `titleize`.

app/services/update_activity_as_programme.rb
```diff
@@ -1,20 +1,21 @@
...
   end
 
   def call
-    activity = Activity.new
-    activity.organisation = Organisation.find(organisation_id)
-    activity.reporting_organisation = activity.organisation
+    activity.extending_organisation = service_owner
 
-    fund = Activity.find(fund_id)
-    fund.child_activities << activity
```
tahb, line 15: Given we never saved the fund, this wasn't ever doing anything.

app/helpers/form_helper.rb
```diff
@@ -60,4 +60,12 @@ def scoped_parent_activities(activity:, user:)
         .call(eager_load_parent: false)
     end
   end
+
+  def create_activity_level_options(user:)
+    authorised_levels = Activity.levels.select { |level|
+      policy = Pundit.policy(user, level.to_sym)
+      policy.create? || policy.update?
+    }
+    authorised_levels.keys.map { |level| OpenStruct.new(level: level, description: level.titleize) }
```
mec, line 69: I am not sure how much we should worry about 'Third-party projects' being shown but `titeize` won't do it, we use the locale elsewhere, something like:

`authorised_levels.keys.map { |level| OpenStruct.new(level: level, description: I18.t("page_content.activity.level.#{level}").capitalise) }`

app/views/staff/activity_forms/parent.html.haml
```diff
@@ -0,0 +1,9 @@
+= render layout: "wrapper" do |f|
+  - scoped_parent_activities = scoped_parent_activities(activity: f.object, user: current_user)
+  = f.hidden_field :parent, value: nil
+  = f.govuk_collection_radio_buttons :parent,
+    scoped_parent_activities,
+    :id,
+    ->(activity){ ActivityPresenter.new(activity).display_title },
+    legend: { tag: 'h1', size: 'xl', text: I18n.t("form.legend.activity.parent", parent_level: f.object.parent_level, level: f.object.level) },
```
mec, line 8: This is another place where 'Third-party project' bites us! setting `level` to something like

`level: t("page_content.activity.level.#{f.object.level}")` is the approach we've used elsewhere.

app/views/staff/activity_forms/level.html.haml
```diff
@@ -0,0 +1,8 @@
+= render layout: "wrapper" do |f|
+  = f.hidden_field :level, value: nil
+  = f.govuk_collection_radio_buttons :level,
+    create_activity_level_options(user: current_user),
+    :level,
+    :description,
+    legend: { tag: 'h1', size: 'xl', text: I18n.t("form.legend.activity.level") },
+    hint_text: I18n.t("form.hint.activity.level").html_safe
```
mec, line 8: What is the concern that drives the use of `html_safe` here?

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -95,6 +109,10 @@ def activity_params
       :aid_type)
   end
 
+  def level
+    params.require(:activity).permit(:level)["level"]
```
mec, line 113: Is this strong parameters on a single param?  Is this instead of adding `:level` to the giant `permit` on line 106? I would love to understand this better.

app/services/create_activity.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    activity.form_state = "level"
+
+    activity.save!
+    activity
+  end
+
+  private
+
+  def service_owner
```
mec, line 21: What does this do?

app/services/update_activity_as_fund.rb
```diff
@@ -22,7 +19,7 @@ def call
     activity.accountable_organisation_reference = "GB-GOV-13"
     activity.accountable_organisation_type = "10"
 
-    activity.save!
+    activity.save if activity.valid?(:parent_step)
```
mec, line 22: I think validation contexts would be nicer and a fraction more efficient here, it would look like this:

`activity.save(context: :parent_step)`

I am wondering about if this save fails for some unknown reason we return the activity regardless, are we okay with that or would a noisey `save!` give us more to go on to debug? What would the user see?

app/services/update_activity_as_programme.rb
```diff
@@ -25,7 +26,7 @@ def call
     activity.accountable_organisation_reference = service_owner.iati_reference
     activity.accountable_organisation_type = service_owner.organisation_type
 
-    activity.save!
+    activity.save if activity.valid?(:parent_step)
```
mec, line 29: Same comment as above?

app/services/update_activity_as_programme.rb
```diff
@@ -1,20 +1,21 @@
...
   def call
-    activity = Activity.new
-    activity.organisation = Organisation.find(organisation_id)
-    activity.reporting_organisation = activity.organisation
+    activity.extending_organisation = service_owner
 
-    fund = Activity.find(fund_id)
-    fund.child_activities << activity
+    activity.parent = begin
+                        Activity.fund.find(parent_id)
```
mec, line 13: I'm interested in the decision to guard against not finding the parent activity here as well, what led you to make that?

app/services/update_activity_as_project.rb
```diff
@@ -27,7 +27,7 @@ def call
     activity.accountable_organisation_reference = service_owner.iati_reference
     activity.accountable_organisation_type = service_owner.organisation_type
 
-    activity.save!
+    activity.save if activity.valid?(:parent_step)
```
mec, line 30: Same here?

app/services/update_activity_as_third_party_project.rb
```diff
@@ -27,7 +27,7 @@ def call
     activity.accountable_organisation_reference = service_owner.iati_reference
     activity.accountable_organisation_type = service_owner.organisation_type
 
-    activity.save(validate: false)
+    activity.save if activity.valid?(:parent_step)
```
mec, line 30: And here! 😛

config/locales/models/activity.en.yml
```diff
@@ -83,7 +89,7 @@ en:
         stage: Stage
         status: Status
         third_party_projects: Third-party projects
-        title: '%{level} name'
+        title: 'Title'
```
mec, line 92: Why has this changed? I have a memory of this being a content change from @duffylouise?

spec/features/staff/users_can_edit_an_activity_spec.rb
```diff
@@ -6,6 +6,67 @@
...
+        within(".identifier") do
+          expect(page).not_to have_content(I18n.t("default.link.add"))
+        end
+      end
+    end
+
+    # Changing the level means the IATI identifier we construct changes
+    # If this is changed after publishing to IATI the effect will be a branch
+    # new IATI publiciation.
+    context "when the activity has a level and a parent" do
```
mec, line 54: Have we thought about how a user will rectify the situation where they create an activity at the wrong level?

spec/services/update_activity_as_project_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      it "uses the service owner as the reporting organisation" do
+        expect(result.reporting_organisation).to eq(beis)
+      end
+    end
+
+    context "when the organisation is a non-governmental organisation" do
+      let(:activity) { create(:project_activity, organisation: ngo) }
+      let(:ngo) { create(:delivery_partner_organisation, organisation_type: 21) }
+
+      it "saves the reporting organisation" do
```
mec, line 23: I like this version better "sets the reporting organisation to be the delivery partner"

app/views/staff/activity_forms/level.html.haml
tahb: Guilty of copy and pasting from another radio step! I presume that one has or had links to IATI in their hints. This can probably go 👍

app/controllers/staff/activity_forms_controller.rb
tahb: It is.

I started by adding it there but found I could change the `level` of an activity again since the `update_activity_attributes_except_dates` is run on every update. It takes all `activity_params` and blindly assigns them. To me that needed a refactor and I decided to work around it for now.

app/services/create_activity.rb
tahb: Doesn't look like anything anymore!

app/services/update_activity_as_fund.rb
tahb: I'll test how `activity.save(context: :parent_step)` fits in, I hadn't thought of that 👍 

The main concern I had here was that the error messages for validations are bubbled back up to the controller, so I removed the bang. I believe it catches unexpected problems and throws a 500 instead - something we shouldn't do by default as it is a jarring user experience.

On another look it seems that `ActivityFormController#update` will render the next step whatever happens:

```
render_wizard @activity, context: :"#{step}_step"
```

Given `update_form_state` guards against changing the form_state unless the activity is valid, perhaps we could guard against thie update taking us to the wrong step (without errors) by making the following change so we can lose the bangs:

```
render_wizard @activity, context: :"#{@activity.form_state}_step"
```

As mentioned the whole action is in dire need of some TLC. I'd be interested in finding a way to unblock this work and set ourselves up to refactor next.

config/locales/models/activity.en.yml
tahb: Ah I thought it was a bug in a context I saw it being used. When looking at an activity details view for a project the title field was:

```
Third-party project name: <title of project>
```

I will take another look to see if it was in fact because level was being injected incorrectly!

spec/features/staff/users_can_edit_an_activity_spec.rb
tahb: We haven't built any of this of course but I was imaginging it would be a similar journey for deleting a record. The user (for now) should contact support and it will end up with the dev team as a request to change the level or remove the record. So long as it remained an edge case it would be the cheapest way to review the implications of deleting/changing hierarchy.

spec/services/update_activity_as_project_spec.rb
tahb: Agreed

app/views/staff/activity_forms/parent.html.haml
tahb: I agree! Will switch to internationalisation.

app/services/update_activity_as_fund.rb
tahb: On further reading `render_wizard` calls `.save` on the object with the given context. Although this highlights that there are 3/4 saves of the same record which is not efficient, the form will not go to the next step unless it's valid. This means we are safe to remove the `!`, I think.

Tests are green when changing to `activity.save(context: :parent_step)` 👍

config/locales/models/activity.en.yml
tahb: My mistake. I added more test coverage to reduce the chances of me accidentally breaking it again https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/515/commits/939f63e8bbdc42a21b27fbe7c296129ee3659bb2

tahb:
@mec thanks for the feedback on this. I've made several changes:

- update services all use `activity.save(context: :parent_step)`
- new tests for activity summary labels
- put back the summary label I broke to how it was
- renamed a test in update a project spec 
- the working out for "third-party project" has been deferred to internationalisation in parent and level views, the level method in presenter and in the assoicated tests
- removed a redundant method called `service_owner` from CreateActivity

PR #518, opened 2020-07-13
(content) Update label on transaction edit

mec:
This has been changed on the table view and this should match.

### DIFF
--- a/config/locales/models/transaction.en.yml
+++ b/config/locales/models/transaction.en.yml
@@ -9,7 +9,7 @@ en:
   form:
     label:
       transaction:
-        curreny: Currency
+        currency: Currency
         description: Describe the transaction
         disbursement_channel: Disbursement channel (optional)
         providing_organisation_reference: International Aid Transparency Initiative (IATI) Reference (optional)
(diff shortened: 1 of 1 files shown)


mec:
[Improve "create programme" content - purpose of programme and value](https://trello.com/c/rI0cKhqL/674-improve-create-programme-content-purpose-of-programme-and-value)

config/locales/models/transaction.en.yml
```diff
@@ -19,6 +19,7 @@ en:
         providing_organisation_type: Providing organisation type
         receiving_organisation_name: Receiving organisation name
         receiving_organisation_type: Receiving organisation type
+        value: Transaction amount
```
lozette, line 22: Where does this get used? Is the label automagically picked out, as I can't find it in any of the views?

lozette: Also on line 12 of this file is `curreny: Currency` (misspelling of the key). I know this isn't in this PR but I'm concerned that this key is misspelled!

mec: Yep, automagic. We have configured the GOVUK form builder to use `form.label.activity` see:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/ffe685f25e24b55b8e6c07686c41a0a6571e8def/config/initializers/form_builder.rb#L10

mec: No  probs on the typo - I'll fix it…

lozette: Love a bit of magic

PR #519, opened 2020-07-13
Remove back links, except on Activity form

lozette:
The back links are trying to mimic the action of the Back button in the browser
but are increasingly confusing as many pages can be accessed in different ways.

This commit removes them, except for the Back links on the activity create/edit
form as there is no other quick way to get back to the activity details page
without a back link. The back link on the Activity form has been amended to
point at the `details` view of the Activity.

**Suggestion for discussion:** on the Activity Form, would it make sense to replace the Back link with a secondary button? E.g. https://design-system.service.gov.uk/components/button/secondary-combo/index.html

Screenshot 2020-07-13 at 15 26 06

## Next steps

### DIFF
--- a/spec/features/staff/users_can_choose_recipient_country_spec.rb
+++ b/spec/features/staff/users_can_choose_recipient_country_spec.rb
@@ -55,28 +55,12 @@ RSpec.feature "Users can choose a recipient country" do
         expect(page).to have_selector "li.autocomplete__option", text: "Afghanistan", visible: true
       end
 
-      scenario "typing a complete country name, clicking it in the list and clicking continue saves the country" do
-        fill_in I18n.t("form.label.activity.recipient_country"), with: "Saint Lucia"
-        find("li.autocomplete__option", text: "Saint Lucia").click
-        click_button I18n.t("form.button.activity.submit")
-        click_on I18n.t("default.link.back")
-        click_on I18n.t("tabs.activity.details")
-
-        within(".recipient_country") do
-          expect(page).to have_content "Saint Lucia"
-        end
-      end
-
       scenario "typing a partial match, clicking on the complete match and clicking continue saves the country " do
         fill_in I18n.t("form.label.activity.recipient_country"), with: "saint"
         find("li.autocomplete__option", text: "Saint Lucia").click
         click_button I18n.t("form.button.activity.submit")
-        click_on I18n.t("default.link.back")
-        click_on I18n.t("tabs.activity.details")
-
-        within(".recipient_country") do
-          expect(page).to have_content "Saint Lucia"
-        end
+        click_link I18n.t("default.link.back")
+        expect(page).to have_content "Saint Lucia"
       end
 
       scenario "choosing a recipient country sets a recipient region associated to that country" do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,6 +204,7 @@
 - Activity XML includes only recipient-country OR recipient-region, not both
 - Activity creation journey changed to ask for a level and parent
 - Add an activities page and navigation
+- Remove Back links, except on the Activity form
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-11...HEAD
 [release-11]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-10...release-11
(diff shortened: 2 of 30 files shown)


db/schema.rb
```diff
@@ -88,6 +88,9 @@
     t.index ["parent_activity_id"], name: "index_budgets_on_parent_activity_id"
   end
 
+  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
tahb, line 91: I'm unsure how this has got in as a new addition! [On develop right now this table is not in the schema](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/db/schema.rb) which is unexpected.

We've been using the data_migrate gem since April https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/12f921d6e9dd5311817de06c547c05f944c0680f and there hasn't been a migration needed.

Any idea where this came from, what it does and if we need it?

spec/features/staff/users_can_choose_recipient_country_spec.rb
```diff
@@ -55,28 +55,11 @@
         expect(page).to have_selector "li.autocomplete__option", text: "Afghanistan", visible: true
       end
 
-      scenario "typing a complete country name, clicking it in the list and clicking continue saves the country" do
-        fill_in I18n.t("form.label.activity.recipient_country"), with: "Saint Lucia"
-        find("li.autocomplete__option", text: "Saint Lucia").click
-        click_button I18n.t("form.button.activity.submit")
-        click_on I18n.t("default.link.back")
```
tahb, line 62: I presume this more specific feature test exists to give greater coverage to the country selection journey which uses a combo of JS and no-JS to provide a different experience.

Instead of removing the test would it be worth changing the way the assertion is made? Perhaps it could navigate through the new "Activities" in the nav, select the new activity and check the content as the user would now be expected to do.

spec/features/staff/users_can_choose_recipient_country_spec.rb
```diff
@@ -55,28 +55,11 @@
...
         fill_in I18n.t("form.label.activity.recipient_country"), with: "saint"
         find("li.autocomplete__option", text: "Saint Lucia").click
         click_button I18n.t("form.button.activity.submit")
-        click_on I18n.t("default.link.back")
-        click_on I18n.t("tabs.activity.details")
-
-        within(".recipient_country") do
-          expect(page).to have_content "Saint Lucia"
-        end
+        expect(activity.reload.recipient_country).to eq("LC") # Saint Lucia
```
tahb, line 62: Although we could make a back-end assertion like this I think we could get even more value by checking that the full string "Saint Lucia" appears on the page. It'd remove the need for the comment too. Related to the previous comment about how we could do this now the back buttons are gone.

db/schema.rb
lozette: No and yikes, I will recreate my db to make sure this doesn't happen again

lozette:
@tahb The back links don't actually go back to the previous step at the moment, they go back to the Details page. But I'll look into making that happen

tahb:
@lozette Ah damn, looks I invented new behaviour in my excitement! I now recall that the 'back' link goes back to the activity page in either create or edit journeys as all actions are technically updates.

What do you think about a 'Back to activity' link at the top and a 'previous step' button as a secondary to account for both needs?

lozette:
@tahb This is proving to be really complex...

Currently, on the *Edit* journey, the "Continue" button on each form step leads you back to the Details page, so having a "Previous step" button that takes you back to the previous step in the form doesn't seem right. It would only make sense if the Continue button took you to the next step of the form.

In the *Create* journey, it makes more sense to have the "Previous step" button take you to the previous step, because the "Continue" button takes you to the next step. But it's not super simple to only show the "Previous step" button on the Create view (but not the Edit view). There is also the question of disabling the Previous step button on steps that can't be edited (e.g. the level & parent).

Given the amount of work this will be, is it worth adding this new functionality in this ticket, or should we add something to the backlog?

tahb:
@lozette I agree, if it's more complex than expected let's spin it out into a new thing and ignore it for now 👍 

Do you think we should rename 'back' to 'back to activity details'?

lozette:
Screenshot 2020-07-15 at 12 01 56

PR #521, opened 2020-07-15
(779) Refactor how users view activity data

mec:
## Changes in this PR
First up, this PR is a bit clunky - so apologies, with hindsight and the luxury of time, I think this work could be broken down to be more managable to review.

The goals of this work were to:

- add a organisation filter to the viiew at `/activities` only for BEIS users, this allows them to view activities for other organiastions
- remove the activities tables from the 'home' page, which is actually the organisation show view

This change feels a bit like growing pains as we shift the UI to a new state so some things are a bit clunky but I am loathe to make this PR larger, so here is a list of things I think we should think about alongside this work:

- how to handle incomplete activities
- show child activities on their own tab in the activity show view
- extend the all activities scope to include programmes that an organisation is the extending org for
- decide how and where a BEIS user downloads IATI XML, they are left on the org show right now

I wanted to keep specs relating to the display of activity tables so I've had to move things around and re-word some specs, some of these have become slightly odd, for example viewing a programme as a delivery partner is no longer exposed in the interface but is valid (in pundit terms) so I kept the spec.

### Before

BEIS organisation page:
sScreenshot_2020-07-15 Organisation Department for Business, Energy and Industrial Strategy — Report your official developme

UKSA organisation page:
aScreenshot_2020-07-15 Organisation UK Space Agency — Report your official development assistance

### After

BEIS organisation page:
Screenshot_2020-07-15 Organisation Department for Business, Energy and Industrial Strategy — Report your official developme 2

UKSA organisation page:
Screenshot_2020-07-15 Organisation UK Space Agency — Report your official development assistance

BEIS activities page:
Screenshot_2020-07-15 Activities — Report your official development assistance

UKSA activities page:
Screenshot_2020-07-15 Activities — Report your official development assistance pnga

## Next steps


mec:
[Refactor how BEIS users view activity data](https://trello.com/c/PlYiQiJF/779-refactor-how-beis-users-view-activity-data)

mec:
I've decided to split this PR into smaller pieces, it will be back soon…

PR #527, opened 2020-07-17
(792) Every delivery partner Organisation has a Submission created for each Fund they are associated with

lozette:
## Changes in this PR

Trello: https://trello.com/c/4qRgkMw5/792-every-delivery-partner-organisation-has-a-submission-created-for-each-fund-they-are-associated-with

In this PR:

- Create a Submission model and associated tests & test fixtures
- Constrain Submissions by Fund and Organisation. At this point, we should not have more than one Submission per Fund/Organisation pair. This may change in future as we add more Submissions for each pair, on a quarterly basis.
- A data migration to seed the database with a Submission per Organisation/Fund pair
- A Pundit policy to control access to Submissions
- A SubmissionPresenter to style the Submission for the front end
- Show Submissions on the organisation home pages. For BEIS users, they will see all Submissions for all Organisation/Fund pairs. Delivery partner users will only see their own Submissions.

### After

Delivery partner user home page:

Screenshot 2020-07-20 at 14 39 56

## Next steps

### DIFF
--- a/app/controllers/staff/organisations_controller.rb
+++ b/app/controllers/staff/organisations_controller.rb
@@ -16,6 +16,7 @@ class Staff::OrganisationsController < Staff::BaseController
     programme_activities = FindProgrammeActivities.new(organisation: organisation, user: current_user).call
     project_activities = FindProjectActivities.new(organisation: organisation, user: current_user).call
     third_party_project_activities = FindThirdPartyProjectActivities.new(organisation: organisation, user: current_user).call
+    submissions = SubmissionPolicy::Scope.new(current_user, Submission).resolve.includes(:organisation, :fund)
 
     respond_to do |format|
       format.html do
@@ -23,6 +24,7 @@ class Staff::OrganisationsController < Staff::BaseController
         @programme_activities = programme_activities.map { |activity| ActivityPresenter.new(activity) }
         @project_activities = project_activities.map { |activity| ActivityPresenter.new(activity) }
         @third_party_project_activities = third_party_project_activities.map { |activity| ActivityPresenter.new(activity) }
+        @submissions = submissions.map { |submission| SubmissionPresenter.new(submission) }
       end
       format.xml do
         @activities = case level
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -1,0 +1,21 @@
+class Submission < ApplicationRecord
+  include PublicActivity::Common
+
+  validates_presence_of :description
+  validates_presence_of :state
+
+  belongs_to :fund, -> { where(level: :fund) }, class_name: "Activity"
+  belongs_to :organisation
+
+  validates_uniqueness_of :fund, scope: :organisation
+  validate :activity_must_be_a_fund
+
+  enum state: [:inactive]
+
+  def activity_must_be_a_fund
+    return unless fund.present?
+    unless fund.fund?
+      errors.add(:fund, I18n.t("errors.models.submission.attributes.fund"))
+    end
+  end
+end
--- a/app/presenters/submission_presenter.rb
+++ b/app/presenters/submission_presenter.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class SubmissionPresenter < SimpleDelegator
+  def state
+    return if super.blank?
+    I18n.t("label.submission.state.#{super.downcase}")
+  end
+end
--- a/db/migrate/20200716130243_create_submissions.rb
+++ b/db/migrate/20200716130243_create_submissions.rb
@@ -1,0 +1,11 @@
+class CreateSubmissions < ActiveRecord::Migration[6.0]
+  def change
+    create_table :submissions, id: :uuid do |t|
+      t.integer :state, default: 0, null: false
+      t.string :description
+      t.references :fund, type: :uuid
+      t.references :organisation, type: :uuid
+      t.timestamps
+    end
+  end
+end
--- a/spec/factories/submission.rb
+++ b/spec/factories/submission.rb
@@ -1,0 +1,9 @@
+FactoryBot.define do
+  factory :submission do
+    description { Faker::Lorem.paragraph }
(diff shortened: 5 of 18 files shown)


app/presenters/submission_presenter.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class SubmissionPresenter < SimpleDelegator
```
leeky, line 3: Is this presenter necessary for a single attribute? I realise this is a pattern we're using throughout the application, so might make sense for consistency.

vendor/data/dp_to_fund_mappings/dp_mappings_072020.csv
```diff
@@ -0,0 +1,18 @@
...
+British Academy,GB-COH-RC000053,true,true
+UKRI - Central,,true,true
+UKRI - ESRC,,true,true
+UKRI - EPSRC,,true,true
+UKRI - AHRC,,true,true
+UKRI - STFC,,true,true
+UKRI - BBSRC,,true,true
+UKRI - NERC,,true,true
+UKRI - MRC,,true,true
+UKRI - InfalsevateUK,,true,true
```
leeky, line 17: Looks like a rogue `no` was turned into `false` in the middle of this string(!)

db/migrate/20200716130243_create_submissions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSubmissions < ActiveRecord::Migration[6.0]
+  def change
+    create_table :submissions, id: :uuid do |t|
+      t.string :state, default: :inactive
```
leeky, line 4: Wondering whether this would be more useful as an `enum`?

If we need to manually create a submission in the console, a typo in the state would cause the `I18n.t` call to blow the page up potentially.

spec/factories/submission.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :submission do
+    description { Faker::Lorem.paragraph }
+    state { :inactive }
+
+    association :fund, factory: :activity
```
leeky, line 6: Maybe use `:fund_activity` here?

app/controllers/staff/organisations_controller.rb
```diff
@@ -16,13 +16,15 @@ def show
     programme_activities = FindProgrammeActivities.new(organisation: organisation, user: current_user).call
     project_activities = FindProjectActivities.new(organisation: organisation, user: current_user).call
     third_party_project_activities = FindThirdPartyProjectActivities.new(organisation: organisation, user: current_user).call
+    submissions = SubmissionPolicy::Scope.new(current_user, Submission).resolve.includes(:organisation).includes(:fund)
```
leeky, line 19: I think you can also write this as:

`submission =  policy_scope(Submission).includes(:organisation, :fund)`

spec/features/staff/organisation_show_page_spec.rb
```diff
@@ -46,7 +48,7 @@
         end
 
         within("##{fund.id}") do
-          expect(page).to_not have_content I18n.t("summary.label.activity.publish_to_iati.yes")
+          expect(page).to_not have_content I18n.t("summary.label.activity.publish_to_iati.true")
```
leeky, line 51: This doesn't seem to belong here? Also, begs the question of whether we should raise on missing translation in tests and whether it's ok to have a generic I18n key for these  booleans? 🐰 🕳️

vendor/data/dp_to_fund_mappings/dp_mappings_072020.csv
lozette: LOLOLOL oh my god well spotted 😆

app/presenters/submission_presenter.rb
lozette: I've added it as future proofing because we'll probably add more things but... maybe overkill for now??

spec/features/staff/organisation_show_page_spec.rb
lozette: You're right this was me tidying up and I should put it in another commit 🤦‍♀️

app/presenters/submission_presenter.rb
lozette: Making an executive decision to keep it ;)

app/models/submission.rb
```diff
@@ -0,0 +1,21 @@
...
+  belongs_to :fund, -> { where(level: :fund) }, class_name: "Activity"
+  belongs_to :organisation
+
+  validates_uniqueness_of :fund, scope: :organisation
+  validate :activity_must_be_a_fund
+
+  enum state: [:inactive]
+
+  def activity_must_be_a_fund
+    return unless fund
```
leeky, line 16: I found this line confusing - I wonder if `return unless fund.present?` might be more explicit. 

A case of too many `funds` !

db/migrate/20200716130243_create_submissions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSubmissions < ActiveRecord::Migration[6.0]
+  def change
+    create_table :submissions, id: :uuid do |t|
+      t.integer :state, default: 0
```
leeky, line 4: Would `null: false` be useful here?

spec/models/submission_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  describe "associations" do
+    it { should belong_to(:fund).class_name("Activity") }
+    it { should belong_to(:organisation) }
+  end
+
+  it "does not allow an association to an Activity that is not level = fund" do
+    programme = create(:programme_activity)
+    submission = build(:submission, fund: programme)
+    expect(submission.valid?).to be false
```
leeky, line 17: We've also used `expect(submission).not_to be_valid` elsewhere (I think `expect(submission).to be_invalid` might also work)

PR #522, opened 2020-07-15
(824) BEIS users can filter activities by organisation

mec:
## Changes in this PR
This PR is part of the larger work to refactor how users view activities. To keep the PRs small and concise this one will come first.

We have the new activities view that shows a user the acitvities associated to their organisation, for BEIS users we introduce a filter that allows them to switch this view to show activities of any other delivery partner organisation.

As this activities view is becoming the single place for managing activities, it make sense to also move the new 'create activity' button to this page.

The activities view was not using the presenter so incomplete activities where not shown correctly, this has been resolved.

With this work in-place we will move on to make further refinements.

## Screenshots of UI changes
image

image

## Next steps

- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [x] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.
- [ ] Do any environment variables need amending or adding?
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/xml-validation.md)

PR #529, opened 2020-07-21
(779) Remove activities from home page

mec:
## Changes in this PR

With #522 and #525 in place we can remove the list of activties from the home page (the organisation show page) as managing activites can now be done via `/activties` for both BEIS and Delivery partner users.

Note: the download IATI xml button remain on the organisation show page for now as moving those is not in scope for this work.

This PR moves a lot of specs around which was un-avoidable, I hope it is not too painful to review.

### Before
image

### After
image

## Next steps


mec:
[Remove activities tables from 'home' page](https://trello.com/c/PlYiQiJF/779-remove-activities-tables-from-home-page)

app/controllers/staff/organisations_controller.rb
```diff
@@ -12,7 +12,7 @@ def show
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
 
-    project_activities = FindProjectActivities.new(organisation: organisation, user: current_user).call
+    project_activities = FindProjectActivities.new(organisation: organisation, user: current_user).call(eager_load_parent: false)
```
lozette, line 15: Ooooo this is clever

mec: It was all @tahb!

app/controllers/staff/organisations_controller.rb
```diff
@@ -12,16 +12,12 @@ def show
...
-    programme_activities = FindProgrammeActivities.new(organisation: organisation, user: current_user).call
-    project_activities = FindProjectActivities.new(organisation: organisation, user: current_user).call
-    third_party_project_activities = FindThirdPartyProjectActivities.new(organisation: organisation, user: current_user).call
+    project_activities = FindProjectActivities.new(organisation: organisation, user: current_user).call(eager_load_parent: false)
+    third_party_project_activities = FindThirdPartyProjectActivities.new(organisation: organisation, user: current_user).call(eager_load_parent: false)
     submissions = SubmissionPolicy::Scope.new(current_user, Submission).resolve.includes(:organisation, :fund)
 
     respond_to do |format|
       format.html do
-        @fund_activities = fund_activities.map { |activity| ActivityPresenter.new(activity) }
```
tahb, line 23: Is there a case where we show any level of activity on the home page? I wonder if all of the finding of these activities is only required when the user requests XML?

mec: We use these to decide whether we show the download xml button, we agreed finding a good place for these buttons is low priority right now, we can definately make further optimisations on these page as it morphs into the home page again or becomes a true `Organisation` show view.

PR #531, opened 2020-07-22
(793) a BEIS user can edit Submissions to set or change a deadline

lozette:
## Changes in this PR

Trello: https://trello.com/c/QfCsh56Z/793-a-beis-user-can-edit-submissions-to-set-or-change-a-deadline

A Deadline date attributes has been added to the Submission model. This attribute is not required but when set, it can only be set to a future date. 

Setting a deadline to a future date moves the Submission from the "Inactive" state to "Active"

Only BEIS users can edit Submissions, Delivery Partner users do not see the Edit link against their submissions.

### After

Screenshot 2020-07-22 at 15 19 13

Screenshot 2020-07-22 at 15 19 24

Screenshot 2020-07-22 at 15 19 34

Screenshot 2020-07-22 at 15 26 07

## Next steps

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -15,7 +15,7 @@ class Submission < ApplicationRecord
   def activity_must_be_a_fund
     return unless fund.present?
     unless fund.fund?
-      errors.add(:fund, I18n.t("errors.models.submission.attributes.fund"))
+      errors.add(:fund, I18n.t("activerecord.errors.models.submission.attributes.fund.level"))
     end
   end
 end
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -9,7 +9,7 @@ class Submission < ApplicationRecord
 
   validates_uniqueness_of :fund, scope: :organisation
   validate :activity_must_be_a_fund
-  validates :deadline, date_not_in_past: true
+  validates :deadline, date_not_in_past: true, date_within_boundaries: true
 
   enum state: [:inactive, :active]
 
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -11,7 +11,7 @@ class Submission < ApplicationRecord
   validate :activity_must_be_a_fund
   validates :deadline, date_not_in_past: true
 
-  enum state: [:inactive]
+  enum state: [:inactive, :active]
 
   def activity_must_be_a_fund
     return unless fund.present?
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -9,6 +9,7 @@ class Submission < ApplicationRecord
 
   validates_uniqueness_of :fund, scope: :organisation
   validate :activity_must_be_a_fund
+  validates :deadline, date_not_in_past: true
 
   enum state: [:inactive]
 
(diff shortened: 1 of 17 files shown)


spec/factories/submission.rb
```diff
@@ -2,6 +2,7 @@
   factory :submission do
     description { Faker::Lorem.paragraph }
     state { :inactive }
+    deadline { Date.today }
```
mec: Is it worthwile making this a date further in the future to better represent what the real deadline would be?

app/models/submission.rb
```diff
@@ -9,6 +9,7 @@ class Submission < ApplicationRecord
 
   validates_uniqueness_of :fund, scope: :organisation
   validate :activity_must_be_a_fund
+  validates :deadline, date_not_in_past: true
```
mec, line 12: Do you think it would be a good idea to re-use our `DateWithinBoundariesValidator` here as well, just to lessen the chance of crazy future dates?

spec/features/staff/beis_users_can_edit_a_submission_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        fill_in "submission[deadline(2i)]", with: "1"
+        fill_in "submission[deadline(1i)]", with: "2021"
+
+        click_on I18n.t("default.button.submit")
+
+        expect(page).to have_content I18n.t("action.submission.update.success")
+        within "##{submission.id}" do
+          expect(page).to have_content("31 Jan 2021")
+        end
+        auditable_events = PublicActivity::Activity.where(trackable_id: submission.id)
```
mec, line 29: I think we often have a seperate spec for the public activity to document the behaviour, like this one:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/30a244379a6640fc57f9ee3ab7b15401fd4d648e/spec/features/staff/users_can_create_a_budget_spec.rb#L21

I like the explicitness of this approach, what do you think?

spec/factories/submission.rb
```diff
@@ -2,9 +2,13 @@
...
     description { Faker::Lorem.paragraph }
     state { :inactive }
-    deadline { Date.tomorrow }
 
     association :fund, factory: :fund_activity
     association :organisation
+
+    trait :active do
+      state { :active }
+      deadline { Date.tomorrow }
```
mec, line 11: Again, I feel like this being a little furhter might reflect reality a little closer, I am stuggling to reason why though!?

spec/features/staff/beis_users_can_edit_a_submission_spec.rb
```diff
@@ -52,6 +52,34 @@
...
+        fill_in "submission[deadline(2i)]", with: "1"
+        fill_in "submission[deadline(1i)]", with: "2021"
+
+        click_on I18n.t("default.button.submit")
+
+        expect(page).to have_content I18n.t("action.submission.update.success")
+        within "##{submission.id}" do
+          expect(page).to have_content I18n.t("label.submission.state.active")
+        end
+        auditable_event = PublicActivity::Activity.find_by(trackable_id: submission.id)
```
mec, line 77: Same thing here, I know it is a drag but I feel like the seperation of the behaviour warrants a seperate spec?

spec/factories/submission.rb
lozette: Agree!

PR #541, opened 2020-07-24
(790) Introduce a new `transparency_identifier` field for the IATI identifier, which cannot be edited

lozette:
## Changes in this PR

Trello: https://trello.com/c/CAXMJuUd/790-problem-users-can-change-the-identifier-after-an-activity-has-been-published-to-iati-which-would-publish-as-a-new-record

The Trello card for this PR shows a wider problem, which is that users can edit their Activity identifiers; if an activity has previously been published to IATI, then editing its identifier and re-publishing would create a new IATI record.

After much discussion it was decided to try to attempt to resolve this problem in slices, so as not to hold up publication of data to IATI. 

This first phase introduces the concept of a `transparency_identifier` field to the Activity. When an activity is created, the identifier entered by the user is copied into this new field. Currently, both the `identifier` and `transparency_identifier` fields are both immutable, but the intention is eventually to have the `identifier` field as editable, so it can be used by BEIS or the DPs as their own internal identifier. 

The changes in this PR:

- identifier field remains the same
- identifier form step includes more guidance on setting the value
- identifier field cannot be edited once a value has been set, rather than completed to simplify it for now
- when the identifier is set, a new transparency_id field is calculated and saved
- transparency_id is used instead of the iati_identifier method to generate the XML 
- activity details page includes a new row that shows the transparency_id to all users and it is uneditable

### After

Screenshot 2020-07-27 at 14 02 25

Screenshot 2020-07-27 at 13 53 27

## Next steps


app/controllers/staff/activity_forms_controller.rb
```diff
@@ -87,6 +87,7 @@ def update
     end
 
     update_form_state
+    add_transparency_identifier
```
tahb, line 90: As a result of the recent refactor here I think you can move this logic to the identifier step https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/541/commits/a4acb9a64d5788984e0ce489a0e44f218a33d481#diff-825521bf19c854ba561043b5a2d09f21R59 and potentially remove the need for guards.

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -197,4 +198,16 @@ def country_to_region_mapping
...
     yaml["data"]
   end
+
+  def add_transparency_identifier
+    return if @activity.identifier.nil?
+    return if @activity.transparency_identifier.present?
+    @activity.update(transparency_identifier: iati_identifier)
+  end
+
+  def iati_identifier
```
tahb, line 208: I wonder if we could move this method to the activity model and move across the existing unit tests from `ActivityXmlPresenter`. This would keep the extra unit test coverage and allow us to remove this heavier logic from the controller which I think are both additional improvements.

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -118,15 +118,15 @@
       end
 
       context "when the activity is a fund activity" do
-        let(:activity) { create(:fund_activity, organisation: organisation, identifier: "IND-ENT-IFIER") }
+        let(:activity) { create(:fund_activity, organisation: organisation, identifier: "IND-ENT-IFIER", transparency_identifier: "GB-GOV-13-IND-ENT-IFIER") }
```
tahb, line 122: Seeing the first change made to the factory makes me question whether we should do this type of work automatically inside FactoryBot, if possible?

```
transparency_identifier: { "#{parent.iati-identifier}-#{identifier}" }
```

Can we reference the parent like this or would we have to do it in an after create blob?

lozette: Done with a trait & after create hook!

db/data/20200727124151_add_transparency_identifiers_to_all_activities.rb
```diff
@@ -0,0 +1,19 @@
+class AddTransparencyIdentifiersToAllActivities < ActiveRecord::Migration[6.0]
+  def up
```
jcoglan, line 2: As @lozette and I discussed, it is unlikely anything else will modify this data while this transaction runs. If there _is_ anything else that could modify the table, then the `transparency_identifier` field or something it's derived from could change between all the records being loaded in `Activity.all.each`, and each iteration of the loop.

Using a compare-and-swap in the loop body would make sure this only updates records where `transparency_identifier` is still blank:

```rb
      transparency_identifier = activity.iati_identifier
      Activity.where(id: activity.id, transparency_identifier: nil).update_all(transparency_identifier: transparency_identifier)
```

jcoglan: I should add the nice thing about this implementation is it's safe if it partially fails -- if some of the records are assigned an ID and then the migration crashes, it can safely be resumed :)

db/data/20200727124151_add_transparency_identifiers_to_all_activities.rb
```diff
@@ -0,0 +1,19 @@
...
+    activities = Activity.all
+    activities.each do |activity|
+      next if activity.transparency_identifier.present?
+      transparency_identifier = activity.iati_identifier
+      activity.update(transparency_identifier: transparency_identifier)
+      activity.save!
+    end
+  end
+
+  def down
```
jcoglan, line 12: This could be done as a single query with `Activity.update_all(transparency_identifier: nil)`, but this implementation raises an exception if any of them fail, which might be safer. I'd only change this if we have a very large number of activities to update.

PR #546, opened 2020-07-29
(843) Allow Budgets to have a negative value

lozette:
## Changes in this PR

Trello: https://trello.com/c/0OZq4dRU/843-users-can-report-negative-budget-values

BEIS have a requirement to allow negative budget values. This has occurred
due to multi-year budget splitting.

Allow Budgets to have a negative value, but NOT a value of zero.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -224,6 +224,8 @@
 
 ## [unreleased]
 
+- Allow budgets to have a negative value (but not zero)
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-12...HEAD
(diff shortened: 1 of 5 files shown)


jcoglan:
Just asking because I'm not familiar: what is done with this budget data? Are there any parts of the system that assume these values are positive and might do something weird with negative values?

lozette:
At the moment the budgets are merely displayed to the users, we're not using them to make any calculations. So there's no impact now, and nothing assumes they're positive. I believe the budgets will be allocated to DPs and then later displayed as part of the reporting journey https://miro.com/app/board/o9J_ksK4E_g=/

jcoglan:
Cool, in that case this looks fine to merge. Also today I learned: the `other_than` option.

PR #555, opened 2020-08-04
(836) Associate Transactions to a Submission

lozette:
## Changes in this PR

Trello: https://trello.com/c/Lasw3YS4/836-associate-transactions-to-a-submission

Associate Transactions to Submissions. A Transaction can have one Submission; a Submission can have many Transactions.

When a Transaction is created, associate it with the current Submission for that Delivery Partner & Fund.

Do not allow Transactions to be created if there is no Submission in existence.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -231,6 +231,8 @@
 
 ## [unreleased]
 
+- Associate Transactions to Submissions
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-13...HEAD
 [release-13]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-12...release-13
 [release-12]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-11...release-12
--- a/app/controllers/staff/transactions_controller.rb
+++ b/app/controllers/staff/transactions_controller.rb
@@ -22,7 +22,6 @@ class Staff::TransactionsController < Staff::BaseController
 
     if result.success?
       @transaction.create_activity key: "transaction.create", owner: current_user
-      associate_transaction_to_submission
       flash[:notice] = I18n.t("action.transaction.create.success")
       redirect_to organisation_activity_path(@activity.organisation, @activity)
     else
@@ -84,13 +83,6 @@ class Staff::TransactionsController < Staff::BaseController
 
   def activity
     @activity ||= Activity.find(activity_id)
-  end
-
-  def associate_transaction_to_submission
-    organisation = activity.organisation
-    fund = activity.associated_fund
-    submission = Submission.active.find_by(organisation: organisation, fund: fund)
-    @transaction.update!(submission: submission) if submission
   end
 
   private def pre_fill_providing_organisation
--- a/app/controllers/staff/transactions_controller.rb
+++ b/app/controllers/staff/transactions_controller.rb
@@ -4,7 +4,7 @@ class Staff::TransactionsController < Staff::BaseController
   include Secured
 
   def new
-    @activity = Activity.find(activity_id)
+    @activity = activity
     @transaction = Transaction.new
     @transaction.parent_activity = @activity
     pre_fill_providing_organisation
@@ -13,7 +13,7 @@ class Staff::TransactionsController < Staff::BaseController
   end
 
   def create
-    @activity = Activity.find(activity_id)
+    @activity = activity
     authorize @activity
 
     result = CreateTransaction.new(activity: @activity)
@@ -22,6 +22,7 @@ class Staff::TransactionsController < Staff::BaseController
 
     if result.success?
       @transaction.create_activity key: "transaction.create", owner: current_user
+      associate_transaction_to_submission
       flash[:notice] = I18n.t("action.transaction.create.success")
       redirect_to organisation_activity_path(@activity.organisation, @activity)
     else
@@ -40,7 +41,7 @@ class Staff::TransactionsController < Staff::BaseController
     @transaction = Transaction.find(id)
     authorize @transaction
 
-    @activity = Activity.find(activity_id)
+    @activity = activity
     result = UpdateTransaction.new(transaction: @transaction)
       .call(attributes: transaction_params)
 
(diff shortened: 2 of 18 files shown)


app/models/submission.rb
```diff
@@ -14,6 +14,8 @@ class Submission < ApplicationRecord
 
   enum state: [:inactive, :active]
 
+  scope :active, -> { where(state: :active) }
```
leeky, line 17: I think you get this scope for free with the `enum`, same as with the `Activity.project` one..?

app/controllers/staff/transactions_controller.rb
```diff
@@ -81,6 +82,17 @@ def id
     params[:id]
   end
 
+  def activity
+    @activity ||= Activity.find(activity_id)
```
leeky, line 86: Should this use the `ActivityPolicy` scope ?

app/controllers/staff/transactions_controller.rb
```diff
@@ -81,6 +82,17 @@ def id
...
+  end
+
+  def associate_transaction_to_submission
+    organisation = activity.organisation
+    fund = activity.fund? ? activity : activity.parent_activities.select { |a| a.level = :fund }
+    submission = Submission.active.where(organisation: organisation, fund: fund).first
```
leeky, line 92: How about `find_by` instead of `where().first`?

app/controllers/staff/transactions_controller.rb
```diff
@@ -81,6 +82,17 @@ def id
     params[:id]
   end
 
+  def activity
+    @activity ||= Activity.find(activity_id)
+  end
+
+  def associate_transaction_to_submission
+    organisation = activity.organisation
+    fund = activity.fund? ? activity : activity.parent_activities.select { |a| a.level = :fund }
```
leeky, line 91: I'm wondering if there's a better way of finding the active submission for the fund + organisation 🤔

app/policies/transaction_policy.rb
```diff
@@ -4,6 +4,11 @@ def create?
   end
 
   def update?
+    parent_activity = record.parent_activity
+    organisation = parent_activity.organisation
+    fund = parent_activity.fund? ? parent_activity : parent_activity.parent_activities.select { |a| a.level = :fund }
```
leeky, line 9: This is the second place I've seen this logic. Think this might belong in the `Activity` model, but not 100% sure 🤔

app/models/submission.rb
lozette: One day I will learn all of Rails' secrets

app/controllers/staff/transactions_controller.rb
lozette: One day I will also learn how to code

jcoglan:
Does this behaviour imply that we should also associate transactions to the current submission when we implement bulk upload for transaction data?

app/controllers/staff/transactions_controller.rb
jcoglan: I can't immediately think of a single-query way of doing this, because there's an unknown number of jumps from the current activity to its containing fund. If the number of jumps was known, you could do this with joins, but as it is you're essentially following a linked list until you find the thing you're looking for. Loading all of `parent_activities` isn't wasteful because the fund will be the last one in the chain, right?

app/policies/transaction_policy.rb
jcoglan: Ah I see what you mean -- turning this into `Activity#fund?`, rather than changing the machinery of what it's doing with the database :)

lozette:
@jcoglan Good question.... @mec what are your thoughts? I have gone with the logic that Transactions don't *have* to be associated with a Submission, and uploaded Transactions currently won't be associated with the active submission (if one exists - it might not because a submission is only active if it has a deadline).

I think maybe we need to flesh this out a bit more? Or address it in another iteration?

CHANGELOG.md
```diff
@@ -228,6 +228,7 @@
 - Customise error messages according to the content review
 - Add a very basic Submission show page and CSV skeleton
 - Migrate AMS GCRF activities from Level C to Level D
+- Associate Transactions to Submissions
```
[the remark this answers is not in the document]
lozette: Argh! Good spot

app/models/activity.rb
```diff
@@ -116,6 +116,11 @@ def parent_activities
     ancestors.reverse
   end
 
+  def associated_fund
+    return self if fund?
+    parent_activities.detect(&:fund?)
```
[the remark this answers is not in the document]
lozette: It does I think

app/policies/transaction_policy.rb
```diff
@@ -4,6 +4,11 @@ def create?
   end
 
   def update?
+    parent_activity = record.parent_activity
```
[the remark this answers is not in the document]
lozette: Good call, I will do that

spec/features/staff/users_can_create_a_transaction_spec.rb
```diff
@@ -13,6 +13,7 @@
 
     scenario "successfully creates a transaction on an activity" do
       activity = create(:fund_activity, organisation: user.organisation)
+      _submission = create(:submission, :active, organisation: user.organisation, fund: activity)
```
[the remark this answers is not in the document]
lozette: I'm not happy with them either, I will have a think (on Monday)

spec/services/ingest_iati_activities_spec.rb
```diff
@@ -254,7 +258,6 @@
         end
 
         it "returns 0 if there is no attribute and the activity does not have an implementing organisation" do
-          uksa = create(:organisation, name: "UKSA", iati_reference: "GB-GOV-EA31")
```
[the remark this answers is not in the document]
lozette: I had to remove it because with the `let` elsewhere, I was getting a "IATI reference has already been used` error!

spec/features/staff/users_can_create_a_transaction_spec.rb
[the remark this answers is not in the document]
lozette: Fixing this with an `after_create` in the Fund factory

app/policies/transaction_policy.rb
```diff
@@ -1,13 +1,22 @@
...
+    Pundit.policy!(user, record.parent_activity).update? && !!associated_submission&.active?
   end
 
   def destroy?
     Pundit.policy!(user, record.parent_activity).destroy?
   end
+
+  private
+
+  def associated_submission
```
jcoglan, line 16: I'm late to reading through this because of other things on my plate -- sorry! I just wanted to flag this logic as something we might want to refactor out as it duplicates some functionality in `CreateTransaction` below, possibly we can have a look at this while doing bulk upload. But even if we don't, just using this policy and `CreateTransaction` to implement bulk upload should make its behaviour consistent :)

PR #559, opened 2020-08-05
(863) Rename submission and add new reports overview page

mec:
## Changes in this PR

🏴‍☠️ controversy ahead shipmates! 🏴‍☠️ 

This work renames the `Submission` model to `Report` to align with the name of the service and attempt to better reflect the purpose, admittedly, from a delivery partner perspective, they are after all 'reporting their ODA spend to BEIS'.

With the refactor in-place we introduce a new `/reports` route and navigation where users can go to view and manage reports.

- For BEIS users we show both active and inactive reporting (the only state we have right now)
- For Delivery partners we only show active reports as that is the only state they should be exposed to right now.

As the submission journey fills out, more tables with reports in other states will appear in this view.

With the new reports view we can then remvoe the old submissions table from the organisation view.

## Data migrations

Renaming a model like this makes me nervous around the data-migrations and how, when and in which order those are run - we have data migrations that depend on the `Submission` but it feels wrong to refactor those to the new name, that said, I feel like this has bitten me before!

I would great appreciate your view on if I should be worried and how we can mitigate this.

### After
BEIS view:
image

Example delivery partner view:
image

## Next steps


app/controllers/staff/reports_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def id
+    params[:id]
+  end
+
+  def report_params
+    params.require(:report).permit(:deadline, :description)
+  end
+
+  def activate_report
```
leeky, line 54: I'd maybe call this `activate!` as the controller is already for Reports. (The bang `!` because it's saving changes)

leeky:
Looks like `spec/validators/date_not_in_past_validator_spec.rb` also mentions `Submission` FYI.

jcoglan:
To reiterate something I mentioned to @mec earlier: to settle the uncertainty around this, I would prefer sticking to the client's terminology unless there's a very pressing reason not to. I didn't get the impression any of the team was strongly opposed to this so let's get it merged. It needs rebasing and checking for any new uses of the `Submission` name on `develop`.

config/locales/models/report.en.yml
```diff
@@ -2,45 +2,45 @@
...
         state: State
         description: Reporting period
         organisation: Organisation
         fund: Level A
         deadline: Deadline
   page_content:
-    submissions:
-      title: Submissions
-      download_as_csv: Download Submission as a CSV
+    report:
```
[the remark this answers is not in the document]
mec: Don't think it matters as long as the right key is used, but I'll up date it. 👍

spec/features/staff/beis_users_can_edit_a_report_spec.rb
```diff
@@ -0,0 +1,190 @@
+require "rails_helper"
+
+RSpec.feature "BEIS users can edit a report" do
+  let(:delivery_partner_user) { create(:delivery_partner_user) }
+  let!(:report) { create(:report, organisation: delivery_partner_user.organisation, deadline: nil, description: "Legacy Report") }
+
+  context "Logged in as a BEIS user" do
+    scenario "they can edit a Report to set the deadline" do
+      user = create(:beis_user)
```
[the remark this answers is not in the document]
mec: I'm still finding where I land on the `let` I definately feel it gets massively overused, as we have the context here though it feels good to place it at the top. 👍

app/controllers/staff/reports_controller.rb
mec: Nice! I'll update this, thank you.

db/migrate/20200805121536_change_submissions_to_reports.rb
```diff
@@ -0,0 +1,16 @@
...
+    remove_reference :transactions, :submission
+    add_reference :transactions, :report, type: :uuid
+  end
+
+  def self.down
+    remove_reference :transactions, :report
+    add_reference :transactions, :submission, type: :uuid
+  end
+
+  def change
```
leeky, line 12: Also, if you remove the reference and add it again, it'll wipe out the values and you'll lose the existing assoications.

I'm not sure if you can mix `up/down` with `change` - try removing `up` and `down` and using:

```ruby
def change
  rename_table :submissions, :reports
  rename_column :transactions, :submission_id, :report_id
  remove_index :transactions, :submission_id
  add_index :transactions, report_id
end

mec: Ah! Nice, thank you! It seems that renaming the column also does the index, so clean, ❤️  Rails! 😄

jcoglan:
🎉🎉🎉

PR #561, opened 2020-08-07
(859) The Report CSV contains Activity details

lozette:
## Changes in this PR

Trello: https://trello.com/c/QLknCpWs/859-the-report-csv-contains-activity-details

In a previous PR we added a basic CSV skeleton for the Submissions (Reports). This PR adds Activity data to this CSV download. Currently all level C & D activities in a Submission are included in the CSV. 

There is also a small refactor here, I have changed the CSV export from HAML to a CSV object sent with event streaming

Note that this should be merged after https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/555 and https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/559 because there is some overlap with functionality (`Activity#associated_fund`) and some changes of names will need to be refactored in.

## Next steps

### DIFF
--- a/app/controllers/staff/submissions_controller.rb
+++ b/app/controllers/staff/submissions_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "csv"
+
 class Staff::SubmissionsController < Staff::BaseController
   include Secured
 
@@ -10,7 +12,17 @@ class Staff::SubmissionsController < Staff::BaseController
     respond_to do |format|
       format.html
       format.csv do
-        response.headers["Content-Disposition"] = "attachment; filename=\"#{ERB::Util.url_encode(@submission.description)}.csv\""
+        fund = @submission.fund
+        @projects = Activity.project.where(organisation: @submission.organisation).select { |activity| activity.associated_fund == fund }
+        @third_party_projects = Activity.third_party_project.where(organisation: @submission.organisation).select { |activity| activity.associated_fund == fund }
+        csv = Activity::CSV_HEADERS.to_csv
+        @projects.each do |project|
+          csv << ExportActivityToCsv.new(activity: project).call
+        end
+        @third_party_projects.each do |project|
+          csv << ExportActivityToCsv.new(activity: project).call
+        end
+        send_data csv, {filename: "#{@submission.description}.csv", type: "text/csv"}
       end
     end
   end
--- a/app/controllers/staff/submissions_controller.rb
+++ b/app/controllers/staff/submissions_controller.rb
@@ -4,6 +4,7 @@ require "csv"
 
 class Staff::SubmissionsController < Staff::BaseController
   include Secured
+  include ActionController::Live
 
   def show
     @submission = Submission.find(id)
@@ -15,14 +16,7 @@ class Staff::SubmissionsController < Staff::BaseController
         fund = @submission.fund
         @projects = Activity.project.where(organisation: @submission.organisation).select { |activity| activity.associated_fund == fund }
         @third_party_projects = Activity.third_party_project.where(organisation: @submission.organisation).select { |activity| activity.associated_fund == fund }
-        csv = Activity::CSV_HEADERS.to_csv
-        @projects.each do |project|
-          csv << ExportActivityToCsv.new(activity: project).call
-        end
-        @third_party_projects.each do |project|
-          csv << ExportActivityToCsv.new(activity: project).call
-        end
-        send_data csv, {filename: "#{@submission.description}.csv", type: "text/csv"}
+        send_csv
       end
     end
   end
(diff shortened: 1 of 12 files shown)


app/controllers/staff/submissions_controller.rb
```diff
@@ -10,7 +12,17 @@ def show
     respond_to do |format|
       format.html
       format.csv do
-        response.headers["Content-Disposition"] = "attachment; filename=\"#{ERB::Util.url_encode(@submission.description)}.csv\""
+        fund = @submission.fund
+        @projects = Activity.project.where(organisation: @submission.organisation).select { |activity| activity.associated_fund == fund }
```
leeky, line 16: It's a pity that `Activity` and `Submission` aren't yet associated, as it would avoid the inefficient `select`:

```ruby
@projects = @submission.activities.project
@third_party_projects = @submission.activities.third_party_project
```

or, if you don't care about the ordering:

```ruby
@projects = @submission.activities.where(level: [:project, :third_party_project])
```

app/controllers/staff/submissions_controller.rb
```diff
@@ -10,7 +12,17 @@ def show
...
+        @third_party_projects = Activity.third_party_project.where(organisation: @submission.organisation).select { |activity| activity.associated_fund == fund }
+        csv = Activity::CSV_HEADERS.to_csv
+        @projects.each do |project|
+          csv << ExportActivityToCsv.new(activity: project).call
+        end
+        @third_party_projects.each do |project|
+          csv << ExportActivityToCsv.new(activity: project).call
+        end
+        send_data csv, {filename: "#{@submission.description}.csv", type: "text/csv"}
```
leeky, line 25: This might eventually need `ActionController::Live` (https://api.rubyonrails.org/classes/ActionController/Live.html) to avoid holding the entire CSV in memory before pushing it out.

app/services/export_activity_to_csv.rb
```diff
@@ -0,0 +1,29 @@
...
+      activity_presenter.status,
+      activity_presenter.planned_start_date,
+      activity_presenter.actual_start_date,
+      activity_presenter.planned_end_date,
+      activity_presenter.actual_end_date,
+      activity_presenter.recipient_region,
+      activity_presenter.recipient_country,
+      activity_presenter.aid_type,
+      activity_presenter.level,
+      "https://#{ENV["DOMAIN"]}#{Rails.application.routes.url_helpers.organisation_activity_details_path(activity_presenter.organisation, activity_presenter)}",
```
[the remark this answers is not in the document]
leeky: Also, might benefit from using the `_url` helper rather than `_path`. Might open the can of `default_url_options`-shaped worms however. But means we could potentially be able to have the links work correctly in development.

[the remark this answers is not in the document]
lozette: Good call, I didn't even think of that!

app/policies/submission_policy.rb
```diff
@@ -15,6 +24,11 @@ def update?
     beis_user?
   end
 
+  def download?
```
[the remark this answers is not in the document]
lozette: Good question, I think it should

app/controllers/staff/submissions_controller.rb
jcoglan: Can this be improved now that #555 has landed?

app/controllers/staff/submissions_controller.rb
```diff
@@ -65,4 +58,17 @@ def activate_submission
       @submission.create_activity key: "submission.activate", owner: current_user
     end
   end
+
+  def send_csv
+    response.headers["Content-Type"] = "text/event-stream"
```
jcoglan, line 63: I like the idea of streaming this row-by-row, that's a nice improvement and I didn't know Rails directly supported streaming now. I'm not sure about the MIME type here as `text/event-stream` is typically used specifically for the Server-Sent Events protocol. Is there a reason to use it here rather than `text/csv`? Does it cause the browser to handle the download differently?

app/controllers/staff/submissions_controller.rb
leeky: @jcoglan Sadly not - that's associating transactions to submissions not activities.

app/controllers/staff/submissions_controller.rb
lozette: Do you know what, I don't know - all the examples I found of this implementation used `event-stream` so I just went with it! But it turns out that `text/csv` works as well so I'll set it back to that.

leeky: I think the controller also needs `include ActionController::Live` added to make this work

(Source: https://api.rubyonrails.org/classes/ActionController/Live.html)

lozette: Oops. Added.

jcoglan: We've talked about this and determined that although that Rails API looks like it's specifically designed for Server-Sent Events, it does allow streaming arbitrary data and does not prefix each line with `data: `, so this implementation looks good 👍

PR #564, opened 2020-08-10
(809) Ingest at other levels

leeky:
Change ingest so that it can create activities at any level specified by the mapping file, rather than assuming project level. This means we'll be able to ingest data for GCRF (whose IATI data represents programme level).

I've used the more generic terms "activity" and "parent" rather than "project" and "programme" throughout to reflect this change.

## Next steps


app/models/activity.rb
```diff
@@ -133,6 +133,13 @@ def parent_level
     Activity.levels.keys[existing_level_index - 1]
   end
 
+  def child_level
+    existing_level_index = Activity.levels.keys.index(level)
```
jcoglan, line 137: The reliance on the ordering of a hash here (`enum :level`) makes me nervous. I know Ruby hashes are insertion order but it's quite easy to break things that rely on this by accident. Is there a more robust way we could encode the hierarchy?

leeky: @jcoglan I tried a couple of approaches, but in the end settled for a [far simpler solution ](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/564/commits/c79e0a0c2c085a5b85fa1d334cff92749ed538e7).

jcoglan: That looks a lot more robust and less easy to break accidentally 👍 

This is also made more robust by the fact we store levels by name, not by implicitly mapping them to integers in the database which would make it more sensitive to the `enum` being re-ordered.

PR #566, opened 2020-08-11
(864) Transaction actuals totals are included in the CSV download

lozette:
## Changes in this PR

Trello: https://trello.com/c/fp8pacDC/864-transaction-actuals-totals-are-included-in-the-csv-download

Add a `transactions_total` method to the Activity model to calculate the sum of all transactions associated to that Activity (thank you @mec for the spike)

Style the total using the ActivityPresenter.

Add the total to the Submission CSV download per Activity line.

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -139,4 +139,8 @@ class Activity < ApplicationRecord
       parent_identifiers << parent.identifier
     }.push(identifier).join("-")
   end
+
+  def transactions_total
+    transactions.sum(:value)
+  end
 end
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -4,7 +4,7 @@ class Activity < ApplicationRecord
 
   STANDARD_GRANT_FINANCE_CODE = "110"
   UNTIED_TIED_STATUS_CODE = "5"
-  CSV_HEADERS = ["Identifier", "Transparency identifier", "Sector", "Title", "Description", "Status", "Planned start date", "Actual start date", "Planned end date", "Actual end date", "Recipient region", "Recipient country", "Aid type", "Level", "Link to activity in RODA"]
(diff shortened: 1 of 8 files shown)


app/models/activity.rb
```diff
@@ -139,4 +139,8 @@ def iati_identifier
       parent_identifiers << parent.identifier
     }.push(identifier).join("-")
   end
+
+  def transaction_total
```
leeky, line 143: How about `transactions_total` to match the association name `transactions` ?

spec/presenters/activity_presenter_spec.rb
```diff
@@ -270,4 +270,15 @@
...
   end
+
+  describe "#transaction_total" do
+    it "returns the transaction total as a formatted number with currency symbol" do
+      project = create(:project_activity)
+      transaction_1 = create(:transaction, parent_activity: project)
+      transaction_2 = create(:transaction, parent_activity: project)
+
+      expect(described_class.new(project).transaction_total)
+        .to eq ActionController::Base.helpers.number_to_currency((transaction_1.value + transaction_2.value), unit: "£")
```
leeky, line 281: My preference is to explicitly set `value` in `create(:transaction, ... )` and then the expectation is looking for a specific value (e.g. "£1234.56"). 

That way if we somehow break `number_to_currency` we know about it.

spec/features/staff/beis_users_can_edit_a_submission_spec.rb
```diff
@@ -132,7 +132,7 @@
         click_on I18n.t("default.button.submit")
 
         auditable_events = PublicActivity::Activity.where(trackable_id: submission.id)
-        expect(auditable_events.map(&:key)).to eq ["submission.update", "submission.activate"]
+        expect(auditable_events.map(&:key)).to match_array ["submission.update", "submission.activate"]
```
leeky, line 135: :+1:

app/models/activity.rb
lozette: Good call 👍

PR #568, opened 2020-08-12
(891) URL encode report CSV filename so it is not truncated

lozette:
## Changes in this PR

Trello: https://trello.com/c/PcSYrR3E/891-the-csv-attachment-filename-is-truncated

URL encode the CSV attachment filename so it does not get truncated. Tested on a Mac in Chrome, Safari & FF; hopefully this covers most bases.

## Next steps

### DIFF
--- a/app/controllers/staff/reports_controller.rb
+++ b/app/controllers/staff/reports_controller.rb
@@ -85,7 +85,7 @@ class Staff::ReportsController < Staff::BaseController
 
   def send_csv
     response.headers["Content-Type"] = "text/csv"
-    response.headers["Content-Disposition"] = "attachment; filename=#{@report.description}.csv"
+    response.headers["Content-Disposition"] = "attachment; filename=#{ERB::Util.url_encode(@report.description)}.csv"
     response.stream.write Activity::CSV_HEADERS.to_csv
     @projects.each do |project|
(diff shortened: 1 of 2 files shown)


spec/features/staff/users_can_view_a_submission_spec.rb
```diff
@@ -31,7 +31,7 @@
       expect(page.response_headers["Content-Type"]).to include("text/csv")
 
       header = page.response_headers["Content-Disposition"]
-      expect(header).to match(/#{submission.description}/)
+      expect(header).to match(/#{ERB::Util.url_encode(submission.description)}/)
```
leeky, line 34: Does the `submission.description` used in this example contain a string that would trigger the truncation issue?

lozette: It has a space in it so yes I think so

lozette: Maybe I should make the test explicitly check for `Legacy%20Submission` which is the actual string 🤔

jcoglan: Yeah, checking for a concrete value rather than some dynamically produced value usually makes me more confident in what the test is doing.

PR #569, opened 2020-08-13
(889) Associate PlannedDisbursements to Reports

lozette:
## Changes in this PR

Trello: https://trello.com/c/GiEQQVOZ/889-associate-planned-disbursements-to-reports

Create an association between Planned Disbursements and Reports, much like we did for Transactions and Reports in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/555

We assume an `active` Report always exists, in order to associate with Planned Disbursements on create and ingest. 

Add a `:with_report` trait to the `project_activity` factory, to create an associated Report after the Project Activity is created.

## Next steps

### DIFF
--- a/app/services/create_planned_disbursement.rb
+++ b/app/services/create_planned_disbursement.rb
@@ -9,6 +9,7 @@ class CreatePlannedDisbursement
     planned_disbursement = PlannedDisbursement.new
 
     planned_disbursement.parent_activity = activity
+    planned_disbursement.report = report
     planned_disbursement.assign_attributes(attributes)
     planned_disbursement.value = sanitize_monetary_string(value: attributes[:value])
 
@@ -25,5 +26,11 @@ class CreatePlannedDisbursement
 
   def sanitize_monetary_string(value:)
     Monetize.parse(value)
+  end
+
+  def report
+    organisation = activity.organisation
+    fund = activity.associated_fund
+    Report.active.find_by(organisation: organisation, fund: fund)
   end
 end
--- a/spec/factories/activity.rb
+++ b/spec/factories/activity.rb
@@ -71,6 +71,13 @@ FactoryBot.define do
       association :extending_organisation, factory: :delivery_partner_organisation
       association :reporting_organisation, factory: :beis_organisation
 
+      trait :with_report do
+        after(:create) do |project|
+          fund = project.associated_fund
+          create(:report, :active, fund: fund, organisation: project.organisation)
+        end
+      end
+
       factory :project_activity_with_implementing_organisations do
         transient do
           implementing_organisations_count { 3 }
--- a/spec/factories/activity.rb
+++ b/spec/factories/activity.rb
@@ -23,6 +23,13 @@ FactoryBot.define do
     association :organisation, factory: :organisation
     association :reporting_organisation, factory: :beis_organisation
 
+    trait :with_report do
+      after(:create) do |activity|
+        fund = activity.associated_fund
+        create(:report, :active, fund: fund, organisation: activity.organisation)
+      end
+    end
+
     factory :fund_activity do
       level { :fund }
       funding_organisation_name { "HM Treasury" }
@@ -35,12 +42,6 @@ FactoryBot.define do
       association :organisation, factory: :beis_organisation
       association :extending_organisation, factory: :beis_organisation
       association :reporting_organisation, factory: :beis_organisation
-
-      trait :with_report do
-        after(:create) do |fund|
-          create(:report, :active, fund: fund, organisation: fund.organisation)
-        end
-      end
     end
 
     factory :programme_activity do
@@ -70,13 +71,6 @@ FactoryBot.define do
 
       association :extending_organisation, factory: :delivery_partner_organisation
       association :reporting_organisation, factory: :beis_organisation
-
-      trait :with_report do
-        after(:create) do |project|
-          fund = project.associated_fund
-          create(:report, :active, fund: fund, organisation: project.organisation)
-        end
-      end
 
       factory :project_activity_with_implementing_organisations do
         transient do
--- a/spec/factories/planned_disbursement.rb
+++ b/spec/factories/planned_disbursement.rb
@@ -11,5 +11,7 @@ FactoryBot.define do
     receiving_organisation_name { Faker::Company.name }
     receiving_organisation_reference { "GB-COH-{#{Faker::Number.number(digits: 6)}}" }
     receiving_organisation_type { "70" }
+
+    association :report
   end
 end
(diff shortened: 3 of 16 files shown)


app/services/create_planned_disbursement.rb
```diff
@@ -9,6 +9,7 @@ def call(attributes: {})
     planned_disbursement = PlannedDisbursement.new
 
     planned_disbursement.parent_activity = activity
+    planned_disbursement.report = report(activity: activity)
```
leeky, line 12: I don't think you need to pass the activity as a parameter here, because of the `attr_accessor`.

app/services/create_planned_disbursement.rb
```diff
@@ -26,4 +27,10 @@ def call(attributes: {})
   def sanitize_monetary_string(value:)
     Monetize.parse(value)
   end
+
+  def report(activity:)
+    organisation = activity.organisation
+    fund = activity.associated_fund
+    Report.active.find_by(organisation: organisation, fund: fund)
```
leeky, line 34: This is definitely calling out for an association between `Report` and `Fund`, and there are several places where we're associating the two without explicitly calling it such (e.g. the `with_report` trait)

Would lead to being able to do `planned_disbursement.report = activity.report` or even `belongs_to :report, through: :parent_activity` 🥳

spec/factories/planned_disbursement.rb
```diff
@@ -11,5 +11,7 @@
     receiving_organisation_name { Faker::Company.name }
     receiving_organisation_reference { "GB-COH-{#{Faker::Number.number(digits: 6)}}" }
     receiving_organisation_type { "70" }
+
+    association :report
```
leeky, line 15: Won't this automatically create an associated report and anything using the trait `with_report` will create an additional one..?

spec/features/staff/users_can_create_a_planned_disbursement_spec.rb
```diff
@@ -54,6 +54,23 @@
...
+
+      visit activities_path
+
+      click_on(project.title)
+      click_on(I18n.t("page_content.planned_disbursements.button.create"))
+
+      fill_in_planned_disbursement_form
+
+      planned_disbursement = PlannedDisbursement.last
+      report = Report.find_by(fund: fund, organisation: project.organisation)
```
leeky, line 70: Should the test explicitly create the report we expect the planned disbursement to be associated with?

lozette:
@leeky I am also wondering that. However I think one Activity can span multiple reports 🤔

app/services/create_planned_disbursement.rb
lozette: 🤦

app/services/create_planned_disbursement.rb
lozette: Yeah we should discuss this with the other devs. Only caveat is that a Fund can (and will) belong to multiple Reports, so we'd need to handle that carefully.

jcoglan: Oh, that makes sense. I was confused about associating `Activity` with `Report` because many reports will include entries for the same activity, but a report is definitely scoped to a single fund.

spec/factories/planned_disbursement.rb
lozette: Possibly; however in (e.g.) spec/features/staff/users_can_create_a_planned_disbursement_spec.rb:57 the Planned Disbursement isn't being created with a factory, it's being created through the UI, so I need to use a `project_activity` `:with_report` to make sure a Report is available for the Planned Disbursement to attach to. IYSIM?

app/services/create_planned_disbursement.rb
jcoglan: Are we still assuming there will _always_ be an active report available? I'd got the impression that there are times there won't be one (e.g. while the latest one is in review) and we'd need to prevent the creation of transactions and planned disbursements at those times.

Is this taken care of somewhere else, e.g. by blocking access to the form when no active report exists?

spec/features/staff/users_can_create_a_planned_disbursement_spec.rb
lozette: Yeah that makes more sense 👍

app/services/create_planned_disbursement.rb
mec: I am wary of jumping in here, but yes I agree, in my head fund plays two roles for Report:

1. As a fund is just an activity, it should be associated to the report in which it was created - however as there are only two funds that already exisit, we don't really see this behaviour in production.
2. A report contains only child activities for a given fund i.e. A report is for GCRF activities or Newton activities.

I'll add a card to discuss this next week so we are clear and can write the stories to do the work - I suspect these will be more technical stories (as opposed to user stories) again.

app/services/create_planned_disbursement.rb
lozette: No this hasn't been taken into account, and I haven't seen any decisions about assuming there will always be an active report available. This logic is in transactions as well (assuming one will always be available). If we need to re-work this then I think we need a separate card to address it for planned disbursements *and* transactions.

mec: @james is correct. There will be a period of time when a report is `inactive` (as well as others to come) and we should stop the creation of new entities for this period of time. I hope we can use the policies to do this?

mec: James is right, there will be a period of time when a report is inactive (there are others as well) and we should stop entities being created - I am hopeful this can be done via the policies?

There are other states as well, we have the basic mapping of this here: 

https://miro.com/app/board/o9J_knq_KVg=/?moveToWidget=3074457349237052629&cot=12

I don't think this needs to happen as part of this work as we still are in this odd 'one giant backfill report' at the mo.

lozette: After discussion with @mec I've added a card to address this for PDs and Transactions https://trello.com/c/Yopp4pKf/906-we-cannot-assume-there-will-always-be-an-active-report

lozette: I'm going to amend this PR to forbid creation of planned disbursements when there is no active report (like Transactions are now) but we'll address the Report's other states in that card I've created.

spec/factories/planned_disbursement.rb
leeky: I see your point, although I'd probably manually create the report and then reference it in `project_activity` in this case.

Just trying to avoid having tests that are creating unnecessary associations - for example, if you created several different planned disbursements, by default they'd each be associated with a brand new project, which doesn't really reflect what would happen in reality.

lozette: Ah OK I didn't realise what you were getting at! I think I was labouring under an assumption that Factories needed to reflect the associations of their models, hence I put it in.

lozette: Although it turns out if I remove the association, all my PlannedDisbursementPolicy tests fail

spec/factories/activity.rb
```diff
@@ -101,6 +108,13 @@
 
       association :extending_organisation, factory: :delivery_partner_organisation
       association :reporting_organisation, factory: :beis_organisation
+
+      trait :with_report do
```
leeky, line 112: Couldn't you move this outside of the `factory ..._activity do` block and have it top-level?

lozette: Done, thanks for the pointer!

PR #570, opened 2020-08-13
(845) Add `programme status` field to activity form

MariluzHerrera:
Trello: https://trello.com/c/jj2Dju4x

## Changes in this PR

After a user research session, users feedback that they are missing some important fields in the activity form in RODA. It was then identified as a priority for them to have the field `programme status` added to the activity form so they can track the status of each activity. Users preferred this status to the previously used `IATI status`, as they are more familiar with the terminology of Programme status. Since IATI status is a requirement for IATI, we will still keep it in the background to ensure the reports follow the IATI standards and pass the IATI validator, but we will not ask the user to introduce an IATI status anymore when they create activities (the form step `status` has been removed and now they are asked for a `programme status` instead). As a result, we now need to map one programme status to a specific IATI status. The client provided us with a mapping chart for this. The mapping logic has been implemented as part of this PR.

This PR also introduces the new field in the form, alongside the mapping logic and a placeholder for future work on ingesting non-IATI data.

### Before

Screenshot 2020-08-13 at 12 11 40

### After

Screenshot 2020-08-13 at 10 59 26

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -12,7 +12,7 @@ class Activity < ApplicationRecord
     :purpose_step,
     :sector_category_step,
     :sector_step,
-    :status_step,
+    :programme_status_step,
     :geography_step,
     :region_step,
     :country_step,
@@ -28,7 +28,7 @@ class Activity < ApplicationRecord
   validates :title, :description, presence: true, on: :purpose_step
   validates :sector_category, presence: true, on: :sector_category_step
   validates :sector, presence: true, on: :sector_step
-  validates :status, presence: true, on: :status_step
+  validates :programme_status, presence: true, on: :programme_status_step
   validates :geography, presence: true, on: :geography_step
   validates :recipient_region, presence: true, on: :region_step, if: :recipient_region?
   validates :recipient_country, presence: true, on: :country_step, if: :recipient_country?
--- a/app/services/ingest_iati_activities.rb
+++ b/app/services/ingest_iati_activities.rb
@@ -33,6 +33,7 @@ class IngestIatiActivities
           add_participating_organisation(delivery_partner: delivery_partner, new_activity: new_activity, legacy_activity: legacy_activity)
           add_title(legacy_activity: legacy_activity, new_activity: new_activity)
           add_description(legacy_activity: legacy_activity, new_activity: new_activity)
+          add_programme_status(legacy_activity: legacy_activity, new_activity: new_activity)
           add_status(legacy_activity: legacy_activity, new_activity: new_activity)
           add_sector(legacy_activity: legacy_activity, new_activity: new_activity)
           add_flow(legacy_activity: legacy_activity, new_activity: new_activity)
@@ -68,6 +69,11 @@ class IngestIatiActivities
   private def add_description(legacy_activity:, new_activity:)
     description = legacy_activity.elements[3].children.detect { |child| child.name.eql?("narrative") }.children.text if legacy_activity.elements[3].name.eql?("description")
     new_activity.description = normalize_string(description)
+  end
+
+  private def add_programme_status(legacy_activity:, new_activity:)
+    # This will need to be replaced once we have a bi-directional mapping between programme_status and IATI status
+    new_activity.programme_status = "Replace me"
   end
 
   private def add_status(legacy_activity:, new_activity:)
--- a/config/locales/codelists/2_03/iati.en.yml
+++ b/config/locales/codelists/2_03/iati.en.yml
@@ -658,6 +658,18 @@ en:
       '920': SUPPORT TO NON- GOVERNMENTAL ORGANISATIONS (NGOs)
       '930': Refugees in Donor Countries
       '998': Unallocated / Unspecified
+    programme_status:
+      '01': Delivery
+      '02': Planned
+      '03': Agreement in place
+      '04': Call/Activity open
+      '05': Review
+      '06': Decided
+      '07': Spend in progress
+      '08': Finalisation
+      '09': Completed
+      '10': Stopped
+      '11': Cancelled
     status:
       '1': Pipeline/identification
       '2': Implementation
--- a/db/migrate/20200803104010_add_programme_status_to_activities.rb
+++ b/db/migrate/20200803104010_add_programme_status_to_activities.rb
@@ -1,0 +1,5 @@
+class AddProgrammeStatusToActivities < ActiveRecord::Migration[6.0]
+  def change
+    add_column :activities, :programme_status, :string
+  end
+end
--- a/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
+++ b/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
@@ -167,22 +167,20 @@ RSpec.feature "Users can create a fund level activity" do
         click_button I18n.t("form.button.activity.submit")
 
         expect(page).to have_content I18n.t("form.legend.activity.sector", sector_category: "Basic Education", level: "programme")
-
         # Don't provide a sector
         click_button I18n.t("form.button.activity.submit")
         expect(page).to have_content I18n.t("activerecord.errors.models.activity.attributes.sector.blank")
 
         choose "Primary education"
         click_button I18n.t("form.button.activity.submit")
-        expect(page).to have_content I18n.t("form.legend.activity.status", level: "programme")
+        expect(page).to have_content I18n.t("form.legend.activity.programme_status", level: "programme")
 
-        # Don't provide a status
+        # Don't provide a programme status
         click_button I18n.t("form.button.activity.submit")
-        expect(page).to have_content I18n.t("activerecord.errors.models.activity.attributes.status.blank")
+        expect(page).to have_content "can't be blank"
 
-        choose("activity[status]", option: "2")
+        choose("activity[programme_status]", option: "07")
         click_button I18n.t("form.button.activity.submit")
-
         expect(page).to have_content I18n.t("page_title.activity_form.show.dates", level: "programme")
 
         click_button I18n.t("form.button.activity.submit")
--- a/spec/lib/programme_to_iati_status_spec.rb
+++ b/spec/lib/programme_to_iati_status_spec.rb
@@ -1,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe "#programme_status_to_iati_status" do
+  context "when the user sets the programme status for an activity" do
+    it "sets the correct IATI status automatically" do
+      status = ProgrammeToIatiStatus.new.programme_status_to_iati_status("01")
+      expect(status).to eq "2"
+
+      status = ProgrammeToIatiStatus.new.programme_status_to_iati_status("02")
+      expect(status).to eq "1"
+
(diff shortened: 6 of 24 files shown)


vendor/data/codelists/BEIS/programme_status_to_iati_status_mapping.yml
```diff
@@ -0,0 +1,23 @@
+data:
```
leeky, line 1: How about defining this mapping in the `ProgrammeToIatiStatus` class itself, e.g.

```ruby
STATUS_MAPPING = {
  "01" => "2",
  "02" => "1",
  "03" => "1"
  ... etc...
}.freeze
```

spec/lib/programme_to_iati_status_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe "#programme_status_to_iati_status" do
+  context "when the user sets the programme status for an activity" do
+    it "the correct iati status will be set automatically" do
```
leeky, line 5: This read strangely when you include the `it`.

How about `it "sets the correct IATI status automatically"` ?

spec/lib/programme_to_iati_status_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      status = ProgrammeToIatiStatus.new.programme_status_to_iati_status("01")
+      expect(status).to eq "2"
+
+      status = ProgrammeToIatiStatus.new.programme_status_to_iati_status("02")
```
leeky, line 9: I'm not sure if testing all the possible inputs and outputs is necessary here.

app/services/ingest_iati_activities.rb
```diff
@@ -67,6 +68,11 @@ def call
     new_activity.description = normalize_string(description)
   end
 
+  private def add_programme_status(legacy_activity:, new_activity:)
+    # This will need to be replaced once we have a bi-directional mapping between programme_status and IATI status
+    new_activity.programme_status = "Replace me"
```
leeky, line 73: What happens if you try to edit an activity whose `programme_status` is set like this?

config/locales/codelists/2_03/iati.en.yml
```diff
@@ -658,6 +658,18 @@ en:
       '920': SUPPORT TO NON- GOVERNMENTAL ORGANISATIONS (NGOs)
       '930': Refugees in Donor Countries
       '998': Unallocated / Unspecified
+    programme_status:
```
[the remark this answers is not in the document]
MariluzHerrera: True. I didn't notice. What would be a good place for it? `vendor/data/codelists/BEIS` ?

app/models/activity.rb
```diff
@@ -13,7 +13,7 @@ class Activity < ApplicationRecord
     :purpose_step,
     :sector_category_step,
     :sector_step,
-    :status_step,
+    :programme_status_step,
```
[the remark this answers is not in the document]
MariluzHerrera: Thanks for pointing this out leeky! I had no idea about the `form_state`. I think I'll need a hand with this issue due to lack of familiarity

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
```diff
@@ -174,15 +174,14 @@
 
         choose "Primary education"
         click_button I18n.t("form.button.activity.submit")
-        expect(page).to have_content I18n.t("form.legend.activity.status", level: "programme")
+        expect(page).to have_content I18n.t("form.legend.activity.programme_status", level: "programme")
 
-        # Don't provide a status
+        # Don't provide an activity status
```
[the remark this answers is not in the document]
MariluzHerrera: 👍

spec/support/form_helpers.rb
```diff
@@ -141,6 +141,11 @@ def fill_in_activity_form(
       month: planned_end_date_month,
       day: planned_end_date_day
     )
+
+    my_activity = Activity.find_by(identifier: identifier)
+    iati_status = ProgrammeToIatiStatus.new.programme_status_to_iati_status(programme_status)
+    expect(my_activity.status).not_to be_nil
```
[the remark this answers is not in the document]
MariluzHerrera: I paired on this with James and I added that line as a suggestion from him. I thought it was a good idea to have an assertion (safety net) that the IATI status won't be nil

vendor/data/codelists/BEIS/programme_status_to_iati_status_mapping.yml
MariluzHerrera: That makes more sense. Thanks leeky!

app/models/activity.rb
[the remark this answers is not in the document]
jcoglan: I also didn't realise the form state was persisted. Would it make sense to keep the existing name as `status_step` and thus not break any existing records, or would this create more confusion over terminology?

spec/support/form_helpers.rb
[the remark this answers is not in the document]
jcoglan: Yes, this was a safety check to make sure the test doesn't pass accidentally. `ProgrammeToIatiStatus.new.programme_status_to_iati_status` can return `nil`, so without this check it's possible for the test to pass when `activity.status` is not set to any value.

db/migrate/20200803104010_add_programme_status_to_activities.rb
```diff
@@ -0,0 +1,5 @@
+class AddProgrammeStatusToActivities < ActiveRecord::Migration[6.0]
+  def change
+    add_column :activities, :programme_status, :string
```
jcoglan, line 3: Should existing activities be given a value for this field, and should it have a default value for new records? Or should the default be left as `NULL`?

spec/lib/programme_to_iati_status_spec.rb
jcoglan: I thought it was valuable to have some representation of the required mappings in the test suite. Part of the motivation for extracting this logic into its own class was so we could test that conversion without going through the whole workflow for each variation.

db/migrate/20200803104010_add_programme_status_to_activities.rb
jcoglan: After reading the final commit I just realised this is one of those tricky things -- `programme_status` is ultimately required, but we save partially-complete activitieis during the multi-step form process, so we need to allow `NULL` in the database. We may still want to set a default value that reflects the concerns in the last commit here.

jcoglan:
I really like how you've broken down this work into different pieces and written each one up, made it easy to follow :)

spec/lib/programme_to_iati_status_spec.rb
MariluzHerrera: I'll use your suggestion here. Sounds definitely better!

spec/lib/programme_to_iati_status_spec.rb
MariluzHerrera: Shall I leave just one or two then? I never know how many tests are too many tests O:-)

app/services/ingest_iati_activities.rb
MariluzHerrera: That's a very good question, and honestly I don't have an answer. What do you reckon the best approach is?

spec/lib/programme_to_iati_status_spec.rb
jcoglan: Personally I would rather have all the mapping cases tested. These test will be fast to run and I don't expect them to cause much maintenance overhead.

MariluzHerrera: Yes, I think I was typing my last message without having seen yours James. I'll leave it as it is :-)

db/migrate/20200803104010_add_programme_status_to_activities.rb
MariluzHerrera: I'm not sure what the best thing to do here is. Same as the conversation above about the `Replace me` placeholder. These both are things I haven't considered but that definitely need attention, and I will need help to make decisions. Probably better to have a conversation if we have time

app/models/activity.rb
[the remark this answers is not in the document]
leeky: I vote for keeping the original `status_step` name here. It's not really mentioned outside of the form handling, unlike say submissions where the rename was more necessary.

@jcoglan The form state is persisted because we have *a lot* of conditional validations which rely on it.

app/services/ingest_iati_activities.rb
leeky: I think any activities where we know the IATI status, but not `programme_status` need to be flagged in a way that we know to go back and fix them later. So I think a data migration that sets this attribute to "Replace me" in cases where it is empty makes sense.

We probably should set the `programme_status` to the proper value in the cases where there is *only* one IATI status that matches it.

Any thoughts, @jcoglan ?

app/models/activity.rb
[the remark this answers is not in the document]
MariluzHerrera: I'll keep the original step name then. Thanks both!

db/migrate/20200803104010_add_programme_status_to_activities.rb
leeky: @MariluzHerrera I think we should leave it with the default set to `NULL` for now, but make a new card to remind us to fix this later (probably once `programme_status` has completely replaced `status`).

Any data manually-inputted through the front end will have this filled in correctly, so I don't think it's an issue.  Probably just need to check that an activity where `programme_status` is set to `Replace me` pre-populates the form with an empty selection when users attempt to edit it (so when they save they're force to choose something!)

config/locales/codelists/2_03/iati.en.yml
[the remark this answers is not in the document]
MariluzHerrera: I'm finally leaving this where it is. This is to avoid losing the logic that allows a user to see the locale value for `programme_status` when they see the activity details

db/migrate/20200803104010_add_programme_status_to_activities.rb
MariluzHerrera: I'll create a new card for this

app/services/ingest_iati_activities.rb
MariluzHerrera: I will create a new card for this so we can agree the best approach for this issue.

app/models/activity.rb
```diff
@@ -28,7 +28,7 @@ class Activity < ApplicationRecord
   validates :title, :description, presence: true, on: :purpose_step
   validates :sector_category, presence: true, on: :sector_category_step
   validates :sector, presence: true, on: :sector_step
-  validates :status, presence: true, on: :status_step
+  validates :programme_status, presence: true, on: :programme_status_step
```
leeky, line 31: Didn't we decide to leave this as `status_step`..?

MariluzHerrera: If I replace line 31 for `validates :programme_status, presence: true, on: :status_step` my validations for the presence of programme status won't work 🤔 When the user creates an activity and does not select any programme status, they don't get the error message that programme status can't be blank. The form just moves onto the next step without complaining...

PR #574, opened 2020-08-14
(890) Reports store their financial quarter and financial year

mec:
## Changes in this PR
I've found the resons for storing the financial quarter and year on a report really hard to articulate, and discussion have taken place to drive this work.

To try and summerise:

- In order to label the actuals column with the FQ for the report
- In order to show the relevent forecasts for the report (these will not have been recorded (associated) in the report as they are from the past)
- in order to show the relevent future forecasts we need to know from where to begin
- in order to display data in ways users find meaningful, users ❤️ financial quarters

We are in a odd place at the moment where we are backfilling data into RODA with many quarters worth of data  – I feel like we know this case is an exception and the fact that the 'current report' has no FQ is alright and something we can deal with.  I do believe we could do the work to slice the legacy data into quarters but am unconvinced the effort is worth it – happy to chat about this if folks feel different.

I wanted the Report to set its own FQ values are early as possible so we do not have to think about, I decided to do this in the intializer after considering other options. As the values are also set to read only, once the value is set and saved Rails will not expect it to change.

This work contains a refactor of where we store the column headers for the csv, as we now generate some of these dynamically having a constant in the Activity model felt awkward.

I wondered about creating a seperate service object for the headers but in the end, decided to keep a single object that can take an activity and give you a csv row or a report and give you the correct set of headers. I was a bit iffy about whether to guard against passing in nil to this service object - thoughts appreciated on that - it feels a little clunky as is.

I also wondered about how to store the values and settled on intgers as they feel low cost and can easily be compared to the relevent part of a Date object as needed.

I suspect we will be doing a lot of calcualting the quarter for a given date and may move that code out to it's own class or helper, but this didn't feel like the time to do that (I am a fan of letting need drive work rather than trying to pre think too much!)

## Next steps

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -1,6 +1,8 @@
 class Report < ApplicationRecord
   include PublicActivity::Common
 
+  attr_readonly :financial_quarter, :financial_year
+
   validates_presence_of :description
   validates_presence_of :state
 
@@ -14,10 +16,35 @@ class Report < ApplicationRecord
 
   enum state: [:inactive, :active]
 
+  def initialize(attributes = nil)
+    super(attributes)
+    self.financial_quarter = current_financial_quarter
+    self.financial_year = current_financial_year
+  end
+
   def activity_must_be_a_fund
     return unless fund.present?
     unless fund.fund?
       errors.add(:fund, I18n.t("activerecord.errors.models.report.attributes.fund.level"))
     end
+  end
+
+  private def current_financial_quarter
+    case Date.today.month
+    when 4, 5, 6
+      1
+    when 7, 8, 9
+      2
+    when 10, 11, 12
+      3
+    when 1, 2, 3
+      4
+    end
+  end
+
+  private def current_financial_year
+    year = Date.today.year
+    return year - 1 if current_financial_quarter == 4
+    year
   end
 end
--- a/spec/models/report_spec.rb
+++ b/spec/models/report_spec.rb
@@ -4,11 +4,21 @@ RSpec.describe Report, type: :model do
   describe "validations" do
     it { should validate_presence_of(:description) }
     it { should validate_presence_of(:state) }
+    it { should have_readonly_attribute(:financial_quarter) }
+    it { should have_readonly_attribute(:financial_year) }
   end
 
   describe "associations" do
     it { should belong_to(:fund).class_name("Activity") }
     it { should belong_to(:organisation) }
+  end
+
+  it "sets the financial_quarter and financial_year when created" do
+    travel_to(Date.parse("01-04-2020")) do
+      report = Report.new
+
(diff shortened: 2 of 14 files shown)


mec:
[Reports store their financial quarter and financial year](https://trello.com/c/is43tpwI/890-reports-store-their-financial-quarter-and-financial-year)

spec/models/report_spec.rb
```diff
@@ -31,4 +41,78 @@
...
+        travel_to(Date.parse("1 November 2020")) do
+          report = Report.new
+          expect(report.financial_year).to eql 2020
+        end
+      end
+    end
+
+    context "when in the fouth quarter" do
+      it "sets the financial year to 2020" do
+        travel_to(Date.parse("1 Feburary 2021")) do
```
leeky, line 111: Typo: `February`

app/models/report.rb
```diff
@@ -1,6 +1,8 @@
 class Report < ApplicationRecord
   include PublicActivity::Common
 
+  attr_readonly :financial_quarter, :financial_year
```
leeky, line 4: :+1:

app/models/report.rb
```diff
@@ -14,10 +16,35 @@ class Report < ApplicationRecord
...
+      2
+    when 10, 11, 12
+      3
+    when 1, 2, 3
+      4
+    end
+  end
+
+  private def current_financial_year
+    today = Date.today
```
leeky, line 46: As you never use other parts of the date, how about just setting `year = Date.today.year` here?

```ruby
year = Date.today.year
year -= 1 if current_financial_quarter == 4
year
```

spec/models/report_spec.rb
```diff
@@ -31,4 +41,78 @@
     report = build(:report, deadline: Date.yesterday)
     expect(report).not_to be_valid
   end
+
+  context "when in the first quarter of the financial year - April to March" do
```
leeky, line 45: I think these need a `describe '#financial_quarter' do` block around them.

spec/models/report_spec.rb
```diff
@@ -31,4 +41,78 @@
...
+      ["1 January 2020", "31 March 2020"].each do |date|
+        travel_to(Date.parse(date)) do
+          report = Report.new
+          expect(report.financial_quarter).to eql 4
+        end
+      end
+    end
+  end
+
+  context "when in the 2020 - 2021 financial year" do
```
leeky, line 89: ... and same here `describe '#financial_year' do`.

Not sure this `context` is needed at all, because it's describing what the test is *doing* rather than what it is *testing*.

mec: I take your point - I do want to make it clear to devlopers who may not know about financial years spanning two calendars years - I've removed the context and add the financial year to the inner context - let me know what you think?

PR #575, opened 2020-08-14
(refactor) Change the way Report state is stored

mec:
Defining the enum as an array has two drawbacks:

1. the values in the DB have no context for developers as they are
   simply numbers
2. less flexibility to allow change as the order of the array matters

By switching to a hash we get around both these drawbacks.

This work also adds an index to state as it is likely the application
will need to sort and display Reports by their state.

We currently have reports in production but they are all 'inactive' so
this work contains a simple data migration to map the stored 0s to
"inactive". Doing this work any later would be far more difficult.

### DIFF
--- a/db/data/20200814135953_report_states_are_strings.rb
+++ b/db/data/20200814135953_report_states_are_strings.rb
@@ -1,0 +1,9 @@
+class ReportStatesAreStrings < ActiveRecord::Migration[6.0]
+  def up
+    Report.update_all state: "inactive"
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/db/migrate/20200814134116_report_state_stored_as_string.rb
+++ b/db/migrate/20200814134116_report_state_stored_as_string.rb
@@ -1,0 +1,6 @@
+class ReportStateStoredAsString < ActiveRecord::Migration[6.0]
+  def change
+    change_column :reports, :state, :string, default: "inactive"
+    add_index :reports, :state
+  end
+end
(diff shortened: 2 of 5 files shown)


db/data/20200814135953_report_states_are_strings.rb
```diff
@@ -0,0 +1,9 @@
+class ReportStatesAreStrings < ActiveRecord::Migration[6.0]
+  def up
+    Report.update_all state: "inactive"
```
leeky, line 3: I'm always a bit wary of having a one-way destructive migrations, but probably ok given how this field isn't really populated right now.

mec: Yeah - this one worries me, I think this is the only time we can do it and my understanding of the data migrations is that this will only evert run once.

leeky: I only just spotted that this was a data migration rather than a regular one! :-(

db/migrate/20200814134116_report_state_stored_as_string.rb
```diff
@@ -0,0 +1,6 @@
+class ReportStateStoredAsString < ActiveRecord::Migration[6.0]
+  def change
+    change_column :reports, :state, :string, default: "inactive"
+    add_index :reports, :state
```
jcoglan, line 4: Will reports be queries purely on `state`? Or are they view in the context of a fund so that a compound index on `[:fund_id, :state]` would make more sense?

mec: I can't think of any places right now where we would want reports by fund - but I can easily imagine the time when we do - can we add the compound index at that point or is it something we need to decide upfront?

jcoglan: If we don't need that functionality right now that's fine. I was just checking whether for BEIS users we list reports out by fund, or just list all submitted reports as a single list.

jcoglan:
Are we confident we know what order the schema and data migration here will run in?

PR #577, opened 2020-08-17
(798) A delivery partner can approve a submission and send it to beis

mec:
## Changes in this PR
Introduces the submit report journey allowing a Delivery partner to submit their report to BEIS.

I went with my jellies™ for this and chose to create a `reports_submissions_controller` to handle this action allowing me to PATCH the report model with the new state and a nested path to redirect to once complete.

An error would redirect to the reports index although we shouldn't see this in use.

The complete page is a bit of a dead end, but I am hopeful we can fix that with content and navigation.

The work here is minimal by design, we want the simplest version of the user completing this task and updating the state of the `Report`.

Content is minimal as there is seperate work for that here: https://trello.com/c/RMw6eeBU

One thing I would like to improve is the methods for the Report index, we need to only include `Oragnisaiton` if the user is BEIS and I could have done somthing like passing `with_organiastions`) but I like the explicitness of wrapping them in methods. I think I could have used scopes on `Report` as well which might be better - thoughts appreciated.

## Screenshots of UI changes
image
image
image
image
image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -247,6 +247,7 @@
   Mapping `programme_status` to `IATI_status` included. Schema migration to replace
   `form_state` at `status` step for `programme_status` step
 - Reports can be submitted
+- Submitted reports are shown to users
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-13...HEAD
 [release-13]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-12...release-13
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -241,10 +241,12 @@
 - Ingest creates new activities at a level below its parent
 - Reports store the relevant financial quarter
 - Associate Planned Disbursements to Reports
+<<<<<<< HEAD
 - Update extending organisation question content
 - `programme status` field added to activity form in exchange for old IATI status.
   Mapping `programme_status` to `IATI_status` included. Schema migration to replace
   `form_state` at `status` step for `programme_status` step
+- Reports can be submitted
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-13...HEAD
 [release-13]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-12...release-13
--- a/app/controllers/staff/reports_controller.rb
+++ b/app/controllers/staff/reports_controller.rb
@@ -12,15 +12,17 @@ class Staff::ReportsController < Staff::BaseController
   end
 
   def show
-    @report = Report.find(id)
-    authorize @report
+    report = Report.find(id)
+    authorize report
+
+    @report_presenter = ReportPresenter.new(report)
 
     respond_to do |format|
       format.html
       format.csv do
-        fund = @report.fund
-        @projects = Activity.project.where(organisation: @report.organisation).select { |activity| activity.associated_fund == fund }
-        @third_party_projects = Activity.third_party_project.where(organisation: @report.organisation).select { |activity| activity.associated_fund == fund }
+        fund = @report_presenter.fund
+        @projects = Activity.project.where(organisation: @report_presenter.organisation).select { |activity| activity.associated_fund == fund }
+        @third_party_projects = Activity.third_party_project.where(organisation: @report_presenter.organisation).select { |activity| activity.associated_fund == fund }
         send_csv
       end
     end
@@ -85,8 +87,8 @@ class Staff::ReportsController < Staff::BaseController
 
   def send_csv
     response.headers["Content-Type"] = "text/csv"
-    response.headers["Content-Disposition"] = "attachment; filename=#{ERB::Util.url_encode(@report.description)}.csv"
-    response.stream.write ExportActivityToCsv.new.headers(report: @report)
+    response.headers["Content-Disposition"] = "attachment; filename=#{ERB::Util.url_encode(@report_presenter.description)}.csv"
+    response.stream.write ExportActivityToCsv.new.headers(report: @report_presenter)
     @projects.each do |project|
       response.stream.write ExportActivityToCsv.new(activity: project).call
     end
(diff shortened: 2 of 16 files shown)


leeky:
I think something like `ReportSubmissionsController` would be a better fit for Rails conventions.

app/controllers/staff/reports_controller.rb
```diff
@@ -12,15 +12,17 @@ def index
...
+
+    @report_presenter = ReportPresenter.new(report)
 
     respond_to do |format|
       format.html
       format.csv do
-        fund = @report.fund
-        @projects = Activity.project.where(organisation: @report.organisation).select { |activity| activity.associated_fund == fund }
-        @third_party_projects = Activity.third_party_project.where(organisation: @report.organisation).select { |activity| activity.associated_fund == fund }
+        fund = @report_presenter.fund
```
leeky, line 24: Since these relate to attributes on the model directly, they should probably still reference `report` rather than the presenter.

app/controllers/staff/reports_controller.rb
```diff
@@ -12,15 +12,17 @@ def index
   end
 
   def show
-    @report = Report.find(id)
-    authorize @report
+    report = Report.find(id)
+    authorize report
+
+    @report_presenter = ReportPresenter.new(report)
```
leeky, line 19: The HTML view references the `@report` instance variable rather than `@report_presenter` so this probably should be updated.

Since `@report` has been renamed, it is no longer be available in the view, so I'm a bit worried no tests broke..!

CHANGELOG.md
```diff
@@ -241,10 +241,12 @@
 - Ingest creates new activities at a level below its parent
 - Reports store the relevant financial quarter
 - Associate Planned Disbursements to Reports
+<<<<<<< HEAD
```
leeky, line 244: Left over from a rebase..?

app/controllers/staff/reports_submissions_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class Staff::ReportsSubmissionsController < Staff::BaseController
```
leeky, line 3: I'd usually write this with `new` and `create` actions because you're "creating a new report submission", even if that doesn't exactly map to the way the underlying model works.

Of course, there's a question around how we implement additional changes in state - having a separate controller for each possible change could get complicated 🤔 

Happy to chat about this!

app/controllers/staff/reports_submissions_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def update
+    @report = Report.find(params[:report_id])
+    authorize @report
+
+    @report.state = :submitted
+    if @report.valid?
+      @report.save!
+      @report.create_activity key: "report.submitted", owner: current_user
+      redirect_to complete_report_submit_path(@report)
```
leeky, line 21: How about rendering the "completed" page here rather than redirecting?

app/controllers/staff/reports_submissions_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    else
```
leeky, line 22: How would a user end up here, given all the policy checking etc beforehand?

app/controllers/staff/reports_submissions_controller.rb
mec: This is definately something I would appreciate talking through as we have two other state changes to add next!

app/controllers/staff/reports_submissions_controller.rb
mec: Even better! Thank you! :)

app/policies/report_policy.rb
```diff
@@ -14,15 +14,24 @@ def create?
     beis_user?
   end
 
-  def update?
+  def edit?
     beis_user?
   end
 
+  def update?
```
jcoglan, line 21: Does this need some additional checks around it? A BEIS user can update a submitted report, but a DP user cannot. Similarly, a DP user can update an active report but a BEIS user cannot. It would be good to include some notion of who "owns" the report in each state so that only the current owner is allowed to edit it.

CHANGELOG.md
jcoglan, line 244: Looks like this crept in during a merge conflict.

app/controllers/staff/reports_submissions_controller.rb
jcoglan: I think it makes sense to have a distinct controller for changing the report's state, as it's a distinct type of change from updating all the other aspects of a report. However we can probably do with a single controller for all state changes, rather than one controller per state.

app/controllers/staff/reports_submissions_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    authorize report
+    @report_presenter = ReportPresenter.new(report)
+    render "staff/reports/submit"
+  end
+
+  def update
+    @report = Report.find(params[:report_id])
+    authorize @report
+
+    @report.state = :submitted
```
jcoglan, line 17: This isn't necessary right now, but as we expand this to handle more states, we should check that the requested state _transition_ is valid -- you can't mark a report `submitted` if it's currently `inactive`, for example. There are gems that can help with this, for example [aasm](https://github.com/aasm/aasm).

app/controllers/staff/reports_submissions_controller.rb
jcoglan: It does seem unlikely this branch could be triggered as the only state change that has been made is something we know to be valid. If we accept `state` as a parameter though, this will become necessary us the user may submit an unrecognised state, or one that is not a valid transition from the report's current state.

app/controllers/staff/reports_submissions_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+      flash[:notice] = I18n.t("action.report.submit.failure")
+      redirect_to reports_path
+    end
+  end
+
+  def complete
+    report = Report.find(params[:report_id])
+    authorize report
+    @report_presenter = ReportPresenter.new(report)
+    render "staff/reports/complete"
```
jcoglan, line 32: Is there an opportunity here to move this template so that it's associated with this controller? Or does it need to be shared between here and `Staff::ReportsController`?

mec:
I am going to close this PR and re-work it.

PR #579, opened 2020-08-17
(875) The CSV download contains aggregated forecasts

lozette:
## Changes in this PR

Trello: https://trello.com/c/n3371hcO/875-aggregated-forecasts-planned-disbursements-are-included-in-the-csv-download

The Report CSV will always contain all Activities associated with a Fund/Organisation pair, whether they had transactions or planned disbursements in the report period or not. But the transaction/planned disburement totals must be scoped to a single Report and its date range, not be the totals for that Activity for all time.
    
1) Scope the `transactions_total` to a Report & that Report's date range
    
The new method is called `actual_total_for_report_financial_quarter` for clarity

2) Calculate the total of all Planned Disbursements in a Report's date range
    
Planned Disbursements scoped to a single Report and that Report's date range are summed and returned in `forecasted_total_for_report_financial_quarter`
    
3) Remove `Activity#transactions_total` and its test as it is no longer used

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -153,4 +153,8 @@ class Activity < ApplicationRecord
   def transactions_total
     transactions.sum(:value)
   end
+
+  def actual_total_for_report_financial_quarter(report:)
+    transactions.where(report: report, date: report.created_at.all_quarter).sum(:value)
+  end
 end
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -157,4 +157,8 @@ class Activity < ApplicationRecord
   def actual_total_for_report_financial_quarter(report:)
     transactions.where(report: report, date: report.created_at.all_quarter).sum(:value)
   end
+
+  def forecasted_total_for_report_financial_quarter(report:)
+    planned_disbursements.where(period_start_date: report.created_at.all_quarter).sum(:value)
+  end
 end
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -150,10 +150,6 @@ class Activity < ApplicationRecord
     }.push(identifier).join("-")
   end
 
-  def transactions_total
-    transactions.sum(:value)
-  end
-
   def actual_total_for_report_financial_quarter(report:)
     transactions.where(report: report, date: report.created_at.all_quarter).sum(:value)
   end
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -636,10 +636,42 @@ RSpec.describe Activity, type: :model do
   describe "#transactions_total" do
     it "returns the total of all the activity's transactions" do
       project = create(:project_activity)
-      _transaction_1 = create(:transaction, parent_activity: project, value: 100.20)
-      _transaction_2 = create(:transaction, parent_activity: project, value: 210)
+      create(:transaction, parent_activity: project, value: 100.20)
+      create(:transaction, parent_activity: project, value: 210)
 
       expect(project.transactions_total).to eq(310.20)
+    end
+  end
+
+  describe "#actual_total_for_report_financial_quarter" do
+    it "returns the total of all the activity's transactions scoped to a report" do
+      project = create(:project_activity, :with_report)
+      report = Report.find_by(fund: project.associated_fund, organisation: project.organisation)
+      create(:transaction, parent_activity: project, value: 100.20, report: report, date: Date.today)
+      create(:transaction, parent_activity: project, value: 50.00, report: report, date: Date.today)
+      create(:transaction, parent_activity: project, value: 210, report: report, date: Date.today - 4.months)
+
+      expect(project.actual_total_for_report_financial_quarter(report: report)).to eq(150.20)
+    end
+
+    it "does not include the totals for any transactions outside the report's date range" do
+      project = create(:project_activity, :with_report)
+      report = Report.find_by(fund: project.associated_fund, organisation: project.organisation)
+      create(:transaction, parent_activity: project, value: 100.20, report: report, date: Date.today - 6.months)
+      create(:transaction, parent_activity: project, value: 210, report: report, date: Date.today - 4.months)
+
+      expect(project.actual_total_for_report_financial_quarter(report: report)).to eq(0)
+    end
+  end
+
+  describe "#actual_total_for_report_financial_quarter" do
+    it "returns the total of all the activity's transactions scoped to a report" do
+      project = create(:project_activity, :with_report)
+      report = Report.find_by(fund: project.associated_fund, organisation: project.organisation)
+      create(:transaction, parent_activity: project, value: 100.20, report: report, date: Date.today)
+      create(:transaction, parent_activity: project, value: 210)
+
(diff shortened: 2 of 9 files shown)


lozette:
Wondering if we need `planned_disbursements_total` and `transactions_total` at all since we don't actually use them? Is it worth removing them to get rid of a few tests?

app/models/activity.rb
```diff
@@ -157,4 +157,12 @@ def transactions_total
   def planned_disbursements_total
     planned_disbursements.sum(:value)
   end
+
+  def actual_total_for_report(report:)
+    transactions.where(report: report, date: report.created_at.all_quarter).sum(:value)
```
lozette, line 162: Is `created_at` the right date to use here? `financial_year` and `financial_quarter` are both created from the date the Report is created, so I assume it's alright to use the `created_at` date to also determine the quarter range?

mec: We've just chatted about this but I think you are right, as the `financial_year` and `financial_quarter` are both derived from the date the Report is created, calculating the date range can be done like this.

mec:
> Wondering if we need `planned_disbursements_total` and `transactions_total` at all since we don't actually use them? Is it worth removing them to get rid of a few tests?

My preference would always be to  loose unused code and tests when possible!

app/models/activity.rb
jcoglan: Why is the date filter necessary? I thought we were associating transactions to reports to make that relationship explicit, rather than inferring it from dates. It feels like a brittle assumption to say that a report relates to the quarter during which it was created in the database.

spec/models/activity_spec.rb
```diff
@@ -652,4 +652,26 @@
       expect(project.planned_disbursements_total).to eq(1570.00)
     end
   end
+
+  describe "#actual_total_for_report" do
+    it "returns the total of all the activity's transactions scoped to a report" do
+      project = create(:project_activity, :with_report)
+      report = Report.find_by(fund: project.associated_fund, organisation: project.organisation)
+      _transaction_1 = create(:transaction, parent_activity: project, value: 100.20, report: report, date: Date.today)
```
jcoglan, line 660: Do these values need to be assigned to variables? If the variable is unused we could remove the assignment entirely.

lozette: They don't! Old habit 😁

app/models/activity.rb
mec: Feels like I am still failing to communicate all this so apologies! :( 

The stance we are taking is that there is only ever one report for a given org and fund. The report is seen as 'the data for a quarter' but I believe it is important to allow transactions that are not in the quarter to be reported at the same time, an example of this would be a correction to a previous reported transaction.

Whilst we know little about how this 'correction' will take place, I think its fair to assume it will be reported via the same mechanism. For this reason, in this context, for transactions, when reviewing, the users will want to see the total transaction value per activity for the quarter in question and this should not include any values that fall outside of the quater date range as this would make the figures inaccurate.

We currently have no validation to ensure the transactions added during a report are only from the quarter in question so this scoping guards against that issue. Once we know more about making corrections we may be able to remove the scope and be confident that transactions associated to a report are only for the date range users care about during the review.

Looking forward we may want to handle corrections in some other way - who knows but for now I feel we need this guard in place.

I agree that the relationship between the creation date and the quarter feeling brittle, but I am confident that as the application is in full control of the creation of the reports (users cannot create them) we can guard against this also and we are mindful of this need going into the work to create the next report.

Sorry for the huge comment - happy (and would rather) talk it though if things are not clear.

app/models/activity.rb
```diff
@@ -157,4 +157,12 @@ def transactions_total
...
+  end
+
+  def forecasted_total_for_report(report:)
+    planned_disbursements.where(report: report, period_start_date: report.created_at.all_quarter).sum(:value)
```
mec, line 166: To fully blow your mind (and possibly make you want to cry!):

Forecasted values for this quarter should not have been submitted in this report (they are from the past, we should not expect what was forecasted to change after we spent the money) so for this scope we only want the date range, as restricting to the report will in theory remove all the values we are interested in.

I would also like to name this one 'forecasted_total_for_report_financial_quarter' if you agree?

app/models/activity.rb
```diff
@@ -157,4 +157,12 @@ def transactions_total
   def planned_disbursements_total
     planned_disbursements.sum(:value)
   end
+
+  def actual_total_for_report(report:)
```
mec, line 161: I think I would like to name this 'actual_total_for_report_financial_quarter' - I feel like we should express what the date range is doing in it as best we can?

mec (changes requested):
I think this is close, thank you for sticking with it!

The scope on the planned_disbursements is not quite right (my fault from showing you before) and I would appreciate a single commit that fixes the transaction scope so we can explain what is going on (if you want a hand to word that I can try and help!) to our future selves.

I am also keen to get the names of the totalling methods as good as we can at expressing the behaviour.

lozette:
@mec Changes done, also updated the PR message and rebased all over

spec/models/activity_spec.rb
```diff
@@ -653,4 +653,15 @@
       expect(project.actual_total_for_report_financial_quarter(report: report)).to eq(100.20)
     end
   end
+
+  describe "#forecasted_total_for_report_financial_quarter" do
+    it "returns the total of all the activity's planned disbursements scoped to a report's financial quarter" do
```
mec, line 648: I really wish there was some way to express/document the behaviour of these totalling methods, I just can't come up with it! I got to this yesterday, but I think it makes it look even more complicated:

````ruby
    it "totals the activity forecasts for the financial quarter of the report" do
      quarter_one_date = Date.parse("1 April 2019")
      quarter_two_date = Date.parse("1 July 2019")

      project = create(:project_activity)
      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: quarter_one_date)
      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: quarter_one_date)

      create(:planned_disbursement, parent_activity: project, value: 2000.00, period_start_date: quarter_two_date)
      create(:planned_disbursement, parent_activity: project, value: 2000.00, period_start_date: quarter_two_date)

      report = create(:report, fund: project.associated_fund)
      expect(project.forecasted_total_for_report_financial_quarter(report: report)).to eq(0)

      travel_to(quarter_one_date) do
        report = create(:report, fund: project.associated_fund)
        expect(project.forecasted_total_for_report_financial_quarter(report: report)).to eq(2000.00)
      end

      travel_to(quarter_two_date) do
        report = create(:report, fund: project.associated_fund)
        expect(project.forecasted_total_for_report_financial_quarter(report: report)).to eq(4000.00)
      end
    end
````

I feel like these are important so I expect more tests, but maybe I am given them too much credit?

Either way, we need to move this work forward and I hope once folks understand what we are aiming for your tests make sense - let me know your thoughts.

lozette: I've included your tests almost verbatim as they make perfect sense to me, much more thorough than the single one I had.

PR #580, opened 2020-08-18
(895) Programme status form step removed for Funds

MariluzHerrera:
Trello: https://trello.com/c/vHswa3bj

## Changes in this PR

Now that `programme status` field has been added to the create activity form, the next step was to remove this form step when users create a level A activity (Fund). Users are only required to add a programme status for levels B, C and D. However, despite not asking the user for a `programme status` on Funds we are still setting a default value of `Spend in progress` in the background, as agreed with the client. This will help us avoid potential errors with validations. As a result, the old IATI `status` also gets set in the background, but the user will not see any of those as they are not relevant for them when handling Funds.

To improve the user experience, only for Funds we have removed the `programme status` step form, the field in the view of activity details and the possibility to edit this field. 

### After

Screenshot 2020-08-18 at 16 56 35

## Next steps

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -34,6 +34,8 @@ class Staff::ActivityFormsController < Staff::BaseController
       skip_step if @activity.fund?
     when :blank
       skip_step
+    when :programme_status
+      skip_step if @activity.fund?
     when :region
       skip_step if @activity.recipient_country?
     when :country
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -3,6 +3,8 @@ class Staff::ActivityFormsController < Staff::BaseController
   include DateHelper
   include ActivityHelper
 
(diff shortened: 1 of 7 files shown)


app/controllers/staff/activity_forms_controller.rb
```diff
@@ -3,6 +3,8 @@ class Staff::ActivityFormsController < Staff::BaseController
   include DateHelper
   include ActivityHelper
 
+  DEFAULT_PROGRAMME_STATUS_FOR_FUNDS = "07"
```
jcoglan, line 6: Glad you've given this a helpful name. Is there a table anywhere that would let us refer to `07` status by its name so we know what it means?

MariluzHerrera: Yes, the values for each code can be found on this file: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/config/locales/codelists/2_03/iati.en.yml

lozette (approved):
Happy with this, great commit messages too 👍

PR #585, opened 2020-08-20
Update the Report policy

mec:
## Changes in this PR

We already have a good understanding of the what state transitions are
valid and which user can make them.

This work updates the report policy and related spec to make it possible
to enforce this understanding and support future work to allow users to
move a report though its state inside the application.

This Miro board speaks to how the state and permission works:

https://miro.com/app/board/o9J_knq_KVg=/

## Next steps

### DIFF
--- a/app/policies/report_policy.rb
+++ b/app/policies/report_policy.rb
@@ -7,24 +7,65 @@ class ReportPolicy < ApplicationPolicy
 
   def show?
     return true if beis_user?
-    record.organisation == user.organisation
+    return true if record.organisation == user.organisation && record.state != "inactive"
+    false
   end
 
   def create?
-    beis_user?
+    false
+  end
+
+  def edit?
+    update?
   end
 
   def update?
     beis_user?
   end
 
-  def download?
-    return true if beis_user?
-    record.organisation == user.organisation
-  end
-
   def destroy?
     false
+  end
+
+  def change_state?
+    case record.state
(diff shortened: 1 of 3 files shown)


app/policies/report_policy.rb
```diff
@@ -7,20 +7,61 @@ def index?
...
+
+  def edit?
+    update?
   end
 
   def update?
     beis_user?
   end
 
+  def change_state?
```
leeky, line 26: I really like this! 🎉

app/policies/report_policy.rb
```diff
@@ -7,20 +7,61 @@ def index?
...
+  end
+
   def download?
-    return true if beis_user?
-    record.organisation == user.organisation
+    show?
+  end
+
+  def activate?
+    return change_state? if record.state == "inactive"
```
leeky, line 48: The explicit `return` makes sense here 👍

app/policies/report_policy.rb
```diff
@@ -7,20 +7,61 @@ def index?
...
+
+  def request_changes?
+    return change_state? if record.state == "in_review"
+  end
+
+  def approve?
+    return change_state? if record.state == "in_review"
   end
 
   def destroy?
```
leeky, line 67: How about moving this RESTful check above the state-related ones?

leeky (approved):
LGTM!  :shipit:

PR #586, opened 2020-08-20
(925) LegacyActivity ignores incomplete activities

leeky:
When the RAEng Newton fund legacy IATI was published, it contained activities which were subsequently cancelled and weren't reference elsewhere.

In this case, `LegacyActivity#find_parent` would try and look up the identifier in the mapping file which didn't exist, causing the lookup to return `nil`. This wouldn't have been a problem, were it not for half-completed activity multi-step forms (missing their identifier) existing in the database. One of these was returned in error.

This lead to ingest raising unexpected error messages, rather than the expected `ParentNotFoundForActivity` exception.

The fix was straightforward enough - use `fetch(value)` instead of `fetch[value]` to retrieve the parent from the mapping file. Instead of returning `nil` when one doesn't exist, it'll raise a `KeyError` which is then caught correctly.


mec (approved):
Beautiful commit message thank you! 🚢

PR #587, opened 2020-08-20
(926) Ingest sets `programme_status` if inferable

leeky:
Set `programme_status` correctly when there is only one possible IATI activity-status that matches it.


MariluzHerrera (approved):
This looks really nice. I like that you've combined both status on the same method `add_statuses` since they are related. The specs are easy to follow, with the sample XMLs.
Thanks for taking the time to do this Leeky! :-)

PR #596, opened 2020-08-24
(Fix) `p` statement removed from spec file

MariluzHerrera:
There was a `p result.call_present` on one of the test files that was accidentally
left on the file. This was printing `Yes` when running the tests. This commit
removes the printing line.


lozette (approved):
We never saw anything!

jcoglan:
Thanks for finding and fixing this :)

PR #598, opened 2020-08-24
(903) The next 4 quarters' forecast totals are included in a report's CSV

lozette:
## Changes in this PR

Trello: https://trello.com/c/Ee0f5VTG/903-a-number-of-future-aggregated-forecast-amounts-are-included-in-the-csv-file

A report has a financial quarter & year, e.g. Q1 2020.

Find the next 4 quarters after this report's quarter, e.g. Q2 2020, Q3 2020, Q4 2020 and Q1 2021

Calculate the forecast totals (planned disbursement totals) for those 4 quarters per Activity. Output these values to the report CSV as future forecast totals. 

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -160,7 +160,7 @@ class Activity < ApplicationRecord
   end
 
   def forecasted_total_for_report_financial_quarter(report:)
-    @forecasted_total_for_report_financial_quarter ||= planned_disbursements.where(period_start_date: report.created_at.all_quarter).sum(:value)
+    @forecasted_total_for_report_financial_quarter ||= forecasted_total_for_date_range(range: report.created_at.all_quarter)
   end
 
   def variance_for_report_financial_quarter(report:)
@@ -169,5 +169,9 @@ class Activity < ApplicationRecord
 
   def requires_call_dates?
     !ingested? && (project? || third_party_project?)
+  end
+
+  def forecasted_total_for_date_range(range:)
+    planned_disbursements.where(period_start_date: range).sum(:value)
   end
 end
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -37,6 +37,19 @@ class Report < ApplicationRecord
     end
   end
 
+  def next_four_financial_quarters
+    case current_financial_quarter
+    when 1
+      ["Q2 #{current_financial_year}", "Q3 #{current_financial_year}", "Q4 #{current_financial_year}", "Q1 #{current_financial_year + 1}"]
+    when 2
+      ["Q3 #{current_financial_year}", "Q4 #{current_financial_year}", "Q1 #{current_financial_year + 1}", "Q2 #{current_financial_year + 1}"]
+    when 3
+      ["Q4 #{current_financial_year}", "Q1 #{current_financial_year + 1}", "Q2 #{current_financial_year + 1}", "Q3 #{current_financial_year + 1}"]
+    when 4
+      ["Q1 #{current_financial_year + 1}", "Q2 #{current_financial_year + 1}", "Q3 #{current_financial_year + 1}", "Q4 #{current_financial_year + 1}"]
+    end
+  end
+
   private def current_financial_quarter
     case Date.today.month
     when 4, 5, 6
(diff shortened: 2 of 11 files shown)


app/helpers/report_helper.rb
```diff
@@ -0,0 +1,20 @@
+module ReportHelper
+  def quarters_to_date_ranges(quarters)
+    quarters.map! do |quarter_string|
+      quarter = quarter_string.match(/(\d) (\d{4})/)
```
leeky, line 4: Interested to know why this is in a helper rather than in `ReportPresenter`..? 🤔

app/models/report.rb
```diff
@@ -37,6 +37,19 @@ def activity_must_be_a_fund
     end
   end
 
+  def next_four_financial_quarters
```
leeky, line 40: Yeah, we should eventually move this into a `FinancialQuarter` class (value object?) perhaps, but this is good for now :)

app/models/activity.rb
```diff
@@ -160,7 +160,7 @@ def actual_total_for_report_financial_quarter(report:)
   end
 
   def forecasted_total_for_report_financial_quarter(report:)
-    @forecasted_total_for_report_financial_quarter ||= planned_disbursements.where(period_start_date: report.created_at.all_quarter).sum(:value)
+    @forecasted_total_for_report_financial_quarter ||= forecasted_total_for_date_range(range: report.created_at.all_quarter)
```
leeky, line 163: I might be confused by this, but given reports `has_many :planned_disbursements`, couldn't you do:  `report.planned_disbursements.sum(:value)` ?

app/services/export_activity_to_csv.rb
```diff
@@ -56,4 +56,17 @@ def headers
...
+  def next_four_quarter_forecasts
+    quarter_date_ranges = quarters_to_date_ranges(next_four_quarters)
+    quarter_date_ranges.map { |range| activity_presenter.forecasted_total_for_date_range(range: range) }
+  end
+
+  private def activity_presenter
+    @activity_presenter ||= ActivityPresenter.new(activity)
+  end
+
+  private def next_four_quarters
```
leeky, line 69: Could this just be included in `next_four_quarter_forecasts` ..?

spec/services/export_activity_to_csv_spec.rb
```diff
@@ -9,6 +9,7 @@
     it "creates a CSV line representation of the Activity" do
       activity_presenter = ActivityPresenter.new(project)
       result = ExportActivityToCsv.new(activity: project, report: report).call
+      next_four_quarter_totals = ExportActivityToCsv.new(activity: project, report: report).next_four_quarter_forecasts
```
leeky, line 12: Is it worth splitting `result` into:

```
export_activity_to_csv = ExportActivityToCsv.new(...)
result = export_activity_to_csv.call
next_four_quarter_totals = result.next_four_quarter_forecasts
```

...so that `result` and `next_four_quarter_totals` comes from the same instance?

leeky (approved):
I have made some minor comments, but LGTM!

app/models/activity.rb
lozette: No you're not confused, I think we originally hung the query off the activity, because originally I was looking for PDs that were for a date range *and* a report, but now we're just looking for ones in a date range. I'll see if I can make it better/more obvious

app/services/export_activity_to_csv.rb
lozette: One is the headers and one is the results but yes maybe we can have a hash rather than an array

app/models/activity.rb
lozette: Sorry ignore my other comment. They need to be constrained to a date range because of https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/579#discussion_r472843701

app/services/export_activity_to_csv.rb
lozette: On second thoughts probably not really - the headers belong to the *report* but the results belong to the *activity* so I don't know if returning both of them in the same object is going to make sense?

PR #601, opened 2020-08-25
(914) Enable a BEIS user to mark a submitted report as "in review"

lozette:
## Changes in this PR

Trello: https://trello.com/c/K6tgKNZv/914-a-beis-user-can-mark-a-report-as-in-review

This is a basic first pass at moving Reports from "submitted" to "in review"

When a Delivery Partner submits a report, it appears on the BEIS user's
dashboard. The BEIS user can then mark as it being "in review" to indicate to
other BEIS users that the report is being looked at.

The UI is very basic and will be iterated upon.

## Screenshots of UI changes

Screenshot 2020-08-25 at 16 24 50

Screenshot 2020-08-25 at 16 25 01

Screenshot 2020-08-25 at 16 25 11

Screenshot 2020-08-25 at 16 25 18

## Next steps

- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [x] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.
- [ ] Do any environment variables need amending or adding?
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/xml-validation.md)

PR #603, opened 2020-08-26
(915) a BEIS user can mark a report as awaiting changes

lozette:
## Changes in this PR

Trello: https://trello.com/c/cqMSIb5w/915-a-beis-user-can-mark-a-report-as-awaiting-changes

#601 Must be merged before this PR.

A BEIS user can move a report from "in review" to "awaiting changes". The DP
user will then see their reports awaiting changes on their dashboard.

Because a report can be moved from "in review" to either "awaiting changes" or
"approved", we have used a parameter in ReportsStateController to indicate
which state the report should be moved to.

The UI for this is still very basic and will be iterated upon.

## Screenshots of UI changes

Screenshot 2020-08-26 at 16 31 40
Screenshot 2020-08-26 at 16 31 50
Screenshot 2020-08-26 at 16 31 58
Screenshot 2020-08-26 at 16 32 05

## Next steps

### DIFF
--- a/app/controllers/staff/reports_controller.rb
+++ b/app/controllers/staff/reports_controller.rb
@@ -7,10 +7,11 @@ class Staff::ReportsController < Staff::BaseController
   include ActionController::Live
 
   def index
-    inactive_reports if current_user.service_owner?
-    current_user.service_owner? ? active_reports_with_organisations : active_reports
-    current_user.service_owner? ? submitted_reports_with_organisations : submitted_reports
-    current_user.service_owner? ? in_review_reports_with_organisations : in_review_reports
+    if current_user.service_owner?
+      reports_for_service_owner
+    else
+      reports_for_delivery_partner
+    end
   end
 
   def show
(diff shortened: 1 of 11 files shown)


app/controllers/staff/reports_controller.rb
```diff
@@ -11,6 +11,7 @@ def index
     current_user.service_owner? ? active_reports_with_organisations : active_reports
     current_user.service_owner? ? submitted_reports_with_organisations : submitted_reports
     current_user.service_owner? ? in_review_reports_with_organisations : in_review_reports
+    current_user.service_owner? ? awaiting_changes_reports_with_organisations : awaiting_changes_reports
```
leeky, line 14: I'm finding this hard to understand at this point - would this be more readable:

```ruby
def index
  if current_user.service_owner?
    reports_for_service_owner
  else
    reports_for_delivery_partner
  end
end

private

def reports_for_service_owner
  active_report_with_organisations
  ...
end

..etc...
```

spec/features/staff/users_can_mark_a_report_as_awaiting_changes_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  context "signed in as a Delivery partner user" do
+    let(:delivery_partner_user) { create(:delivery_partner_user) }
+
+    before do
+      authenticate!(user: delivery_partner_user)
+    end
+
+    scenario "they cannot mark a report as in awaiting changes" do
+      report = create(:report, state: :in_review)
```
leeky, line 39: Shouldn't `Report` belong to an organisation that `delivery_partner_user` is a member of, as per the test below..?

leeky (approved):
I just added a couple of small suggestion, but otherwise LGTM! :shipit:

spec/features/staff/users_can_mark_a_report_as_awaiting_changes_spec.rb
lozette: Good point!

PR #605, opened 2020-08-27
(924) Store partner and RODA IDs

jcoglan:
## Changes in this PR

This set of changes adds the ability to store RODA Identifiers for activities. RODA Identifiers are defined hierarchically, so the ID of a fund becomes a prefix for the IDs of its programmes, and so on down the tree. The complete RODA ID schema is:

    {A}-{B}-{C}{D}

where, `{A}`, `{B}`, `{C}`, and `{D}` are the Identifier _fragments_ belonging to level-A, B, C and D activities. These fragments may contain letters, digits, dashes, underscores, and slashes. The combined length of `{A}-{B}` is 18 characters maximum, and `{C}{D}` is at most 21 characters, giving a maximum total length of 40.

Users can add RODA IDs to activities starting with funds, then moving to programmes and so on down the tree. Since validation relies on the length of parent's ID fragment, we only allow an ID to be added once all the ancestors of an activity have an ID fragment.

## Screenshots of UI changes

To begin with, a fund has a blank Roda Identifier, with a link to add one.

Screenshot 2020-08-27 at 17 48 23

A programme sitting under this fund does not yet have an 'Add' link, because its parent does not have a RODA ID.

Screenshot 2020-08-27 at 17 48 44

On clicking the 'Add' button, the user sees a step for adding the RODA Identifier fragment. This step is also added to the creation journer, after the Delivery Partner ID step.

Screenshot 2020-08-27 at 17 49 03

The step validates that the ID fragment only contains acceptable characters.

Screenshot 2020-08-27 at 17 49 26

A length validation appropriate to the activity's level is enforced.

Screenshot 2020-08-27 at 17 49 52

On succfully saving the ID, we return to the fund's page and the RODA Identifier is displayed.

Screenshot 2020-08-27 at 17 50 13

If we now look at the details for a programme under this fund, we have an 'Add' button for entering the RODA Identifier at this level.

Screenshot 2020-08-27 at 17 50 28

At this level, a length limit is enforced dynamically based on the length of the parent ID.

Screenshot 2020-08-27 at 17 50 54

On saving the record, the full ("compound") RODA Identifier is displayed, and cannot be edited.

Screenshot 2020-08-27 at 17 51 39

## Next steps

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -10,6 +10,7 @@ class Staff::ActivityFormsController < Staff::BaseController
     :level,
     :parent,
     :identifier,
+    :roda_identifier,
     :purpose,
     :sector_category,
     :sector,
@@ -74,6 +75,9 @@ class Staff::ActivityFormsController < Staff::BaseController
     when :identifier
       @activity.assign_attributes(delivery_partner_identifier: delivery_partner_identifier)
       add_transparency_identifier
+    when :roda_identifier
+      @activity.assign_attributes(roda_identifier_fragment: roda_identifier_fragment)
+      @activity.cache_roda_identifier!
     when :purpose
       @activity.assign_attributes(title: title, description: description)
     when :sector_category
@@ -133,6 +137,10 @@ class Staff::ActivityFormsController < Staff::BaseController
 
   def delivery_partner_identifier
     params.require(:activity).permit(:delivery_partner_identifier).fetch("delivery_partner_identifier", nil)
+  end
+
+  def roda_identifier_fragment
+    params.require(:activity).permit(:roda_identifier_fragment).fetch("roda_identifier_fragment", nil)
   end
 
   def sector_category
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -35,6 +35,8 @@ class Staff::ActivityFormsController < Staff::BaseController
     case step
     when :parent
       skip_step if @activity.fund?
+    when :roda_identifier
+      skip_step unless @activity.can_set_roda_identifier?
     when :blank
       skip_step
     when :programme_status
--- a/app/validators/roda_identifier_validator.rb
+++ b/app/validators/roda_identifier_validator.rb
@@ -1,0 +1,50 @@
+class RodaIdentifierValidator < ActiveModel::Validator
+  VALILD_IDENTIFIER = /^[A-Za-z0-9\-\_\/\\]+$/
+
+  def validate(activity)
+    fragment = activity.roda_identifier_fragment
+    return if fragment.blank?
+
+    unless VALILD_IDENTIFIER.match?(fragment)
+      activity.errors.add(:roda_identifier_fragment,
+        I18n.t("activerecord.errors.models.activity.attributes.roda_identifier_fragment.invalid_characters"))
+    end
+
+    validate_for_fund(activity, fragment) if activity.fund?
+    validate_for_programme(activity, fragment) if activity.programme?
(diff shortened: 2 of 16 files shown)


jcoglan:
[Store both partner and RODA IDs for activities](https://trello.com/c/46dPlfAL/924-store-both-partner-and-roda-ids-for-activities)

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -74,6 +77,9 @@ def update
     when :identifier
       @activity.assign_attributes(delivery_partner_identifier: delivery_partner_identifier)
       add_transparency_identifier
+    when :roda_identifier
+      @activity.assign_attributes(roda_identifier_fragment: roda_identifier_fragment)
+      @activity.cache_roda_identifier!
```
leeky, line 82: Did you consider tying setting the compound column to the `roda_identifier_fragment=` setter method (and maybe making the compound attribute read-only)?

There's quite a few behaviours that only exist within the multi-step form that has to be duplicated into ingest, CSV import etc, and they're easy to miss.

jcoglan: I did consider putting this logic in the model -- tying an update of `roda_identifier_compound` whenever `roda_identifier_fragment` is changed. However, I thought implementing it that way would remove a lot of context that's important for this change. If we put this logic in the model, it needs to respond to any situation when `roda_identifier_fragment` is changed, e.g. updating the `roda_identifier_compound` of all the child/descdendent activities.

I'd rather move this logic up a layer and embed it inside the workflow that checks whether you're allowed to set the RODA ID, which restricts the kinds of updates we're expecting and means `roda_identifier_compound` only needs to be updated on the current model.

I was worried that a more general-purspoe implementation could cause unexpected behaviour if someone actually needed to go in and change the `roda_identifier_fragment` via the console. Does that make sense?

Making `roda_identifier_compound` read-only is appealing -- is there a way of doing this so it can be set if it's `nil`, but not after that? Activities are not usually new records when this attribute is set.

jcoglan: I've now added `Activity#roda_identifier_compound=` to catch and code that's mistakenly setting this field.

app/validators/roda_identifier_validator.rb
```diff
@@ -0,0 +1,51 @@
...
+    parent_fragment = activity.parent.roda_identifier_fragment
+    limit = 21 - parent_fragment.size
+
+    check_size(activity, fragment, limit, :level_d_too_long)
+  end
+
+  def check_size(activity, fragment, limit, error_type)
+    if fragment.size > limit
+      activity.errors.add(:roda_identifier_fragment,
+        I18n.t("activerecord.errors.models.activity.attributes.roda_identifier_fragment.#{error_type}", limit: limit))
```
leeky, line 48: Since this error is in a standard ActiveRecord error scopes, couldn't you also do:

`activity.errors.add(:roda_identifier_fragment, error_type)`

config/locales/models/activity.en.yml
```diff
@@ -97,6 +99,7 @@ en:
         flow: Flow type
         geography: Benefitting recipient geography
         delivery_partner_identifier: Identifier
+        roda_identifier: RODA Identifier
```
leeky, line 102: Should probably be a lowercase `i` - "RODA identifier"

app/validators/roda_identifier_validator.rb
jcoglan: Good spot, I've just amended this.

config/locales/models/activity.en.yml
jcoglan: Just amended this.

leeky (approved):
LGTM! :shipit:

PR #608, opened 2020-09-01
(927) Show the actual vs forecast variance on the report show page

mec:
## Changes in this PR
As a first step towards showing all the changes made in a report in the application itself (rather than just the csv download) we want to show the total forecasted and actual spend and the variance for the reporting financial quarter.

This is also the place where we will allow users to add comments at the activity level in future work.

As we already calculate all the values for the csv this is a small change.

## Screenshots of UI changes

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -275,6 +275,7 @@
   `Add activity` now.
 - RODA Identifiers can be added to activities on creation and when updating
 - Reports in 'awaiting changes' can be submitted.
+ - Forecasted and actual spend and variance is shown on the report show page
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-14...HEAD
 [release-14]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-13...release-14
@@ -288,5 +289,3 @@
 [release-6]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-5...release-6
 [release-5]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-4...release-5
 [release-4]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-3...release-4
-[release-3]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-2...release-3
-[release-2]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-1...release-2
--- a/app/controllers/staff/reports_controller.rb
+++ b/app/controllers/staff/reports_controller.rb
@@ -17,14 +17,13 @@ class Staff::ReportsController < Staff::BaseController
   def show
     @report = Report.find(id)
     authorize @report
+
     @report_presenter = ReportPresenter.new(@report)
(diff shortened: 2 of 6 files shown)


app/controllers/staff/reports_controller.rb
```diff
@@ -147,12 +146,14 @@ def send_csv
...
-    @third_party_projects.each do |project|
-      response.stream.write ExportActivityToCsv.new(activity: project, report: @report).call
+    @report_activities.each do |activity|
+      response.stream.write ExportActivityToCsv.new(activity: activity, report: @report).call
     end
     response.stream.close
   end
+
+  def level_c_and_d_activities_for_report(report:)
+    return [] if report.nil?
```
leeky, line 158: Probably makes no different here, but you could also do `return Activity.none if report.nil?` which returns an empty set of activities.

app/controllers/staff/reports_controller.rb
```diff
@@ -147,12 +146,14 @@ def send_csv
...
+    Activity.where(level: :project).or(Activity.where(level: :third_party_project)).where(organisation: report.organisation).select { |activity| activity.associated_fund == report.fund }
```
leeky, line 159: I think you don't need the `or` here, i.e:

```ruby
Activity.where(level: [:project, :third_party_project], organisation: report.organisation)...
```

spec/features/staff/users_can_view_reports_spec.rb
```diff
@@ -112,6 +112,35 @@
...
+      scenario "the report shows the total forecasted and actual spend and the variance" do
+        quarter_one_2019 = Date.parse("2019-4-1")
+        quarter_two_2019 = Date.parse("2019-7-1")
+
+        activity = create(:project_activity, organisation: delivery_partner_user.organisation)
+
+        report = create(:report, :active, organisation: delivery_partner_user.organisation, fund: activity.associated_fund, financial_quarter: 1, financial_year: 2019, created_at: quarter_one_2019)
+        report_presenter = ReportPresenter.new(report)
+
+        _forecasted_value = create(:planned_disbursement, parent_activity: activity, period_start_date: quarter_one_2019, value: 1000)
```
leeky, line 124: Should the planned disbursement also have `report: report` like the transaction below?

leeky (approved):
I've added some minor comments, but LGTM! :shipit:

CHANGELOG.md
```diff
@@ -268,12 +268,11 @@
 - Submitted reports can be moved into the review state 
 - Reports in review can be moved into the awaiting changes state   
 - Transactions & Planned Disbursements cannot be edited if they are associated with an approved Report
-- BEIS users can move a Report into the approved state
```
leeky, line 271: Should this have been removed here?

app/controllers/staff/reports_controller.rb
mec: Pah! I went looking for that but settled! Thank you!

CHANGELOG.md
mec: Nope! Thanks!

spec/features/staff/users_can_view_reports_spec.rb
mec: No, I don't believe so. The planned disbursement for this financial quarter would not have been added in this report, rather in a previous one so I've just let the factory sort that.

PR #610, opened 2020-09-02
Reports that are awaiting changes can be submitted

mec:
## Changes in this PR

When we introduced the awaiting changes state we missed handling the
transition from `awaiting_changes` to `submitted` in the report state
controller.

Also removes a redundant expectation in the users can submit a
report spec.

## Next steps

### DIFF
--- a/spec/features/staff/users_can_submit_a_report_spec.rb
+++ b/spec/features/staff/users_can_submit_a_report_spec.rb
@@ -6,49 +6,69 @@ RSpec.feature "Users can submit a report" do
       authenticate!(user: delivery_partner_user)
     end
 
-    scenario "they can submit a report" do
-      report = create(:report, :active, organisation: delivery_partner_user.organisation)
-      report_presenter = ReportPresenter.new(report)
+    context "when the report is active" do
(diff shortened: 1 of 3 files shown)


spec/features/staff/users_can_submit_a_report_spec.rb
```diff
@@ -6,49 +6,69 @@
...
+
+        visit report_path(report)
         click_link t("action.report.submit.button")
-        click_button t("action.report.submit.confirm.button")
 
-        auditable_events = PublicActivity::Activity.all
-        expect(auditable_events.last.key).to include("report.submitted")
-        expect(auditable_events.last.owner_id).to include delivery_partner_user.id
-        expect(auditable_events.last.trackable_id).to include report.id
+        click_button "Confirm submission"
```
leeky, line 51: Should this use an `t(..)` helper rather than the text itself?

leeky (approved):
Well spotted - LGTM! :shipit:

mec:
> Well spotted - LGTM! :shipit:

It was Odeta not me! 😄

PR #611, opened 2020-09-02
(930) Add `intended beneficiaries` to activity form

MariluzHerrera:
Trello: https://trello.com/c/RzKvjBtR

## Changes in this PR

`intended_beneficiaries` is a new form field that was requested by the client. In this field, the user can add other benefitting countries for a specific activity. Please note this is for BEIS internal use and it does not get reported to IATI.

This field comes with some validations: the maximum number of intended beneficiaries a user can select is 10, this field is required when the user selects a region as recipient in the geography step, but it will be optional if the user selects a country on the geography step. In this case, the app will show an intermediate step asking the user whether they want to add additional benefitting countries or not. Based on what they answer, the `intended_beneficiaries` page will or will not be shown.

When the user selects either a region or a country on the `geography` step, only the countries within that region will be displayed as potential `intended beneficiaries`. If the region selected is `developing countries, unspecified`, then the full list of countries will be presented.

For the countries within each region list, I have used the latest mapping list that the client provided us with. Some of the yml files needed amendment as a result. Now some regions have different codes and others have been reorganised into different regions. 

This PR also includes a temporary placeholder on `intended_beneficiaries` field for ingesting existing data. We will be able to change this once we have the real data.

### After

Screenshot 2020-09-04 at 15 00 58

Screenshot 2020-09-02 at 14 43 42

Screenshot 2020-09-04 at 15 35 21

## Next steps


spec/factories/activity.rb
```diff
@@ -15,6 +15,7 @@
     geography { :recipient_region }
     recipient_region { "489" }
     recipient_country { nil }
+    intended_beneficiaries { ["CU", "DM", "DO"] }
```
leeky, line 18: How about using the simplest value possible here - maybe even the "not applicable" state, if that is valid?

spec/features/staff/users_can_add_benefitting_countries_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+          expect(page).to have_content t("form.legend.activity.intended_beneficiaries")
+          expect(page).to have_content("Argentina")
+          expect(page).to have_content("Venezuela")
+          check "Argentina"
+          check "Colombia"
+          check "Peru"
+          click_button t("form.button.activity.submit")
+          activity.reload
+          expect(activity.intended_beneficiaries).to eq(["AR", "CO", "PE"])
```
leeky, line 27: Will the array always be stored in alphabetical order?

app/services/ingest_iati_activities.rb
```diff
@@ -279,6 +280,11 @@ def call
     end
   end
 
+  private def add_intended_beneficiaries(legacy_activity:, new_activity:)
+    # To be ingested once we have the information
+    new_activity.intended_beneficiaries = ["Replace me"]
```
leeky, line 285: Will there be any side effects when editing an activity whose `intended_beneficiaries` has been set like this?

Do we need to set this value for the activities already ingested?

leeky (dismissed):
I've left a few comments, but otherwise this is really great work @MariluzHerrera and looked like it was a mammoth task to get the codelists in order 🎉

config/locales/codelists/2_03/iati.en.yml
```diff
@@ -286,24 +286,30 @@ en:
...
       ZM: Zambia
       ZW: Zimbabwe
+      'N/A': Not Applicable
     recipient_region:
+      '89': Europe, regional
       '189': North of Sahara, regional
       '289': South of Sahara, regional
       '298': Africa, regional
       '380': West Indies, regional
-      '389': North & Central America, regional
```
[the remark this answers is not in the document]
MariluzHerrera: Are those activities `recipient_region?` or `recipient_country?`. If the user selected the country and then the region got set in the background it _might_ be okay, but I'm not sure.
The regions I removed don't appear in the mapping list Odeta gave me. I guess I can ask her

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -211,6 +214,10 @@ def recipient_country
     params.require(:activity).permit(:recipient_country).fetch("recipient_country", nil)
   end
 
+  def intended_beneficiaries
+    params.require(:activity).permit(intended_beneficiaries: []).fetch("intended_beneficiaries", nil)
```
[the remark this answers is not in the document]
MariluzHerrera: I used the `drop(1)` because I was getting an empty string as first item of the array `intended_beneficiaries`, regardless of having other countries selected, and this was giving me problems with views and specs. I thought that empty string was generated by the default empty hidden field on the checkbox fieldset. Am I right? I've not used govUK checkboxes before

app/models/activity.rb
```diff
@@ -37,6 +38,7 @@ class Activity < ApplicationRecord
   validates :geography, presence: true, on: :geography_step
   validates :recipient_region, presence: true, on: :region_step, if: :recipient_region?
   validates :recipient_country, presence: true, on: :country_step, if: :recipient_country?
+  validates :intended_beneficiaries, presence: true, length: {maximum: 10, message: I18n.t("activerecord.errors.models.activity.attributes.intended_beneficiaries.too_many")}, on: :intended_beneficiaries_step, if: :requires_intended_beneficiaries?
```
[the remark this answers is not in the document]
MariluzHerrera: I didn't know about the automated translations, thanks! If I want to have a customised translation, is it okay to leave it as it is now?

config/locales/models/activity.en.yml
```diff
@@ -160,6 +164,7 @@ en:
         button:
           new: Add implementing organisation
         heading: Implementing organisations
+      # intended_beneficiaries: Add other benefitting countries
```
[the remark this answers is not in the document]
MariluzHerrera: I forgot about that. Not sure why I commented it out. I'll double check and delete it if not needed

config/locales/codelists/2_03/iati.en.yml
```diff
@@ -286,24 +286,30 @@ en:
       ZA: South Africa
       ZM: Zambia
       ZW: Zimbabwe
+      'N/A': Not Applicable
```
jcoglan, line 289: I'm not sure this should be listed as country, if it's not a valid country value in most contexts. Instead, this could be a switch on this specific step to determine whether intended beneficiaries are required or not.

spec/factories/activity.rb
MariluzHerrera: I selected a random array of countries to have more than one country on the factory. I'm happy to reduce the array but "Not Applicable" cannot be used here because `geography` is set to `recipient_region`, hence at least a benefitting country is needed. Do you want me to reduce the array to just one random country?

spec/features/staff/users_can_add_benefitting_countries_spec.rb
MariluzHerrera: The countries will be displayed in the UI in alphabetical order, and I _think_ that's the way they'll get stored

app/services/ingest_iati_activities.rb
MariluzHerrera: For activities with a value of `recipient_country` on `geography` step we could use "Not Applicable". For activities with `recipient_region` I don't have a better idea than this placeholder. Open to suggestions, though

app/controllers/staff/activity_forms_controller.rb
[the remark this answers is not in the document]
jcoglan: This might be an artifact of how Rails does checkboxes and radio buttons -- it generates something like this so the browser sends `0` if the checkbox is unchecked and `1` if it is set:

```html
<input type="hidden" name="model[field]" value="0">
<input type="checkbox" name="model[field]" value="1">
```

I'm not sure how this applies on these checkbox lists though -- how do I get a page in development where I can see this working?

app/helpers/codelist_helper.rb
```diff
@@ -66,6 +67,31 @@ def country_name_from_code(code)
     country.name
   end
 
+  def intended_beneficiaries_checkbox_options
+    recipient_region = @activity.recipient_region
+    list = load_yaml(entity: "activity", type: "intended_beneficiaries")
+    show_list = if recipient_region == DEVELOPING_COUNTRIES_CODE
```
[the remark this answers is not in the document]
jcoglan: I think we wrote it like this at first and this might be a rubocop-mandated change.

config/locales/codelists/2_03/iati.en.yml
MariluzHerrera: I know this shouldn't be there. The reason I put it in there is because of the way the `activity_presenter` works. If it wasn't there, when you are in the `activity details` page on an activity that has "Not Applicable" as `intended beneficiaries` you will not see the "Not Applicable" value. You'll see something like `translation missing: en.iati....etc` I couldn't find another way to display this value in the activity details without putting it on that file/list

app/helpers/codelist_helper.rb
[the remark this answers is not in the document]
MariluzHerrera: Yes, I remember trying different ways to write that logic and having standardrb complaining about them. I'll try to give it another go tomorrow in making it more readable

app/helpers/codelist_helper.rb
```diff
@@ -66,6 +67,31 @@ def country_name_from_code(code)
...
+      list.values.flatten
+    else
+      list[recipient_region]
+    end
+    objects = if @activity.recipient_country?
```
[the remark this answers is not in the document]
jcoglan: I think we discussed this when we spoke about it earlier -- like my earlier comment about including `N/A` in the country list, this should probably not be part of the list of country checkboxes. It should be a separate checkbox which controls whether any items from the country list need to be selected.

app/models/activity.rb
```diff
@@ -212,4 +214,8 @@ def requires_call_dates?
   def forecasted_total_for_date_range(range:)
     planned_disbursements.where(period_start_date: range).sum(:value)
   end
+
+  def requires_intended_beneficiaries?
+    recipient_region? || (recipient_country? && intended_beneficiaries != "Not Applicable")
```
jcoglan, line 219: Isn't `intended_beneficiaries` an array? As I've alluded to above, this would be more cleanly handled by adding a `boolean` field to the model to indicate whether intended beneficiaries are required, and then wiring the `Not applicable` checkbox up to that field.

jcoglan:
Though I've left a few nit-picky comments about how `Not applicable` is handled, this is a really solid piece of work and quite a challenging thing to implement nicely, and I really like how you've gone about it. It's well-tested and written up :)

app/models/activity.rb
MariluzHerrera: Not sure I fully understand your suggestion. We can pair on that on our 1:1 tomorrow if that's okay?

app/models/activity.rb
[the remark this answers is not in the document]
leeky: What I meant is - if the locale files have a key at `activerecord.errors.models.activity.attributes.intended_beneficiaries.too_long`, Rails will know where to find it without you having to explicitly write it here.

spec/factories/activity.rb
leeky: Actually, on second thought, this is stored as an array, so I'd leave it be :+1:

app/models/activity.rb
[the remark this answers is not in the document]
MariluzHerrera: Oh, I see!! Sorry, I misunderstood you. Thanks for that. I'll make the change :-)

config/locales/codelists/2_03/iati.en.yml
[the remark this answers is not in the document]
MariluzHerrera: The client suggested that they could go on production and manually update the region of each of these activities to the relevant existing one. I have given them a list of the activities that need amendment. They are happy with this plan.

app/presenters/activity_presenter.rb
```diff
@@ -68,6 +68,16 @@ def recipient_country
     I18n.t("activity.recipient_country.#{super}")
   end
 
+  def requires_additional_benefitting_countries
+    return if super.nil?
+    I18n.t("activity.requires_additional_benefitting_countries.#{super}")
+  end
+
+  def intended_beneficiaries
+    return if super.blank?
```
jcoglan, line 77: Do we need to return `"Not application"` here if the beneficiaries list is empty, or if `requires_additional_benefitting_countries` is `false`?

MariluzHerrera: I hadn't noticed this, thanks. The partial for activity details view has some logic in it so it won't display the field `intended beneficiaries` in the situations you've mentioned

MariluzHerrera: The only situation I can think of is in the case of an incomplete activity where `intended beneficiaries` haven't been set yet. In that case the field will be display on the activity details page but the value will be blank. I think if we display "Not applicable" could be confusing as the user wouldn't know if it's because the activity is incomplete or because the activity does not require any additional benefitting countries

PR #612, opened 2020-09-02
(924) Allow Delivery Partner Identifier to be edited

jcoglan:
## Changes in this PR

These commits replace our existing method of generating the transparency identifier so that it's generated from the RODA identifier, not from the Delivery Partner identifier. With that change in place, we can make it possible to edit Delivery Partner identifiers again.

## Screenshots of UI changes

When viewing an activity, it is possible to edit its delivery partner identifier.

Screenshot 2020-09-02 at 16 14 18

After setting the RODA identifier, the Transparency identifier is generated.

Screenshot 2020-09-02 at 16 14 40

## Next steps

### DIFF
--- a/app/views/staff/shared/xml/_activity.xml.haml
+++ b/app/views/staff/shared/xml/_activity.xml.haml
@@ -1,8 +1,5 @@
 %iati-activity{"default-currency" => activity.default_currency, "xml:lang" => activity.organisation.language_code }
-  - if activity.previous_identifier?
-    %iati-identifier= activity.previous_identifier
-  - else
-    %iati-identifier= activity.transparency_identifier
+  %iati-identifier= activity.iati_identifier
   %reporting-org{"type" => activity.reporting_organisation.organisation_type, "ref" => activity.reporting_organisation.iati_reference }
     %narrative= activity.reporting_organisation.name
   %title
--- a/db/data/20200903140938_correct_transparency_identifiers.rb
+++ b/db/data/20200903140938_correct_transparency_identifiers.rb
@@ -1,0 +1,20 @@
+class CorrectTransparencyIdentifiers < ActiveRecord::Migration[6.0]
+  def up
+    Activity.transaction do
+      Activity.update_all(transparency_identifier: nil)
+
+      Activity.lock.where.not(roda_identifier_compound: nil).find_each do |activity|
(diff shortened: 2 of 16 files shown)


app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -1,8 +1,5 @@
 %iati-activity{"default-currency" => activity.default_currency, "xml:lang" => activity.organisation.language_code }
-  - if activity.previous_identifier?
-    %iati-identifier= activity.previous_identifier
-  - else
-    %iati-identifier= activity.transparency_identifier
+  %iati-identifier= activity.iati_identifier
```
leeky, line 2: :+1:

db/data/20200903140938_correct_transparency_identifiers.rb
```diff
@@ -0,0 +1,20 @@
+class CorrectTransparencyIdentifiers < ActiveRecord::Migration[6.0]
+  def up
+    Activity.transaction do
+      Activity.update_all(transparency_identifier: nil)
+
+      Activity.lock.where.not(roda_identifier_compound: nil).each do |activity|
+        activity.transparency_identifier = [
+          activity.reporting_organisation.iati_reference,
```
leeky, line 8: I'm was wondering if this needed  to be run in a specific order - i.e. `Fund -> Programme -> Project -> Third Party Project`, but possibly not because you're constructing these from the individual activity `roda_identifier_fragment` 🤔

leeky (approved):
LGTM! :shipit: 

That looked like a confusing puzzle to solve - just have some small comments.

db/data/20200903140938_correct_transparency_identifiers.rb
```diff
@@ -0,0 +1,20 @@
+class CorrectTransparencyIdentifiers < ActiveRecord::Migration[6.0]
+  def up
+    Activity.transaction do
+      Activity.update_all(transparency_identifier: nil)
+
+      Activity.lock.where.not(roda_identifier_compound: nil).each do |activity|
```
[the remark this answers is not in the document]
jcoglan: Possibly. Would load fewer into memory at once but still hold a write lock on all affected records until the end of the transaction. Seems like a good idea though.

PR #614, opened 2020-09-03
Fixes to the report journey

mec:
## Changes in this PR

Whilst working to create the next report automatically (https://trello.com/c/kllPs5cG) I came accross a few issues I felt needed addressing first:

- We had a constraint on Report that meant reports had to be unique for the level A activity in each organisation, this is incorrect and would stop new reports being created an organisation has many reports for the same level A activity.
- Once you get a few reports in different states the report index page becomes visually cluttered, adding empty states to the tables eases this (hopefully)
- When editing a report (adding the description and/or deadline) it is super helpful for BEIS users to also see the organisation, Level A actvity (fund) and financial quarter.
- Because of the way reports work a BEIS user can try to activate an invalid one resulting in the application failing (500), we need to do more work on this possibly as it feels a bit clunky, as a stop gap handle the invalid report and show an error, whilst not perfect, it is better than the application blowing up.

## Screenshots of UI changes
image

image

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -288,6 +288,8 @@
 - Reports no longer have to be unique
 - Reports cannot be unique for the Level A activity
 - Add empty states for report tables
+- Show Level A (Fund), organisation and financial quarter on the report edit
+  page
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-15...HEAD
 [release-15]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-14...release-15
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -290,6 +290,7 @@
 - Add empty states for report tables
 - Show Level A (Fund), organisation and financial quarter on the report edit
   page
+- Handle attempts to activate invalid reports
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-15...HEAD
 [release-15]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-14...release-15
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -284,6 +284,8 @@
 - Delivery Partner Identifiers can be edited
 
 ## [unreleased]
+=======
+- Reports no longer have to be unique
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-15...HEAD
 [release-15]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-14...release-15
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -286,6 +286,8 @@
 ## [unreleased]
 =======
 - Reports no longer have to be unique
+- Reports cannot be unique for the Level A activity
+- Add empty states for report tables
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-15...HEAD
 [release-15]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-14...release-15
--- a/app/controllers/staff/reports_state_controller.rb
+++ b/app/controllers/staff/reports_state_controller.rb
@@ -47,10 +47,15 @@ class Staff::ReportsStateController < Staff::BaseController
 
   private def change_report_state_to_active
     authorize report, :activate?
-    report.update!(state: :active)
-    report.create_activity key: "report.activated", owner: current_user
-    @report_presenter = ReportPresenter.new(report)
-    render "staff/reports_state/activate/complete"
+    if report.valid?
+      report.update!(state: :active)
+      report.create_activity key: "report.activated", owner: current_user
+      @report_presenter = ReportPresenter.new(report)
+      render "staff/reports_state/activate/complete"
+    else
+      flash[:error] = t("action.report.activate.failure")
+      redirect_to report_path(report)
+    end
   end
 
   private def confirm_submission
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -11,7 +11,6 @@ class Report < ApplicationRecord
   has_many :transactions
   has_many :planned_disbursements
 
-  validates_uniqueness_of :fund, scope: :organisation
   validate :activity_must_be_a_fund
   validates :deadline, date_not_in_past: true, date_within_boundaries: true
 
--- a/app/controllers/staff/reports_controller.rb
+++ b/app/controllers/staff/reports_controller.rb
@@ -32,18 +32,22 @@ class Staff::ReportsController < Staff::BaseController
   end
 
   def edit
-    @report = Report.find(id)
-    authorize @report
+    report = Report.find(id)
+    authorize report
+
+    @report_presenter = ReportPresenter.new(report)
   end
 
   def update
-    @report = Report.find(id)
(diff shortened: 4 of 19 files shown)


CHANGELOG.md
```diff
@@ -279,6 +279,7 @@
 - Ingest RS GCRF data from IATI
 - Ingest BA GCRF data from IATI
 - Ingest RAEng GCRF data from IATI
+- Reports cannot be unique for the Level A activity
```
leeky, line 282: This might need rewriting with what the change *does* do...

jcoglan: Yes, I found this wording hard to parse in the changelog and commit message. Maybe "Reports no longer have to be unique" rather than "Reports cannot be unique".

app/models/report.rb
```diff
@@ -11,7 +11,6 @@ class Report < ApplicationRecord
   has_many :transactions
   has_many :planned_disbursements
 
-  validates_uniqueness_of :fund, scope: :organisation
```
jcoglan, line 14: This makes sense to remove but makes me wonder why this constraint was here to begin with. Do we know?

app/controllers/staff/reports_state_controller.rb
```diff
@@ -47,10 +47,15 @@ def update
 
   private def change_report_state_to_active
     authorize report, :activate?
-    report.update!(state: :active)
-    report.create_activity key: "report.activated", owner: current_user
-    @report_presenter = ReportPresenter.new(report)
-    render "staff/reports_state/activate/complete"
+    if report.valid?
```
jcoglan, line 50: Would it make sense to perform this check on all other state changes? This is the first change that's made one `change_report_state` method structurally different from any of the others, and as I'd considered refactoring all these methods down to a single implementation I'm curious about the need for them to differ.

app/models/report.rb
mec: I put it down as a simple mistake, there is no way a report can be unique for a fund inside the scope of an organsiation,  

I imagine it was missed becase our tests don't get setup with prior reports and we only have one report per fund and organisation right now, as soon as I approved one and tried to create the new one the error was clear - and that drove this change.

app/controllers/staff/reports_state_controller.rb
mec: I've had a go a drying up refactor - to be honest, I think I prefer the more verbose way, let me know what you think and how I might improve this if I have missed the mark!?

mec: (I've just seen you've got something similar in

app/controllers/staff/reports_state_controller.rb
```diff
@@ -3,18 +3,26 @@
 class Staff::ReportsStateController < Staff::BaseController
   include Secured
 
+  STATE_TO_POLICY = {
```
jcoglan, line 6: The policies here could have the `?` be part of the policy name, as in `active: :activate?`, rather than appending the `?` inside the controller methods.

app/controllers/staff/reports_state_controller.rb
```diff
@@ -24,89 +32,40 @@ def edit
...
-    @report_presenter = ReportPresenter.new(report)
-    render "staff/reports_state/request_changes/confirm"
-  end
-
-  private def change_report_state_to_awaiting_changes
-    authorize report, :request_changes?
-    report.update!(state: :awaiting_changes)
-    report.create_activity key: "report.awaiting_changes", owner: current_user
+  private def confirm_state(state)
+    authorize report, state + "?"
```
jcoglan, line 51: Should this use `STATE_TO_POLICY` instead of assuming the state name is the same as the policy name? This looks like a bug to me, so I'd expect some tests to be failing because of this.

app/controllers/staff/reports_state_controller.rb
```diff
@@ -24,89 +32,40 @@ def edit
...
     @report_presenter = ReportPresenter.new(report)
-    render "staff/reports_state/request_changes/complete"
+    render "staff/reports_state/#{state}/confirm"
```
jcoglan, line 53: There's a similar problem here with assuming the state name is the same as the names of the views, which I'm not sure is true. This has a lot of overlap with the changes I've made in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/616/commits/ae79d1a2519305075aaccaaa8859a9c62992f56c, which extracts policy names, log actions, and view names into constants so they can be looked up by state name.

mec: As you have taken this further why don't I drop the refactor and you add your work from #616 ontop of this?

app/controllers/staff/reports_state_controller.rb
```diff
@@ -24,89 +32,40 @@ def edit
...
-  private def change_report_state_to_approved
-    authorize report, :approve?
-    report.update!(state: :approved)
-    report.create_activity key: "report.approve", owner: current_user
-    @report_presenter = ReportPresenter.new(report)
-    render "staff/reports_state/approve/complete"
+    authorize report, policy + "?"
+    if report.valid?
+      report.update!(state: state)
+      report.create_activity key: "report.state.changed_to.#{state}", owner: current_user
```
jcoglan, line 62: We've changed the names of the events sent to the audit log here -- is that a problem or is it fine to change those?

app/controllers/staff/reports_state_controller.rb
jcoglan: Ah sorry, this was my mistake in reading the code: the parameter to this method is named `state`, but it's not a state name, it's an action, e.g. `review` rather than `in_review`. That means the `authorize` and `render` calls are correct but the code could be clarified, either by renaming the `state` parameter here, or making `confirm_state` and `change_report_state_to` _both_ take state names.

mec: Yeah, I was going to say that 'confirm' methods needs better langauge, but I think dropping my refactor and applying yours is the better option? Thoughts?

jcoglan: I think it would make sense to back this refactor out of this PR, and combine it with some of my changes in #616. I like what you've done here, would just benefit from a bit of renaming and being reviewed separately from all the functional changes in this PR.

jcoglan: (It's actually less complicated than what I attempted in #616, it doesn't require quite as much config/constants mapping states to other things as I've done.)

mec: Okay cool, I only glanced at #616, but I am interested in the race condition (plus I just love to try and use that phrase!) I've dropped the refactor on here. I think without the refactor this still stands and could go in as long as we follow up by making sure we handle your concerns.

jcoglan (approved):
This looks great now -- really nice set of UX improvements, thanks @mec :)

app/controllers/staff/reports_state_controller.rb
jcoglan: Absolutely :) Do open your refactoring as a new PR and we'll land that before I have another look at #616. I'm sorry, I should have checked in with you before I started my PR, I'd assumed that controller was "finished" and wouldn't be being changed much.

mec: No probs at all! 👍 I'll open a PR but I'm interested in why you want to pass the state as a parm so it may be #616 makes more sense!

jcoglan: Although I had the idea for this refactoring while turning `state` into a form parameter, it really doesn't have to be done in my PR, it can be done separately. Let's get your refactoring of the `confirm` and `change` methods in and then I'll rebase #616.

PR #616, opened 2020-09-04
Pass report state as a parameter

jcoglan:
While thinking through some consistency problems in relation to how identifiers and bulk actions are handled, it occurred to me that the `ReportsStateController` is vulnerable to a couple of race conditions that coule occur if two or more people are working on the same report. Descriptions of those conditions are in the commit messages.

### DIFF
--- a/app/controllers/staff/reports_state_controller.rb
+++ b/app/controllers/staff/reports_state_controller.rb
@@ -30,11 +30,15 @@ class Staff::ReportsStateController < Staff::BaseController
   end
 
   def update
-    if STATE_TO_POLICY_ACTION.key?(params[:state])
-      change_report_state_to(params[:state])
-    else
-      authorize report
-      redirect_to report_path(report)
+    state = params[:state]
+
+    Report.transaction do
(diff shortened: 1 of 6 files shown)


app/controllers/staff/reports_state_controller.rb
```diff
@@ -47,20 +51,23 @@ def update
...
-      redirect_to report_path(report)
     end
+
+    @report_presenter = ReportPresenter.new(report)
+    render "staff/reports_state/#{policy_action}/complete"
   end
 
   private def report
-    @report ||= Report.find(params[:report_id])
+    @report ||= Report.lock.find(params[:report_id])
```
mec, line 71: What led you to decide to lock this record, because you expect it to be changed?

mec (approved):
Very nice! I find this super interesting for lots of reasons! I left one question about the `lock` for my own learning.

I guess I am also interested in the expectaion this sets for Bob - they think they submitted when in-fact it was Alice - is the 'You are not authorised' a more accurate response? I don't think there is a right or wrong answer and don't expect one, just enjoying the thought process! 😄 

Happy to approve this! 🚢

PR #617, opened 2020-09-04
Refactor reports state controller

mec:
## Changes in this PR

There was quite a bit of duplication in this controller.

I really like having the policy name and the report state different, I
think it makes things nice and clear. Not wanting to change this meant
having some kind of mapping from state names to policy names – I went
with a hash.

I've renamed the confirm_<state> method to try and better describe the
behaviour and align it with the argument that is passed in (a policy
action name)

I've also changed the key for the Public Activity record, I am confident
this is not an issue as we do not expose public activity records as yet
and we have no report state changes in production either (they are all
inactive).

I added all the state change failure messages to the report.yml locale
file, due to the linear nature of the state change I can't see how any
of these will be triggered but they are in for completeness.

## Next steps

### DIFF
--- a/app/controllers/staff/reports_state_controller.rb
+++ b/app/controllers/staff/reports_state_controller.rb
@@ -3,18 +3,26 @@
 class Staff::ReportsStateController < Staff::BaseController
   include Secured
 
+  STATE_TO_POLICY_ACTION = {
+    active: :activate,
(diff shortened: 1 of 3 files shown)


jcoglan (approved):
Really like the names you've settled on for the methods and parameters here, it's a lot clearer :)

PR #621, opened 2020-09-07
(964) New reports are created automatically

mec:
## Changes in this PR

When the current report is approved we go ahead and create the next report
in the inactive state, ready for BEIS to activate.

The application relies on there only ever being one active report for a
given delivery partner organisation and level A activity so we want the
application to handle the creation of the new one to avoid error.

We also know that due to the nature of the service a report will have to
be approved either in the financial quarter in which it was created or
the following one - in both cases creating the new report upon approval
means that the new report has the correct financial year.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -301,6 +301,7 @@
 - Update activity policy to account for the report state
 
 ## [unreleased]
+- New reports are created when the prior is approved
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-16...HEAD
 [release-16]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-15...release-16
--- a/app/controllers/staff/reports_state_controller.rb
+++ b/app/controllers/staff/reports_state_controller.rb
@@ -61,6 +61,9 @@ class Staff::ReportsStateController < Staff::BaseController
     if report.valid?
       report.update!(state: state)
       report.create_activity key: "report.state.changed_to.#{state}", owner: current_user
+
+      find_or_create_new_report(organisation_id: report.organisation.id, fund_id: report.fund.id) if state == :approved
+
       @report_presenter = ReportPresenter.new(report)
       render "staff/reports_state/#{policy_action}/complete"
     else
@@ -71,5 +74,9 @@ class Staff::ReportsStateController < Staff::BaseController
 
   private def report
     @report ||= Report.find(params[:report_id])
+  end
+
+  private def find_or_create_new_report(organisation_id:, fund_id:)
+    Report.where.not(state: :approved).find_or_create_by!(organisation_id: organisation_id, fund_id: fund_id)
   end
 end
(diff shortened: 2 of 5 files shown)


app/controllers/staff/reports_state_controller.rb
```diff
@@ -111,10 +111,17 @@ def update
...
+    create_new_report(organisation_id: report.organisation.id, fund_id: report.fund.id)
+
     render "staff/reports_state/approve/complete"
   end
 
   private def report
     @report ||= Report.find(params[:report_id])
   end
+
+  private def create_new_report(organisation_id:, fund_id:)
```
jcoglan, line 124: The only thing I'd worry about here is how it's enforced that only one "modifiable" report (i.e. not approved, so is either in a state where the DP can edit it, or can be moved to such a state) exists per fund and DP at a time. Rather than implementing this as "create a new `Report` and save it", it might be better as "make sure a `Report` with the right properties and state exists", in case this action ends up being executed more than once somehow.

mec: I am going to pair with @lozette on this so I'll move this to draft until we've done that.

mec: Spoken to you as well now @jcoglan - great talk! Thanks for explaining it all, I've updated this a bit now to defend against duplicate reports being created – let me know what you think.

CHANGELOG.md
```diff
@@ -292,11 +292,13 @@
 - Handle attempts to activate invalid reports
 - `Recipient_region` codelist modified
 - `Intended beneficiaries` form field added, including validations, and country to region mapping
+<<<<<<< HEAD
```
jcoglan, line 295: Looks like a merge conflict marker, just needs to be removed.

mec: not my day for this! Thanks!

app/controllers/staff/reports_state_controller.rb
```diff
@@ -72,4 +75,8 @@ def update
   private def report
     @report ||= Report.find(params[:report_id])
   end
+
+  private def find_or_create_new_report(organisation_id:, fund_id:)
+    Report.where.not(state: :approved).find_or_create_by!(organisation_id: organisation_id, fund_id: fund_id)
```
jcoglan, line 80: I didn't realise you could chain `where.not(...)` with `find_or_create_by!` like this, but I've checked the SQL and is is doing the right thing:

```sql
SELECT "reports".* FROM "reports"
WHERE "reports"."state" != 'approved'
AND "reports"."organisation_id" = '...'
AND "reports"."fund_id" = '...'
LIMIT 1;
```

If this returns nothing, it creates a new `Report` with the given `fund_id` and `organisation_id`, and relies on the database to set the default value of `state = 'inactive'`. Nice :)

jcoglan (approved):

app/controllers/staff/reports_state_controller.rb
mec: Yeah - it was new to me as well but I ❤️ it!

mec: I think it's a new Rails 6 one…

PR #622, opened 2020-09-07
(898)`ODA eligibility` form step added to activity

MariluzHerrera:
Trello: https://trello.com/c/Kt1vrYSS

## Changes in this PR

As part of the new fields on create activity journey requested by the client, the ODA eligibility has been added to every level activity. Worth noting that `ODA eligibility` is a requirement only for levels C and D activities, but the client suggested we implemented this on all levels for now. They are happy with refining this later if needed.

This new field has two options: `Eligible` and `No longer eligible`. A defaulted to `Eligible` has been implemented, as requested by the client. This field does not get reported to IATI or included in any XML.

This commit introduces the new form step, as well as tests to assert we achieve the desired behaviour.

### After

Screenshot 2020-09-07 at 11 35 44

Screenshot 2020-09-07 at 11 36 07

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -23,6 +23,7 @@ class Activity < ApplicationRecord
     :intended_beneficiaries_step,
     :flow_step,
     :aid_type,
+    :oda_eligibility_step,
   ]
 
   strip_attributes only: [:delivery_partner_identifier, :roda_identifier_fragment]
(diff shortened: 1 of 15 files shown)


app/models/activity.rb
```diff
@@ -23,6 +23,7 @@ class Activity < ApplicationRecord
     :intended_beneficiaries_step,
     :flow_step,
     :aid_type,
+    :oda_eligibility,
```
jcoglan, line 26: For consistency with the other values here, could we name this `oda_eligibility_step`? It helps us tell these validation events apart from the `form_state` values. (I just noticed `aid_type` should have `_step` appended as well.)

MariluzHerrera: Oops! I was meant to write `oda_eligibility_step`. I'll amend it. Shall we do a quick new card for the `aid_type` step?

jcoglan: I'll just make a quick PR to do that after this one is merged. If I do this at the same time, it will conflict with this one.

jcoglan (approved):

PR #623, opened 2020-09-07
(976) Amend the data in three columns on the Report CSV export

lozette:
## Changes in this PR

Trello: https://trello.com/c/BrJD86Sg/976-amendments-to-csv-export-data

1. Show the underlying code on Aid Type & Sector in CSV

In the CSV export, show the underlying IATI code number alongside the Aid Type
and Sector information for activities

2. Export programme_status from Activity, not status 

We were previously exporting the `status` attribute of activities into the CSV
report, this was incorrect and it should show the `programme_status`

## Next steps


app/presenters/activity_presenter.rb
```diff
@@ -8,11 +8,21 @@ def aid_type
     I18n.t("activity.aid_type.#{super.downcase}")
   end
 
+  def aid_type_with_code
+    return if aid_type.blank?
+    "#{aid_type} (#{to_model.aid_type})"
```
MariluzHerrera, line 13: Interesting! I've never come across `to_model` before. I just read the documentation that says that often acts like `self` (although in the examples I've seen `to_model` appears as a method for an object, and not the other way around). Do you mind me asking what is `to_model` doing exactly in this case?

lozette: It's accessing the original model pre-decoration (or before it's been through the presenter). So if I called `aid_type` here I get the result of the presented `aid_type` (which is going to be an internationalised string). If I want to get the original un-presented value I call `to_model`. `self` in this case would be the presented object, we want the original object.

MariluzHerrera: I see! That makes sense. Thanks for the explanation! In that case, I think the PR looks good to me

MariluzHerrera (approved):
🚀

PR #625, opened 2020-09-07
(970) Update the organisation_id of programmes to be the BEIS organisation ID

lozette:
## Changes in this PR

Trello: https://trello.com/c/2bPBRojL/970-all-level-b-activities-should-have-the-organisation-association-set-to-beis

During a recent data migration, some ingested activities were moved from
"Level C" to "Level B". However their organisation_id was not updated to reflect
the fact that Level B activities always belong to the BEIS organisation, not a
Delivery Partner organisation.

This data migration corrects that oversight; all Level B activities now have
the BEIS organisation as their owning organisation.

## Next steps

### DIFF
--- a/db/data/20200907144609_update_organisation_on_all_programme_activities.rb
+++ b/db/data/20200907144609_update_organisation_on_all_programme_activities.rb
@@ -1,0 +1,10 @@
+class UpdateOrganisationOnAllProgrammeActivities < ActiveRecord::Migration[6.0]
+  def up
+    beis_organisation_id = Organisation.find_by(service_owner: true).id
+    Activity.programmes.update_all(organisation_id: beis_organisation_id)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -284,7 +284,6 @@
 - Delivery Partner Identifiers can be edited
 
 ## [unreleased]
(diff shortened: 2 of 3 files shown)


db/data/20200907144609_update_organisation_on_all_programme_activities.rb
```diff
@@ -0,0 +1,14 @@
+class UpdateOrganisationOnAllProgrammeActivities < ActiveRecord::Migration[6.0]
+  def up
+    programmes = Activity.programmes
+    beis_organisation_id = Organisation.find_by(service_owner: true).id
+    programmes.each do |programme|
```
jcoglan, line 5: Is there a reason to loop over these in memory and do an `UPDATE` for each one, rather than using `Activity.programmes.update_all`?

lozette: You're right there isn't, I was using a pattern from when I updated all the `transparency_id` attributes on Activities before, but for that change we had a different `transparency_id` for each record 🤦

MariluzHerrera (approved):
It looks good to me! 🚀

jcoglan (approved):

db/data/20200907144609_update_organisation_on_all_programme_activities.rb
jcoglan: That makes sense :)

PR #628, opened 2020-09-08
(906) Update activity policy

mec:
## Changes in this PR
This work started as trying to update the transaction policy but I soon realised it makes more sense to do the Activity policy first as Budgets, Transactions and PlannedDisbursments are all 'added to' an Activity.

The idea is to set out to make the policy spec document what actions a given user can perform on a given Activity under certain conditions.  The new condition here is the state of the 'current active report' but it turned out we have others that I hope have now been covered.

Activities are a bit wierd because they can be created without a level or parent activity, without which we have limited options for calculating the permissions, ultimately though, users should only edit activities that are 'owned' by their own organisation (through the `organisation` association) so this is general rule.

I've taken the stance that BEIS users can see everything and edit their own acitivites at any time - we have no concept of a 'internal BEIS report assurance' so this seems the only way to handle this right now.

For Delivery partners there are a lot of factors that influence what they can do and when, and the hope is that these are now documented in the spec…

The commit that introduces the policy changes breaks the test suite as lots of specs now need a report in the correct state to work correctly, I kept the fixes for these in a seperate commit. The fix used was to just setup a generic report with FactoryBot as I felt, generally speaking, a report in a spec setup will need the organisation and fund so further traits on the factory wouldn't be helpful.

We have more work to do after this, but I have more confidence now that changes cannot be made to the models even when the UI might be a bit behind and make it seem like you can.

We need to followup with at least updates to the transaction and planned disbursement policies, we should also look at how activites and budget relate to reports if possible.

Apologies for the first large commit - I found the metal agitlity to write the new spec hard to come by and wasn't able to also split the work at sensible points! 😞 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -298,6 +298,7 @@
   associated organisation. All Level B activities should belong to BEIS.
 - `GDI` form step added to create activity journey
 - Show Aid Type and Sector codes in Report CSV export for activities
+- Update activity policy to account for the report state
 
 ## [unreleased]
 
@@ -313,5 +314,3 @@
 [release-8]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-7...release-8
 [release-7]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-6...release-7
 [release-6]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-5...release-6
-[release-5]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-4...release-5
-[release-4]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-3...release-4
--- a/spec/features/staff/users_can_create_a_budget_spec.rb
+++ b/spec/features/staff/users_can_create_a_budget_spec.rb
@@ -134,6 +134,7 @@ RSpec.describe "Users can create a budget" do
         project_activity = create(:project_activity,
           parent: programme_activity,
           organisation: user.organisation)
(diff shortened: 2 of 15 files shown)


spec/features/staff/users_can_create_a_budget_spec.rb
```diff
@@ -134,6 +134,7 @@
         project_activity = create(:project_activity,
           parent: programme_activity,
           organisation: user.organisation)
+        _report = create(:report, state: :active, organisation: user.organisation, fund: fund_activity)
```
lozette, line 137: Agree that this is a bit lumpy but I'm not sure of a better way we can do it right now given the Report needs to belong to a specific fund/org. We can probably put this on the to refactor list ;)

lozette (approved):
Nice work

spec/features/staff/users_can_create_a_budget_spec.rb
mec: Yeah, I feel like we'll be setting the org and fund every time so having a factory won't help.

Also, that indentation!! I'll sort it!

mec: Ooopsie! Indentation was right the first time! 🤣

CHANGELOG.md
```diff
@@ -295,8 +295,10 @@
 - `ODA eligibility` form step added to create activity journey
 - Update any ingested Level B activities that do not have the BEIS organisation as their
   associated organisation. All Level B activities should belong to BEIS.
+<<<<<<< HEAD
```
leeky, line 298: ^^^

leeky:
This won't affect ingest, as it doesn't check policies at all.

CHANGELOG.md
mec: ah no... thanks!

PR #629, opened 2020-09-09
Release 16

lozette:
- Reports no longer have to be unique
- Reports cannot be unique for the Level A activity
- Add empty states for report tables
- Show Level A (Fund), organisation and financial quarter on the report edit
  page
- Handle attempts to activate invalid reports
- `Recipient_region` codelist modified
- `Intended beneficiaries` form field added, including validations, and country to region mapping
- `ODA eligibility` form step added to create activity journey
- Update any ingested Level B activities that do not have the BEIS organisation as their
  associated organisation. All Level B activities should belong to BEIS.
- `GDI` form step added to create activity journey
- Show Aid Type and Sector codes in Report CSV export for activities

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -283,7 +283,8 @@
 - Transparency Identifiers are set based on RODA Identifiers
 - Delivery Partner Identifiers can be edited
 
-## [unreleased]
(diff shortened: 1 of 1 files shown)


leeky (approved):
:shipit:

PR #630, opened 2020-09-09
(850) Add `total applications` and `total awards` to create activity 

MariluzHerrera:
Trello: https://trello.com/c/ETRlZE9w

## Changes in this PR

`Total applications` and `Total awards` are combined in a new field when users create level C or D activities. They will exist only if there is a call present on the activity. Applications refer to organisations applying for a grant/fund related to a specific activity. Awards is related to the number of organisations that have been awarded the grant/fund they applied for.

This field is optional for activities when `ingested: true`. It does not get reported to IATI.

Both fields (applications and awards) only accept integers. We agreed with the client that the value can be set to `0` whenever the application process is still ongoing, or when an activity has been suspended before the call.

### After

Screenshot 2020-09-09 at 11 43 09

Screenshot 2020-09-09 at 11 43 46

## Next steps


db/migrate/20200908155632_add_total_applications_to_activities.rb
```diff
@@ -0,0 +1,5 @@
+class AddTotalApplicationsToActivities < ActiveRecord::Migration[6.0]
+  def change
+    add_column :activities, :total_applications, :integer
```
leeky, line 3: As this is an integer, perhaps this could default to `0` and not allow NULLs..?

db/migrate/20200908155742_add_total_awards_to_activities.rb
```diff
@@ -0,0 +1,5 @@
+class AddTotalAwardsToActivities < ActiveRecord::Migration[6.0]
+  def change
+    add_column :activities, :total_awards, :integer
```
leeky, line 3: ... and same with this column.

spec/models/activity_spec.rb
```diff
@@ -496,11 +496,29 @@
         end
       end
 
+      context "when there is a call and total applications is blank" do
+        subject(:activity) { build(:project_activity, call_present: true, total_applications: nil) }
+        it "should not be valid" do
+          expect(activity.valid?(:total_applications_and_awards_step)).to be_falsey
```
leeky, line 502: `valid?` will only ever return `true` or `false`, so you could us `be false` here.

`truthy` and `falsey` [aren't exactly the same](https://eddyluten.com/rspec-be_truthy-exists-or-be-true) as `true` and `false`

leeky (approved):
A couple of minor comments that you can feel free to ignore where appropriate.

 LGTM! :shipit:

spec/models/activity_spec.rb
MariluzHerrera: I used `be_falsey` to keep this test inline with the rest of the tests in this file. Almost all use the terms `truthy` or `falsey`. Shall I change mine to `be_false`?

db/migrate/20200908155632_add_total_applications_to_activities.rb
MariluzHerrera: Yeah, I wasn't sure about defaults here. Is defaulting to `0` the convention? I guess the reason I didn't default in the end is because I wanted to be sure the user enters a number (by having validations of presence on this step), instead of clicking `continue` by mistake and saving the wrong value. I'm unsure what the best option is in this case. What do you think?

spec/models/activity_spec.rb
leeky: I'd stick with consistency 👍

db/migrate/20200908155632_add_total_applications_to_activities.rb
leeky: Good point there about the forms - if we preset this to `0` does the form get pre-filled with that value? I note further down that `0` has a special (temporary) meaning anyway, so maybe leave this be.

It's probably not a convention, but it's something I often do - possibly because I've been bitten by databases that can't `sum` columns that contain `NULL` in the past 🤷

MariluzHerrera: That was my assumption, only because we I defaulted a boolean to `true`, the radio button for the true value gets pre-filled. I thought defaulting to `0` would have the same behaviour (I might be wrong, I'll check). 
Happy to add not allowing `NULL`. Don't want to cause trouble with databases 😅

PR #633, opened 2020-09-10
(fix) Transaction and Planned disbursements can be added

mec:
## Changes in this PR
When there are more than one reports for an organisation and fund we can
be in the position that the policies for transactions and planned
disbursements stop them from being created.

We have work planned to refactor both of these policies, but we also
have a need to add some today for a training session.

To fix the issue short term we make the associated_report only return
reports that are in the editable state of either active or awaiting
changes and not any state that has potential to return reports in other
states that would stop the transactions being added.

This work does not replace the refactor, merely enables us to run
training session today (10 Sep 2020)


lozette (approved):
Works for me!

PR #634, opened 2020-09-10
(1002) Show the deadline on the report edit form

mec:
## Changes in this PR
When the financial year and quarter were added to the edit form as
read-only values we had to use a presenter rather than the report
itself.

Something about doing so stops the value of the deadline field showing
in the form. If we switch to the report object itself,everything works
as expected so we instead pass in the report presenter as a local to the
form partial and use that to render the financial quarter and year.

## Next steps

### DIFF
--- a/app/views/staff/reports/_form.html.haml
+++ b/app/views/staff/reports/_form.html.haml
@@ -1,15 +1,15 @@
 = f.govuk_error_summary
 .govuk-form-group
   = label_tag "organisation", t("form.label.report.organisation"), class: "govuk-label"
-  = content_tag :p, f.object.organisation.name, class: "govuk-body", id: "organisation"
+  = content_tag :p, report_presenter.organisation.name, class: "govuk-body", id: "organisation"
 
 .govuk-form-group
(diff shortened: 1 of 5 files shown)


app/views/staff/reports/_form.html.haml
```diff
@@ -1,15 +1,15 @@
...
 
 .govuk-form-group
   = label_tag "financial-quarter-and_year", t("form.label.report.financial_quarter_and_year"), class: "govuk-label"
-  = content_tag :p, f.object.financial_quarter_and_year, class: "govuk-body", id: "finacial-quarter-and-year"
+  = content_tag :p, report_presenter.financial_quarter_and_year, class: "govuk-body", id: "finacial-quarter-and-year"
 
 .govuk-form-group
   = label_tag "level-a-activity", t("form.label.report.level_a_activity"), class: "govuk-label"
-  = content_tag :p, f.object.fund.title, class: "govuk-body", id: "level-a-acitivty"
+  = content_tag :p, report_presenter.fund.title, class: "govuk-body", id: "level-a-acitivity"
```
lozette, line 12: Typo here - `acitivity`

lozette (approved):
Approved as long as you fix the typo :)

PR #635, opened 2020-09-11
(994) Activity hierarchy levels displayed on UI

MariluzHerrera:
Trello: https://trello.com/c/09vlNklC

## Changes in this PR

As a request from the client, in order to ensure consistency in how people view the service, we have updated the terminology for activity levels. They now show both the former terminology (`Fund`, `Programme`, `Project`, and `Third-party project`) along with their respective levels (A, B, C, and D), which the users are more familiar with.

This is only a change on the UI, mainly in the translation files, so we can still keep our code without causing much disruption at this moment.

This PR includes changes on the translation files, amendments on the specs, and a custom method for capitalisation when presenting the activity levels in some of the views. It also introduces the updated definitions for each activity level that will be displayed when users click on `Add activity` and choose the level of activity they want to create.

### Before

Screenshot 2020-09-11 at 10 04 32

Screenshot 2020-09-11 at 10 05 04

### After

Screenshot 2020-09-11 at 10 02 17

Screenshot 2020-09-11 at 10 01 35

Screenshot 2020-09-11 at 10 03 28

Screenshot 2020-09-11 at 10 02 49

## Next steps

### DIFF
--- a/app/helpers/activity_helper.rb
+++ b/app/helpers/activity_helper.rb
@@ -17,4 +17,8 @@ module ActivityHelper
     return parent.title if parent.fund? && user.delivery_partner?
     link_to parent.title, organisation_activity_path(parent.organisation, parent), {class: "govuk-link govuk-link--no-visited-state"}
   end
+
+  def custom_capitalisation(level)
+    "#{level.chars.first.upcase}#{level[1..-1]}"
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -304,6 +304,8 @@
 - New reports are created when the prior is approved
 - `Total applications` and `Total awards` form step added to create activity journey
 - Report deadline value is shown on the edit form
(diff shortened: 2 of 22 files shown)


app/helpers/activity_helper.rb
```diff
@@ -17,4 +17,8 @@ def link_to_activity_parent(parent:, user:)
     return parent.title if parent.fund? && user.delivery_partner?
     link_to parent.title, organisation_activity_path(parent.organisation, parent), {class: "govuk-link govuk-link--no-visited-state"}
   end
+
+  def custom_capitalisation(level)
```
lozette, line 21: I'd like to see a spec for this in `activity_helper_spec` to make it obvious what it's doing

lozette (changes requested):

app/helpers/activity_helper.rb
MariluzHerrera: That's true. I forgot about doing one. Thanks!

lozette (approved):

PR #636, opened 2020-09-11
(982) Update transaction policy

mec:
## Changes in this PR
Inline with the report and activity policy this PR updates the Transaction policy. We take the same approach where we test the policy thoroughly and try to use the spec to document to developers under which conditions a transation can be added or edited.

The UI is then updated to reflect the new policy by showning add and edit UI elements only when they can be used.

I added `EDITABLE_STATES` to the Report model as we have started to use this in multiple places and it feels safer and neater to define which states are considered editable in the Report model instead of when they get used.

## Next steps

### DIFF
--- a/app/policies/transaction_policy.rb
+++ b/app/policies/transaction_policy.rb
@@ -38,6 +38,6 @@ class TransactionPolicy < ApplicationPolicy
 
   private def editable_report_for_organisation_and_fund
     activity = record.parent_activity
-    Report.find_by(organisation: activity.organisation, fund: activity.associated_fund, state: [:active, :awaiting_changes])
+    Report.find_by(organisation: activity.organisation, fund: activity.associated_fund, state: Report::EDITABLE_STATES)
   end
 end
--- a/app/policies/transaction_policy.rb
+++ b/app/policies/transaction_policy.rb
@@ -1,23 +1,43 @@
 class TransactionPolicy < ApplicationPolicy
+  def show?
+    return true if beis_user?
+    record.parent_activity.organisation == user.organisation
+  end
+
   def create?
-    Pundit.policy!(user, record.parent_activity).create? && associated_report.present?
+    return false if record.parent_activity.level.nil?
+
+    if beis_user?
+      return true if record.parent_activity.fund? || record.parent_activity.programme?
+    end
+
+    if delivery_partner_user?
+      return true if editable_report_for_organisation_and_fund.present?
+    end
+
+    false
   end
 
   def update?
-    return false if associated_report&.approved?
-    Pundit.policy!(user, record.parent_activity).update? && associated_report.present?
+    return false if record.parent_activity.level.nil?
+
+    if beis_user?
+      return true if record.parent_activity.fund? || record.parent_activity.programme?
+    end
+
+    if delivery_partner_user? && editable_report_for_organisation_and_fund.present?
+      return true if editable_report_for_organisation_and_fund == record.report
+    end
+
+    false
   end
 
   def destroy?
-    Pundit.policy!(user, record.parent_activity).destroy?
+    false
   end
 
-  private
-
-  def associated_report
-    parent_activity = record.parent_activity
-    organisation = parent_activity.organisation
-    fund = parent_activity.associated_fund
-    Report.find_by(organisation: organisation, fund: fund, state: [:active, :awaiting_changes])
+  private def editable_report_for_organisation_and_fund
+    activity = record.parent_activity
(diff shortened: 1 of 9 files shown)


app/policies/transaction_policy.rb
```diff
@@ -1,23 +1,43 @@
 class TransactionPolicy < ApplicationPolicy
+  def show?
+    return true if beis_user?
+    record.parent_activity.organisation == user.organisation
+  end
+
   def create?
-    Pundit.policy!(user, record.parent_activity).create? && associated_report.present?
+    return false if record.parent_activity.level.nil?
```
MariluzHerrera, line 8: In which cases activities would have no level? Those that are incomplete and the `form_state` is before the level step?

spec/policies/transaction_policy_spec.rb
```diff
@@ -1,75 +1,186 @@
...
+              it { is_expected.to permit_action(:create) }
+              it { is_expected.to permit_action(:edit) }
+              it { is_expected.to permit_action(:update) }
+
+              it { is_expected.to forbid_action(:destroy) }
+            end
+          end
+        end
+      end
+    end
```
MariluzHerrera, line 184: Amazing coverage of all possible scenarios on this spec!! Very clear and easy to follow

spec/features/staff/users_can_edit_a_transaction_spec.rb
```diff
@@ -67,4 +67,36 @@
...
+
+      scenario "shows the edit link" do
+        visit organisation_activity_path(activity.organisation, activity)
+
+        expect(page).to have_link t("default.link.edit"), href: edit_activity_transaction_path(activity, transaction)
+      end
+    end
+
+    context "when the transaction cannot be edited" do
+      before { report.update(state: :active) }
```
MariluzHerrera, line 93: Is the reason this transaction cannot be edited because the report is not the one associated to the activity?

app/policies/transaction_policy.rb
mec: You got it, you can hit the create activity button and we go ahead and create a new activity without a level or parent.

spec/policies/transaction_policy_spec.rb
mec: 👍 were you able to get a sense of this documenting the behaviour? I usually avoid this amount of nested contexts but I feel like this case actually warrants the complexity because there are so many factors.

spec/features/staff/users_can_edit_a_transaction_spec.rb
mec: Yep, it's not super clear here is it – any suggestions on how I might improve this?

MariluzHerrera (commented):
Hi! Sorry I didn't submit the review properly! I left some comments and a couple of questions on some code that I'm not entirely sure about (not because of the way you implement it, but due to my lack of knowledge). The rest looks really good to me, and once I fully understand those bits of code I asked about I think I'd be happy to approve. Thanks! :-)

spec/features/staff/users_can_edit_a_transaction_spec.rb
MariluzHerrera: Not sure. Maybe linking the transaction to a complete different report on the `before` block? But if feels redundant to do that because at the beginning of the bigger context you set up the variables, including the `report` one, and you link that to the correct fund and organisation. So I should've assumed that the reason why this transaction is not editable is because you're not updating the transaction to reflect this report as the report the transaction belongs to (if that makes sense?). In summary, I think it's clear enough as it is, and it was just me that I'm too junior 😅

spec/policies/transaction_policy_spec.rb
MariluzHerrera: I was. It's true that there's a fair amount of nesting but everything was so well organised that it was easy to follow. And as you said, in this case it needed to be done this way. I wouldn't change anything.

MariluzHerrera (approved):

spec/features/staff/users_can_edit_a_transaction_spec.rb
mec: Don't say that! It's a fair comment and has nothing to do with your job title! 

I think when I wrote this one I only saw it as 'testing if the edit link shows up' rather than 'testing if this user can edit this thing in this state' (as that is taken care of by the policy and other specs. So I figured as long as this spec shows that the transaction cannot be edited, it doesn't have to be as clear why.

MariluzHerrera: That makes sense. Thanks for clarifying!

PR #640, opened 2020-09-14
(1010) BEIS users can add entities to activities without a report

mec:
BEIS users should be able to add transactions and planned disbursements
to an activity when there is no editable report for the activity. Because the report
association is required they cannot as the object is considered invalid.

Make the association optional but validate the presence of the value if
the activity is for a delivery partner.

This change removes the shoulda spec for the report association as it
is now covered by the validations spec i.e. we check that the
association is set or not based on the activity organisation.

## Next steps

### DIFF
--- a/app/models/planned_disbursement.rb
+++ b/app/models/planned_disbursement.rb
@@ -5,8 +5,9 @@ class PlannedDisbursement < ApplicationRecord
   strip_attributes only: [:providing_organisation_reference, :receiving_organisation_reference]
 
   belongs_to :parent_activity, class_name: "Activity"
-  belongs_to :report
+  belongs_to :report, optional: true
 
+  validates_presence_of :report, unless: -> { parent_activity&.organisation&.service_owner? }
   validates_presence_of :planned_disbursement_type,
     :period_start_date,
     :currency,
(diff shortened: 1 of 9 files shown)


spec/models/transaction_spec.rb
```diff
@@ -14,10 +14,24 @@
     it { should validate_presence_of(:providing_organisation_type) }
     it { should validate_presence_of(:receiving_organisation_name) }
     it { should validate_presence_of(:receiving_organisation_type) }
-  end
 
-  describe "associations" do
-    it { should belong_to(:report) }
+    context "when the activity belongs to a delivery partner organisation" do
+      before { activity.update(organisation: build_stubbed(:delivery_partner_organisation)) }
```
MariluzHerrera, line 19: (TIL) I hadn't seen this `build_stubbed` used before. I've read about it now and it seems a good idea to use it here.

app/models/planned_disbursement.rb
```diff
@@ -5,8 +5,9 @@ class PlannedDisbursement < ApplicationRecord
   strip_attributes only: [:providing_organisation_reference, :receiving_organisation_reference]
 
   belongs_to :parent_activity, class_name: "Activity"
-  belongs_to :report
+  belongs_to :report, optional: true
 
+  validates_presence_of :report, unless: -> { parent_activity&.organisation&.service_owner? }
```
MariluzHerrera, line 10: I like how this is written :-)

MariluzHerrera (approved):
It looks very good, very clear. I also learnt something new (the `build_stubbed` from Factory Bot) that I could use myself in future tests, so thanks for that! :-)

spec/models/transaction_spec.rb
mec: Yeah, I wish I was better at A: remembering to use it and B: had a little more depth to my understanding of it and when to use it!

PR #642, opened 2020-09-15
(911) Users can view budgets added in a report

mec:
## Changes in this PR
As budgets do not appear in the report CSV download (or the trackers) we need a way to review the budgets that have been added in a report.

I've added a tabbed view to the report show page so the variance and budgets can live alongside each other, as per this mockup:

https://miro.com/app/board/o9J_kpLoc7Y=/?moveToWidget=3074457349383556686&cot=12

The Budget policy is an exact copy of the transaction one, I have kept the duplication as I can live with it right now, there are changes that will makes these diverge to come.

I've added a new scope to Activity with a super long name! This is used accross the two tab views so made sense to have it in a single place and as it returns Activities, this felt right.

Like the tabs on Activities, I've kept things really simple so there is some duplication of the tab markup - again, I am happy to live with this tradeoff right now.

Delivery partners can edit their budgets right from the report view, but there is no budget#show for BEIS users, I think that is good for now to get a minimum feature out.

## Screenshots of UI changes
image

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -321,6 +321,7 @@
   The comments are exported to the report CSV
 - For delivery partners, Budgets relate to the report that creates them
 - Ingest can handle the new budget/report relationship
+- Update the budget policy
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-16...HEAD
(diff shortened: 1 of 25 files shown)


lozette (approved):
One tiny speeling error but otherwise 👍  That was a lot of work for something that seems small on the surface

PR #645, opened 2020-09-16
(1015) Add the report financial quarter & year to the CSV filename

lozette:
Trello: https://trello.com/c/D4NFGkAA/1015-bug-report-csv-filename-is-empty-if-the-report-has-no-description

While testing, we found that some reports did not have descriptions, and
therefore when downloaded did not have a filename (other than `.csv`).

This commit adds the report financial quarter & year to the filename, so if the
report description is empty there is still *something* in the filename.

### DIFF
--- a/app/controllers/staff/reports_controller.rb
+++ b/app/controllers/staff/reports_controller.rb
@@ -150,7 +150,7 @@ class Staff::ReportsController < Staff::BaseController
 
   def send_csv
     response.headers["Content-Type"] = "text/csv"
-    response.headers["Content-Disposition"] = "attachment; filename=#{ERB::Util.url_encode(@report.description)}.csv"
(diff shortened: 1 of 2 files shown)


jcoglan (approved):
This is consistent with the descriptions we're automatically generating for transactions 👍

app/controllers/staff/reports_controller.rb
```diff
@@ -150,7 +150,7 @@ def approved_reports
 
   def send_csv
     response.headers["Content-Type"] = "text/csv"
-    response.headers["Content-Disposition"] = "attachment; filename=#{ERB::Util.url_encode(@report.description)}.csv"
+    response.headers["Content-Disposition"] = "attachment; filename=#{ERB::Util.url_encode(@report_presenter.financial_quarter_and_year)}-#{ERB::Util.url_encode(@report.description)}.csv"
```
mec, line 153: Do you think it is worth adding the fund title as well, that really makes the name complete for a DP without the description, we already have the `fund` in the ReportPresenter?

mec (approved):
Looks good - wondered about adding the fund title as the icing on the cake, leave that up to you.

PR #646, opened 2020-09-16
(987) Ingest additional fields from CSV

leeky:
Given a CSV with the following columns, update the relevant activities with additional information:

 - delivery_partner_identifier
 - programme_status
 - call_open_date
 - call_close_date
 - gdi
 - oda_eligibility
 - intended_beneficiaries (containing a `|`-separated list of country names)
 - total_applications
 - total_awards

**NB: The CSV dataset is *NOT* included in this PR as it contains some data that shouldn't be publicly accessible. Instead, this will be manually copied onto the server before it is manually run.**

This code exists *ONLY* for the purposes of preparing production for real users and will likely be removed once this work is completed.

### DIFF
--- a/app/services/ingest_csv.rb
+++ b/app/services/ingest_csv.rb
@@ -1,0 +1,57 @@
+require "csv"
+
+class IngestCsv
+  ACCEPTABLE_INVALID_ATTRIBUTES = [:gdi].freeze
+
+  attr_accessor :csv, :filename
+
+  def initialize(filename)
+    self.filename = filename
+    self.csv = CSV.foreach(filename, headers: true, skip_blanks: true, encoding: "bom|utf-8", header_converters: ->(h) { h.strip.downcase })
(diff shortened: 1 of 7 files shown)


app/services/ingest_csv.rb
```diff
@@ -38,6 +38,8 @@ def call
         else
           # Skip row and write message
           write_log("Skipping Activity #{row}: #{activity.errors.full_messages}")
+          write_log("attributes: #{attributes.inspect}")
+          # binding.pry
```
mec: Don't know if you are intending to keep this?

jcoglan (approved):

mec (approved):
🚢

app/services/ingest_csv.rb
leeky: Yep, it's there for a reason 👍

PR #648, opened 2020-09-17
(876) Upload transactions as CSV

jcoglan:
## Changes in this PR

This allows delivery partner users to upload transactions for a report by uploading a CSV. They work by filling out a template we give them, which contains the required column headings, and the IDs of all the activities they need to report on. The RODA ID is used to identify activities on upload, while the Delivery Partner ID and Name are there for the user's information and are ignored on upload.

This is quite a large PR but the code is mostly fairly boring validation and translation of CSV fields onto database columns, plus a little glue at the end to expose the functionality via the service. I'd recommend reading each commit rather than all the changes at once.

I'd like to check that all the logic for validation and translation the fields is correct, that the right associations between transactions, activities and reports are created, and that any permissions-related checks are correct.

## Screenshots of UI changes

When viewing a report, a new button is available in the menu to "Upload actuals":

Screenshot 2020-09-17 at 09 56 58

This links to a page where the user can download a CSV template, and upload their finished spreadsheet:

Screenshot 2020-09-17 at 09 58 16

The download gives them a CSV with the required column headings, and the IDs of each activity they need to provide actuals for in this report.

Screenshot 2020-09-17 at 09 57 52

The user fills this spreadsheet out and re-exports as CSV. If what they upload contains any errors, then the transactions are not created, and the errors are displayed.

Screenshot 2020-09-17 at 09 58 33

If they don't upload a file at all, an error message is shown:

Screenshot 2020-09-17 at 14 52 35

If all the data is valid and the transactions are saved, a success message is shown.

Screenshot 2020-09-17 at 14 54 24

## Next steps

### DIFF
--- a/app/controllers/staff/transaction_uploads_controller.rb
+++ b/app/controllers/staff/transaction_uploads_controller.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+require "csv"
+
+class Staff::TransactionUploadsController < Staff::BaseController
+  include Secured
+
+  before_action :authorize_report
+
+  def new
+    @report_presenter = ReportPresenter.new(@report)
+  end
+
+  def show
+    response.headers["Content-Type"] = "text/csv"
+    response.headers["Content-Disposition"] = "attachment; filename=transactions.csv"
+
+    response.stream.write(CSV.generate_line(csv_headers))
+    @report.reportable_activities.each do |activity|
+      response.stream.write(CSV.generate_line(csv_row(activity)))
+    end
+    response.stream.close
+  end
+
+  private def authorize_report
+    @report = Report.find(params[:report_id])
+    authorize @report, :show?
+  end
+
+  private def csv_headers
+    ["Activity Name", "Activity Delivery Partner Identifier"] + ImportTransactions.column_headings
+  end
+
+  private def csv_row(activity)
+    [
+      activity.description,
+      activity.delivery_partner_identifier,
+      activity.roda_identifier_compound,
+    ]
+  end
+end
--- a/app/controllers/staff/transaction_uploads_controller.rb
+++ b/app/controllers/staff/transaction_uploads_controller.rb
@@ -22,6 +22,25 @@ class Staff::TransactionUploadsController < Staff::BaseController
     response.stream.close
   end
 
+  def update
+    @report_presenter = ReportPresenter.new(@report)
+    rows = parse_transactions_from_upload
+
+    if rows.nil?
+      @errors = []
+      flash.now[:error] = t("action.transaction.upload.file_missing")
+      return
+    end
+
+    importer = ImportTransactions.new(report: @report, uploader: current_user)
+    importer.import(rows)
+    @errors = importer.errors
+
+    if @errors.empty?
+      flash.now[:notice] = t("action.transaction.upload.success")
+    end
+  end
+
   private def authorize_report
     @report = Report.find(params[:report_id])
     authorize @report, :show?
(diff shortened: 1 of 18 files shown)


jcoglan:
[Delivery partner users can upload a csv template to report actuals](https://trello.com/c/eGncQv5X/876-delivery-partner-users-can-upload-a-csv-template-to-report-actuals)

jcoglan:
[Delivery partner users can download a csv template for reporting actuals](https://trello.com/c/90x4g6h3/944-delivery-partner-users-can-download-a-csv-template-for-reporting-actuals)

app/controllers/staff/transaction_uploads_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def new
+    @report_presenter = ReportPresenter.new(@report)
+  end
+
+  def show
+    response.headers["Content-Type"] = "text/csv"
+    response.headers["Content-Disposition"] = "attachment; filename=transactions.csv"
+
+    csv = CSV.generate { |table|
```
leeky, line 18: Did you consider using `CSV.generate_row` so you can write the CSV out to the stream one line at a time, rather than generating the whole thing and *then* streaming it out (which removes the streaming benenfit) ?

mec (changes requested):
Wow! This is an epic piece of work! I feel pretty bad about saying anything given how much is here and how much value it adds!  Reviewing this makes me realise how far I have yet to go as a develper.

There are just two tiny things I wondered about in the comments.

app/controllers/staff/transaction_uploads_controller.rb
jcoglan: I was principally using the streaming API because this data is dynamically generated, whereas the Rails `send_file` interface only seems to like sending a file from disk, although I suppose there's nothing to stop me just using `render` normally to send a blob of text.

jcoglan: Where is `generate_row` documented? `ri` doesn't seem to know anything about it.

leeky: I misremembered - it's `generate_line` - was previously `generate_row` back in Ruby 1.8.x days!

app/controllers/staff/transaction_uploads_controller.rb
```diff
@@ -24,6 +24,17 @@ def show
     response.stream.close
   end
 
+  def update
+    importer = ImportTransactions.new(report: @report, uploader: current_user)
+    file = params[:report][:transaction_csv]
```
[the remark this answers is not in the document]
jcoglan: That's a good catch, I've pushed a test and fix for this and will update the screenshots.

app/views/staff/transaction_uploads/update.html.haml
```diff
@@ -0,0 +1,12 @@
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %h1.govuk-heading-xl
+        = t("page_title.transaction.upload")
+
+  - if @errors.empty?
+    %p.govuk-body= t("action.transaction.upload.success")
```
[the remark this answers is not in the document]
jcoglan: Thanks, I really struggle with the GOV.UK frontend stuff. Which bit did you mean to wrap -- jsut the `<p>`, or one of its containers?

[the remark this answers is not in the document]
jcoglan: Ah I've actually removed this now in favour of a flash message but I'll make the other change to wrap the error table.

app/controllers/staff/transaction_uploads_controller.rb
jcoglan: Cool, I've amended that, thank you :)

jcoglan:
@mec don't feel bad :) You've already caught some things I missed, any feedback is useful!

Also, if any of the team don't understand something I've done, I'd rather make it clearer than just leave something unmaintainable for others, so do shout if you have questions.

mec (approved):

PR #649, opened 2020-09-17
(1009) Associate all ingested Transactions, Budgets and Planned Disbursements with a report

lozette:
## Changes in this PR

Trello: https://trello.com/c/122KiJmo/1009-should-we-associate-all-the-ingested-entities-with-the-current-reports

All ingested Transactions, Budgets and Planned Disbursements need to be associated
with a report in order for them to be editable by Delivery Partners.

This data migration finds all ingested entities, then finds the "Historic data
tracker" report associated with those entities (via their associated activity)
and then assigns them to the correct report.

## Next steps

### DIFF
--- a/db/data/20200916151130_associate_all_ingested_entities_with_reports.rb
+++ b/db/data/20200916151130_associate_all_ingested_entities_with_reports.rb
@@ -1,0 +1,18 @@
+class AssociateAllIngestedEntitiesWithReports < ActiveRecord::Migration[6.0]
+  def up
+    Organisation.where(service_owner: false).each do |organisation|
+      ingested_activities = Activity.where(ingested: true, organisation: organisation)
+      ingested_activities.each do |activity|
+        fund = activity.associated_fund
+        report = Report.where(financial_quarter: nil).find_by(fund: fund, organisation: organisation)
+        activity.transactions.update_all(report: report)
+        activity.budgets.update_all(report: report)
+        activity.planned_disbursements.update_all(report: report)
+      end
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end


db/data/20200916151130_associate_all_ingested_entities_with_reports.rb
```diff
@@ -0,0 +1,18 @@
+class AssociateAllIngestedEntitiesWithReports < ActiveRecord::Migration[6.0]
+  def up
+    Organisation.where(service_owner: false).each do |organisation|
+      ingested_activities = Activity.where(ingested: true, organisation: organisation)
+      ingested_activities.each do |activity|
+        fund = activity.associated_fund
+        report = Report.where("description like ?", "%Historic tracker data%").find_by(fund: fund, organisation: organisation)
```
mec, line 7: Do we need the `like` part - it feels a bit flakey and we know there is only one report for each organisation and fund in production?

mec (commented):
I was expecting this to be way more complex! I wonder if we need the like though?

db/data/20200916151130_associate_all_ingested_entities_with_reports.rb
lozette: Yeah I wasn't sure about this either - I think we can drop it given there's only one Report per fund/org on production. I was just mindful because the ticket mentions the name of the report.

mec: Would `where(financial_quarter: nil)` help becasuse any report the app creates will have this set, these first ones are the only ones we shoul let in without it?

lozette: Yes good call, this should work 👍

mec (approved):
🚢

PR #650, opened 2020-09-17
(998) Display the parent activity's RODA Identifier

jcoglan:
When entering a RODA Identifier, there's nothing to tell the user where
they are, and the data they enter here is final, so it's both easy and
very costly to make a mistake and enter the wrong thing.

As a quick improvement, we can add the parent activity's RODA identifier
to the page, if the activity has a parent. You can only enter a RODA ID
if the parent activity has one completed, so this is guaranteed to
always have a value to display.

## Screenshots of UI changes

Screenshot 2020-09-17 at 16 18 11

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -326,6 +326,7 @@
 - Show budgets added in a report on the reports view
 - `Collaboration type` form field added to create activity journey and activity XML
 - Upload transaction data in bulk as CSV
+- Display the parent activity's RODA identifier when adding RODA IDs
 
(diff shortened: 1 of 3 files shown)


jcoglan:
[Users can see RODA ID prefix, so that they know which bit they are entering](https://trello.com/c/jDmSA5xw/998-users-can-see-roda-id-prefix-so-that-they-know-which-bit-they-are-entering)

lozette (approved):

PR #651, opened 2020-09-17
(988) Ingest new activity data from CSV

leeky:
Given a CSV, which includes a `delivery_partner_identifier` and `parent_roda_indentifier_compound` column, ingest activities.

In addition to the above, the following data columns should be provided:

- title
- description
- sector
- programme_status
- planned_start_date
- planned_end_date
- actual_start_date
- actual_end_date
- recipient_country
- flow
- aid_type
- call_open_date
- call_close_date
- gdi
- oda_eligibility
- intended_beneficiaries
- total_applications
- total_awards

NB: The existence of the `parent_roda_identifier_compound` column in the CSV causes us to create activities associated with a parent vs. the existing 'update existing activities with non-IATI fields' behaviour.

### DIFF
--- a/app/services/ingest_csv.rb
+++ b/app/services/ingest_csv.rb
@@ -62,21 +62,27 @@ class IngestCsv
       parent = Activity.find_by(roda_identifier_compound: row["parent_roda_identifier_compound"])
       return :missing_parent if parent.nil?
 
-      Activity.new(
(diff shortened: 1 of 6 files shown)


mec (approved):
Feels a bit odd given we've already used this to great effect, but 🚢 !

PR #652, opened 2020-09-17
(981) Update report state content

MariluzHerrera:
Trello: https://trello.com/c/6A0C6a5U

## Changes in this PR

The content on the text provided to the users when they are about to change the state of a report or they have just changed it has been updated following the content guidance provided by the client. This gives all users a clear indication
of what will happen and what happens next when changing the state of a report.

At each report state change we have a confirmation page and then a complete page. In order to deliver the functionality, at first we went ahead and built the feature before we had final content for the mini journeys. We are now adding the content. Some examples of final content can be found on the screenshots below.

There is a couple of test examples that needed amendment after the changes in the content. Those changes are also included in this PR, as well as the capitalisation of the extension for the acronym ODA on the heading of the app.

### After

Screenshot 2020-09-17 at 11 36 57

Screenshot 2020-09-17 at 11 46 26

Screenshot 2020-09-17 at 12 09 27

Screenshot 2020-09-17 at 12 42 03

Screenshot 2020-09-17 at 15 16 30

## Next steps

### DIFF
--- a/app/views/staff/reports_state/review/complete.html.haml
+++ b/app/views/staff/reports_state/review/complete.html.haml
@@ -1,17 +1,22 @@
-= content_for :page_title_prefix, t("page_title.report.in_review.complete", report_description: @report_presenter.description, report_financial_quarter: @report_presenter.financial_quarter_and_year)
+= content_for :page_title_prefix, t("page_title.report.in_review.complete", report_organisation: @report.organisation.name, report_financial_quarter: @report_presenter.financial_quarter_and_year)
 
 %main.govuk-main-wrapper#main-content{ role: "main" }
   .govuk-grid-row
     .govuk-grid-column-two-thirds
       .govuk-panel.govuk-panel--confirmation
         %h1.govuk-panel__title
-          = t("action.report.in_review.complete.title")
-
-        .govuk-panel__body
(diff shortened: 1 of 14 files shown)


app/views/staff/reports_state/review/complete.html.haml
```diff
@@ -1,17 +1,22 @@
...
-          = t("action.report.activate.complete.body", report_description: @report_presenter.description, report_financial_quarter: @report_presenter.financial_quarter_and_year)
+          = t("action.report.in_review.complete.title", report_financial_quarter: @report_presenter.financial_quarter_and_year, report_organisation: @report.organisation.name)
 
       %h2.govuk-heading-m
         What happens next
 
       %p.govuk-body
-        This report is now in review
+        The report is now in review with BEIS.
+        %br
```
mec, line 15: I would have just made this a new paragraph?

mec (approved):
Thank you for taking  care of this @MariluzHerrera!

I am not going to comment on the content as I know the client has reviewed and approved it themselves.

I left one tiny comment that you can decide on!

🚢

app/views/staff/reports_state/review/complete.html.haml
MariluzHerrera: I had the same thought, but I am following the way the client gave me the text, which contains a line break but not a new paragraph 🤷‍♀️

PR #661, opened 2020-09-22
(861) Sign out link not active on Users page

mec:
## Changes in this PR
It was 🐞’n me and there was a card.

## Screenshots of UI changes
Screenshot 2020-09-22 at 15 17 35

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -338,6 +338,7 @@
   import
 - Do not automatically strip letters from numeric value fields; instead reject
   the values as invalid and show an error to the user  
+- the sign out navigation link is not active on the users page
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-17...HEAD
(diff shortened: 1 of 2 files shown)


lozette (approved):

jcoglan:
If this was a bug, is it possible to add a test that confirms the expected behaviour?

mec:
> If this was a bug, is it possible to add a test that confirms the expected behaviour?

@jcoglan I guess so, feels a bit like over testing for my tastes, but happy to admit it's just taste? There is a test for the method that adds the active class https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/8048b656ca04d8bab8759a61ca777f871e4972f9/spec/helpers/application_helper_spec.rb#L16

It was just being passed the wrong url?

jcoglan:
@mec That's fair enough, I think if this was a _behaviour_ bug e.g. people not being able to log out, a test would be necessary but it's more of a judgement call for presentational concerns. Do go ahead and merge :)

PR #662, opened 2020-09-22
Reject non-numeric values in bulk transaction import

jcoglan:
Monetary values entered in the `Value` column in bulk upload are passed
through the `CreateTransaction` service, which uses `Monetize.parse` to
interpret them. This is much more permissive than we want; it allows
various currency symbols and many different kinds of formatting, whereas
we want to specifically allow this format:

- an optional `£` symbol, followed by
- an optional minus (`-`) symbol, followed by
- at least 1 digit, with comma (`,`) allowed for readability
- an optional decimal point followed by 2 digits

We also cannot just remove the Monetize call, as passing the string
directly through to ActiveRecord results in truncation; a string like
"3a45" is truncated to the value 3.0. We would like any non-numeric
characters in the input treated as an error, as they may represent a
mistake on the user's part. This is especially important in bulk upload,
where a user is adding a lot of data at once and not necessarily
checking all of it afterwards.

The ConvertFinancialValue service implements the desired format, and
ImportTransactions can then report any parsing errors back to the user.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -334,6 +334,8 @@
 ## [unreleased]
 
 - Add and amend Activity data fields in the Report CSV export
+- Accept strictly numeric values in the `Value` column for bulk transaction
+  import
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-17...HEAD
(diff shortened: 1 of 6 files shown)


lozette (approved):
LGTM! I will utilise this in the other value fields, cheers 👍

PR #664, opened 2020-09-23
Swap order of Actuals & Forecast so Forecast comes first in the CSV columns

lozette:
## Changes in this PR

Swap the order of Forecast & Actuals in the Report CSV, so that Forecast comes first.


leeky (approved):
LGTM :shipit:

PR #665, opened 2020-09-23
(1014) Do not strip alphabetical characters from values, reject the value instead

lozette:
## Changes in this PR

Trello: https://trello.com/c/ZU0SW2pm/1014-when-a-monetary-value-is-supplied-with-letters-in-it-do-not-strip-lettters-and-accept-the-numbers-fail

Previously, letters accidentally entered in the `value` fields for budgets,
transactions and planned disbursements were being stripped and the numbers
accepted.

We would prefer these entries to be rejected as an error instead of silently
stripped and passed through.

Use the `ConvertFinancialValue` to sanitize the values before they are added
to their parent entities, and show a relevant error message if any letters are
present.

## Next steps

### DIFF
--- a/app/services/convert_financial_value.rb
+++ b/app/services/convert_financial_value.rb
@@ -1,5 +1,5 @@
 class ConvertFinancialValue
-  VALUE_FORMAT = /^(?:£ *)?((?:- *)?[0-9,]+(?:\.[0-9]{2})?)$/
+  VALUE_FORMAT = /^(?:£ *)?((?:- *)?[0-9,]+(?:\.[0-9]{1,2})?)$/
 
   Error = Class.new(StandardError)
 
--- a/app/services/create_budget.rb
+++ b/app/services/create_budget.rb
@@ -9,7 +9,8 @@ class CreateBudget
     budget = Budget.new
     budget.parent_activity = activity
     budget.assign_attributes(attributes)
-    budget.value = sanitize_monetary_string(value: attributes[:value])
+
+    convert_and_assign_value(budget, attributes[:value])
 
     unless activity.organisation.service_owner?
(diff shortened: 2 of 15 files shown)


app/services/convert_financial_value.rb
```diff
@@ -1,5 +1,5 @@
 class ConvertFinancialValue
-  VALUE_FORMAT = /^(?:£ *)?((?:- *)?[0-9,]+(?:\.[0-9]{2})?)$/
+  VALUE_FORMAT = /^(?:£ *)?((?:- *)?[0-9,]+(?:\.[0-9]{1,2})?)$/
```
jcoglan, line 2: Apologies, I overlooked this as a requirement. Thanks for correcting it!

app/services/create_budget.rb
```diff
@@ -24,11 +25,16 @@ def call(attributes: {})
...
+
+  def editable_report_for_activity(activity:)
     Report.find_by(organisation: activity.organisation, fund: activity.associated_fund, state: Report::EDITABLE_STATES)
   end
 
-  private def sanitize_monetary_string(value:)
-    Monetize.parse(value)
+  def convert_and_assign_value(budget, value)
+    budget.value = ConvertFinancialValue.new.convert(value.to_s)
+    budget
```
jcoglan, line 36: What's the reason for returning the record in each of these methods? The result of `convert_and_assign_value` does not appear to be used.

lozette: Good point, it's updated in place isn't it. I've removed them.

jcoglan: Cool cool :shipit:

jcoglan (approved):

PR #669, opened 2020-09-24
(1036) BEIS Users can download Programme (Level B) xml

mec:
## Changes in this PR
The service needs to publish Programme (Level B) activities to IATI.

I first allowed the `FindProgrammeService` to optionally take a fund id and if provided, returns only the programmes for that fund and organisation - I also tidied up this class a little as it felt like it was working harder than  it needed to.

I ended up with quite a 🧙 magaical bit of code to my eye, so I am on the fence about it, the commit message explains, but basically `where` can take nothing as it's condition and not be run (TIL a 'no-op'), so we provide the fund id in a hash or nil and let Rails figure it all out. The method was pretty tough to name so I am wondering if something more approachable would be better - love to hear your views.

Then we move on to the super odd 'organisation controller that returns activity xml' - this really irks me but now doesn't feel like the time to fix it.

I refactored it a bit, story in the commit, but basically working a bit harder than it might need to checking for projects and third-party projects and in the remote channce no activities are publishable was going to return an empty xml file to the user.

With that one in the bag it was simple to work out which funds to show buttons for and provides the activities using the updated service.

Once I got a good set of activities in my DB, I was able to export the programmes as xml and validate it with IATI ✅ 

## Screenshots of UI changes

image

## Next steps

### DIFF
--- a/app/controllers/staff/organisations_controller.rb
+++ b/app/controllers/staff/organisations_controller.rb
@@ -12,20 +12,24 @@ class Staff::OrganisationsController < Staff::BaseController
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
 
-    project_activities = FindProjectActivities.new(organisation: organisation, user: current_user).call(eager_load_parent: false)
-    third_party_project_activities = FindThirdPartyProjectActivities.new(organisation: organisation, user: current_user).call(eager_load_parent: false)
+    @project_activities = iati_publishable_project_activities(
+      organisation: organisation,
+      user: current_user
+    )
+
+    @third_party_project_activities = iati_publishable_third_party_project_activities(
+      organisation: organisation,
+      user: current_user
+    )
 
     respond_to do |format|
-      format.html do
-        @project_activities = project_activities.map { |activity| ActivityPresenter.new(activity) }
-        @third_party_project_activities = third_party_project_activities.map { |activity| ActivityPresenter.new(activity) }
-      end
+      format.html
       format.xml do
         @activities = case level
         when "project"
-          project_activities.publishable_to_iati
+          @project_activities
         when "third_party_project"
-          third_party_project_activities.publishable_to_iati
+          @third_party_project_activities
         else
           []
         end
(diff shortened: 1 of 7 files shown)


app/services/find_programme_activities.rb
```diff
@@ -17,11 +18,12 @@ def call(eager_load_parent: true)
...
 
-    programmes = if organisation.service_owner
-      programmes.all
-    else
-      programmes.where(extending_organisation_id: organisation.id)
-    end
-    programmes
+    return programmes if organisation.service_owner
+
+    programmes.where(extending_organisation_id: organisation.id).where(fund_query_condition(fund_id))
```
jcoglan, line 23: If you're worried about this being too magical, another option would be to make the whole `.where` call conditional on having a `fund_id`, i.e.:

```rb
programmes = programmes.where(extending_organisation_id: organisation.id)
programmes = programmes.where(parent_id: fund_id) if fund_id.present?

programmes
```

Or, dynamically build the conditions, as in:

```rb
conditions = { extending_organisation_id: organisation.id }
conditions[:parent_id] = fund_id if fund_id.present?

programmes.where(conditions)
```
```

Comment (mec) [reply app/services/find_programme_activities.rb:23]: For the first one, does Rails run just one query?

File: app/controllers/staff/organisations_controller.rb [hunk 2]
```diff
@@ -12,13 +12,35 @@ def show
...
+
+    @project_activities = FindProjectActivities.new(
+      organisation: organisation,
+      user: current_user
+    ).call(eager_load_parent: false).publishable_to_iati
+
+    @third_party_project_activities = FindThirdPartyProjectActivities.new(
+      organisation: organisation,
+      user: current_user
+    ).call(eager_load_parent: false).publishable_to_iati
```
jcoglan, line 29: These variables are well-named, and it's probably worth pulling these queries out into methods so the `show` method becomes a high-level description of what's being displayed, rather than the implementation details of fetching that data.

app/services/find_programme_activities.rb
jcoglan: None of these snippets execute a query -- `where` does not trigger a database request on its own. The query is only executed when you attempted to read the result, e.g. by calling `each` or `first`. So you can build a query by incrementally adding `where` calls to a chain, and it effectively merges their conditions into a single query.

mec: @jcoglan thanks - that is a super fundamentail bit of knowledge to learn!

mec: @jcoglan I could not get the first way to work at all, not sure what wasn't right, I've settled on the second way, I feel like it is the most approachable.

app/controllers/staff/organisations_controller.rb
```diff
@@ -12,6 +12,8 @@ def show
 
     @organisation_presenter = OrganisationPresenter.new(organisation)
 
+    @funds = funds_for_organisation_programes(organisation_id: organisation.id)
```
lozette, line 15: Tiny thing but `programes` is misspelled here

app/controllers/staff/organisations_controller.rb
```diff
@@ -78,7 +87,21 @@ def update
     end
   end
 
-  private
+  private def funds_for_organisation_programes(organisation_id:)
```
lozette, line 90: Misspelled `programes`

lozette (approved):
Approving apart from one typo!

PR #671, opened 2020-09-25
(1013) Report the user type to analytics

mec:
The team want to use Google Analytics and understand which users are
doing what actions to help better understand how users are interacting
with the service.

As we use Google tag manager, the first step is to add the which type of
user is signed in (BEIS or non-BEIS) to the dataLayer so the tag manager
can pick it up and pass it to Google Analytics as a 'dimension'.

The principal is to create a dataLayer object before the tag manager
code is fired and add the data to that.

Although we are adding further identifiers to the tracking, I don't
believe it to be personally identifiable, I have however, made it clear
that BEIS must understand this change and any associated risk.

I originally had this as a boolean 'service_owner` but once in GA it
just would not segment the data the way the team wanted, so I ended up
using the strings in the commit.

There is a load of work in Google Tag manager and Google Analytics that goes along with this (its horrible).

Once merge we should:

- publish the changes to the staging and production environments in Tag Manager
- create the dimension in Analytics for staging and production
- create the segments in Analytics for staging and production
- confirm it is working in staging and production

## Next steps

- [x] Update Tag Manager and Analytics

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -341,6 +341,7 @@
 - the sign out navigation link is not active on the users page
 
 ## [unreleased]
+- The user type is tracked in Google Analytics
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-18...HEAD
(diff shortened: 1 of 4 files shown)


lozette (approved):
Looks good as far as I'm familiar with GA!

PR #672, opened 2020-09-25
Release 18

lozette:
- Add and amend Activity data fields in the Report CSV export
- Accept strictly numeric values in the `Value` column for bulk transaction
  import
- Do not automatically strip letters from numeric value fields; instead reject
  the values as invalid and show an error to the user
- the sign out navigation link is not active on the users page


mec (approved):
🚢

PR #673, opened 2020-09-25
Correct GDI value in Activity factory

lozette:
## Changes in this PR

Trello: https://trello.com/c/2jy2pVG2/1041-activity-factory-sets-incorrect-value-for-gdi

In https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/627
we added the GDI step to the Activity forms. The Activity factory had the
incorrect value of "No" for GDI, when it should be 1-4.

We have amended this to be "4", the equivalent of "No".

## Next steps


mec (approved):
🚢 💥 smashed it out!

PR #680, opened 2020-09-29
(1052) Intended Beneficiaries should be separated by a semicolon in report CSV

lozette:
## Changes in this PR

Trello: https://trello.com/c/ACZgsbLv/1052-in-report-csv-separate-intended-beneficiary-countries-with-a-semicolon

The Intended Beneficiaries on an activity are stored as an array. In the UI we
display the countries as "Country, Country and Country". However in the Report
CSV these need to be expressed as "Country; Country; Country".

This commit introduces an ActivityCsvPresenter which inherits from ActivityPresenter
and overrides `#intended_beneficiaries` to generate the list of countries
separated by a semicolon.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -344,6 +344,7 @@
 ## [unreleased]
 - The user type is tracked in Google Analytics
 - `providing_organisation_reference` is set when the user uploads transactions
+- Separate the list of intended beneficiaries in the report CSV with semicolons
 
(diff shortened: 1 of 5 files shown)


MariluzHerrera (approved):
Nice to see another use of `to_model` now that I understand what it does :-) 
Looks good. Thanks!

PR #683, opened 2020-10-02
Users can edit invalid activities

MariluzHerrera:
Trello: https://trello.com/c/VCBQEBMr

## Changes in this PR

Updating activities through the wizard should behave in the same way as creation - only validating attributes relevant to the current form step. Currently the wizard behaves differently when creating and updating activities:

- When creating a brand new activity, the wizard validates only the attributes that appear on the current form step, which makes sense

- When updating an activity, we validate against all validation rules, rather than those relevant to the current step.

This was fine when activities could only be entered through the wizard, as each step required the previous steps to all validate before moving on. Now, we have incomplete/invalid activities - so you can be on one page of the wizard, fix the data, and will receive an error message triggered by a validation error caused by an attribute on a completely different form step.

This PR introduces changes so when the user needs to edit an invalid activity they will be able to do so without getting any errors. Eventually, being able to fill empty activity fields that are required will make the activity valid.

To achieve this new behaviour from the Wizard, we have modified the `update_form_state` method to return only if the current step is invalid (rather than the whole activity being invalid). After introducing this change, we got 3 failing tests in relation to the geography step (`recipient_country` and `recipient_region`) and the sector step. They have a slightly different behaviour since when a user sets the geography to, for example, a country, it is required that they select what country the activity is going to be related to. And this is done on a separate Wizard step. Similar situation with Sector step: once the user selects a sector category, they need to select a specific sector within that category. This last selection happens on a different Wizard step. We have preserved that behaviour by introducing an if-elsif-else statement on `update_form_state` method 

Co-authored-by: Leeky <leeky@leeky.org.uk>

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -345,6 +345,7 @@
 - The user type is tracked in Google Analytics
 - `providing_organisation_reference` is set when the user uploads transactions
 - Separate the list of intended beneficiaries in the report CSV with semicolons
+- Users can now edit fields on invalid completed activities
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-18...HEAD
 [release-18]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-17...release-18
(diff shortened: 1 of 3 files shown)


lozette (approved):
This is mercifully less complex than I thought it might be!

PR #684, opened 2020-10-02
Release 19

lozette:
- The user type is tracked in Google Analytics
 - `providing_organisation_reference` is set when the user uploads transactions
 - Separate the list of intended beneficiaries in the report CSV with semicolons
 - Users can now edit fields on invalid completed activities


MariluzHerrera (approved):
🚀

PR #690, opened 2020-10-05
(1070) Fix: report object is valid when deadline passed

mec:
## Changes in this PR
Our Report objects can only change state if they are vaild and they were only vaild when the deadlne was not in the past. This meant after 12:00 on the day of the deadline, delivery partners could not submit their reports.

This has been fixed here by only validating the deadline when submtted the edit form.

The PR also passes a validation context based on the report state changing, this lets us keep the behaviour that a report must have a description in order to be activated without causes validation issues later.

The second commit adds a way to show Rails flashes as GOVUK error summarys - which in turns allows us to give the user the reason a report could not be activated, see the screen shots.

### Before
image

### After
image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -348,7 +348,7 @@
 - Users can now edit fields on invalid completed activities
 
 ## [unreleased]
-
+- Fix bug that prevented delivery partners from submitting a report.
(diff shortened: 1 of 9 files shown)


lozette (approved):
Nice 🚀

PR #692, opened 2020-10-06
(1075) Collect financial quarter and year on planned disbursements

mec:
## Changes in this PR
This is the first step toward making planned disbursements (forecasts) a complete feature in the application.

We based planned disbursements on IATI fields, which was fine to a point but refinement is needed now. The concept of an 'orginal' vs 'revised' planned disbursement has not been modelled in the application as yet.

To achieve this we first want to make all new planned disbursements 'original' rather than offering a choice to the user. We also want to make planned disbursements a total value for the financial quarter so we no longer need to collect the disburement date or date range. Instead we collect a financial quarter and year the same way Report is modelled.

From the financial quarter and year we can calculate the start and end date and fill those in in order to publish to IATI. For the moment we've kept the columns in the DB as we have data in production and I we would rather approach that data as a seperate concern to this work.

We agreed a first attempt at the UI for collecting the financial quarter and year and this represents that thinking. Further refinement may or may not be needed.

As we set the start and end date based on the collected financial quarter and year, we do not need to show these fields in the UI so they have been removed.

The module this work introduces it a plain old ruby object that ended up in the `lib` directory, to support this we've configured the app the eager load from this directory. Eager loading is not performed in testing so I checked that this works by setting the test environment to `config.eager_load = true` and ran the test suite.

Work will follow this to further change how planned disbursents are collected.

### Before
image

### After
image

## Next steps


mec:
[Collect financial quarter and year on planned disbursements](https://trello.com/c/nU5d2eJZ/1075-collect-financial-quarter-and-year-on-planned-disbursements)

config/application.rb
```diff
@@ -37,6 +37,8 @@ class Application < Rails::Application
       g.orm :active_record, primary_key_type: :uuid
     end
 
+    config.eager_load_paths += %W[#{config.root}/lib]
+
```
MariluzHerrera, line 41: I have a question about this line of code. Is it here where you are telling Rails to eager load from the file you created on `lib` that contains your `FinancialPeriod` module?

db/migrate/20201006091457_add_financial_quarter_and_year_to_planned_disbursements.rb
```diff
@@ -0,0 +1,21 @@
+class AddFinancialQuarterAndYearToPlannedDisbursements < ActiveRecord::Migration[6.0]
+  def change
+    add_column :planned_disbursements, :financial_quarter, :integer, index: true
+    add_column :planned_disbursements, :financial_year, :integer, index: true
+
```
MariluzHerrera, line 5: Do we want to allow `NULL` in those columns?

db/migrate/20201006091457_add_financial_quarter_and_year_to_planned_disbursements.rb
```diff
@@ -0,0 +1,21 @@
...
+      change.up do
+        planned_disbursements = PlannedDisbursement.all
+
+        planned_disbursements.each do |planned_disbursement|
+          financial_quarter = FinancialPeriod.quarter_from_date(planned_disbursement.period_start_date)
+          financial_year = FinancialPeriod.year_from_date(planned_disbursement.period_start_date)
+
+          planned_disbursement.financial_quarter = financial_quarter
+          planned_disbursement.financial_year = financial_year
+          planned_disbursement.save(validate: false)
```
MariluzHerrera, line 16: I see here about not validating this action, as you explained in the commit message. Good point.

lib/financial_period.rb
```diff
@@ -0,0 +1,24 @@
...
+      "2"
+    when 10, 11, 12
+      "3"
+    end
+  end
+
+  def self.year_from_date(date)
+    month = date.month
+    year = date.year
+    if (4..12).cover?(month)
```
MariluzHerrera, line 20: TIL about `.cover?` method. Didn't come across it before (very useful!). I like the way you implemented the logic to get the year of the forecast :-)

app/assets/stylesheets/partials/_forms.css.scss
```diff
@@ -5,3 +5,4 @@
 .date-fieldset {
   margin: 0 0 30px;
 }
+
```
MariluzHerrera, line 8: I guess this got here accidentally?

app/services/update_planned_disbursement.rb
```diff
@@ -8,6 +8,18 @@ def initialize(planned_disbursement:)
...
+        FinancialPeriod.start_date_from_quarter_and_year(attributes.fetch(:financial_quarter), attributes.fetch(:financial_year))
+      planned_disbursement.period_end_date =
+        FinancialPeriod.end_date_from_quarter_and_year(attributes.fetch(:financial_quarter), attributes.fetch(:financial_year))
+    elsif attributes.key?(:period_start_date) && attributes.key?(:period_end_date)
+      planned_disbursement.financial_quarter =
+        FinancialPeriod.quarter_from_date(attributes.fetch(:period_start_date).to_date)
+      planned_disbursement.financial_year =
+        FinancialPeriod.year_from_date(attributes.fetch(:period_start_date).to_date)
+    end
+
```
MariluzHerrera, line 17: I was wondering whether it could be worth wrapping this logic in a method or similar, since we seem to be repeating the same code in here and in the `create_planned_disbursement`. Is there a reason not to do it?

app/views/staff/planned_disbursements/new.html.haml
```diff
@@ -6,5 +6,5 @@
       %h1.govuk-heading-xl
         = t("page_title.planned_disbursement.new")
 
-        = form_with model: @planned_disbursement, url: activity_planned_disbursements_path(@activity) do |f|
-          = render partial: "form", locals: { f: f }
+      = form_with model: @planned_disbursement, url: activity_planned_disbursements_path(@activity) do |f|
+        = render partial: "form", locals: { f: f }
```
MariluzHerrera, line 10: I wasn't sure what was going on with the indentation here, so I changed it back on my local and had a look at the interface to see the difference. Well spotted!

lib/financial_period.rb
```diff
@@ -21,4 +22,30 @@ def self.year_from_date(date)
...
+  end
+
+  def self.current_year_string
+    year_from_date(Date.today).to_s
+  end
+
+  def self.next_ten_years
+    this_year = Date.today.year
+    tenth_year = this_year + 9
+    (this_year..tenth_year).step.to_a
```
MariluzHerrera, line 37: Quick question... So if you don't specify anything between brackets after `step`, Ruby will default to one whole number at a time, right?

spec/services/create_planned_disbursement_spec.rb
```diff
@@ -13,6 +13,26 @@
...
+      context "when start and end date are provided" do
+        it "sets the financial quarter and year" do
+          start_date = "1 April 2020"
+          end_date = "30 June 2020"
+          result = described_class.new(activity: activity).call(attributes: {period_start_date: start_date, period_end_date: end_date})
+          expect(result.object.financial_quarter).to eq 1
+          expect(result.object.financial_year).to eq 2020
+        end
+      end
+
```
MariluzHerrera: What would the outcome be in the case of providing a start date that's on quarter 1, for example, and an end date that is on quarter 2?

MariluzHerrera (commented):
This is a really good PR, with well organised commits that helped me to understand the whole process. I've made some comments. Most of them are questions just so I can learn and understand a bit more about things I found interesting. Thanks for that!

I also have a general question: I've seen in the Trello card that there are some values regarding planned disbursements that we can infer from somewhere else (values like currency, for example), and also other values that we would be defaulting in the backend (original vs reviewed planned disbursements), or others that we don't report anymore (such as the receiving organisation). Despite of this, the planned disbursements form still includes fields for the user to enter those values, so I was wondering if these changes (I mean, removing the unnecessary fields from the form) would come on following PRs, or maybe they need a bit of UR first?

config/application.rb
mec: Yep. We using `+=` so we are adding a path the one already configured.

lib/financial_period.rb
mec: It was new to me as well - I spent far too long trying to craft these methods and I still find them a bit odd to look at, but they do get the job done!

db/migrate/20201006091457_add_financial_quarter_and_year_to_planned_disbursements.rb
mec: I think I was worried about the data already in production, but looking at it now the migration will populate these values so I can set them to not null - thanks!

app/assets/stylesheets/partials/_forms.css.scss
mec: I did fiddle with some styles - can't remember why now! I'll clean this up!

app/views/staff/planned_disbursements/new.html.haml
mec: Yeah - it was a good spot!

lib/financial_period.rb
mec: Yep, the step here is a method call on the `(this_year..tenth_year)` which is a Range, so basically `Range.step` the docs for Range are here https://ruby-doc.org/core-2.6.2/Range.html#method-i-step you can see that when nothing is passed, step simply calls `succ` which is the equivilent of +1 each time. 

It probably works with `each` do you think that would improve this?

spec/services/create_planned_disbursement_spec.rb
mec: It's been a while since I looked at this code, but I suspect it would blow up! 😱 That said I am not sure you can provide dates that far apart through the validation rules - let me go check…

app/services/update_planned_disbursement.rb
mec: I think I see what you are saying, something to make it clearer what is going on here? I am not sure how I would approach sharing anything between the two service objects though - do you have an approach in mind?

It's a bit naughty but the `elsif` block dissappears in the next PR I have lined up if that helps ease you mind?! :)

db/migrate/20201006091457_add_financial_quarter_and_year_to_planned_disbursements.rb
mec: Yeah, okay - I tried to set this to not allow null, but the columns are added and then then the database complains that we have planned disbursements with null values and it all fails - I guess we would have to ensure the values are set first and then add the constraint after?

How do you think the database constriant differs from the Rails validations - I have these two attributes as required there always?

spec/services/create_planned_disbursement_spec.rb
mec: Okay, took me a while to get my head back into this - sorry!

This spec is for the create_planned_disbursement service object which takes the start and end dates - the code that creates the financial quarter and financial year only uses the start date so would never go wrong.

The rule is 'if we have a start and end date, we'll take the start date and calculate the financial quarter and year from that ignoring the end date'.

Worth noting that both this test and the code are removed by the end of this PR.

mec:
Thank you @MariluzHerrera, I am glad you found it interesting! 👍 

About your second point, yes indeed - there is more work following this and the next one removes all the things you mention from the UI and infer the values! It all needs rebasing ontop of this, but if you want a sneak peek, the branch is here:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/tree/feature/1056-new-forecasts-are-original

I like to keep my PRs at a managable size for the reviewer so decided to break this here - how did you find it? Could you have coped with a larger anount of change - I'd be intereseted to hear your thoughts?

db/migrate/20201006091457_add_financial_quarter_and_year_to_planned_disbursements.rb
MariluzHerrera: That's interesting. I had a very similar problem with a couple of columns I created on `activities`. They were integers as well and at first I didn't specify the not allow NULL. When I went to change it I couldn't either for the same reason as you, I think.

As you said, we are going to populate the data with a migration so it should be alright. 

About the difference in the constraints from the DB to the validations on the model, I wasn't sure but I've found this thread on StackOverflow that explains the advantages of having both: https://stackoverflow.com/questions/2367281/ruby-on-rails-is-it-better-to-validate-in-the-model-or-the-database 

Having said that, I think we tend to use more the model validation in this project. I haven't seen many `null: false` on our schema. Maybe it's something worth talking about with the wider team?

lib/financial_period.rb
MariluzHerrera: I think they look elegant!

app/services/update_planned_disbursement.rb
MariluzHerrera: Alright! That makes sense. I was surprised to find repeated code on your code because I know you are very thorough. I thought there would be a reason (and there is :-) ). 
I always struggle with scope. I still trip over things like thinking I have access to certain method or certain variable from a file just to find out that I don't. I wasn't sure you could easily share a method between services. Happy to leave it as it is then

lib/financial_period.rb
MariluzHerrera: I think using `each` wouldn't make much difference. I like having `step` here (I learned about it thanks to this line of code)

spec/services/create_planned_disbursement_spec.rb
MariluzHerrera: Okay. I think I went to the develop branch and tried to put start date and end date in different quarters and it allowed me. But as you said, in order to get the financial quarter we only need the start date. That solves the problem with the dates (plus this is outdated, so it doesn't matter much. I was just being curious, sorry!)

MariluzHerrera:
> I like to keep my PRs at a managable size for the reviewer so decided to break this here - how did you find it? Could you have coped with a larger anount of change - I'd be intereseted to hear your thoughts?

I agree with you about the size of PRs. I find very long and complex PRs a bit off-putting when it comes to reviewing them, if you know what I mean. I found this one small enough to not feel overwhelmed. I found some complexity and I've learnt a few Rails things out of it, which is great. It was manageable overall. Thanks!

MariluzHerrera (approved):
I found your explanations and rationales quite useful. Thanks for taking my comments on board and for trying to implement some of the changes I suggested (like the null in the DB one), even if they didn't work in the end. 🚀

db/migrate/20201006091457_add_financial_quarter_and_year_to_planned_disbursements.rb
mec: Interesting link! Thank you! I strikes me if this was a completely new table we could do both constraints, but with a change it can be hard as we have to ensure the data is in a fit state before we make the change - that wouldn't be easy here (but not impossible) another one we have to strike the right balance on?

Thanks for calling it out and the discussion.

spec/services/create_planned_disbursement_spec.rb
mec: Curious = good! 😄

mec:
@MariluzHerrera always a pleasure! Thanks for the review!

PR #695, opened 2020-10-09
(1066) Users can see a list of invalid activities in the index page

MariluzHerrera:
Trello: https://trello.com/c/kJRwvB56

## Changes in this PR

**Update at 22/10/2020**

We have decided to take a different approach and avoid the changes on the UI (see explanation in commit message). We are now doing a rake task that will get us the list of invalid activities we need to provide the client with. We think this is less risky as it doesn't involve changing the codebase for the UI, and less time-consuming, as we won't need to come back to revert those UI changes.

--------------------------------------------------------------------------------------------------------------------

After adding new fields to the create activity journey, and given the fact that some of these fields were mandatory, we found that there are some activities on production that were created through RODA before these fields were added. Now
these activities are invalid because they are missing the data on the mandatory new fields. 

In order to make easier for the client to identify invalid activities from a long list, we have now added a *temporary* change that will allow us to display all the invalid activities on their own list, in the activities index page. This way, the user will be able to click on each activity and amend/fill in the missing data. Once we are certain all the activities in production are valid, we will revert these changes.

During the implementation of this feature, we got a bunch of random Unoptimised query errors from the gem `Bullet`. It would throw errors when running the entire test suite but not when running the tests individually. After some investigation with senior members of the dev team we found out that some code related to Bullet whitelists introduced on different tests was conflicting with our tests (see full explanation on the first commit of this PR). The modifications on the code needed to make all the tests pass and Bullet to behave as we expect are done on this separate commit, since we probably want to keep these changes after we revert the rest.

## Next steps

### DIFF
--- a/doc/utilities.md
+++ b/doc/utilities.md
@@ -1,0 +1,20 @@
+# Invalid_activities rake task
+
+## Context
+
+We made this task to be able to provide the client with a detailed list of
+invalid activities so they could amend each of them. The information on
+each CSV row is:
+  - Activity organisation name
+  - Activity title
+  - Activity level
+  - Activity URL
+
+The purpose of this task is to be run in production, where the invalid activities
+live at the moment.
+
+## Running the task
+
+You can run this task on the console by typing `rake invalid_activities`. This
+command will run the task and, once it's finished, you will find the file
+`invalid_activities.csv` in the `tmp` folder of this project.
--- a/spec/lib/tasks/invalid_activities_spec.rb
+++ b/spec/lib/tasks/invalid_activities_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+require "csv"
+Rails.application.load_tasks
+
+describe "invalid_activities.rake" do
+  it "creates a csv file on tmp folder with a list of invalid activities in RODA" do
+    activities = create_list(:project_activity, 5)
+    activities.first.update_columns(gdi: nil, collaboration_type: nil)
+    activities.last.update_columns(title: nil, programme_status: nil)
+
+    run_task(task_name: "invalid_activities")
+    invalid_activities_from_csv = CSV.read("tmp/invalid_activities.csv")
+
+    expect(invalid_activities_from_csv.count).to eql(2)
+    expect(invalid_activities_from_csv.first).to include(activities.first.title)
+    expect(invalid_activities_from_csv.first).to include(activities.first.organisation.name)
+    expect(invalid_activities_from_csv.first).to include(activities.first.level)
+    expect(invalid_activities_from_csv.first).to include(Rails.application.routes.url_helpers.organisation_activity_details_url(activities.first.organisation, activities.first.id, host: ENV["DOMAIN"]))
+  end
+end
+
+def run_task(task_name:)
+  Rake::Task[task_name].invoke
+end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -63,6 +63,7 @@ Rails.application.configure do
     Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :organisation
     Bullet.add_whitelist type: :unused_eager_loading, class_name: "Transaction", association: :provider
     Bullet.add_whitelist type: :unused_eager_loading, class_name: "Transaction", association: :receiver
+    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :parent
   end
 
   config.hosts = [
--- a/lib/tasks/invalid_activities.rake
+++ b/lib/tasks/invalid_activities.rake
@@ -1,0 +1,14 @@
+require "csv"
+
+desc "Creates a csv file with a list of completed but invalid activities in RODA"
+task invalid_activities: :environment do
+  activities = Activity.where(form_state: "complete")
+  invalid_activities_array = activities.reject { |activity| activity.valid? }
+
+  CSV.open("tmp/invalid_activities.csv", "wb") do |csv|
+    invalid_activities_array.each do |activity|
+      activity_url = Rails.application.routes.url_helpers.organisation_activity_details_url(activity.organisation, activity, host: ENV["DOMAIN"])
+      csv << [activity.organisation.name, activity.title, activity.level, activity_url]
+    end
+  end
+end
--- a/spec/services/find_programme_activities_spec.rb
+++ b/spec/services/find_programme_activities_spec.rb
@@ -8,14 +8,6 @@ RSpec.describe FindProgrammeActivities do
   let!(:extending_organisation_programme) { create(:programme_activity, extending_organisation: other_organisation) }
   let!(:other_programme) { create(:programme_activity) }
 
-  before(:all) do
-    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Activity", association: :parent
-  end
-
-  after(:all) do
-    Bullet.delete_whitelist(type: :unused_eager_loading, class_name: "Activity", association: :parent)
-  end
-
   describe "#call" do
     it "eager loads the organisation and parent activity" do
       expect_any_instance_of(ActiveRecord::Relation)
--- a/spec/services/find_project_activities_spec.rb
+++ b/spec/services/find_project_activities_spec.rb
@@ -8,14 +8,6 @@ RSpec.describe FindProjectActivities do
   let!(:organisation_project) { create(:project_activity, organisation: other_organisation) }
   let!(:other_project) { create(:project_activity) }
 
-  before(:all) do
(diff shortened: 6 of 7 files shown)


mec (dismissed):
This is a super interesting PR for me! The work looks good, I have a couple of questions out of the gate if that is okay:

- I see the value adding this, but it makes me worry when we add something with the intent of taking it out again - we rarely get the time to do so. 

As the motivation for this is to help the team fix acitivites that are in production in certain conditions, did we consider a rake task that can be run that spits back the invalid activities and perhaps links to them? This would give the team almost a 'check list' to work through to fix the problems, be global across the app, supplied to BEIS as a spreadsheet and mean we don't have to revert anything?

- Bullet always makes me nervous, I think becasue I don't understand it enough! I do wonder if as we are looking at fixing these bullet specs under the right light? To my knowledge they are used to test a set of 'FindActivities' services that I am unconvinced about as a pattern and which form a quick look, we only use in one controller? Do we think there might be value in refactoring the controller that uses the services and removing them entirely?

I think the fact no one else has attempted to use them shows one of their weaknesses? I actually remember the work that added them and being concerned about this bullet issue then:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/494#discussion_r448853183

There is a lot here so happy to talk about this if preferred.

jcoglan:
@mec My commit message explains why we were observing weird effects from Bullet in terms of what it's doing behind the scenes. I discussed this with @MariluzHerrera and my decision went as follows.

Using `Bullet.add_whitelist` in _both_ the environment config and in individual tests is not advisable since it can lead to conflicts, where a test turns a particular whitelisted thing on and off, changing the base state of the app's config, and those are hard to debug, as we've seen here. The `Bullet.{add,delete}_whitelist` API does not seem designed for this type of use as it does not provide a reliable way of restoring whatever the prior state was. We need to pick one or the other.

Putting it in individual tests means people would need to add `Bullet.{add,delete}_whitelist` config to any test that hits a problematic controller action. This particular action is hit by a lot of tests, and it won't be obvious that one needs to add this config in order to silence the warnings Bullet produces, or even if that is a good idea. Getting warnings about code you didn't write is confusing.

Putting it in the environment config means we're globally ignoring a certain kind of optimisation problem -- you cannot. In general, I would rather be alerted to these, but I think ignoring an unused eager load is much less dangerous than ignoring an N+1 query, so I'm happy with the trade-off. In this case, the unused eager load comes from the fact that not all `Activity` objects use the `parent` association during validation -- `fund` activities don't require a parent. The fact that some but not all loaded activities use the `parent` association causes Bullet to raise an error _whether you eager load the association or not_ -- it's impossible to make this code pass Bullet inspection.

I'm open to making the opposite decision and putting Bullet whitelist logic in individual tests, and not in the environment config, I just think we need to pick one or the other.

mec:
@jcoglan I didn't think I was proposing removing the global bullet config, that must be poor communication on my part, apologies, I'll try again:

I stand by my first comment about having a task instead of UI that meets this need and am interested in the motivation - I wasn't part of the discussion and so am interested to learn.

These find activities services we have that are covered here don't seem to do much and I wondered about removing them entirely - it was more a question for my own learning:

We have these four find activitiy services that are not used much and look like they could be a scope on activity, perhaps I have missed something that means they can't be and I guess that is the motivation for my question - trying to understand why we have these service objects, that had bad bullet config and that since we surfaced them him, could we take this opportunity to remove them?

As usual, happy to talk as clearly I am not great at communicating here, but also aware that here can work better for other folks.

MariluzHerrera:
> 
> I stand by my first comment about having a task instead of UI that meets this need and am interested in the motivation - I wasn't part of the discussion and so am interested to learn.
> 

Regarding this point, I don't remember being a specific reason to make changes on the app instead of a rake task for example. Do you, @jcoglan? I just simply didn't consider that could be an option. I guess I can ask Dan whether a list of invalid activities with their URLs would be equally useful, because I agree that the changes in this PR require having to come back and revert them, whereas the list won't require any changes on the codebase neither now nor later.

spec/lib/tasks/invalid_activities_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+require "csv"
+Rails.application.load_tasks
+
+describe "invalid_activities.rake" do
+  it "creates a csv file on tmp folder with a list of invalid activities in RODA" do
+    activities = create_list(:project_activity, 5)
+    activities.first.update_columns(gdi: nil, collaboration_type: nil)
```
mec, line 8: I'm interested to learn what led to you choosing `update_columns` here, I've never seen it?!

spec/lib/tasks/invalid_activities_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    activities.last.update_columns(title: nil, programme_status: nil)
+
+    run_task(task_name: "invalid_activities")
+    invalid_activities_from_csv = CSV.read("tmp/invalid_activities.csv")
```
mec, line 12: Do we need to worry about ordering of the activities? I assume `CSV.read` retains the order so the first record is always the one we expect?

mec (approved):
This looks good, I feel loads better about the simplicity of this task vs adding temporary changes in the application.

I've approved this and left some questions that I don't think block it, I would really appreciate a tiny bit of documentation so our future selves know this task exists, it doesn't need much, perhaps a `utilities.md` in our doc folder with a couple of sentances about this and how to use it, what do you think?

spec/lib/tasks/invalid_activities_spec.rb
MariluzHerrera: I learnt about `update_columns` recently too (thanks to Leeky). Here I try to modify two activities to make them invalid by changing the values of some mandatory fields to `nil`. I tested on Rails console first, and the usual methods for updating these fields were failing when I tried to save the activity (the validation rules made these changes fail and the DB got rollback). `update_columns` will allow me to change the values of those columns because you do it in the database directly, bypassing validations and callbacks.

https://apidock.com/rails/ActiveRecord/Persistence/update_columns

spec/lib/tasks/invalid_activities_spec.rb
MariluzHerrera: That was my assumption too. When I run this on the console, with mock data it always kept activities in the same order.

MariluzHerrera:
> This looks good, I feel loads better about the simplicity of this task vs adding temporary changes in the application.
> 
> I've approved this and left some questions that I don't think block it, I would really appreciate a tiny bit of documentation so our future selves know this task exists, it doesn't need much, perhaps a `utilities.md` in our doc folder with a couple of sentances about this and how to use it, what do you think?

Thank you for your feedback! I agree this rake task is less messy than changing the UI. 
I think documentation is a good idea. Never seen a `utilities.md` before. Do we have any examples somewhere? I'd like to get an idea of how they're written, and what information is relevant (I'm not an expert on rake tasks 😊 )

mec:
> I think documentation is a good idea. Never seen a `utilities.md` before. Do we have any examples somewhere? I'd like to get an idea of how they're written, and what information is relevant (I'm not an expert on rake tasks 😊 )

I just made it up, I don't think it's a pattern!? We have a `doc` directory where we put dev documentation and this doesn't feel like it fits in any of the existing files so I figured why not create something? I don't think we need loads, just put ourselves in the positiion of someone who has been asked to get a list of invalid activities by the client 'like we've done before' and write out the steps they need to run the rake task and get the data?

MariluzHerrera:
@mec What do you think of the documentation? Did I miss something or did I over-explain anything? If you're still happy with this PR (and the documentation), I'll merge it

doc/utilities.md
```diff
@@ -0,0 +1,24 @@
+# Invalid_activities rake task
+
+## Context
+
+`invalid_activities` is a rake task that produces a CSV file that contains
```
mec, line 5: This paragraph feels more like a commit message for me, not sure we need it?

mec (commented):
@MariluzHerrera I would suggest loosing the first paragraph, otherwise 👍

doc/utilities.md
MariluzHerrera: Okay. I wasn't sure how much of an explanation of this rake task to give to future devs. I agree that is repetitive, considering that the commit message of the task is quite similar. I'll remove it and merge the PR if that's okay?

jcoglan:
> Regarding this point, I don't remember being a specific reason to make changes on the app instead of a rake task for example. Do you, @jcoglan?

The main reason I had in mind was to let end users get at this information directly rather than needing to ask devs to prepare a fresh list for them, but the latter might be a workable solution.

PR #702, opened 2020-10-23
(1056) New forecasts only collect the required information

mec:
## Changes in this PR
Following on from #692, we remove the fields from the form that collects values we can infer ourselves:

- planned disbursement type is always original (when creating an new one)
- currency is always GBP
- providing organisation is always BEIS
-  we do not collect a recieving organisation

There is a refactor here to store the planned_disbursement_type as an enum, this gives us nice helper methods to use with planned disbursements like `original?` that makes the code a little more readable.

### Keeping the columns
The general principal here is to not remove the columns of data we are now infering, the main reason being we might have values for them in production and it seems wise to keep those for moment.

### Before
image

### After
image

## Next steps

### DIFF
--- a/app/models/planned_disbursement.rb
+++ b/app/models/planned_disbursement.rb
@@ -2,7 +2,7 @@ class PlannedDisbursement < ApplicationRecord
   include PublicActivity::Common
   PLANNED_DISBURSEMENT_BUDGET_TYPES = {"1": "original", "2": "revised"}
 
-  strip_attributes only: [:providing_organisation_reference, :receiving_organisation_reference]
+  strip_attributes only: :receiving_organisation_reference
 
   belongs_to :parent_activity, class_name: "Activity"
   belongs_to :report, optional: true
@@ -17,7 +17,8 @@ class PlannedDisbursement < ApplicationRecord
     :receiving_organisation_name,
     :receiving_organisation_type,
     :financial_quarter,
-    :financial_year
+    :financial_year,
+    :providing_organisation_reference
   validates :value, inclusion: {in: 0.01..99_999_999_999.00}
 
   def unknown_receiving_organisation_type?
--- a/app/models/planned_disbursement.rb
+++ b/app/models/planned_disbursement.rb
@@ -2,8 +2,6 @@ class PlannedDisbursement < ApplicationRecord
   include PublicActivity::Common
   PLANNED_DISBURSEMENT_BUDGET_TYPES = {"1": "original", "2": "revised"}
 
-  strip_attributes only: :receiving_organisation_reference
-
   belongs_to :parent_activity, class_name: "Activity"
   belongs_to :report, optional: true
 
@@ -14,11 +12,9 @@ class PlannedDisbursement < ApplicationRecord
     :value,
     :providing_organisation_name,
     :providing_organisation_type,
-    :receiving_organisation_name,
-    :receiving_organisation_type,
+    :providing_organisation_reference,
     :financial_quarter,
-    :financial_year,
-    :providing_organisation_reference
+    :financial_year
   validates :value, inclusion: {in: 0.01..99_999_999_999.00}
 
   def unknown_receiving_organisation_type?
--- a/app/models/planned_disbursement.rb
+++ b/app/models/planned_disbursement.rb
@@ -17,8 +17,4 @@ class PlannedDisbursement < ApplicationRecord
     :financial_quarter,
     :financial_year
   validates :value, inclusion: {in: 0.01..99_999_999_999.00}
-
-  def unknown_receiving_organisation_type?
-    receiving_organisation_type == "0"
-  end
 end
--- a/app/models/planned_disbursement.rb
+++ b/app/models/planned_disbursement.rb
@@ -1,6 +1,7 @@
 class PlannedDisbursement < ApplicationRecord
   include PublicActivity::Common
-  PLANNED_DISBURSEMENT_BUDGET_TYPES = {"1": "original", "2": "revised"}
+
+  enum planned_disbursement_type: {original: "1", revised: "2"}
 
   belongs_to :parent_activity, class_name: "Activity"
   belongs_to :report, optional: true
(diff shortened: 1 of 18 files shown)


app/models/planned_disbursement.rb
```diff
@@ -2,7 +2,7 @@ class PlannedDisbursement < ApplicationRecord
   include PublicActivity::Common
   PLANNED_DISBURSEMENT_BUDGET_TYPES = {"1": "original", "2": "revised"}
 
-  strip_attributes only: [:providing_organisation_reference, :receiving_organisation_reference]
+  strip_attributes only: :receiving_organisation_reference
```
MariluzHerrera: I hadn't seen `strip_attributes` before (TIL). Nice one!

app/models/planned_disbursement.rb
```diff
@@ -17,7 +17,8 @@ class PlannedDisbursement < ApplicationRecord
     :receiving_organisation_name,
     :receiving_organisation_type,
     :financial_quarter,
-    :financial_year
+    :financial_year,
+    :providing_organisation_reference
```
MariluzHerrera: I see we validate the presence of name, and type, but we didn't validate the presence of `providing_organisation_reference` before. Out of curiosity, was there a reason for that?

app/services/create_planned_disbursement.rb
```diff
@@ -36,11 +40,13 @@ def call(attributes: {})
     result
   end
 
-  private
-
-  def convert_and_assign_value(planned_disbursement, value)
+  private def convert_and_assign_value(planned_disbursement, value)
```
MariluzHerrera, line 43: I see you tend to change the way we define private methods from indicating `private` on a line of code, followed by a list of methods underneath that are understood to be private. You put `private def` on every method. I guess it's a matter of preference. Since I was taught the other way during my bootcamp, I'd like to have a different point of view on whether you think using `private def` is better/more readable, if you don't mind sharing your opinion?

config/locales/models/planned_disbursement.en.yml
```diff
@@ -9,21 +9,18 @@ en:
...
       planned_disbursement:
-        providing_organisation: Providing organisation
         receiving_organisation: Receiving organisation
     hint:
       planned_disbursement:
-        providing_organisation: The organisation where this planned disbursement is coming from
-        providing_organisation_reference_html: For example, GB-GOV-13. To lookup codes or for more infomation see <a href="http://org-id.guide/" target="_blank" class="govuk-link">the organisation finder service (Opens in new window)</a>
+        receiving_organisation: Receiving organisation
+    hint:
+      planned_disbursement:
```
MariluzHerrera, line 19: This file now looks this way: 
`hint:
      planned_disbursement:
        receiving_organisation: Receiving organisation
    hint:
      planned_disbursement:
        period_end_date: End date
        period_start_date: Start date
    hint:
      planned_disbursement:`

Is this intentional? Would it be better to have all the hints grouped together?

app/models/planned_disbursement.rb
```diff
@@ -14,11 +12,9 @@ class PlannedDisbursement < ApplicationRecord
...
-    :receiving_organisation_name,
-    :receiving_organisation_type,
+    :providing_organisation_reference,
     :financial_quarter,
-    :financial_year,
-    :providing_organisation_reference
+    :financial_year
   validates :value, inclusion: {in: 0.01..99_999_999_999.00}
 
   def unknown_receiving_organisation_type?
```
MariluzHerrera: If we don't collect `receiving_organisation` anymore. Do we want to keep this method?

MariluzHerrera: Nevermind. I just saw you removed it on your last commit :-)

app/models/planned_disbursement.rb
mec: Generally speaking, the IATI reference of an organisation is optional, whilst IATI recommend providing one as it uniquely identifies an organisation, it can be hard to provide one.  Prior to this work we pre-filled in the providing org with BEIS but we thought there might be cases where it was not BEIS and the user may not be able to provide a reference in that case so it was optional.

Now we just set the value to BEIS we know we'll get the reference so we can expect every record to have one.

I feel like we are in a bit of an odd place right now, in theory we could drop all the columns where we only use the values to pusblish to IATI but the value is always the same, I think I prefer this odd place for now as we have data in these columns and I would hate to loose that, just in case.

app/services/create_planned_disbursement.rb
mec: LOL! It's funny you say that! This is exactly what I said to the dev that I first saw this as well! 😄 

I used the single `private` approach before this project, we have both all over this code base, the dev using the single private explained their reasoning and I couldn't argue with it (with a larger number of private methods you can loose sight of which methods are private) and I made the decision to make a attempt at consistency across the code and settled on the multiple privates – I wish standardrb declared a rule for this so we don't have to think about it, but it doesn't and I am not sure if it can and just haven't had time to check!

config/locales/models/planned_disbursement.en.yml
mec: Good spot! Thank you!

MariluzHerrera (approved):
Looks good to me! Thanks for answering my questions :-) 
🚀

mec:
@MariluzHerrera I've updated two of the specs to use some nice rspec style in cd22a2b04c32541d34a82e6fcffaf53ab893e41e are you still okay to approve?

app/services/create_planned_disbursement.rb
```diff
@@ -11,6 +11,8 @@ def call(attributes: {})
 
     planned_disbursement.parent_activity = activity
     planned_disbursement.assign_attributes(attributes)
+    planned_disbursement.planned_disbursement_type = "1"
```
leeky, line 14: FYI you can pass a symbol i.e. `:original` or `:revised` when setting this which is a little more readable.

MariluzHerrera:
> @MariluzHerrera I've updated two of the specs to use some nice rspec style in [cd22a2b](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/cd22a2b04c32541d34a82e6fcffaf53ab893e41e) are you still okay to approve?

Thanks for the heads-up! I've seen the changes and they look good. Also, if you want to follow Leeky's suggestion about the symbols that's fine for me. Little perks from `enum` :-)

PR #704, opened 2020-10-26
(1099) GDI options have been modified

MariluzHerrera:
Trello: https://trello.com/c/IidBaA1Y

## Changes in this PR

This PR follows a request from the client. They needed to include one option on GDI for activities that benefit either India or China (or both) but that were created before the GDI policy existed.

Likewise, the option `No` is now presented as `GDI not applicable` as requested by the client. This will be the option users can select for activities benefitting any of the other countries (no China or India).

Specs and factories have also been modified to reflect these changes. There is one spec related to the ingest work, which modification I decided to add on a separate commit. The reason for this is that this work was done by another developer from the team, and now I have introduced an option for `GDI not applicable`, which could conflict with the values this developer created for nil, which are `not applicable` or `n/a`. Ideally I'd like this to be reviewed by that developer or someone senior who is familiar with the ingest work.

### After

Screenshot 2020-10-26 at 16 31 46

## Next steps

### DIFF
--- a/config/locales/codelists/2_03/iati.en.yml
+++ b/config/locales/codelists/2_03/iati.en.yml
@@ -43,7 +43,8 @@ en:
       '1': Yes - China
       '2': Yes - India
       '3': Yes - China and India
-      '4': 'No'
+      '4': GDI not applicable
+      '5': No - This activity benefits China or India but GDI does not apply
     geography:
       recipient_country: Country
       recipient_region: Region
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -362,6 +362,7 @@
   retained for existing records
 - Planned disbursements do not include receiving organisation in the IATI xml
   export
+- Answers for GDI form step have been modified
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-20...HEAD
 [release-20]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-19...release-20
--- a/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
+++ b/spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
@@ -257,7 +257,7 @@ RSpec.feature "Users can create a fund level activity" do
         click_button t("form.button.activity.submit")
         expect(page).to have_content t("activerecord.errors.models.activity.attributes.gdi.blank")
 
-        choose "No"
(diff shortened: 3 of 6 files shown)


config/locales/codelists/2_03/iati.en.yml
```diff
@@ -43,7 +43,8 @@ en:
       '1': Yes - China
       '2': Yes - India
       '3': Yes - China and India
-      '4': 'No'
+      '4': No - This activity benefits China or India but GDI does not apply
```
leeky, line 46: Isn't this change splitting the single "No" answer into two?

A number of existing activities that have GDI set to `4` should now presumably be `5`, but we have no way of knowing.

leeky (dismissed):
LGTM! Just have concerns about the existing activities.

config/locales/codelists/2_03/iati.en.yml
MariluzHerrera: It is kind of a split. It defines the "No" answer better. Option 4 will be for activities that benefit China/India but were created before the GDI policy existed (not sure about when that policy was introduced). Option 5 is for those activities that benefit any other countries that aren't China or India. 

It is a very good point and a good reason to be concerned. Assuming users stated "No" on activities benefitting other countries than China or India, I could swap the values for 4 and 5. But if they used the former option "No" for activities benefitting China or India _before_ the GDI policy, these activities would be allocated the wrong code :/

leeky: Probably worth flagging with @mec just to be on the safe side.

mec: It's a good point, thanks for raising it. 

It looks like this is the field where we ingested 'replace me' - is that correct?

As with most cases like this BEIS have introduce the new response and will have to decide how to handle the transition of data - I'll try and pick this up with them this morning.

Let's hold this PR until we can tease a little more out of BEIS - ultimately they will have to decide if they can accept the blanket no as I don't think there is any other option right now.

MariluzHerrera: The decision made by the client is to map any existing "No" option to "GDI not applicable". For this, I have decided to swap the values of codes 4 and 5. This saves us a migration, since existing activities with value "No" (code 4), will now keep their code, which equals to "GDI not applicable". I've pushed the new code again and I'll be requesting a re-review

mec (approved):
I think the switching of of the codes is a good move right now. 👍

PR #706, opened 2020-10-27
(1112) Update aid type

CristinaRO:
## Changes in this PR

https://trello.com/c/B6JVmnyT/1112-update-aid-type
https://trello.com/c/DS4wCbIn/868-update-aid-type-radio-buttons

- Remove unused aid types.
- Replace the hints in the aid type form with shorter, more accessible copy.
- Rename the aid activity with more user-friendly text

### After
Screenshot 2020-10-27 at 16 02 33

## Next steps

### DIFF
--- a/app/helpers/codelist_helper.rb
+++ b/app/helpers/codelist_helper.rb
@@ -2,6 +2,16 @@
 
 module CodelistHelper
   DEVELOPING_COUNTRIES_CODE = "998"
+  ALLOWED_AID_TYPE_CODES = [
+    "B02",
+    "B03",
+    "C01",
+    "D01",
+    "D02",
+    "E01",
+    "G01",
+  ]
+
   def yaml_to_objects(entity:, type:, with_empty_item: true)
     data = load_yaml(entity: entity, type: type)
     return [] if data.empty?
@@ -111,7 +121,9 @@ module CodelistHelper
   end
 
   def aid_type_radio_options
-    yaml_to_objects_with_description(entity: "activity", type: "aid_type", code_displayed_in_name: true)
+    yaml_to_objects_with_description(entity: "activity", type: "aid_type", code_displayed_in_name: true).select { |a|
+      ALLOWED_AID_TYPE_CODES.include?(a.code)
+    }
   end
 
   def load_yaml(entity:, type:)
--- a/app/helpers/codelist_helper.rb
+++ b/app/helpers/codelist_helper.rb
@@ -34,7 +34,9 @@ module CodelistHelper
 
     data = data.collect { |item|
       name = code_displayed_in_name ? "#{item["name"]} (#{item["code"]})" : item["name"]
-      OpenStruct.new(name: name, code: item["code"], description: item["description"])
+      description = t("form.label.#{entity}.#{type}.#{item["code"]}", default: item["description"])
+
+      OpenStruct.new(name: name, code: item["code"], description: description)
     }
 
     data.sort_by(&:code)
@@ -121,9 +123,13 @@ module CodelistHelper
   end
 
   def aid_type_radio_options
-    yaml_to_objects_with_description(entity: "activity", type: "aid_type", code_displayed_in_name: true).select { |a|
-      ALLOWED_AID_TYPE_CODES.include?(a.code)
-    }
+    options = yaml_to_objects_with_description(
+      entity: "activity",
+      type: "aid_type",
+      code_displayed_in_name: true
+    )
+
+    options.select { |a| ALLOWED_AID_TYPE_CODES.include?(a.code) }
   end
 
   def load_yaml(entity:, type:)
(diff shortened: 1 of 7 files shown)


CristinaRO:
[Update Aid type](https://trello.com/c/B6JVmnyT/1112-update-aid-type)

vendor/data/codelists/IATI/2_03/activity/aid_type.yml
```diff
@@ -4,37 +4,6 @@ attributes:
   name: AidType
   complete: '1'
 data:
-  - category: A
```
leeky, line 7: These codelists are maintained by IATI and are occasionally updated - probably not for this PR, but I'm wondering if there's any value in these `.yml` files keeping the complete data set, and then us using some other method to hide certain options from the user?

Whilst not applicable in this case, I've seen errors caused by the database containing values that *previously* were valid and are no longer...

spec/support/form_helpers.rb
```diff
@@ -197,7 +197,7 @@ def fill_in_activity_form(
     click_button t("form.button.activity.submit")
 
     expect(page).to have_content t("form.legend.activity.aid_type")
-    expect(page).to have_content "A code for the vocabulary aid-type classifications. International Aid Transparency Initiative (IATI) descriptions can be found here (Opens in new window)"
+    expect(page).to have_content t("form.hint.activity.aid_type")
```
leeky, line 200: 👍

leeky (approved):
LGTM! :shipit: 

Just needs a fix to pass the linter.

vendor/data/codelists/IATI/2_03/activity/aid_type.yml
CristinaRO: I don't know! I would appreciate some guidance about the best way to approach this within the time constraints. @mec ?

mec: I think we should keep anything in the `vendor` directory intact. 

If you are looking for an approach, perhaps we could set a const that stores the aid types BEIS want to use and use that to produce the list shown to the user in the UI, this along with a helper that can add the custom descriptions (from the locales we already have for aid types) feels like it would work?

mec:
@CristinaRO has had to go onto support so we will pick this up in her abscence, I think the comments still stand for whomever does the work so we'll leave this PR in draft.

MariluzHerrera:
I was checking how it looks on the UI and I saw this on the activity details page: 

Screenshot 2020-11-05 at 15 10 45

pezholio:
Yikes! Will look at that

pezholio:
@MariluzHerrera Fixed that!

app/helpers/codelist_helper.rb
```diff
@@ -24,7 +34,9 @@ def yaml_to_objects_with_description(entity:, type:, code_displayed_in_name: fal
 
     data = data.collect { |item|
       name = code_displayed_in_name ? "#{item["name"]} (#{item["code"]})" : item["name"]
-      OpenStruct.new(name: name, code: item["code"], description: item["description"])
+      description = t("form.label.#{entity}.#{type}.#{item["code"]}", default: item["description"])
```
MariluzHerrera, line 37: I'm not sure I understand why we have this translation here. I see it takes the values from `activity.en.yml` form.label.activity.aid_type. Is there an advantage of doing it this way versus what we did before in this method? 
Descriptions tend to be stored on their own files inside vendor (as in `aid_type.yml` inside the vendor folder), not here. That's why I'm a bit confused.

I see you already filter the options on the `aid_type_radio_options` to show only the ones in the `ALLOWED_AID_TYPE_CODES` const, which looks good to me. Having those same options on `activity.en.yml` feels redundant? Also, on `form.label.activity.aid_type` there is one type missing that is in the const (`E01`). Sorry for the brain dump here, I just feel I'm missing the key behind all this so I thought I'd better ask 😅

MariluzHerrera (commented):
I've tried to leave a couple of questions on the code, after explaining myself horribly. Sorry about that. Let me know if there's anything I said that you don't understand. Happy to try to explain it in a better way :-)

app/helpers/codelist_helper.rb
pezholio: The reason I did it this way was because I thought the ambition was that we left the vendored files alone. See the [discussion here](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/706#discussion_r513534024). If it's OK to alter the descriptions, then yeah, I think changing the description in the vendored files would be a neater solution.

MariluzHerrera: That's true. I forgot about the custom descriptions being a change on the vendored file. Sorry about that.
I've also seen that the reason `E01` is not on `activity.en.yml` is because we don't use a custom description for that code. That all makes sense now. Thanks.

In that case I think it's fine as it is since we want to keep the vendor intact. The only thing I would suggest is to correct a couple of missing spaces between words in the custom description of `D01` (there's a `volunteersand`, and a `developingcountries`), and it should be good to go. Thank you @pezholio!

MariluzHerrera (approved):
🚀

PR #708, opened 2020-10-30
(1147) Only allow one original planned disbursement

mec:
## Changes in this PR
There should only ever be one original planned disbursement for a given
activity, financial quarter and year. This commit enforces that with a
validation on planned disbursement type that only apples when creating a
new planned disbursement.

This also meant updating a number of specs that are creating multiple
planned disbursements in the same financial quarter.

As there is no field in the UI to link an error to we add the error to
the base and use the GOVUK form builders link_base_errors_to option to
make the link the financial quarter, however there seems to be a bug
with this that the id for the link is not generated correctly, it is
missing the `--error` part- I don't feel that this should block this
work however.

## Next steps

### DIFF
--- a/app/models/planned_disbursement.rb
+++ b/app/models/planned_disbursement.rb
@@ -6,6 +6,8 @@ class PlannedDisbursement < ApplicationRecord
   belongs_to :parent_activity, class_name: "Activity"
   belongs_to :report, optional: true
 
+  validate :only_one_original, on: :create
+
   validates_presence_of :report, unless: -> { parent_activity&.organisation&.service_owner? }
   validates_presence_of :planned_disbursement_type,
     :period_start_date,
@@ -17,4 +19,10 @@ class PlannedDisbursement < ApplicationRecord
     :financial_quarter,
     :financial_year
   validates :value, inclusion: {in: 0.01..99_999_999_999.00}
+
+  def only_one_original
+    if PlannedDisbursement.find_by(financial_quarter: financial_quarter, financial_year: financial_year, parent_activity: parent_activity, planned_disbursement_type: :original).present?
+      errors.add(:base, I18n.t("activerecord.errors.models.planned_disbursement.attributes.planned_disbursement_type.only_one_original", financial_quarter: financial_quarter, financial_year_start: financial_year, financial_year_end: financial_year + 1))
+    end
+  end
 end
--- a/app/views/staff/planned_disbursements/_form.html.haml
+++ b/app/views/staff/planned_disbursements/_form.html.haml
@@ -1,4 +1,4 @@
-= f.govuk_error_summary
+= f.govuk_error_summary link_base_errors_to: :financial_quarter
 
 = f.govuk_fieldset legend: { text: nil } do
   .govuk-grid-row
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -1045,22 +1045,21 @@ RSpec.describe Activity, type: :model do
 
   describe "#forecasted_total_for_report_financial_quarter" do
     it "returns the total of all the activity's planned disbursements scoped to a report's financial quarter only" do
-      project = create(:project_activity, :with_report)
-      report = Report.find_by(fund: project.associated_fund, organisation: project.organisation)
+      report = create(:report, financial_quarter: 1, financial_year: 2020, state: :active, created_at: Date.parse("2020-04-01"))
+      project = create(:project_activity)
 
-      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: Date.today)
-      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: Date.today)
-      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: 4.months.ago)
+      create(:planned_disbursement, parent_activity: project, report: report, value: 1000.00, period_start_date: Date.parse("2020-04-01"), financial_quarter: 1, financial_year: 2020)
+      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: Date.parse("2020-01-01"), financial_quarter: 4, financial_year: 2019)
 
-      expect(project.forecasted_total_for_report_financial_quarter(report: report)).to eq(2000.00)
+      expect(project.forecasted_total_for_report_financial_quarter(report: report)).to eq(1000.00)
     end
 
     it "does not include totals for any planned disbursements outside the report's date range" do
-      project = create(:project_activity, :with_report)
-      report = Report.find_by(fund: project.associated_fund, organisation: project.organisation)
+      report = create(:report, financial_quarter: 3, financial_year: 2020, state: :active, created_at: Date.parse("2020-10-01"))
+      project = create(:project_activity)
 
-      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: 6.months.ago)
-      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: 4.months.ago)
+      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: Date.parse("2020-04-01"), financial_quarter: 1, financial_year: 2020)
+      create(:planned_disbursement, parent_activity: project, value: 1000.00, period_start_date: Date.parse("2020-07-01"), financial_quarter: 2, financial_year: 2020)
 
       expect(project.forecasted_total_for_report_financial_quarter(report: report)).to eq(0)
     end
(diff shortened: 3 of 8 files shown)


app/views/staff/planned_disbursements/_form.html.haml
```diff
@@ -1,4 +1,4 @@
-= f.govuk_error_summary
+= f.govuk_error_summary link_base_errors_to: :financial_quarter
```
MariluzHerrera, line 1: I've read the [documentation](https://govuk-form-builder.netlify.app/introduction/error-handling/#linking-errors-on-the-base-object-to-specific-fields) on the govuk form builder regarding linking the error to the base. I have a question about this. If there are any other errors on this form (apart from an attempt to create a second `:original` planned disbursement, all the errors will be linked to `:financial_quarter` because that's the recommendation from govuk, right? (They suggest to link them to the first field)

spec/models/activity_spec.rb
```diff
@@ -1073,17 +1072,16 @@
...
     end
   end
 
   describe "#forecasted_total_for_date_range" do
     it "returns the total of all the activity's planned disbursements scoped to a date range" do
       project = create(:project_activity, :with_report)
-      _disbursement_1 = create(:planned_disbursement, parent_activity: project, value: 200, period_start_date: Date.parse("13 April 2020"))
-      _disbursement_2 = create(:planned_disbursement, parent_activity: project, value: 1000, period_start_date: Date.parse("20 November 2020"))
+      _disbursement_1 = create(:planned_disbursement, parent_activity: project, value: 200, period_start_date: Date.parse("13 April 2020"), financial_quarter: 1, financial_year: 2020)
+      _disbursement_2 = create(:planned_disbursement, parent_activity: project, value: 1000, period_start_date: Date.parse("20 November 2020"), financial_quarter: 3, financial_year: 2020)
```
MariluzHerrera, line 1085: Do these date ranges only work within the same financial quarter?

spec/models/planned_disbursement_spec.rb
```diff
@@ -9,6 +9,40 @@
...
+        report: report
+      )
+
+      error_message = t("activerecord.errors.models.planned_disbursement.attributes.planned_disbursement_type.only_one_original",
+        financial_quarter: first_planned_disbursement.financial_quarter,
+        financial_year_start: first_planned_disbursement.financial_year,
+        financial_year_end: first_planned_disbursement.financial_year + 1)
+
+      expect(second_planned_disbursement).to be_invalid
+      expect(second_planned_disbursement.errors[:base]).to eq [error_message]
```
MariluzHerrera, line 36: Never seen an expectation with `eq [something_between_square_brackets]` before. Is this because `error_message` is a variable that has several items in it? (I mean, the translation file of the error needs a financial_quarter, financial_year start and end)

spec/presenters/activity_presenter_spec.rb
```diff
@@ -491,14 +491,17 @@
   describe "#forecasted_total_for_date_range" do
     it "returns the planned disbursement total for a date range as a formatted number" do
       project = create(:project_activity, :with_report)
-      _disbursement_1 = create(:planned_disbursement, parent_activity: project, value: 200.20, period_start_date: Date.today)
-      _disbursement_2 = create(:planned_disbursement, parent_activity: project, value: 1500, period_start_date: 3.months.ago)
+      current_financial_quarter = Date.parse("2020-07-01")
+      last_financial_quarter = Date.parse("2020-04-01")
+      next_financial_quarter = Date.parse("2020-10-01")
+      _disbursement_1 = create(:planned_disbursement, parent_activity: project, value: 200.20, period_start_date: current_financial_quarter, financial_quarter: 2, financial_year: 2020)
+      _disbursement_2 = create(:planned_disbursement, parent_activity: project, value: 1500, period_start_date: last_financial_quarter, financial_quarter: 1, financial_year: 2019)
```
MariluzHerrera, line 498: I think I might have misunderstood something here... quick question: If `last_financial_quarter` starts on April 2020 based on your `Date.parse("2020-04-01")` Would the financial_year be 2020 then?

spec/services/export_activity_to_csv_spec.rb
```diff
@@ -52,7 +52,7 @@
 
   describe "#next_four_quarter_forecasts" do
     it "gets the forecasted total for the date ranges of the next four quarters" do
```
MariluzHerrera, line 54: I guess this answers my previous question about the date ranges working within the same financial quarter only :-)

MariluzHerrera (approved):
This is the icing on the cake of your previous PRs on `planned_disbursement` :-) Very thorough amendment of the specs after introducing a relatively small change. 

As usual, I've left a few questions about things that caught my eye. Nothing blocks this PR though, so feel free to merge when you want 🚀 
Thank you!

app/models/planned_disbursement.rb
```diff
@@ -17,4 +19,10 @@ class PlannedDisbursement < ApplicationRecord
     :financial_quarter,
     :financial_year
   validates :value, inclusion: {in: 0.01..99_999_999_999.00}
+
+  def only_one_original
+    if PlannedDisbursement.find_by(financial_quarter: financial_quarter, financial_year: financial_year, parent_activity: parent_activity, planned_disbursement_type: :original).present?
+      errors.add(:base, I18n.t("activerecord.errors.models.planned_disbursement.attributes.planned_disbursement_type.only_one_original", financial_quarter: financial_quarter, financial_year_start: financial_year, financial_year_end: financial_year + 1))
```
jcoglan, line 25: I'm wondering if it's currently possible to implement this by putting a unique index on these columns, rather than implementing as a validation rule? As I suggested in https://gist.github.com/jcoglan/dcb4e87652fd172a83f9263973cad2da, we have a use case where we need to enter "revisions" which requires either creating a new `PlannedDisbursement` for the relevant activity and quarter, or amending the existing one. In this case the caller needs to check which of these it should be doing, or using `upsert` or `create_or_find_by` to produce the right result. Trying to create a new one, getting a validation error, and then updating the existing one is a more awkward way of implementing that and doesn't ensure the data is consistent.

I'm also confused about the `planned_disbursement_type: :original` condition -- shouldn't there be a single planned disbursment per activity/quarter _of any type_, with the first one in the history implicitly being the original?

app/views/staff/planned_disbursements/_form.html.haml
jcoglan: Re: my comment above, I'm wondering if this is an error that should ever be shown to the user. My expectation here was the user states what the value for the planned disbursement for a certain activity/quarter should be, and we either create or update `PlannedDisbursement` records to reflect that. Does the current workflow assume this always creates new records, and this is a stepping stone to the design in https://gist.github.com/jcoglan/dcb4e87652fd172a83f9263973cad2da? What is a user expected to do if they see this error on submitting the form?

mec: @MariluzHerrera As far as I understand it the errors on `base` will all be linked to `financial_quarter` (if it actually worked!) errors that relate to a specific field would still be linked to the relevent fields as they would be associated in the `errors` object.

@jcoglan I found very easy to see this errors the way the UI works right now, we have am 'add forecast' button that lets the user add a new original forecast, if they were to select a quarter and year of an existing original planned disbursement and submit the form, this would be the error state.

Yes this is the 3rd step towrds a complete forecast workflow, sorry I should have been clearer in the PR message.

I agree it is a dead end for the user right now, but I do feel the priority is to get forecasts working so we can collect them right now.

spec/models/activity_spec.rb
mec: This work is a bit odd, I should have been clearer in the PR message (apologies) at the moment the code that sorts the forecasts for display still relies on the `period_start_date` so we still have to include it, but this 'only one original' validation uses the financial quarter and year so for this 'interim' piece of work we need both - the work to fix the display and sorting is to come on this card: https://trello.com/c/Uw6hDK2r

spec/models/planned_disbursement_spec.rb
mec: No you haven't you are right! last_financial_quarter should be 1 2020. Thank you! I hope this is all a bit clearer when we can remove the dates in the next work.

spec/presenters/activity_presenter_spec.rb
mec: No you haven't you are right! last_financial_quarter should be 1 2020. Thank you! I hope this is all a bit clearer when we can remove the dates in the next work.

app/models/planned_disbursement.rb
mec: I was hesitent to apply a restriction on the DB as I know we have data in production that would not meet the condition (so my expectation is the migration would fail), in my head the plan is to introduce the changes to make forecasts work as expected, fix the data in production and then be in a postion to apply tighter DB level restrictions - happy to admit this may not be the best approach, but it is the one I settled on - I would love to talk through another approach if we think there is a better one?

This validation is used to prevent a user adding what they believe is a new forecast (by clicking the add new forecast button) and accidentally tyring to add it to a financial quarter that already has a forecast - I believe this is different to the user who is actively looking to revise an existing forecast (for which there is UI in the next piece of work https://trello.com/c/WB5Q9F9d)

Not sure I agree with your second statement, for a given activity and financial quarter I expect only one 'original' (the first) and then any number of revisions (a maximum of one per reporting quarter) for the life of the activity?

Happy to talk this through as a wider piece of work to try and address your concerns, but I do think it's important to see all the steps that we have planned?

spec/presenters/activity_presenter_spec.rb
MariluzHerrera: Yes, it is clear now. I was a bit confused with so many dates and quarters. Thanks for the [other explanation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/708#discussion_r516524462) too. It made clear why we need dates now that we won't need later :-)

spec/models/planned_disbursement_spec.rb
mec: I think it is because `errors[:base]` is an array of error messages, but we have setup the spec so that we can expect only one, you can probably wrtie this expecation a number of ways, perhaps `include?(error_message)` would be clearer?

Because I wanted to show the financial quarter and year that already has an original, I had to jump through the hoops of creating the expect error string so I can compare in the final expectation - I can see how this might make you think there is more going on - not sure what to do about that, I was keen to test that the message was correct.

Here is the guide tor the errors part:
https://guides.rubyonrails.org/active_record_validations.html#errors

MariluzHerrera: No, that's absolutely fine. I wasn't expecting you to change the way you wrote this expectation. It was just curiosity. Glad to see that my guess (square brackets because it's an array) was right. Thanks for the supporting documentation and explanation though!

mec:
@jcoglan  is kindly taking this work off my hands so it get's done effectively (I am struggling to get give it the time it needs) and we need this forecast work completed as a whole for Q3 reporting.

This PR may or may not be needed depending on the approach @jcoglan settles on so I'll mark it draft and wait for him to decide.

mec:
@MariluzHerrera thanks for all the work on reviewing this up to now!

jcoglan (approved):

PR #709, opened 2020-11-02
(1103) Add `No` to ODA eligibility

MariluzHerrera:
https://trello.com/c/a0PhOuZx

## Changes in this PR

We already had `oda_eligibility` as one of the fields in the activity form. On the codebase, this used to be a boolean since we only handled two types of values( `Eligible` and `No longer eligible`). Recently, the client requested to add a new option for this field (`No - was never eligible`), since this allows them to report activities that were never ODA eligible. 

For this we have had to change the type of column `oda_eligibility` is on the database. We have chosen to make it an integer, so we can make this field an enum, which allows us to include the three different options and it also gives us helper methods that could simplify future queries on the database. 

Please, note that the presence of existing data in this column prior to convert it into an integer made it very difficult to make this change. I tried to change the type following the Rails documentation but I kept getting data type mismatch errors warning me that the column couldn't be cast to a type integer. The way around it that ended up working was to create a column as an integer, migrating the existing records from one column to the newly created one, deleting the old column, and renaming the new column to match the name of the deleted column. This worked well and the final result is that `oda_eligibility` is an integer with a default of 1, and the existing data have been migrated to reflect the new way we store this value.

Apart from adding the new option for `oda_eligibility`, we have added hints on the radio buttons to give users more context and help them decide what category best reflect the eligibility of their activities. 

Specs and ingest work have also been modified to reflect the new situation.

### Before

Screenshot 2020-11-02 at 13 45 00

### After

Screenshot 2020-11-02 at 09 26 25

## Next steps

### DIFF
--- a/app/services/ingest_csv_row.rb
+++ b/app/services/ingest_csv_row.rb
@@ -53,9 +53,10 @@ class IngestCsvRow
   end
 
   def process_oda_eligibility(value)
-    return false if value.nil?
+    value = value.to_s.strip.downcase
+    return :skip if value.blank? || value.downcase == "n/a" || value.downcase == "not applicable"
 
-    value.downcase == "eligible"
+    oda_eligibility_mapping[value] || :skip
   end
 
   def process_gdi(value)
@@ -275,6 +276,14 @@ class IngestCsvRow
   def aid_type_mapping
     @aid_type_mapping ||= begin
       yaml_to_objects(entity: "activity", type: "aid_type")
+        .map { |status| [status["name"].downcase, status["code"]] }
+        .to_h
+    end
+  end
+
+  def oda_eligibility_mapping
+    @oda_eligibility_mapping ||= begin
+      yaml_to_objects(entity: "activity", type: "oda_eligibility")
         .map { |status| [status["name"].downcase, status["code"]] }
         .to_h
     end
--- a/db/migrate/20201102105419_change_oda_eligibility_type.rb
+++ b/db/migrate/20201102105419_change_oda_eligibility_type.rb
@@ -1,0 +1,19 @@
+class ChangeOdaEligibilityType < ActiveRecord::Migration[6.0]
+  def up
+    add_column :activities, :oda_eligibility_integer, :integer, default: 1, null: false
+    Activity.where(oda_eligibility: nil).update_all(oda_eligibility_integer: 1)
+    Activity.where(oda_eligibility: true).update_all(oda_eligibility_integer: 1)
+    Activity.where(oda_eligibility: false).update_all(oda_eligibility_integer: 2)
+    remove_column :activities, :oda_eligibility, :boolean
+    rename_column :activities, :oda_eligibility_integer, :oda_eligibility
+  end
+
+  def down
+    add_column :activities, :oda_eligibility_boolean, :boolean, default: true
+    Activity.where(oda_eligibility: 0).update_all(oda_eligibility_boolean: false)
+    Activity.where(oda_eligibility: 1).update_all(oda_eligibility_boolean: true)
+    Activity.where(oda_eligibility: 2).update_all(oda_eligibility_boolean: false)
+    remove_column :activities, :oda_eligibility, :integer
+    rename_column :activities, :oda_eligibility_boolean, :oda_eligibility
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -371,6 +371,8 @@
 ## [unreleased]
 - Fix allow application to create new users in Auth0
 
+- Option `No - was never eligible` added to ODA eligibility form step
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-21...HEAD
 [release-21]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-20...release-21
 [release-20]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-19...release-20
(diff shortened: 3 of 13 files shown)


db/migrate/20201102105419_change_oda_eligibility_type.rb
```diff
@@ -0,0 +1,10 @@
+class ChangeOdaEligibilityType < ActiveRecord::Migration[6.0]
+  def change
+    add_column :activities, :oda_eligibility_integer, :integer, default: 1
+    Activity.where(oda_eligibility: nil).update_all(oda_eligibility_integer: 1)
+    Activity.where(oda_eligibility: true).update_all(oda_eligibility_integer: 1)
+    Activity.where(oda_eligibility: false).update_all(oda_eligibility_integer: 2)
+    remove_column :activities, :oda_eligibility, :boolean
+    rename_column :activities, :oda_eligibility_integer, :oda_eligibility
+  end
```
jcoglan, line 9: Although it is annoying the database won't let you do this type change directly, this is exactly the way to go about this sort of change. Sometimes it's done over multiple deploys so that the changes are backward compatible, but I'm not too worried about that in this case.

app/services/ingest_csv_row.rb
```diff
@@ -280,6 +281,14 @@ def aid_type_mapping
     end
   end
 
+  def oda_eligibility_mapping
+    @oda_eligibility_mapping ||= begin
+      yaml_to_objects(entity: "activity", type: "oda_eligibility")
```
jcoglan, line 286: Just to check I understand: does this load data from the file `vendor/data/codelists/IATI/2_03/activity/oda_eligibility.yml` that was added in the previous commit?

db/migrate/20201102105419_change_oda_eligibility_type.rb
MariluzHerrera: Well, that explains that despite following the documentation and reading several threads on StackOverflow I wasn't able to make it work. Glad to hear this is the way to do it. I was a bit worried this would look too "handcrafted" and someone would point it out (hence, the probably overexplanation on the PR message). 
Regarding the possibility of a rollback, Meyric suggested that probably we wanted this change irreversible to avoid losing any data. For that reason I decided to put everything on the same migration file.

app/services/ingest_csv_row.rb
MariluzHerrera: Yes, it takes the names and the codes from that file.

db/migrate/20201102105419_change_oda_eligibility_type.rb
jcoglan: That's a good point about reversibility -- Rails won't know how to reverse this operation. We can either mark it as irreversible, or we could implement a `down` method. There's no data loss here, because the new integer values can be mapped back to the old boolean ones. You would have ambiguous data once any records with value `3` are added though.

pezholio: Forgive my new-to-the-project naivety, but could we not put lines 4 - 6 in a data migration, and then remove the  `oda_eligibility` column at a later date? (once this code is confirmed working and deployed to prod) Would keep things a bit cleaner, and would save any data loss if the migrations choke for some reason on deploy.

jcoglan: A problem with the `data_migrations` gem (and a reason I'd like to move away from it) is that migrations are not ordered relative to schema migrations. Here, the data changes are part of changing the type of the column so I think they should be part of this migration, not a separate one.

jcoglan: "Remove the `oda_eligibility` column at a later date" is what I was getting at with doing this over multiple deploys. Actually doing this kind of change in a fully backward/forward compatible way is a lot more steps and I'm not sure it's worth it in this case.

pezholio: Yeah, I'm not a fan of the `data_migrations` gem either. We moved towards just having the data migrations as plain old Ruby scripts that we just run after deploy

leeky: As `nil` is being remapped to `1` could the integer field have `null: false` in the migration now?

MariluzHerrera: I'll give it a go and see it nothing breaks

MariluzHerrera: It worked leeky! 🎉

MariluzHerrera:
@jcoglan @pezholio I have modified the migration file with an `up` and a `down` method. I ran this migration and rollback on my local and it seems to work well. Let me know if you're happy with this or you think it needs more changes. Thank you!

db/migrate/20201102105419_change_oda_eligibility_type.rb
pezholio: Yeah, this is ace. Much happier with this 👍

pezholio (approved):
Looks great 👍

MariluzHerrera:
@pezholio I've just pushed some changes on the migration file. Mainly I added a line on the `down` method to update records with a value of `0` to be `false` because it is what aligns best with reality. I have got rid of the line that updates nil values to true, since the column as an integer does not allow `null` so we wouldn't have any `nil` values on `oda_eligibility` (if that makes any sense, sorry for the crappy explanation). 
Are you still happy to approve this PR?

pezholio (approved):
Yep, looks good 👍

PR #710, opened 2020-11-02
Add scripts to rule them all

pezholio:
This adds scripts using the [Scripts to Rule Them All](https://github.com/github/scripts-to-rule-them-all) pattern to make it easier to get started with the app.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+worker: bundle exec sidekiq -C config/sidekiq.yml
+web: bundle exec rails server -p 3000
--- a/script/console
+++ b/script/console
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+# script/console: Launch a console for the application.
+
+set -e
+
+cd "$(dirname "$0")/.."
+
+echo "==> Updating..."
+script/update
+
+echo "==> Starting a local Rails console..."
+bundle exec rails console "$@"
--- a/script/setup
+++ b/script/setup
@@ -1,0 +1,65 @@
+#!/bin/sh
+
+# script/setup: Set up the application for the first time after cloning, or set
+#               it back to the initial unused state.
+
+set -e
+
+cd "$(dirname "$0")/.."
+
+if [ -d vendor/bundle ]; then
+  echo "==> Cleaning installed Ruby dependencies..."
+  git clean -x --force -- vendor/bundle
+fi
+
+echo "==> Bootstrapping..."
+script/bootstrap
+
+if [ ! -f .env.development.local ]; then
+  echo "==> Copying default environment config..."
+  cp .env.example .env.development.local
+fi
+
+echo "==> Dropping database..."
+set +e
+bundle exec rails db:drop
+DB_DROP_RESULT=$?
+set -e
+
+if [ -z "$CI" ] && [ "$DB_DROP_RESULT" -ne "0"  ]; then
+  printf "\nDatabase drop failed. Continue anyway? [y/N] "
+  read -r
+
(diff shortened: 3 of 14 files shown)


Procfile
```diff
@@ -0,0 +1,2 @@
+worker: bundle exec sidekiq -C config/sidekiq.yml
```
leeky, line 1: NB: There aren't currently any jobs for the worker to run and I'm not actually sure *when* we'll need Sidekiq.

script/console
```diff
@@ -0,0 +1,25 @@
...
+      echo "Unknown environment '$ENVIRONMENT'"
+      exit 1
+      ;;
+  esac
+else
+  echo "==> Updating..."
+  script/update
+
+  echo "==> Starting a local Rails console..."
+  bundle exec rails console
```
leeky, line 24: Is it worth adding `--sandbox` support to this?

script/test
```diff
@@ -0,0 +1,38 @@
...
+if [ -n "$TEST_FILE" ]; then
+  echo "==> Running the tests matching '$TEST_FILE'..."
+  bundle exec rspec --pattern "$TEST_FILE"
+else
+  if [ -n "$CI" ]; then
+    echo "==> Linting Ruby..."
+    bundle exec standardrb
+  else
+    echo "==> Linting Ruby in fix mode..."
+    bundle exec standardrb --fix
```
leeky, line 30: This is nice! :+1:

script/console
pezholio: Oooh, good plan 👍

script/test
pezholio: I nicked it off the Rails template, so can't claim credit 😄

Procfile
pezholio: Ooooh. Is it worth keeping this just in case?

script/update
```diff
@@ -0,0 +1,16 @@
...
+cd "$(dirname "$0")/.."
+
+echo "==> Bootstrapping..."
+script/bootstrap
+
+echo "==> Running database migrations..."
+bundle exec rails db:migrate
+
+echo "==> Running test database migrations..."
+RAILS_ENV="test" bundle exec rails db:migrate
```
leeky, line 16: Is this right? I thought the test database should use `bundle exec rails db:schema:load` as it'll truncate any static data added in the migrations?

pezholio: Yeah, you're probably right actually. I prefer not to add data in migrations, but this is probably safer

script/setup
```diff
@@ -0,0 +1,65 @@
...
+  echo "==> Cleaning installed Ruby dependencies..."
+  git clean -x --force -- vendor/bundle
+fi
+
+echo "==> Bootstrapping..."
+script/bootstrap
+
+if [ ! -f .env.development.local ]; then
+  echo "==> Copying default environment config..."
+  cp .env.example .env.development.local
```
leeky, line 20: This is great - I've often accidentally copied this to `.env` and wondering after running the test suite where my development data has gone!

script/test
```diff
@@ -0,0 +1,38 @@
...
+
+cd "$(dirname "$0")/.."
+
+if [ -n "$DEBUG" ]; then
+  set -x
+fi
+
+set +e
+echo "==> Setting up test user..."
+createuser --super test >/dev/null 2>&1
```
leeky, line 16: Should `DATABASE_URL` in `.env.test` be updated to use this `test` user account?

leeky (approved):
LGTM! :shipit:

PR #714, opened 2020-11-04
(1180) Fix Auth0 create_user call

mec:
## Changes in this PR
We recently updated the Auth0 gem and I missed a breaking change.

The signature of the `create_user` method has changed; the connection
name, which is just a string that matches the default database
connection in Auth0, comes first and is required and the name of the
user is passed in the options as `name`.

Without this change the application cannot create new users in Auth0.
The application handles the error gracefully as the API simply returns
an error code 'connection not found' that we display to the user.

I settled on not creating a test for this case as I should have spotted the change in the API when I was looking at the dependabot update, I feel the system responded correctly as we were passing the wrong value as the connection. If folks feel differently, please shout and I'll have a go at a test.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -369,6 +369,7 @@
 - Use scripts to rule them all for development tasks
 
 ## [unreleased]
+- Fix allow application to create new users in Auth0
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-21...HEAD
 [release-21]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-20...release-21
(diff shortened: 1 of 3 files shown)


pezholio:
The change seems OK, but I'm confused as to why the tests didn't catch this? I'd feel more comfortable if we could recreate the behaviour we currently see in the tests, then fix that. It would guard against this happening again in the future. Happy to pair on this if you've got time 👍

mec:
@pezholio yeah, essentially we were sending values to the API but they were the wrong ones, it's a bit of an odd test to write but happy to pair and have a go - be in touch.

mec:
@pezholio thanks for the help - this is loads better. I've checked the other calls we make and they look good as they update exisiting entities in Auth0 rather than creating new ones.

pezholio (approved):

pezholio:
Looks good to me!

pezholio:
(Soz for the delay, was waiting for Travis)

PR #717, opened 2020-11-05
(1192) Adding RODA IDs to invalid activities list

MariluzHerrera:
Trello: https://trello.com/c/r1Tn9w2s

## Changes in this PR

After supplying the client with a list of invalid activities, we were asked to add the RODA ID of each activity, so the users could group them up and make it easier for them to work on them.

This PR adds this RODA ID for each activity on the csv file, and adds the expectation on the spec file.

## Next steps


mec (approved):
🚢

PR #720, opened 2020-11-06
I can report free standing technical cooperation

pezholio:
## Changes in this PR

This adds a new field to report free standing technical cooperation. Some of this could be gleaned from the aid type further down the line, but this is probably good enough for now.

## Screenshots of UI changes

image

### DIFF
--- a/app/services/ingest_iati_activities.rb
+++ b/app/services/ingest_iati_activities.rb
@@ -43,6 +43,7 @@ class IngestIatiActivities
           add_requires_additional_benefitting_countries(legacy_activity: legacy_activity, new_activity: new_activity)
           add_intended_beneficiaries(legacy_activity: legacy_activity, new_activity: new_activity)
           add_gdi(legacy_activity: legacy_activity, new_activity: new_activity)
+          add_fstc_applies(new_activity: new_activity)
 
           new_activity.form_state = :complete
           new_activity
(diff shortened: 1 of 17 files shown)


db/schema.rb
```diff
@@ -55,15 +55,16 @@
     t.boolean "call_present"
     t.date "call_open_date"
     t.date "call_close_date"
-    t.string "intended_beneficiaries", array: true
     t.string "roda_identifier_fragment"
     t.string "roda_identifier_compound"
+    t.string "intended_beneficiaries", array: true
```
mec, line 60: I always forget if we are supposed to worry about these lines shifting in the schema?

app/services/ingest_iati_activities.rb
```diff
@@ -305,6 +306,21 @@ def call
...
+    when "C01"
+      true
+    when "E01"
+      true
+    when "D01"
+      true
+    when "GO1"
+      false
+    else
+      false
```
mec, line 320: I agree, this will keep the test suite happy, we shouldn't be using the ingest code again.

mec (approved):
Looks good. 

I did expect to see some coverage on the activtiy details over editing the value and coming back to the details page, but I can't see that for the existing steps beyond the `assert_all_edit_links_go_to_the_correct_form_step` – I might be missing something, what are your thoughts?

You have a couple of commit messages with the git comments tops left in - I would be taking those out, but happy to let you decide on your happy place for that!

I'll hit the approve button but I would appreciate hearing your thoughts.

MariluzHerrera:
I only have a comment/question (sorry it's a bit nitpicking). The acronym on the UI is STFC (according to the screenshot), but I've seen this before as FSTC, which aligns better with Free-Standing Technical Cooperation. Not sure it's a typo or that I've missed something on the way...

pezholio:
@MariluzHerrera Yeah, that was a typo. I've fixed it now - had to change all the code and migrations too 😆

pezholio:
@mec I'll get rid of the Git comments too - I missed those! To the point on the coverage - I must've missed where that stuff gets tested. I'll have a look 👍

db/schema.rb
pezholio: I've got rid of it, just in case. Noticed it jumping back on me, so let's keep it where it is

PR #726, opened 2020-11-10
(1195) Start building out the Activity import machinery

pezholio:
This adds a couple of classes that start to build out the code that will eventually backfill missing activity data. It's heavily influenced by work already done by @jcoglan 

So far it accepts an array of hashes (which will eventually come from a CSV) to update an existing Activity (which is fetched by its RODA ID). 

If a field is blank, it is ignored. If any activities in the array of attribute hashes are missing, then the entire operation fails and the main classes error array is populated.

### DIFF
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -72,6 +72,7 @@ module Activities
         title: "Title",
         description: "Description",
         recipient_region: "Recipient Region",
+        delivery_partner_identifier: "Delivery partner identifier",
       }
 
       def initialize(row)
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -1,0 +1,130 @@
+module Activities
+  class ImportFromCsv
+    Error = Struct.new(:row, :column, :value, :message) {
+      def csv_row
+        row + 2
+      end
+    }
+
+    attr_reader :errors
+
+    def initialize(organisation:)
+      @organisation = organisation
+      @errors = []
+    end
+
+    def import(activities)
+      ActiveRecord::Base.transaction do
+        activities.each_with_index { |row, index| import_row(row, index) }
+        raise ActiveRecord::Rollback unless @errors.empty?
+      end
+    end
+
+    def import_row(row, index)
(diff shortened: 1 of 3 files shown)


mec:
@pezholio I am sorry this is so nitpiking, I would love for language on these to be more 'import' and less 'ingest', mainly to align with what we have - 'import_transactions'?

pezholio:
No, that's cool. I was a bit iffy on the language too

jcoglan:
I also struggled with what to name things. I'm not 100% on this but I'd mentally filed "ingest" as meaning gather data from external historical data sources like IATI, and have been using "import" or "upload" for data submitted directly by RODA users. These terms could probably be clearer or better defined.

app/services/activities/import_from_csv.rb
```diff
@@ -0,0 +1,133 @@
...
+          "activity/recipient_region.yml",
+          I18n.t("importer.errors.activity.invalid_region"),
+        )
+      end
+
+      def validate_from_codelist(code, codelist_file, message)
+        return nil if code.blank?
+
+        codelist_path = codelist_root.join(codelist_file)
+        codelist = YAML.safe_load(File.read(codelist_path))
```
leeky, line 121: Did you consider using `CodelistHelper.load_yaml` here?

leeky: Not sure if they'll be any overhead in multiple code list YAML files over and over for each row, or this should be memoised at some point?

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    subject.import([activity_attributes])
+
+    expect(subject.errors.count).to eq(0)
+
+    expect(existing_activity.reload.title).to eq(activity_attributes["Title"])
+    expect(existing_activity.description).to eq(activity_attributes["Description"])
+    expect(existing_activity.recipient_region).to eq(activity_attributes["Recipient Region"])
+  end
+
+  it "ignores any blank columns" do
```
leeky, line 42: Hopefully we'll never see a case where a column *should* be empty and that should be updated on the Activity(!)

leeky (approved):
LGTM! :shipit:

PR #728, opened 2020-11-11
(1105) Add policy markers to activity form

MariluzHerrera:
Trello: https://trello.com/c/hKQ5j2uG

## Changes in this PR

Policy markers are a new field on the activity form requested by the client. IATI considers 12 different policy markers, but BEIS only use 8 of them. These 8 have been added to the form. They share the same 5 answers/values so I have decided to make them `enum` so we can get helper methods that can come handy for future queries. In order to be able to share the enum methods among multiple database columns I am using prefixes as recommended in the [ActiveRecord documentation](https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html)

These markers get reported to IATI, so they have been included on the activity XML and I have checked them against the IATI validator and they pass. Please note, there is only one value we do not report to IATI (`not assessed`) since this is a custom BEIS value they use for internal purposes only. 

Despite these 8 policy markers being different entries on the database, I have grouped them up on the same form step/page for a better user experience, defaulting the value on the select box to the one BEIS uses most (`not assessed`).
I have also created a `details` section that the users can access to get an explanation of the different responses they can select for each marker.

Lastly, an ingest spec needed amendment due to validation errors. This is on a separate commit.

### After

Screenshot 2020-11-11 at 16 28 43

Screenshot 2020-11-11 at 16 29 18

## Next steps

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -29,6 +29,7 @@ class Staff::ActivityFormsController < Staff::BaseController
     :flow,
     :aid_type,
     :fstc_applies,
+    :policy_markers,
     :oda_eligibility,
   ]
 
@@ -65,6 +66,8 @@ class Staff::ActivityFormsController < Staff::BaseController
     when :collaboration_type
       skip_step if @activity.fund?
       assign_default_collaboration_type_value_if_nil
+    when :policy_markers
+      skip_step unless @activity.is_project?
     end
 
     render_wizard
@@ -145,6 +148,17 @@ class Staff::ActivityFormsController < Staff::BaseController
       @activity.assign_attributes(aid_type: aid_type)
     when :fstc_applies
       @activity.assign_attributes(fstc_applies: fstc_applies)
+    when :policy_markers
+      @activity.assign_attributes(
+        policy_marker_gender: policy_markers_iati_codes_to_enum(policy_marker_gender),
+        policy_marker_climate_change_adaptation: policy_markers_iati_codes_to_enum(policy_marker_climate_change_adaptation),
+        policy_marker_climate_change_mitigation: policy_markers_iati_codes_to_enum(policy_marker_climate_change_mitigation),
+        policy_marker_biodiversity: policy_markers_iati_codes_to_enum(policy_marker_biodiversity),
+        policy_marker_desertification: policy_markers_iati_codes_to_enum(policy_marker_desertification),
+        policy_marker_disability: policy_markers_iati_codes_to_enum(policy_marker_disability),
+        policy_marker_disaster_risk_reduction: policy_markers_iati_codes_to_enum(policy_marker_disaster_risk_reduction),
+        policy_marker_nutrition: policy_markers_iati_codes_to_enum(policy_marker_nutrition),
+      )
     when :oda_eligibility
       @activity.assign_attributes(oda_eligibility: oda_eligibility)
     end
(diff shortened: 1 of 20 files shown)


vendor/data/codelists/IATI/2_03/activity/policy_markers.yml
```diff
@@ -0,0 +1,26 @@
+---
```
pezholio, line 1: On this, I think we should mirror what comes from IATI directly, otherwise it could cause confusion. See https://iatistandard.org/reference_downloads/203/codelists/downloads/clv3/json/en/PolicySignificance.json

config/locales/models/activity.en.yml
```diff
@@ -75,6 +83,21 @@ en:
         parent: Select the parent activity
         planned_end_date: Planned end date
         planned_start_date: Planned start date
+        policy_markers:
```
pezholio, line 86: I might be misunderstanding something, but could we not use the codelist from https://iatistandard.org/en/iati-standard/203/codelists/policymarker/ for this?

pezholio (changes requested):
Code-wise it looks great - but I'm just a bit confused about the IATI codelists. I think we should be consistent with what we use. I could equally be barking up the wrong tree here! 😆

spec/support/form_helpers.rb
```diff
@@ -227,6 +268,16 @@ def fill_in_activity_form(
     expect(page).to have_content gdi
     expect(page).to have_content flow
     expect(page).to have_content t("activity.aid_type.#{aid_type.downcase}")
+    if level == "project" || level == "third_party_project"
+      expect(page).to have_content policy_marker_gender
```
leeky, line 272: These probably need `within` blocks as multiple policy markers could contain the same value.

This is currently looking for "Not assessed" somewhere on the page, which would still be true even if the summary page was missing some of the markers.

MariluzHerrera: That's true. I'll change that

config/locales/models/activity.en.yml
MariluzHerrera: I _think_ we don't use the IATI codelists on the locales files, only in the vendor. `activity.en.yml` is for translation files for the UI. It's a RODA thing rather than an IATI thing, if you know what I mean?

vendor/data/codelists/IATI/2_03/activity/policy_markers.yml
MariluzHerrera: I saw you requested changes on this PR. Is this the part you want to see some changes? I'm not very clear with what I should change or not. 
I did it this way because we don't use all the options from IATI for our users. We have removed the last option that IATI considers because it does only affect a policy marker that BEIS don't work with, and also BEIS had added a custom option that I haven't put on this file because it's not IATI related. 
Do you want me to add the last one even if we don't use it and then modify the code so we don't offer that option to the users on the UI?

pezholio: Yeah, I think that would be best. It kinda chimes with what we did with the Aid Type in this PR https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/706 

Apologies if I got the wrong end of the stick!

MariluzHerrera: I think I managed to put everything in the right place. Have a look at the new changes when you have time and let me know what you think. Thanks @pezholio!!

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -280,6 +294,38 @@ def fstc_applies
...
+  end
+
+  def policy_marker_disaster_risk_reduction
+    params.require(:activity).permit(:policy_marker_disaster_risk_reduction).fetch("policy_marker_disaster_risk_reduction", nil)
+  end
+
+  def policy_marker_nutrition
+    params.require(:activity).permit(:policy_marker_nutrition).fetch("policy_marker_nutrition", nil)
+  end
+
```
pezholio, line 328: Probably not something for now, but I am a bit concerned that this controller is getting a bit unweildy, see my comment on @leeky's PR here https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/731#discussion_r522834283

app/helpers/codelist_helper.rb
```diff
@@ -132,6 +139,14 @@ def aid_type_radio_options
     options.select { |a| ALLOWED_AID_TYPE_CODES.include?(a.code) }
   end
 
+  def policy_markers_select_options
+    objects = yaml_to_objects(entity: "activity", type: "policy_markers", with_empty_item: false)
+    not_assessed_option = OpenStruct.new(name: "Not assessed", code: "1000")
+
+    filtered_list = objects.select { |object| ALLOWED_POLICY_MARKERS_SIGNIFICANCES.include?(object.code) }.sort_by(&:code)
```
pezholio, line 146: 👍

vendor/data/codelists/IATI/2_03/activity/policy_markers.yml
pezholio: Looks good! 👍

pezholio (approved):
LGTM 👍

app/controllers/staff/activity_forms_controller.rb
MariluzHerrera: I see what you mean. I had a similar thought, and agree with you that's probably worth having a look in the future to refactor this a bit

PR #729, opened 2020-11-11
(1199) Move programme activity validation specs

leeky:
Whilst adding Sustainable Development Goals I came across a problem where some feature specs for programme-level validations was appearing in a fund-level test.

This caused confusion when I was trying to test the SDG form step being skipped when adding a fund.

I've copied the specs into the correct file and amended the existing ones to test at the right level.


MariluzHerrera (approved):
Nice piece of work! That's also something that I wondered the first time I saw that spec. I thought the reason for having the programme created on that file was the association with the Fund as parent of the Programme, but it's true that makes more sense and it looks quite tidy the way you've done it now. Thanks for that!

PR #730, opened 2020-11-12
(1198) Allow the activity importer to create activities

pezholio:
This allows the `Activities::ImportFromCsv` importer can now create activities which:

a) Do not have a RODA ID specified; and:
b) Have a RODA Fragment and RODA Parent ID

If the above is true, the row is passed to the ActivityCreator class, which initializes a new activity, assigns the attributes, saves it, and gathers up any errors.

If there is no RODA ID, or RODA Fragment and RODA Parent ID, then an error is added to the class instance, and no subsequent or previous Activities are imported.

I've also added a new converter method that fetches the parent Activity by its RODA ID, and returns the UUID from the database. If the activity cannot be found, an error is added to the errors hash.


pezholio:
[Allow the Activity Importer to create Activities](https://trello.com/c/P7IZGyzb/1198-allow-the-activity-importer-to-create-activities)

app/services/activities/import_from_csv.rb
```diff
@@ -21,31 +21,44 @@ def import(activities)
...
-        add_error(index, :roda_id, row["RODA ID"], I18n.t("importer.errors.activity.not_found"))
+      if row["RODA ID"].blank? && row["RODA ID Fragment"].present? && row["Parent RODA ID"].present?
+        action = create_activity(row, index)
+      elsif row["RODA ID"].present?
+        action = update_activity(row, index)
       else
-        updater = ActivityUpdater.new(activity, @organisation, row)
-        updater.update
+        add_error(index, :roda_id, row["RODA ID"], I18n.t("importer.errors.activity.cannot_create_or_update")) && return
+      end
```
jcoglan, line 30: It might be worth tightening the ambiguity detection here. This will catch the case where `RODA ID` is blank and either `RODA ID Fragment` or `Parent RODA ID` is blank, but it won't catch some other cases, for example both `RODA ID` and `RODA ID Fragment` are present. I'd suggest:

- create an activity if `RODA ID` is blank, `RODA ID Fragment` is present, and `Parent RODA ID` is present
- update an activity if `RODA ID` is present, `RODA ID Fragment` is blank, and `Parent RODA ID` is blank (i.e. exact inverse of previous case)
- raise an ambiguity error in all other cases

Alternatively I'd select create vs update based purely on the presence of `RODA ID` and treat the absence or presence of the other fields as errors, but this might produce less informative errors than the above.

app/services/activities/import_from_csv.rb
```diff
@@ -61,6 +74,54 @@ def update
...
+
+      def find_activity_by_roda_id(roda_id)
+        activity = Activity.by_roda_identifier(roda_id)
+        @errors[:roda_id] ||= [roda_id, I18n.t("importer.errors.activity.not_found")] if activity.nil?
+
+        activity
+      end
+    end
+
+    class ActivityCreator
```
jcoglan, line 86: This class does exactly what I was expecting in terms of using of identifiers: passing `Parent RODA ID` to `Activity.by_roda_identifier` to get the parent, assigning `roda_identifier_fragment` and using `cache_roda_identifier` to generate the full ID 👍 

A restriction implied in this design is that a CSV import can only create activities whose parent is already in the system, and that parent must have a RODA ID assigned. That sidesteps a bunch of problems to do with ordering the import rows -- to assign a RODA ID to an activity, its parent must already have an ID, so these IDs must be assigned in activity-level order.

By using the RODA ID to identify the parent, that means this system can only create activities where assigning a RODA ID is safe, because if the parent exists then it must have a RODA ID already.

Not sure if I've explained that very well, the point is this avoids a bunch of nasty problems and I like it :)

app/services/activities/import_from_csv.rb
jcoglan: In fact in the case where `RODA ID` and `RODA ID Fragment` are both present, this could update an existing activity and overwrite its `roda_identifier_fragment` field, so I'd treat this as a logic bug rather than just unclear error messaging.

pezholio: @jcoglan I've added another commit with some explanations to try and tighten up this ambiguity detection 👍

jcoglan: @pezholio That looks correct to me and I believe all cases have test coverage. Looks good 👍

pezholio:
@jcoglan Are you OK to approve this?

pezholio:
Whoop!

PR #731, opened 2020-11-12
(1101) I can report Sustainable Development Goals (SDG) where applicable

leeky:
Users can report up to 3 Sustainable Development Goals (SDG) for Level B-D activities, in order of relevance. 

### Sustainable Development Goals form step

sdgs_does_not_apply

#### Goals selectable when SDGs apply
sdgs_apply

#### Error when first most relevant goal not provided
first_sdg_required

### Activity details page
Screenshot 2020-11-12 at 16 16 07

## Next steps

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -166,11 +166,9 @@ class Staff::ActivityFormsController < Staff::BaseController
     when :covid19_related
       @activity.assign_attributes(covid19_related: covid19_related)
     when :sustainable_development_goals
-      @activity.assign_attributes(sdgs_apply: sdgs_apply)
+      @activity.assign_attributes(sustainable_development_goals)
 
-      if @activity.sdgs_apply?
-        @activity.assign_attributes(sdg_1: sdg_1, sdg_2: sdg_2, sdg_3: sdg_3)
-      else
+      unless @activity.sdgs_apply?
         @activity.assign_attributes(sdg_1: nil, sdg_2: nil, sdg_3: nil)
       end
     when :oda_eligibility
@@ -300,20 +298,8 @@ class Staff::ActivityFormsController < Staff::BaseController
     params.require(:activity).permit(:flow).fetch("flow", nil)
   end
 
-  def sdg_1
-    params.require(:activity).permit(:sdg_1).fetch("sdg_1", nil)
-  end
-
-  def sdg_2
-    params.require(:activity).permit(:sdg_2).fetch("sdg_2", nil)
-  end
-
-  def sdg_3
-    params.require(:activity).permit(:sdg_3).fetch("sdg_3", nil)
-  end
-
-  def sdgs_apply
-    params.require(:activity).permit(:sdgs_apply).fetch("sdgs_apply", false)
+  def sustainable_development_goals
+    params.require(:activity).permit(:sdg_1, :sdg_2, :sdg_3, :sdgs_apply)
   end
 
   def aid_type
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -27,6 +27,7 @@ class Staff::ActivityFormsController < Staff::BaseController
     :gdi,
     :collaboration_type,
     :flow,
+    :sustainable_development_goals,
     :aid_type,
     :fstc_applies,
     :policy_markers,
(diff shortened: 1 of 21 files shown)


pezholio:
Not done a proper review yet, but the one thing that sticks out is the way the goals are added seems a bit odd. I would have expected for this to happen over two pages, with the first page asking if Sustainable Development Goals apply, and the second page asking which ones apply (assuming the answer to the previous question is "yes"). The wording "The sustainable development goals apply to this activity with the three most relevent in decsending order being:" seems a bit clunky to me. Happy to be told other wise though 👍

app/helpers/codelist_helper.rb
```diff
@@ -14,7 +16,7 @@ module CodelistHelper
 
   def yaml_to_objects(entity:, type:, with_empty_item: true)
     data = load_yaml(entity: entity, type: type)
-    return [] if data.empty?
+    raise UnreadableCodelist if data.empty?
```
pezholio, line 20: 👍

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -272,6 +275,18 @@ def flow
...
+    params.require(:activity).permit(:sdg_1).fetch("sdg_1", nil)
+  end
+
+  def sdg_2
+    params.require(:activity).permit(:sdg_2).fetch("sdg_2", nil)
+  end
+
+  def sdg_3
+    params.require(:activity).permit(:sdg_3).fetch("sdg_3", nil)
+  end
```
pezholio, line 296: I dunno what you think, but I'm wondering if it might be better if we fetched all the sustainable development goals in one go, rather than having them as seperate parameters. I know it's a bit of a departure from what we've done elsewhere, but as they're all being collected on the one page, I don't think it's a huge deal. Would help keep this controller from getting too long and bloated. Might be worth actually passing the param fetching stuff to its own class at some point. Anyway, just a brain dump

app/presenters/activity_presenter.rb
```diff
@@ -109,6 +109,15 @@ def flow_with_code
...
   end
 
+  def sustainable_development_goals
+    goals = [sdg_1, sdg_2, sdg_3].compact
+    return if goals.blank?
+
+    goals.map { |goal|
+      I18n.t("form.label.activity.sdg_options.#{goal}")
+    }.join(" / ")
+  end
```
pezholio, line 119: I wonder if it might be better to display these as a bulleted (or numbered) list to better reflect the fact that the order of them matters

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -147,7 +147,13 @@ def update
     when :fstc_applies
       @activity.assign_attributes(fstc_applies: fstc_applies)
     when :sustainable_development_goals
-      @activity.assign_attributes(sdgs_apply: sdgs_apply, sdg_1: sdg_1, sdg_2: sdg_2, sdg_3: sdg_3)
+      @activity.assign_attributes(sdgs_apply: sdgs_apply)
+
+      if @activity.sdgs_apply?
+        @activity.assign_attributes(sdg_1: sdg_1, sdg_2: sdg_2, sdg_3: sdg_3)
+      else
+        @activity.assign_attributes(sdg_1: nil, sdg_2: nil, sdg_3: nil)
```
pezholio, line 157: 👍 Good to have belt and braces here

app/controllers/staff/activity_forms_controller.rb
leeky: I actually went back and forth on this a few times - did this for consistency, but actually these three fields are related, so make sense having the params being permitted through one method.

pezholio: Yeah, I think there's a bigger piece of work to do around getting all this into an ActivityUpdater class of its own, which I've already got some ideas about. Let's leave it as it is now, in the interests of getting it out.

leeky:
@pezholio I've added a couple of commits to address your comments. Would be grateful if you could give them a look :)

config/locales/models/activity.en.yml
```diff
@@ -52,6 +52,31 @@ en:
         requires_additional_benefitting_countries:
           "true": "Yes"
           "false": "No"
+        sdg_1: First most relevant goal
+        sdg_2: Second most relevant goal (optional)
+        sdg_3: Third most relevant goal (optional)
+        sdgs_apply_options:
+          "false": "This is a broad programme/delivery activity so the sustainable development goals do not apply"
+          "true": "The sustainable development goals apply to this activity with the three most relevent in descending order being:"
```
pezholio, line 60:
```suggestion
          "true": "The sustainable development goals apply to this activity with the three most relevant in descending order being:"
```

pezholio (approved):
Other than one tiny typo, it's good to go 👍

PR #735, opened 2020-11-16
Change location of overriden hints

pezholio:
@MariluzHerrera noticed a bug where hints on some questions were being overridden accidentally (introduced in #706). I've moved the location of hints that override descriptions given in the codelist to somewhere that better describes what they do.

### Before

image

### After

image


MariluzHerrera (approved):
This looks great. I like having the descriptions on the hint section of the yaml file. 
Thanks for fixing this @pezholio! 🚀

PR #736, opened 2020-11-16
(1123) Covid-19 related research

CristinaRO:
## Changes in this PR

- Add a field to report whether the activity has any relation to the Covid-19 pandemic.
- Include the Covid-19 field in the CSV report.
- Append the text "COVID-19" to the activity description in the IATI XML export, where applicable.

### After
Screenshot 2020-11-16 at 16 42 16
Screenshot 2020-11-16 at 16 43 31
Screenshot 2020-11-16 at 16 43 55
Screenshot 2020-11-16 at 16 44 43

## Next steps

The file I generated failed the validation for a field to which I had not made any changes:
`Element 'activity-status': The attribute 'code' is required but missing.`

There are no validations on the `Activity` model to ensure that the `status` is populated. However, this is out of the scope of the current story.


CristinaRO:
[Covid-19 related research](https://trello.com/c/YNcx4ntC/1123-covid-19-related-research)

app/helpers/codelist_helper.rb
```diff
@@ -147,6 +147,13 @@ def policy_markers_select_options
     filtered_list.unshift(not_assessed_option)
   end
 
+  def covid19_related_radio_options
+    yaml = YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/BEIS/covid19_related_research.yml"))
+    yaml["data"].collect { |item|
+      OpenStruct.new(code: item["code"], description: item["description"])
+    }.compact.sort_by(&:code)
+  end
```
pezholio, line 155: Can we get a test for this please?

config/locales/codelists/BEIS/activity.en.yml
```diff
@@ -0,0 +1,8 @@
+---
+en:
+  covid19_related:
+    "0": Not related
+    "1": New activity full focus on COVID-19
+    "2": Existing activity adapted to fully focus on COVID-19
+    "3": New activity that will somewhat focus on COVID-19
+    "4": Existing activity adapted to somewhat focus on COVID-19
```
pezholio, line 8: Can we not use the Codelist introduced here somehow? They seem largely the same. https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/736/commits/9fec4002a851591798a2d1e193e6514fbd22750c

leeky: Might be a point to share with the rest of the team - there's lots of duplication between codelists and I18n which might be unnecessary.

@CristinaRO This probably belongs in the main `activity.en.yml`, for consistency.

CHANGELOG.md
```diff
@@ -382,6 +382,7 @@
 - Policy markers added to activity form, including BEIS custom answer `not assessed`
 - Add a field to report whether the activity has any relation to the Covid-19 pandemic.
 - Include the Covid-19 field in the CSV report.
+- Append the text "COVID-19" to the activity description in the IATI XML export, where applicable.
```
leeky, line 385: FYI - I usually try to stick to one CHANGELOG entry per PR.

leeky (approved):

app/helpers/codelist_helper.rb
```diff
@@ -147,6 +147,13 @@ def policy_markers_select_options
     filtered_list.unshift(not_assessed_option)
   end
 
+  def covid19_related_radio_options
```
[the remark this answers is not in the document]
CristinaRO: Possibly! From folklore, "codelists" were initially only used for IATI stuff, hence the harcoding that assumes IATI as a source of YAML.

The reality that BEIS have some of their own codes, which could be called "codelists", emerged in time. Is it here to stay? Would it be premature optimisation? I definitely didn't want to go off on a sidequest and lose track of the main narrative.

CHANGELOG.md
CristinaRO: I was told that the changelog entry should belong to the most relevant commit, and I like small commits, they're much easier to review than one giant pull request with a hundred changed files. This sort of leads into there being a few "main" commits where the spirit of the "thing being changed" is split between. 🤷🏻‍♀️

config/locales/codelists/BEIS/activity.en.yml
CristinaRO: > Can we not use the Codelist introduced here somehow? They seem largely the same.

@pezholio The pattern seems to be to keep "vendor data" such as codelists separate from "interface data" such as translation strings.

I interpreted this to mean that the "vendor" data is assumed to be a rigid set that we should keep intact so we can assimilate any changes to it without having to redo our tweaks, whereas we could choose (or have to) display the interface data in some other ways, with some edits.

> This probably belongs in the main activity.en.yml, for consistency.

@leeky I would counter that, for consistency with translations of the IATI codelists, it belongs in its own file in a BEIS structure. I didn't just slap a new file in, I thought about it and checked how it was done for the other codelists. If in fact there is no consistent pattern in the app, I propose we make a separate card for that, where we can compare and choose an approach that seems the most consistent at this point in time.

app/helpers/codelist_helper.rb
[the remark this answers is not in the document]
leeky: I'd leave it be in that case, and we can tidy it up when we encounter a second BEIS "codelist"

config/locales/codelists/BEIS/activity.en.yml
leeky: @CristinaRO My mistake, I've been doing so much work with model transations and forgot that the IATI code list version existed (I know you didn't just "slap a new file in") :-/

I note that that iati.en.yml contains translations that aren't part of the standard, so we should migrate these over at some point.

app/helpers/codelist_helper.rb
CristinaRO: I've added a test in the latest push!

config/locales/codelists/BEIS/activity.en.yml
CristinaRO: BTW, I'm keen to add these enhancements in the current sprint (in general, I know this is the last day of the current one). I have the experience of "refactoring sprints" being perpetually moved to the future, and I don't want to repeat it on this project! I just (you know, "just") don't want them to muddle the scope of the current PR.

CristinaRO:
Thank you for the reviews, @leeky and @pezholio ! I appreciate you going over the code and suggesting improvements, even if I push back on some of the suggestions.

PR #739, opened 2020-11-17
(1135) Fix intended beneficiaries for q3

MariluzHerrera:
Trello: https://trello.com/c/M9FaDIxY

## Changes in this PR

A decision has been made with the client that in order to facilitate Q3 reporting we are going to try replicate the behaviour we have seen on the trackers for the geography/intended beneficiaries step. For this we are now making `intended_beneficiaries` an optional step on all cases (included when user selects `recipient_region`) and we will now display a full list of all countries approved by OECD-DAC regardless of the region the user has selected. This is to reflect what we have seen on the trackers where users will select countries from different regions even if they have selected a specific region.

We are keeping the limit to 10 countries as this is the expectation we have set with BEIS.

We are also adding the missing regions back into the service, so now users are able to select any of those.

After Q3 reporting we will re-design this step from scratch, so these changes are just a quick fix to allow a less troubled reporting period this time.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -392,6 +392,7 @@
 - Policy markers added to the CSV report file
 - Activity Objectives added to the activity form, IATI XML and CSV report
 - Add missing regions back on RODA and open scope of country list for `intended_beneficiaries`
+- `Intended_beneficiaries` is now optional in all cases
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-22...HEAD
 [release-22]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-21...release-22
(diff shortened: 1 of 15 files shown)


CristinaRO (changes requested):
Thank you for walking me through the changes and explaining the context!

As we talked about on Slack, I would prefer to have fixes to the specs together with the commit that would have broken them. I think they belong together from a logical standpoint, and it makes it easier to follow the changes.

MariluzHerrera:
> Thank you for walking me through the changes and explaining the context!
> 
> As we talked about on Slack, I would prefer to have fixes to the specs together with the commit that would have broken them. I think they belong together from a logical standpoint, and it makes it easier to follow the changes.

No worries @CristinaRO! Thanks for reviewing my PR. I've placed the fix for the failing tests on the commit with the changes that broke them. Hope it's easier to follow now! :)

CristinaRO (approved):

CristinaRO:
Requires rebasing.

PR #740, opened 2020-11-17
Release 22

mec:
## Changes in this PR

- Fix allow application to create new users in Auth0
- Activity capital-spend is always exported as 0 to the IATI xml
- Fix allow application to create new users in Auth0
- Filter out unused aid types.
- Replace the hints in the aid type form with shorter, more accessible copy if available.
- Option `No - was never eligible` added to ODA eligibility form step
- Two original planned disbursements for the same activity, financial quarter
  and year cannot be created.
- Add a field to report Free Standing Technical Cooperation
- Policy markers added to activity form, including BEIS custom answer `not assessed`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -368,7 +368,7 @@
 - Answers for GDI form step have been modified
 - Use scripts to rule them all for development tasks
 
-## [unreleased]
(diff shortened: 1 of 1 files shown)


MariluzHerrera (approved):
🚀

PR #743, opened 2020-11-18
Remove legacy ingest code

leeky:
Remove the `IngestIatiActivities`, `IngestCsv` and `IngestCsvRow` classes used to ingest our initial dataset.

This has the additional benefit of removing three particularly slow tests.


pezholio (approved):
Amazing. I’m all for deleting slow tests too! :shipit:

PR #744, opened 2020-11-18
(1202) Add a Rake task to run the CSV importer

pezholio:
This adds a Rake task to run an Activity import with the command:

```
rake activites:import CSV=path/to/file.csv ORGANISATION_ID=8c3b69ec-1e9c-49ae-8e04-7c5d3826b253
```

This parses the CSV from the path and runs the importer. If the import is successful, the number of added / updated Activities is presented to the user. If there are any errors, these are played back to the user.

I've kept the tests pretty light touch, and only testing the code that gets introduced in the Rake task, as we don't want a whole bunch of CSV files clogging up the test suite, and we'd only be repeating ourselves anyway, as the everything is already tested in the `ImportFromCsv` class.

### DIFF
--- a/spec/support/rake_helpers.rb
+++ b/spec/support/rake_helpers.rb
@@ -1,0 +1,38 @@
+require "rake"
+
+module TaskExampleGroup
+  extend ActiveSupport::Concern
+
+  included do
+    def silent_warnings
+      old_stderr = $stderr
+      $stderr = StringIO.new
+      yield
+    ensure
+      $stderr = old_stderr
+    end
+
+    let(:task_name) { self.class.top_level_description.sub(/\Arake /, "") }
+    let(:tasks) { Rake::Task }
+
+    # Make the Rake task available as `task` in your examples:
+    subject(:task) { tasks[task_name] }
+
+    # Silence any errors that get sent to stderr when the task gets run
+    around(:example) do |example|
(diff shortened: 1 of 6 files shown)


pezholio:
[Add a Rake task to run the CSV importer](https://trello.com/c/HT1j4mFF/1202-add-a-rake-task-to-run-the-csv-importer)

spec/support/rake_helpers.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    # Silence any errors that get sent to stderr when the task gets run
+    around(:example) do |example|
+      silent_warnings { example.run }
+    end
+  end
+end
+
+RSpec.configure do |config|
+  config.define_derived_metadata(file_path: %r{/spec/tasks/}) do |metadata|
```
leeky, line 29: If you change this to `/spec/lib/tasks` then you can omit `type: :task` from the `describe` block and RSpec will infer this from the path 😃

leeky (approved):
Just one minor suggestion, but otherwise LGTM! :shipit:

spec/support/rake_helpers.rb
pezholio: Ah, good shout. You can tell I cargo-culted this from somewhere else can't you? 😆

PR #745, opened 2020-11-18
(1246) Ensure tests run in a random order each time

leeky:
I recently noticed that my tests were always using the same seed, instead of running in a different order each time, as per the RSpec configuration.

This fix ensures the seed is always regenerated every time, unless the `--seed` switch is passed.

Source: https://github.com/rails/spring/issues/113


pezholio (approved):
LGTM

PR #747, opened 2020-11-19
(1247) Fix intermittent test failures

leeky:
Move the `ClimateControl.modify` from the `around` block and inline them into the relevant tests.

I believe this is interfering with hooks that database_cleaner sets up in order to ensure that the database is cleared between test runs, as I found this also fixed things:

```
config.before(:each, type: :task) do
  DatabaseCleaner.strategy = :truncation
end
```

NB: I tried using `--bisect` to find the smallest number of tests needed to recreate the failure and it kept locking up after the first pass. I've fixed this by adding the `shell` bisect runner rather than the default `fork` runner.

### DIFF
--- a/spec/lib/tasks/import_activities_spec.rb
+++ b/spec/lib/tasks/import_activities_spec.rb
@@ -43,12 +43,6 @@ describe "rake activities:import", type: :task do
       CSV
     end
 
-    around do |example|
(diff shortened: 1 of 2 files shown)


pezholio:
Oooh, that would make sense, as we've only just started to see these intermittent fails

pezholio (approved):

PR #748, opened 2020-11-19
(1057) Forecast revision service

jcoglan:
## Changes in this PR

Apologies that this is fairly large PR. The functionality is explained below, and the core user-facing changes are the fairly small commit named `Revise planned disbursements`. Everything leading up to that is putting the services in place so that changing the UI-level functionality was a small change. There's a few commits after that that add some DB constraints and do some post-change refactoring.

It will probably be easier to read this commit-by-commit, and I'd appreciate a proof read of the commit messages to check they explain what's going on adequately. This code puts quite a few assumptions in place about how history works in this service and it's important we all understand them so that we keep the data in the service consistent.

Some design documents that went into this, if you need to refer to them:

- https://gist.github.com/jcoglan/dcb4e87652fd172a83f9263973cad2da
- https://gist.github.com/jcoglan/fbc49818accb78420433496b20b8adda
- https://gist.github.com/jcoglan/aa392f636a7304d0453992d5c642c170

### Before

Currently, the "financials" section for an activity lists all its planned disbursements with their type: "Original" or "Revised".

Screenshot 2020-11-19 at 11 10 26

The "add forecasted spend" button goes to a form where a financial quarter and value can be entered.

Screenshot 2020-11-19 at 11 10 52

The newly created forecast appears in the list.

Screenshot 2020-11-19 at 11 11 33

The "edit" link next to each item goes to an identical form where the quarter and value can be changed.

Screenshot 2020-11-19 at 11 11 52

We can change the quarter the forecast relates to, and this is reflected in the dates shown in the list.

Screenshot 2020-11-19 at 11 12 36

The system prevents us creating two forecasts for the same quarter; we get an error if we atempt this via the "add forecasted spend" action.

Screenshot 2020-11-19 at 11 17 22

However, it is possible to edit a forecast so that its quarter matches another, creating two forecasts for the same quarter. This data makes no sense.

Screenshot 2020-11-19 at 11 18 15

### After

The above exposes the problem of allowing the user to make arbitrary edits to data for which we store a historical log: they can edit it so that it no longer makes sense. The job of the services introduced in this PR is to store values entered by the user in a way that makes historical sense, without the user having to think about it.

They can add a new forecast, as before:

Screenshot 2020-11-19 at 11 21 46

This creates an "original" entry in the database with the quarter and value as given.

Screenshot 2020-11-19 at 11 24 12

The "edit" link now goes to a form where only the value can be changed; the quarter is fixed. A forecast relates to an activity and quarter, and we're updating the value for it, not changing what it relates to.

Screenshot 2020-11-19 at 11 24 44

The updated value is reflected in the list, and nothing else has changed, and no new records have been created.

Screenshot 2020-11-19 at 11 25 07

In the database, a single record still exists, and its value has been updated.

Screenshot 2020-11-19 at 11 25 33

Let's say we add some more forecasts for other quarters:

Screenshot 2020-11-19 at 11 26 29

We now have three records in the database, all of them "original", all belonging to the same activity but at different quarters. Note that all are linked to the same report -- the current active one for the activity's organisation.

Screenshot 2020-11-19 at 11 26 56

Now let's imagine that reporting cycle ends, the report is approved, and a new one is opened. We now edit the second of the above forecasts:

Screenshot 2020-11-19 at 11 31 37

This new value is reflected in the list, which still only shows three values:

Screenshot 2020-11-19 at 11 32 03

But in the database, we have four records. Because a new reporting cycle started, this value is associated with the new report, and the ones from the previous report are all retained.

Screenshot 2020-11-19 at 11 33 01

This demonstrates the core logic: a forecast linked to the current report can be modified, but those linked to older approved reports are persisted. "Editing" those values means adding new records that supersede them, not modifying the approved data.

## Next steps


jcoglan:
[Forecasts can be revised](https://trello.com/c/WB5Q9F9d/1057-forecasts-can-be-revised)

jcoglan:
[Update the view of forecasts to show the financial quarter and value of the most recent revision if there is one](https://trello.com/c/Uw6hDK2r/1055-update-the-view-of-forecasts-to-show-the-financial-quarter-and-value-of-the-most-recent-revision-if-there-is-one)

jcoglan:
In light of https://gist.github.com/jcoglan/aa392f636a7304d0453992d5c642c170, there are some data migrations I would also like to put in place before this is released, so that the production makes sense with the model implemented here, so please _DO NOT MERGE OR RELEASE THIS_ before I've had a chance to address those.

pezholio (commented):
Left a couple of comments, but I don't think I've got enough context to go any deeper

app/models/report.rb
```diff
@@ -29,11 +27,12 @@ class Report < ApplicationRecord
     where(state: [:active, :awaiting_changes])
   end
 
+  scope :for_activity, ->(activity) do
```
leeky, line 54: 👍

app/services/planned_disbursement_history.rb
```diff
@@ -55,16 +55,19 @@ def revise_entry(entry, value)
...
   end
 
   def update_entry(entry, value)
     entry.update!(value: value)
   end
 
-  def record_history?
-    !@activity.organisation.service_owner?
+  def current_report
+    unless @activity.organisation.service_owner?
```
leeky: I wonder if `Organisation` needs a `delivery_partner?` method so this could be `if @activity.organisation.delivery_partner?`, especially since this is the more common option.

spec/support/reporting_cycle.rb
```diff
@@ -3,14 +3,21 @@ def initialize(activity, financial_quarter, financial_year)
     @activity = activity
     @financial_quarter = financial_quarter
     @financial_year = financial_year
+    @report = nil
   end
 
   def tick
```
leeky, line 10: I'm wondering if there's a better name for this method. It's not obvious to me why calling it means subsequent `history.set_value()` calls are logged against future quarters.

leeky (approved):
LGTM 👍

app/controllers/staff/reports_controller.rb
```diff
@@ -83,67 +83,67 @@ def reports_for_delivery_partner
   end
 
   def inactive_reports
-    inactive_reports = policy_scope(Report.where(state: :inactive)).includes([:fund, :organisation])
+    inactive_reports = policy_scope(Report.inactive).includes([:fund, :organisation])
```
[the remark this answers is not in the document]
jcoglan: Thanks for spotting that, my T key is generating a lot of phantom keystrokes!

CHANGELOG.md
```diff
@@ -376,8 +376,6 @@
 - Filter out unused aid types.
 - Replace the hints in the aid type form with shorter, more accessible copy if available.
 - Option `No - was never eligible` added to ODA eligibility form step
-- Two original planned disbursements for the same activity, financial quarter
-  and year cannot be created.
```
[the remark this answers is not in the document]
jcoglan: Ah if it's gone into a release then yes, I'll amend that.

app/services/planned_disbursement_history.rb
```diff
@@ -0,0 +1,94 @@
...
+    latest_entry = find_latest_entry
+
+    return if latest_entry&.value == value
+
+    if latest_entry&.revised?
+      update_entry(latest_entry, value)
+    elsif latest_entry&.original?
+      revise_entry(latest_entry, value)
+    else
+      create_original_entry(value)
```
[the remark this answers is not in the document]
jcoglan: Yep that makes sense, I've just revised the control flow here.

app/services/planned_disbursement_history.rb
jcoglan: I wasn't sure if the categories "not a service owner" and "delivery partner" were the same tthing. If they are, I'm happy to add this method. I think the intent here is that reports are required from any organisation that isn't BEIS, and therefore needs their spending reviewed by BEIS.

spec/support/reporting_cycle.rb
jcoglan: I had the same problem and went with `tick` just for the purpose of making progress. What it actually does is simulate the turn-over of the reporting cycle where the current report gets approved, and a new one is opened for the DP to start adding stuff to. Any ideas for better words to use to capture this?

PR #749, opened 2020-11-19
(1102) Users can add activity objectives

MariluzHerrera:
Trello: https://trello.com/c/xSDokkYQ

## Changes in this PR

As part of the additional fields to be included in the activity form before Q3 reporting period, we are now adding the Aims/Objectives field on this PR. This is a free text area with no character limit, so the user can give a detailed explanation of the activity objectives. It comes after the Description field (same order in the trackers). It is mandatory for activities at levels B, C, and D. 

Apart from having it in the activity form, we need it on the CSV report file (see dedicated commit), and it gets reported to IATI (on another commit within this PR). The XML files have been check agains the IATI validator, passing with no errors.

The client provided us with the hint texts and titles that you can find in the translation files.

### After

Screenshot 2020-11-19 at 15 42 39

Screenshot 2020-11-19 at 11 40 55

## Next steps


CristinaRO (approved):

CristinaRO:
Requires rebase.

PR #751, opened 2020-11-20
(1203) Add missing fields to the activity importer

pezholio:
This adds the remaining missing fields to the activity importer. I've added each field (or group of fields) on a commit by 
commit basis, so there's a lot of commits, but should hopefully make things easier to follow.

### DIFF
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -113,6 +113,7 @@ module Activities
         :sustainable_development_goals_step,
         :covid19_related_step,
         :oda_eligibility_step,
+        :programme_status_step,
       ]
 
       attr_reader :errors, :activity
@@ -170,6 +171,7 @@ module Activities
         sdg_3: "SDG 3",
         covid19_related: "Covid-19 related research",
         oda_eligibility: "ODA Eligibility",
+        programme_status: "Programme Status",
       }
 
       def initialize(row)
@@ -276,6 +278,14 @@ module Activities
           oda_eligibility,
           :oda_eligibility,
           I18n.t("importer.errors.activity.invalid_oda_eligibility"),
+        )
+      end
+
+      def convert_programme_status(programme_status)
+        validate_from_codelist(
+          programme_status,
+          :programme_status,
+          I18n.t("importer.errors.activity.invalid_programme_status"),
         )
       end
 
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -121,6 +121,7 @@ module Activities
         :collaboration_type_step,
         :flow_step,
         :aid_type_step,
+        :fstc_applies_step,
       ]
 
       attr_reader :errors, :activity
@@ -191,6 +192,7 @@ module Activities
         collaboration_type: "Collaboration type (Bi/Multi Marker)",
         flow: "Flow",
         aid_type: "Aid type",
+        fstc_applies: "Free Standing Technical Cooperation",
       }
 
       def initialize(row)
@@ -340,6 +342,12 @@ module Activities
           :aid_type,
           I18n.t("importer.errors.activity.invalid_aid_type"),
         )
+      end
+
+      def convert_fstc_applies(fstc_applies)
+        raise I18n.t("importer.errors.activity.invalid_fstc_applies") unless ["1", "0"].include?(fstc_applies)
+
+        fstc_applies
       end
 
       def convert_call_open_date(call_open_date)
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -119,6 +119,7 @@ module Activities
         :sector_category_step,
         :sector_step,
         :collaboration_type_step,
+        :flow_step,
       ]
 
       attr_reader :errors, :activity
@@ -187,6 +188,7 @@ module Activities
         actual_end_date: "Actual end date",
         sector: "Sector",
         collaboration_type: "Collaboration type (Bi/Multi Marker)",
+        flow: "Flow",
       }
 
       def initialize(row)
@@ -319,6 +321,14 @@ module Activities
           collaboration_type,
           :collaboration_type,
           I18n.t("importer.errors.activity.invalid_collaboration_type"),
+        )
+      end
+
+      def convert_flow(flow)
+        validate_from_codelist(
+          flow,
+          :flow,
+          I18n.t("importer.errors.activity.invalid_flow"),
         )
       end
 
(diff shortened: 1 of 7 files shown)


CristinaRO:
[Add missing fields to the Activity importer](https://trello.com/c/RKw1YV9I/1203-add-missing-fields-to-the-activity-importer)

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -255,6 +259,21 @@
       expect(subject.errors.first.message).to eq(I18n.t("importer.errors.activity.invalid_intended_beneficiaries"))
     end
 
+    it "has an error if the GDI is invalid" do
+      new_activity_attributes["GDI"] = "Here is a string"
```
CristinaRO, line 332: When you say "we're given a code, not a string" - the code is still presented as a string, right?

app/services/activities/import_from_csv.rb
```diff
@@ -259,6 +261,12 @@ def convert_parent_id(roda_id)
         parent.id
       end
 
+      def convert_covid19_related(covid19_related)
+        raise I18n.t("importer.errors.activity.invalid_covid19_related") unless covid19_related_radio_options.map(&:code).include?(covid19_related.to_i)
```
CristinaRO, line 271: Ruby helpfully converts (some?) non-digit strings into 0, thus making the input appear valid. For example:

```
 [0, 1, 2, 3, 4].include?("bla".to_i)
=> true
```

Are we OK with that? Would it be ok to convert the options to strings and then check that the input is included in the stringified options?

CristinaRO (commented):
Partial review covering commits up to and including "Add ODA Eligibility".

config/locales/models/activity.en.yml
```diff
@@ -424,6 +424,11 @@ en:
         invalid_oda_eligibility: The ODA Eligibility option is invalid
         invalid_programme_status: The Programme status option is invalid,
         invalid_call_open_date: The Call Open Date option is in an invalid format
+        invalid_call_close_date: The Call Close Date option is in an invalid format
```
CristinaRO, line 427: Oh, I see, the Call Close string ended up in the ensuing commit. Up to you if you want to move it - I appreciate very, very much that you've split the big change into smaller commits, so I feel bad about asking to rewrite history to move one line. ;)

CristinaRO (commented):
Partial review covering from "Add programme status field" up to and including "Add aid type".

app/services/activities/import_from_csv.rb
pezholio: Good shout. I've reversed this, so we cast both sides to strings, so it's more reliable.

spec/services/activities/import_from_csv_spec.rb
pezholio: Oh yeah, this is just me putting any old rubbish in the field. I'll change it out, so it's clearer.

config/locales/models/activity.en.yml
pezholio: Nah, it's fine. That commit would have resulted in tests failing too (`spec/i18n_spec.rb`) and I like to make sure every commit has a green test status, so that helps keep me honest 🙂

config/locales/models/activity.en.yml
```diff
@@ -417,6 +417,7 @@ en:
         not_found: Cannot be found
         invalid_region: The region cannot be found
         invalid_country: The country code cannot be found
+        invalid_intended_beneficiaries: The intended beneficiaries are invalid
```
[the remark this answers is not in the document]
pezholio: I think it's fine for now, it's only to alert us that something's not quite right when we import. We can interate text when/if this becomes a user-facing class.

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -259,6 +262,7 @@
       expect(new_activity.sector).to eq(new_activity_attributes["Sector"])
       expect(new_activity.sector_category).to eq("112")
       expect(new_activity.collaboration_type).to eq(new_activity_attributes["Collaboration type (Bi/Multi Marker)"])
+      expect(new_activity.flow).to eq(existing_activity_attributes["Flow"])
```
CristinaRO: I see that you've fixed this to use `new_activity` in the next commit, but in the interest of green commits and whatnot, would you move the fix to this commit? :)

app/services/activities/import_from_csv.rb
```diff
@@ -121,6 +121,7 @@ class ActivityCreator
         :collaboration_type_step,
         :flow_step,
         :aid_type_step,
+        :ftsc_step,
```
CristinaRO: Should this be `fstc_applies_step` instead?

app/services/activities/import_from_csv.rb
```diff
@@ -144,8 +119,7 @@ def create
         @activity.level = calculate_level
         @activity.cache_roda_identifier
 
-        # TODO: This will eventually need to validate against all contexts (Activity::VALIDATION_STEPS)
-        return if @activity.save(context: IMPORT_VALIDATION_STEPS)
+        return if @activity.save(context: Activity::VALIDATION_STEPS)
```
CristinaRO, line 126: I just noticed that in the activity model, the `VALIDATION_STEPS` have an aid type step that doesn't follow the naming convention

```
    :sustainable_development_goals_step,
    :aid_type,
    :fstc_applies_step,
```
Line https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/models/activity.rb#L39

This wouldn't usually be part of the scope of this PR, but if we encounter bugs from that step, this is where we'd run into them here too. 😬

CristinaRO (changes requested):
One last hurdle! Thank you for your patience and perseverance!

CristinaRO:
There are also conflicts in the changelog, so presumably now is the time to rebase and fix them, when the PR is so close to being merged. \o/

app/services/activities/import_from_csv.rb
pezholio: Yeah, that's fair. I've added another step that addresses this 👍

app/services/activities/import_from_csv.rb
pezholio: 👍

spec/services/activities/import_from_csv_spec.rb
pezholio: Done!

pezholio:
@CristinaRO Have addressed and fixed everything. Should be good to go now!

CristinaRO (approved):

PR #752, opened 2020-11-20
(1249) Add BEIS ID and export it to the report csv

mec:
## Changes in this PR

To aid BEIS users in the period whilst the service is introduced and
runs along with the tracker spreadsheets we should capture and export
what is known as the BEIS ID.

This is an identifier used inside BEIS to track a single activity. Some
Delivery partner identifiers will use the BEIS ID and some will not, so
we should keep it separately to the delivery partner identifier to cover
all instances.

This field will not be visible in the application at all, it is simply
'passed through' to the report csv so up-stream systems can match a RODA
ID to an existing BEIS ID, with the goal of replacing BEIS IDs with RODA
IDs long term.

This fields value will only ever be set by historic data imported to the
service by the team during delivery partner on-boarding, never from an
end user.

Once the migration of all delivery partners to RODA is complete, this
field becomes redundant, but could remain and simply no longer appear in
the report csv.

The choice to neither restrict or validate this field is intentional as we
cannot guarantee the uniqueness of these IDs or their format and some
activities may not have one at all.

The work to import values into this column does not form part of this PR 
and will follow later.

## Next steps


mec:
[Capture and store the 'BEIS ID' to facilitate BEIS migration to using the RODA ID](https://trello.com/c/GzvOm347/1249-capture-and-store-the-beis-id-to-facilitate-beis-migration-to-using-the-roda-id)

app/presenters/activity_csv_presenter.rb
```diff
@@ -3,4 +3,9 @@ def intended_beneficiaries
     return if super.blank?
     to_model.intended_beneficiaries.map { |item| I18n.t("activity.recipient_country.#{item}") }.join("; ")
   end
+
+  def beis_id
+    return "" if super.blank?
```
pezholio, line 8: Do we need to do this? `nil.to_s` is functionally the same as doing this

pezholio (approved):
Looks good, just one question on the presenter, but not a blocker

app/presenters/activity_csv_presenter.rb
mec: 🤦‍♀️ yes, thanks for pointing this out! 😄

jcoglan:
Where did the docs for identifiers go? I have my [gist][1] but I don't recall where this got incorporated into the service's docs. It would be worth adding this new identifier to that document and sharing it; I've been pointing people to the gist in order to explain IDs and that doesn't feel sustainable in terms of maintaining the document.

[1]: https://gist.github.com/jcoglan/e525c9059e38f0c812069f6171c37652

PR #754, opened 2020-11-23
(1104) I can report ODA eligibility lead

pezholio:
## Changes in this PR

This adds the ODA eligibility lead to the Activity form. This is personal data, and we'll be dealing with the privacy implications of storing this later.

## Screenshots of UI changes

image

image

## Next steps


pezholio:
[I can report ODA Eligibility Lead](https://trello.com/c/3hZbWoXu/1104-i-can-report-oda-eligibility-lead)

app/models/activity.rb
```diff
@@ -78,6 +79,7 @@ class Activity < ApplicationRecord
   validates :policy_marker_disaster_risk_reduction, presence: true, on: :policy_markers_step, if: :requires_policy_markers?
   validates :policy_marker_nutrition, presence: true, on: :policy_markers_step, if: :requires_policy_markers?
   validates :oda_eligibility, presence: true, on: :oda_eligibility_step
+  validates :oda_eligibility_lead, presence: true, on: :oda_eligibility_lead_step, if: :requires_oda_eligibility_lead?
```
MariluzHerrera, line 82: I think it should be a `is_project?` method on the Activity model that does the same thing, if you want to use that one instead since it's there

config/locales/models/activity.en.yml
```diff
@@ -170,7 +172,7 @@ en:
...
           climate_change_adaptation: This activity intends to reduce the vulnerability of human or natural systems to the impacts of climate change and climate related risks, by maintaining or increasing adaptive capacity and resilience
           climate_change_mitigation: This activity contributes to the objective of stabilisation of greenhouse gas (GHG) concentrations in the atmosphere at a level that would prevent dangerous anthropogenic interference with the climate system by promoting efforts to reduce or limit GHG emissions or to enhance GHG sequestration
-          biodiversity: 'This activity promotes at least one of the three objectives of the Convention: the conservation of bio-diversity, sustainable use of its components (ecosystems, species or genetic resources), or fair and equitable sharing of the benefits of the utilisation of genetic resources'
+          biodiversity: "This activity promotes at least one of the three objectives of the Convention: the conservation of bio-diversity, sustainable use of its components (ecosystems, species or genetic resources), or fair and equitable sharing of the benefits of the utilisation of genetic resources"
```
MariluzHerrera, line 175: I just saw this. I wrote that and now I'm not sure there should be any quotation marks in there. I'll double check, but thanks for flagging :-)

spec/support/form_helpers.rb
```diff
@@ -273,6 +274,13 @@ def fill_in_activity_form(
     choose oda_eligibility
     click_button t("form.button.activity.submit")
 
+    if level == "project" || level == "third_party_project"
+      expect(page).to have_content t("form.label.activity.oda_eligibility_lead")
+      expect(page).to have_content t("form.hint.activity.oda_eligibility_lead")
+      fill_in "activity[oda_eligibility_lead]", with: oda_eligibility_lead
+      click_button t("form.button.activity.submit")
+    end
+
```
MariluzHerrera, line 283: Shall we include an assertion that the content of `oda_eligibility_lead` is displayed on the activity details page? (probably somewhere after line 333).

config/locales/models/activity.en.yml
pezholio: Yeah, this is my IDE - it makes changes using Prettier on save 🙂

MariluzHerrera (commented):
Thanks for this PR @pezholio! This looks good to me. I made a couple of suggestions: One to add an extra spec example that could be useful to have, and another suggestion to replace your `requires_oda_eligibility_lead?` method with `is_project?`, which a method we already have on the code for fields that only apply to levels C and D (for consistency, more than anything else). Other than that, it's good to go! :-)

config/locales/models/activity.en.yml
MariluzHerrera: I thought that might be the reason :)

spec/support/form_helpers.rb
pezholio: Oh, good shout 👍

MariluzHerrera (approved):
🚀

PR #756, opened 2020-11-25
(853) I can report the fund pillar on my Newton-funded activities

pezholio:
## Changes in this PR

This adds the Fund Pillar question to the Activity form. This only applies to Newton-funded Activities.

## Screenshots of UI changes

image

image

## Next steps


CristinaRO:
[I can report the fund pillar on my Newton funded activities](https://trello.com/c/I6bc2O9z/853-i-can-report-the-fund-pillar-on-my-newton-funded-activities)

app/models/activity.rb
```diff
@@ -325,10 +327,18 @@ def is_project?
...
+    parent.present? && associated_fund.is_gcrf_fund?
   end
 
   def is_newton_funded?
-    parent.present? && associated_fund.roda_identifier_fragment == "NF"
+    parent.present? && associated_fund.is_newton_fund?
+  end
+
+  def is_newton_fund?
+    roda_identifier_fragment == "NF"
```
CristinaRO, line 338: @leeky  Could I check with you that this way of identifying the funds is OK long-term, and better than checking that the string is contained within the fragment?

spec/factories/activity.rb
```diff
@@ -129,6 +138,14 @@
           create_list(:implementing_organisation, evaluator.implementing_organisations_count, activity: project_activity)
         end
       end
+
+      trait :newton_funded do
+        parent factory: [:fund_activity, :newton]
```
CristinaRO, line 143: I thought the parent of a project must be a programme. Shouldn't the parent of a `trait :newton_funded` project be a `:programme_activity, :newton_funded`?

spec/support/form_helpers.rb
```diff
@@ -337,6 +346,7 @@ def fill_in_activity_form(
         expect(page).to have_content policy_marker_nutrition
       end
     end
+    expect(page).to have_content fund_pillar if parent&.roda_identifier_compound&.include?("NF")
```
leeky, line 349: This is risky - should be using`parent&.associated_fund&.roda_identifier_fragment == "NF"`

If I created a GCRF-funded programme whose `roda_identifier_fragment` was, for example, `INFO01` this would pass.

CristinaRO, line 349: Could you use one of the newer ways to identify that the parent is either the Newton fund or Newton-funded? Also check with @leeky about the inclusion check possibly being too generous. https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/770#discussion_r535067144

CristinaRO (changes requested):
Couple of comments, definitely worth checking with @leeky around the fund identification. 😬

app/models/activity.rb
```diff
@@ -325,10 +327,18 @@ def is_project?
...
-    parent.present? && associated_fund.roda_identifier_fragment == "GCRF"
+    parent.present? && associated_fund.is_gcrf_fund?
   end
 
   def is_newton_funded?
-    parent.present? && associated_fund.roda_identifier_fragment == "NF"
+    parent.present? && associated_fund.is_newton_fund?
+  end
+
+  def is_newton_fund?
```
leeky, line 337: @pezholio I can see that this is useful in tests, however I can having this exposed in the `Activity` model risks folks using `is_newton_fund?` by mistake when they mean `is_newton_funded?`.

app/models/activity.rb
leeky: @CristinaRO Well, `roda_identifier` for a fund should never change (and we override `Activity#roda_identifier_compound=` to prevent it) as every child activity would also need to be updated if it did.

spec/support/form_helpers.rb
pezholio: Oh, good call

spec/factories/activity.rb
pezholio: Ah, you're right. Put that down to ignorance on my part. I missed the distinction between project and programme early on.

app/models/activity.rb
pezholio: Yeah, you're right. I'll try and think of some nicer wording.

spec/support/activity_helpers.rb
```diff
@@ -51,4 +51,10 @@ def page_displays_an_activity(activity_presenter:)
     expect(page).to have_content t("summary.label.activity.aid_type")
     expect(page).to have_content activity_presenter.aid_type
   end
+
+  def parent_is_newton?(parent)
```
CristinaRO, line 55: I found this method's name and argument name very confusing. In the context of this method, `parent` is just an activity that is being tested.

Could we use `associated_fund` for the naming instead?

We already use the notion of an associated fund https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/models/activity.rb#L218

 `associated_fund` returns itself if the activity is a fund, or it goes up the lineage until it finds the fund.

What do you think about this version?

`def associated_fund_is_newton?(activity)
  return if activity.nil?

  activity.associated_fund.roda_identifier_fragment == "NF"
end
```

CristinaRO (approved):
One non-blocking comment. (The naming of things!)

PR #758, opened 2020-11-25
Refactor/enforce report constraints

jcoglan:
## Changes in this PR

This PR includes a data migration on the `planned_disbursements` table and two new indexes on the `reports` table. These are described in full in the commit messages; the overall intent is to make sure data in production aligns with assumptions made by the `PlannedDisbursementHistory` and `PlannedDisbursementOverview` services. We do this either by correcting existing data or by putting indexes in place that ensure we cannot create data that breaks these assumptions.

## Next steps

### DIFF
--- a/db/migrate/20201124155809_enforce_one_historic_report_per_series.rb
+++ b/db/migrate/20201124155809_enforce_one_historic_report_per_series.rb
@@ -1,0 +1,10 @@
+class EnforceOneHistoricReportPerSeries < ActiveRecord::Migration[6.0]
+  def change
+    change_table :reports do |t|
+      t.index [:fund_id, :organisation_id],
+        where: "financial_quarter IS NULL OR financial_year IS NULL",
+        unique: true,
+        name: "enforce_one_historic_report_per_series"
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


db/migrate/20201124155809_enforce_one_historic_report_per_series.rb
```diff
@@ -0,0 +1,10 @@
+class EnforceOneHistoricReportPerSeries < ActiveRecord::Migration[6.0]
+  def change
+    change_table :reports do |t|
+      t.index [:fund_id, :organisation_id],
+        where: "financial_quarter IS NULL OR financial_year IS NULL",
```
leeky, line 5: Partial indexes are new to me 🤔

leeky (approved):
Some great work here to improve our data integrity 🎉 

LGTM! :shipit:

db/migrate/20201124155809_enforce_one_historic_report_per_series.rb
jcoglan: I only just learned about them but they're really neat for being able to scope uniqueness constraints.

jcoglan:
[Investigate any impact on forecasts already in production that changing how we collect them might bring.](https://trello.com/c/QldEUrcL/1062-investigate-any-impact-on-forecasts-already-in-production-that-changing-how-we-collect-them-might-bring)

PR #759, opened 2020-11-25
Mark all "historic" reports as approved

jcoglan:
We have a set of reports with no financial quarter; these represent the
sets of "historic" data we imported for an organisation during
onboarding, before they began using RODA. They are considered to predate
all quarterly reports, for the purposes of recording the history of our
data.

New data entered via RODA should _never_ be assigned to these reports.
Many objects like transactions and forecasts are automatically linked to
the current "editable" (active or awaiting_changes) report for that
organisation, and having these "historic" reports in an editable state
allows data entered via RODA to be linked to them.

By marking them as approved, we prevent the RODA service linking new
data to them, so we can only add data to them by explicitly choosing to
do so via an ingest process.

## Next steps

### DIFF
--- a/db/data/20201125123809_approve_historic_reports.rb
+++ b/db/data/20201125123809_approve_historic_reports.rb
@@ -1,0 +1,10 @@
+class ApproveHistoricReports < ActiveRecord::Migration[6.0]
+  def up
+    historic_imports = Report.where(financial_quarter: nil).or(Report.where(financial_year: nil))
+    historic_imports.update_all(state: "approved")
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end


leeky (approved):
LGTM! :shipit: 

I just wonder if we need to extract out another scope `historic_imports` just so we can include it in the migration, if it's not used anywhere else..?

jcoglan:
@leeky it's a fair point, I think I included it because I keep using it to check data in production, but it would be less necessary once some additional indexes are there that mean I don't need to check it. I keep wanting a guarantee that either `financial_quarter` and `financial_year` are both `NULL`, or neither of them are, but since the model/DB doesn't guarantee that yet I keep seaching for `financial_quarter IS NULL (AND|OR) financial_year IS NULL`.

I think I'll take it out and try to find a way that means I don't need it.

jcoglan:
For example, the indexes added in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/758 mean you can only have one report per `(fund_id, organisation_id)` for which _either_ `financial_quarter` or `financial_year` is `NULL`, which restricts the need for the sort of queries I've been doing.

jcoglan:
Alright, I've removed that model code and inlined the logic into the migration.

jcoglan:
[Investigate any impact on forecasts already in production that changing how we collect them might bring.](https://trello.com/c/QldEUrcL/1062-investigate-any-impact-on-forecasts-already-in-production-that-changing-how-we-collect-them-might-bring)

PR #760, opened 2020-11-25
(1100) Report GCRF Challenge Area

leeky:
## Changes in this PR

This adds the GCRF Challenge Area step to the Activity form, which only applies to GCRF funded programmes, projects and third-party projects.

#### Activity form
gcrf_challenge_area_form

#### Activity details page
gcrf_challenge_area_details

## Next steps


app/controllers/staff/activity_forms_controller.rb
```diff
@@ -75,6 +75,8 @@ def show
       skip_step unless @activity.is_project?
     when :sustainable_development_goals
       skip_step if @activity.fund?
+    when :gcrf_challenge_area
```
jcoglan, line 78: Is this code required to make the tests in the previous commit pass? i.e. does the build pass on the previous commit or not?

app/models/activity.rb
```diff
@@ -327,4 +327,8 @@ def requires_policy_markers?
   def is_project?
     project? || third_party_project?
   end
+
+  def is_gcrf_funded?
+    associated_fund&.roda_identifier_fragment == "GCRF"
```
jcoglan, line 332: Is there a situation where `associated_fund` returns `nil`? When does that happen?

app/models/activity.rb
```diff
@@ -79,7 +79,7 @@ class Activity < ApplicationRecord
   validates :policy_marker_disability, presence: true, on: :policy_markers_step, if: :requires_policy_markers?
   validates :policy_marker_disaster_risk_reduction, presence: true, on: :policy_markers_step, if: :requires_policy_markers?
   validates :policy_marker_nutrition, presence: true, on: :policy_markers_step, if: :requires_policy_markers?
-  validates :gcrf_challenge_area, presence: true, on: :gcrf_challenge_area_step, if: :is_gcrf_funded?
+  validates :gcrf_challenge_area, presence: true, on: :gcrf_challenge_area_step, if: -> { is_gcrf_funded? && !fund? }
```
jcoglan, line 82: Would it make sense to make `!fund?` part of the `is_gcrf_funded?` method? Or should that method also return `true` for the GCRF fund itself?

app/models/activity.rb
leeky: Yes, this was deliberate - see https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/760/files/2ac93b03a09e738c9120f9c0fa868cf7b2e4aff4#diff-cc9fe353a052960d819ee781d541ddc49305a1982a64712769f1d8a17b96ed6fR1210-R1214

When an activity is first created it is missing basic attributes which determine which `level` it is and what `parent` it belongs to. It makes sense for `associated_fund` to be `nil` in these cases, which is why we need the safe navigation operator here.

leeky: Of course, I could be more explicit, i.e.

```ruby
def is_gcrf_funded?
  return false if associated_fund.nil?

  associated_fund.roda_identifier_fragment == "GCRF"
end
```

app/models/activity.rb
leeky: I kept them separate to avoid having to refactor [these tests](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/760/files/144a311195f54c6c6b3c4cef19d7d3502898ab9a#diff-6b85c30d286a1b9fda4df24c1e2ed295ebcc21cd045ad03fc96939d9cf51a25dR273-R278), but perhaps I should tidy those up instead.

leeky: Ended up fixing the tests, and everything feels more logical now, so thanks 👍

app/services/activities/import_from_csv.rb
```diff
@@ -239,6 +240,15 @@ def convert_gdi(gdi)
...
       end
 
+      def convert_gcrf_challenge_area(area)
+        area = area.to_s
+        area = "Not applicable" if area == "N/A"
+
+        raise I18n.t("importer.errors.activity.invalid_gcrf_challenge_area") unless gcrf_challenge_area_options.values.map(&:to_s).include?(area)
+
+        gcrf_challenge_area_options.invert[area]
+      end
```
pezholio, line 250: I'm a little bit iffy about this. On the other fields of this nature, we accept codes, rather than text, becuase text is a bit more error prone, because of differences in punctuation, capitalisation etc. I think it might be better to be consistent.

spec/controllers/staff/activity_forms_controller_spec.rb
```diff
@@ -0,0 +1,64 @@
+require "rails_helper"
+
+RSpec.describe Staff::ActivityFormsController do
```
pezholio, line 3: I'm finding these specs tricky to follow. Can we have the expectations inline, rather than using custom matchers? Something like:

```
context "when editing a fund" do
...
  it "skips the parent step" do
    expect { get_step :parent }.to redirect_to(next_step)
  end
...
```

leeky: This is what I get for trying to adopt nested `let` blocks (which I generally loathe for readability!)

leeky: @pezholio On second thoughts, I'm wondering if using inline expectations coupled with custom matchers is best here - the important point to these tests is that *either* the current step is rendered *or* we redirect to the next step (whatever that might be), and I'd like to avoid yet another place where the order of the steps matters.

i.e. would prefer `expect { ... }. to redirect_to_next_wizard_step` over `redirect_to(wizard_path(... :identifier))`

pezholio: OK, cool. Let's do that. I think having the custom matcher more visible would help readability too 👍

pezholio (approved):
LGTM 👍

PR #762, opened 2020-11-25
Prevent forecast scopes from being NULL

jcoglan:
The "scope" of a forecast (planned disbursement) is the thing it relates
to: the object you plan to spend money on. A forecast is a plan to spent
money on a certain activity in a certain quarter, and so the fields
describing that object should not be nullable.

This migration prevents the parent_activity_id, financial_quarter, and
financial_year from being NULL. No records have NULL in these columns in
production, so this migration should deploy successfully.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_11_24_160000) do
+ActiveRecord::Schema.define(version: 2020_11_25_160306) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
@@ -173,12 +173,12 @@ ActiveRecord::Schema.define(version: 2020_11_24_160000) do
     t.string "receiving_organisation_type"
     t.string "receiving_organisation_reference"
     t.boolean "ingested", default: false
-    t.uuid "parent_activity_id"
+    t.uuid "parent_activity_id", null: false
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.uuid "report_id"
-    t.integer "financial_quarter"
-    t.integer "financial_year"
+    t.integer "financial_quarter", null: false
+    t.integer "financial_year", null: false
     t.index ["parent_activity_id", "financial_year", "financial_quarter", "planned_disbursement_type"], name: "unique_type_per_unversioned_item", unique: true, where: "(report_id IS NULL)"
     t.index ["parent_activity_id", "financial_year", "financial_quarter", "report_id"], name: "unique_report_per_versioned_item", unique: true, where: "(report_id IS NOT NULL)"
     t.index ["parent_activity_id"], name: "index_planned_disbursements_on_parent_activity_id"
--- a/db/migrate/20201125160306_prevent_null_planned_disbursement_scope.rb
+++ b/db/migrate/20201125160306_prevent_null_planned_disbursement_scope.rb
@@ -1,0 +1,7 @@
+class PreventNullPlannedDisbursementScope < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :planned_disbursements, :parent_activity_id, false
+    change_column_null :planned_disbursements, :financial_quarter, false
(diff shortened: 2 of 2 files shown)


db/schema.rb
```diff
@@ -195,7 +195,7 @@
     t.date "deadline"
     t.integer "financial_quarter"
     t.integer "financial_year"
-    t.index ["fund_id", "organisation_id"], name: "enforce_one_editable_report_per_series", unique: true, where: "((state)::text = ANY ((ARRAY['active'::character varying, 'awaiting_changes'::character varying])::text[]))"
+    t.index ["fund_id", "organisation_id"], name: "enforce_one_editable_report_per_series", unique: true, where: "((state)::text = ANY (ARRAY[('active'::character varying)::text, ('awaiting_changes'::character varying)::text]))"
```
jcoglan, line 198: I don't know why running `db:migrate` has changed this index definition. Does anyone know if that's a problem? This expression is something Rails has generated from the migration source, which has `where: "state IN ('active', 'awaiting_changes')"`.

pezholio: What version of Postgres are you using? Wonder whether that's related?

jcoglan: I'm using 13.0.

mec: @jcoglan @pezholio I've run `db:setup` and then `db:schema:load` from this branch on Postgres 11.6 (production runs 11.6 I think?) running in a container so it was clean and it seems to be happy - can I do anything else to confirm?

jcoglan: Ah, should I be running that version on my dev machine? Is this possible to do with Homebrew -- I imagine running multiple versions of Postgres isn't possible without containers, and trying to run 11 after migration my data files to 13 won't work...?

pezholio: Yeah, I dunno how easy that'd be. I have started to prefer using Docker for dev for this reason, but this also comes with its own problems! As long as everything seems fine, I think we can proceed without any problems.

mec: @jcoglan we haven't enforced it yet but I think it is a wise move to a least think about, given Homebrew seems to often just update postgres without really asking (in my experience) and, as you said, once your data goes up I don't think it can go back down?

You can have multiple installations of Postgres with Homebrew but you will have to manage which is started (with `brew services`) and you still have the internal data probelm (not sure how they map the data stores), I wonder if you can `pg_dump` the v13 and import to the v11, as in theory, that is just sql and run 11 for now?

I"ve re-run this migration against Postgres 11.6 and it reverts this change of syntax to:

`"((state)::text = ANY ((ARRAY['active'::character varying, 'awaiting_changes'::character varying])::text[]))"`

Given it is v11 in production (https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/d6b515655aa602f54447aeae73678109fe2d7bef/terraform/postgres.tf#L15) we should go with that flavour?

mec: Aside: @pezholio @jcoglan wisest approach I've seen is to run the 'backing services' (Postgres and Redis) in Docker and keep the Rails app as normal (out of Docker) - as we have pretty robust ways to mange Rails, Ruby and Gem versions already in place?

pezholio: That's a decent shout. The biggest pain around running Docker locally is managing Rails dependencies and having to rebundle every time there's a new or updated dependency

jcoglan: If we're running different Postgres versions then this bit of the schema is going to flip between different syntaxes whenever any of us runs `db:migrate`. I change this index again in #768 -- I'm wondering whether someone running 11.6 could regenerate the schema in that PR, we merge it, and then ignore changes to it in future PRs until we can put Postgres in Docker, which is probably the most reliable way to pin its version in development.

jcoglan: I'll take this change out of this PR as well.

jcoglan: @pezholio I'm on board with putting Postgres in Docker but running the app on the host. Having multiple versions of a DB on the same machine is much harder than having multiple versions of ruby and gems.

mec: I have PG 11 so I can run the mirgation on #762 let's get this one in - it looks good.

mec (approved):

PR #765, opened 2020-11-26
(1111) Report UK delivery partner named contact for project-level activities

CristinaRO:
## Changes in this PR

- Capture a new field, UK Delivery partner named contact, for project and third-party project level activities that are Newton funded

### After

TBA

## Next steps


CristinaRO:
[I can report UK DP Named Contact](https://trello.com/c/CblFACuD/1111-i-can-report-uk-dp-named-contact)

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -76,6 +77,8 @@ def show
       skip_step if @activity.fund?
     when :oda_eligibility_lead
       skip_step unless @activity.is_project?
+    when :uk_dp_named_contact
+      skip_step unless @activity.is_project? && @activity.is_newton_funded?
```
pezholio, line 81: I don't think we need `is_project?` if we have `is_newton_funded?` do we?

app/views/staff/shared/activities/_activity.html.haml
```diff
@@ -88,6 +88,15 @@
       - if policy(activity_presenter).update? && step_is_complete_or_next?(activity: activity_presenter, step: :sector)
         = a11y_action_link(t("default.link.#{activity_presenter.call_to_action(:sector)}"), activity_step_path(activity_presenter, :sector_category), t("summary.label.activity.sector"))
 
+  - if @activity.is_project? && @activity.is_newton_funded?
```
pezholio, line 91: Same here. I think a thing that's Newton funded is, by definition, a project

spec/support/form_helpers.rb
```diff
@@ -332,6 +339,9 @@ def fill_in_activity_form(
     end
     expect(page).to have_content oda_eligibility
     expect(page).to have_content oda_eligibility_lead if level == "project" || level == "third_party_project"
+    if (level == "project" || level == "third_party_project") && parent.is_newton_funded?
```
pezholio, line 342: I'm not sure this will work for projects, because (as I understand it) their parent is the fund itself, so `parent.is_newton_funded?` will return false, even if the direct parent is the Newton fund.

app/controllers/staff/activity_forms_controller.rb
CristinaRO: Programmes can also be Newton funded, and they sit above projects.

spec/support/form_helpers.rb
pezholio: Just when I think I understand everything! 😆

pezholio (approved):
LGTM - Sorry I got the wrong end of the stick on those comments!

PR #768, opened 2020-11-27
Restrict all potentially concurrent report states to one per series

jcoglan:
The state machine that reports operate looks like this:

                                     +-- awaiting_changes <--+
                                     |                       |
                                     |                       |
                                     V                       |
    inactive ------> active ------> submitted ------> in_review
                                                             |
                                                             |
                                                             V
                                                         approved

During the `active` and `awaiting_changes` states, the delivery partner
is allowed to change the report's contents -- these states are
"editable".

We added an index to allow only one report to exist in this state at a
time, for each set of reports scoped by (fund_id, organisation_id). The
purpose of this is to ensure the report sequence represents a totally
ordered history, because you can't have two editable reports open
concurrently that partner-entered data could be assigned to.

The problem is that this doesn't fully ensure the history is totally
ordered. For example, it allows the following sequence of state
transitions to occur on a pair of reports in the same series:

    Report 1            | Report 2
    --------------------+--------------------
    inactive            | inactive
    active              |
    submitted           |
                        | active
                        | submitted
    awaiting_changes    |
                        | approved

When Report 1 is active, new data is assigned to it. It is then
submitted, which allows Report 2 to enter an editable state -- the
current index only covers `active` and `awaiting_changes`. New data is
then assigned to Report 2.

Once Report 2 is submitted, it is possible under the current index for
Report 1 to go into awaiting changes, and therefore have more data
assigned to it. We've now broken the total ordering of report data by
adding data to Report 1, after we've added some to Report 2. Once we
start adding data in Report 2, it should be _impossible_ to further
change Report 1.

So, once a report enters the loop of states where it could become
editable again -- the states `active`, `submitted`, `in_review`, and
`awaiting_changes` -- we must not allow any other report to enter those
states, until the first report becomes `approved` and will never be
editable again.

The safest way to do that is to change the index condition so that any
state that's _not_ `inactive` or `approved` is covered by the uniqueness
rule. This does allow BEIS to create a new `inactive` report while the
current one is in the review cycle; they just can't activate it until
that review is concluded.

This constraint is currently obeyed in production and so this migration
should run successfully:

    > Report.where.not(state: %w[inactive approved]).count
    => 13
    > Report.where.not(state: %w[inactive approved]).map { |r| [r.fund_id, r.organisation_id] }.uniq.size
    => 13

## Next steps

### DIFF
--- a/db/migrate/20201127105312_restrict_report_states.rb
+++ b/db/migrate/20201127105312_restrict_report_states.rb
@@ -1,0 +1,12 @@
+class RestrictReportStates < ActiveRecord::Migration[6.0]
+  def change
+    remove_index :reports,
+      column: [:fund_id, :organisation_id],
+      name: "enforce_one_editable_report_per_series"
+
+    add_index :reports, [:fund_id, :organisation_id],
+      where: "state NOT IN ('inactive', 'approved')",
+      unique: true,
+      name: "enforce_one_editable_report_per_series"
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_11_25_172731) do
+ActiveRecord::Schema.define(version: 2020_11_27_105312) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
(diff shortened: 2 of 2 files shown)


jcoglan:
@mec This was the PR that needs the schema regenerating with Postgres 11.6.

mec:
@jcoglan I can't redo this migration unless we change it to:

````
remove_index :reports, column: [:fund_id, :organisation_id],  name: "enforce_one_editable_report_per_series"
````

Once I can redo it, the syntax of that index doesn't change from your commit - perhaps you ran this on an earlier version of Postgres or this particular syntax has not changed?

If you are happy to make this change, I think this can go ahead now - let me know?

jcoglan:
I'll change the `remove_index` line. That's surprising, the migration was reversible on my machine, and that's handled by Rails, not by Postgres, so I wonder why we're getting different behaviour.

Okay, let's leave the generated SQL as is and see if it causes diff noise after it's merged.

mec:
@jcoglan is it possible we are running different versions of Rails?

jcoglan:
@mec Not as far as I know -- `bundle` always complains if any of my installed gems aren't up to date with `Gemfile.lock`.

mec:
Okay, I've done it again. The steps I took, all running against PG 11.10 in docker:

- switch to this branch
- load the schema into the db from this branch with `rails db:schema:load`
- check the migration status, this migration is up
- take off this migration with `rails db:rollback`
- confirm the migration is down with `rails db:migrate:status`
- run the migration
- confirm the migration is up
- check the diff

No changes. 

I am not sure if this is the right thing or enough to give us the confidence we want on this issue?

Other than this oddity, I think this work is good to go in - so let me know what you would like to do?

jcoglan:
Yeah if you've done those steps, that should have regenerated `db/schema.rb` from your own database, so if it's not changed then I think we should merge it and then see if anyone else gets diff noise from it.

mec (approved):
🚢

PR #769, opened 2020-11-27
(1055) calculate variance by report history

jcoglan:
## Changes in this PR

This PR changes how `Activity#forecasted_total_for_report_financial_quarter` and therefore `Activity#variance_for_report_financial_quarter` work. Previously they were summing all the `planned_disbursements` values for the given activity and date range, by generating a date range from the given `Report`'s `created_at` field, and then calling `Activity#forecasted_total_for_date_range` with that date range.

Now, we use the `Report`'s `financial_{quarter,year}` fields, not `created_at`, to determine which forecasts we're interested in, and we use the version of that forecast that was current at the time of the given report. As we're now storing multiple versions of a forecast for a given activity and quarter, we need to grab the latest version as of the given report, not sum the value over all stored versions.

My main questions here are about whether these changes in terms of how we model time are OK. Assuming that `created_at` (the time a record is written) is the same as the logical time period it relates to makes me nervous, and makes things harder to test, so I'm happy to get rid of that, but I'm less certain about the move from date-based to quarter-based operation here.

I'm also not sure what this means for `Activity#forecasted_total_for_date_range` -- should this still exist? Should its functionality be changed?

#### Update

I discussed the above questions with @mec who helped me understand how `Activity#forecasted_total_for_date_range` is used in the CSV export code. This PR now also includes updates so that all date-based forecast logic is replaced with quarter-based logic that understands how forecasts are versioned. This affects how variances are calculated, and the "next four quarters" forecast functionality in the CSV export.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -401,6 +401,7 @@
 ## [unreleased]
 
 - Show only the financial quarter and value for forecasts on the activity page
+- Use the latest versions of forecasts from the report history to calculate variance
 - Add a new activity status: "Paused"
 - Update importer date format to be MM/DD/YYYY
 - Separate intended beneficiaries with a pipe
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -308,10 +308,6 @@ class Activity < ApplicationRecord
     !ingested? && is_project?
   end
 
-  def forecasted_total_for_date_range(range:)
-    planned_disbursements.where(period_start_date: range).sum(:value)
-  end
-
   def comment_for_report(report_id:)
     comments.find_by(report_id: report_id)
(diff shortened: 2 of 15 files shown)


jcoglan:
[Update the view of forecasts to show the financial quarter and value of the most recent revision if there is one](https://trello.com/c/Uw6hDK2r/1055-update-the-view-of-forecasts-to-show-the-financial-quarter-and-value-of-the-most-recent-revision-if-there-is-one)

mec (approved):
This looks great! I am glad to see the back of all the wraggling of dates, this feels much simpler to reason about now we use the financial quarter and year.

PR #773, opened 2020-11-30
(1213) developers can import forecasts

jcoglan:
## Changes in this PR

This PR adds a script that lets us import forecast data into the Q2 reports using a command-line script and CSV files. The behaviour of the import is described in detail in the commits. We plan to split the input data into one file per fund per delivery partner, so that each file's data can be assigned to a specific report. The interface for specifying which report to use is given in the last commit.

## Screenshots of UI changes

Screenshot 2020-11-30 at 11 26 25

## Next steps

### DIFF
--- a/app/services/import_planned_disbursements.rb
+++ b/app/services/import_planned_disbursements.rb
@@ -51,18 +51,23 @@ class ImportPlannedDisbursements
 
   def lookup_activity(roda_identifier)
     activity = Activity.by_roda_identifier(roda_identifier)
-    return activity if activity
 
-    @errors << "The RODA identifier '#{roda_identifier}' was not recognised."
-    nil
+    unless activity
+      @errors << "The RODA identifier '#{roda_identifier}' was not recognised."
+      return nil
+    end
+
+    unless Report.for_activity(activity).find_by(id: @report.id)
+      @errors << "The activity #{activity.roda_identifier} is not related to the report, which belongs to #{report_fund} and #{report_organisation}."
+      return nil
+    end
+
+    activity
   end
 
   def log_report_not_latest_error(latest_report)
-    organisation = @report.organisation.name
-    fund = @report.fund.roda_identifier
-
     message = [
-      "The report #{@report.id} (#{organisation}, #{quarter @report} for #{fund},",
+      "The report #{@report.id} (#{report_organisation}, #{quarter @report} for #{report_fund},",
       "#{@report.state}) is not the latest for that organisation and fund.",
       "The latest is #{latest_report.id}, for #{quarter latest_report} (#{latest_report.state}).",
     ]
@@ -72,5 +77,13 @@ class ImportPlannedDisbursements
 
   def quarter(report)
     "Q#{report.financial_quarter} #{report.financial_year}"
+  end
+
+  def report_fund
+    @report.fund.roda_identifier
+  end
+
+  def report_organisation
+    @report.organisation.name
   end
 end
--- a/app/services/import_planned_disbursements.rb
+++ b/app/services/import_planned_disbursements.rb
@@ -1,0 +1,34 @@
+require "csv"
+
+class ImportPlannedDisbursements
+  RODA_ID_KEY = "RODA identifier"
+  FORECAST_COLUMN_HEADER = /FC +(\d{4})\/\d{2} +FY +Q([1-4])/
+
+  def initialize(report:)
+    @report = report
+  end
+
+  def import(forecasts)
+    forecasts.each { |row| import_row(row) }
+  end
+
+  def import_row(row)
+    roda_identifier = row[RODA_ID_KEY]
+    activity = Activity.by_roda_identifier(roda_identifier)
+
+    row.each do |key, value|
+      match = FORECAST_COLUMN_HEADER.match(key)
+      next unless match
+
+      year = match[1].to_i
+      quarter = match[2].to_i
+
+      import_forecast(activity, quarter, year, value)
+    end
+  end
+
+  def import_forecast(activity, quarter, year, value)
+    history = PlannedDisbursementHistory.new(activity, quarter, year, report: @report)
+    history.set_value(value)
+  end
+end
--- a/app/services/import_planned_disbursements.rb
+++ b/app/services/import_planned_disbursements.rb
@@ -4,13 +4,27 @@ class ImportPlannedDisbursements
   RODA_ID_KEY = "RODA identifier"
   FORECAST_COLUMN_HEADER = /FC +(\d{4})\/\d{2} +FY +Q([1-4])/
 
+  attr_reader :errors
+
   def initialize(report:)
     @report = report
+    @errors = []
   end
 
   def import(forecasts)
-    forecasts.each { |row| import_row(row) }
+    latest_report = Report
+      .where(fund: @report.fund_id, organisation: @report.organisation_id)
+      .in_historical_order
+      .first
+
+    ActiveRecord::Base.transaction do
+      log_report_not_latest_error(latest_report) unless @report == latest_report
+      forecasts.each { |row| import_row(row) }
+      raise ActiveRecord::Rollback unless @errors.empty?
+    end
   end
+
+  private
 
   def import_row(row)
     roda_identifier = row[RODA_ID_KEY]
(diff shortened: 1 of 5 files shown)


jcoglan:
[dxw can import forecasts](https://trello.com/c/oc3e5aEA/1213-dxw-can-import-forecasts)

app/services/import_planned_disbursements.rb
```diff
@@ -0,0 +1,37 @@
...
+    forecasts.each { |row| import_row(row) }
+  end
+
+  def import_row(row)
+    roda_identifier = row[RODA_ID_KEY]
+    activity = Activity.by_roda_identifier(roda_identifier)
+
+    row.each do |key, value|
+      match = FORECAST_COLUMN_HEADER.match(key)
+      next unless match
```
pezholio, line 39: Should we raise an error if a column doesn't match this pattern? (assuming we throw away and use the RODA ID)

app/services/import_planned_disbursements.rb
```diff
@@ -34,4 +48,21 @@ def import_forecast(activity, quarter, year, value)
...
+
+  def log_report_not_latest_error(latest_report)
+    organisation = @report.organisation.name
+    fund = @report.fund.roda_identifier
+
+    message = [
+      "The report #{@report.id} (#{organisation}, #{quarter @report} for #{fund},",
+      "#{@report.state}) is not the latest for that organisation and fund.",
+      "The latest is #{latest_report.id}, for #{quarter latest_report} (#{latest_report.state}).",
+    ]
```
pezholio: Can this be in a translation? Will make it easier and cleaner to test

app/services/import_planned_disbursements.rb
```diff
@@ -28,7 +28,8 @@ def import(forecasts)
 
   def import_row(row)
     roda_identifier = row[RODA_ID_KEY]
-    activity = Activity.by_roda_identifier(roda_identifier)
+    activity = lookup_activity(roda_identifier)
+    return unless activity
```
pezholio, line 35: 👍

app/services/import_planned_disbursements.rb
```diff
@@ -58,6 +58,13 @@ def lookup_activity(roda_identifier)
     nil
   end
 
+  def convert_value(value, header:, roda_identifier:)
+    ConvertFinancialValue.new.convert(value)
+  rescue
```
pezholio, line 74: Do we want to rescue a specific error here?

script/import_forecasts.rb
```diff
@@ -0,0 +1,62 @@
...
+  args.on "-q", "--quarter QUARTER", Integer
+  args.on "-y", "--year YEAR", Integer
+  args.on "-i", "--input FILE"
+}
+
+options = {}
+parser.parse!(into: options)
+
+fund = Activity.fund.by_roda_identifier(options.fetch(:fund))
+organisation = Organisation.find_by(name: options.fetch(:organisation))
```
pezholio, line 16: I like fetching by name, rather than ID. Even though we have to find the exact name, it seems clearer

script/import_forecasts.rb
```diff
@@ -0,0 +1,62 @@
+require "optparse"
```
pezholio, line 1: Big fan of using scripts, rather than Rake BTW - may nick this approach for my import work 👍

pezholio: Can we get some tests for this script though? The lack of tests makes me a bit nervous

pezholio:
Just a couple of thoughts on this. I really like the general approach though 👍

app/services/import_planned_disbursements.rb
jcoglan: We can't do that in this case because the input files contain a lot of columns we should ignore -- they're there for informational purposes to make compiling the data easier. I did wonder about how to avoid situations where a column is intended to contain forecast values but they end up being ignored b/c they don't match this pattern.

app/services/import_planned_disbursements.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize(report:)
+    @report = report
+  end
+
+  def import(forecasts)
+    forecasts.each { |row| import_row(row) }
+  end
+
+  def import_row(row)
+    roda_identifier = row[RODA_ID_KEY]
```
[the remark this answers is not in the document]
jcoglan: I'm not a huge fan of mutating objects that were passed in by the caller, unless the caller can guarantee it will never use those objects afterward. In this case it doesn't help us guarantee all the columns are processed, per my earlier comment.

app/services/import_planned_disbursements.rb
jcoglan: In general I'd prefer that, although in this case I wanted to spefically test the exact message emitted, and not hide it behind a `t()` call, especially not one that takes 8 parameters.

app/services/import_planned_disbursements.rb
jcoglan: I'm not sure. Anything that causes an exception will cause the transaction to rollback, and will be printed to the shell where the user is running the importer. I'm happy to make this more user-friendly if we think it's worth the time investment -- I was trading off the fact I'm under a tight deadline, and this is code for a one-off task for devs to run, which will be thrown away immediately afterward.

script/import_forecasts.rb
jcoglan: My thinking here is I want the person running this to be really sure about which report they're adding the data to, and so I did this by having them type in the names of the fund and organisation, and then confirming that the selected report matches what they were intending. I'm sure there's lot of other UI designs one could do here and I want to make this as safe as possible, so happy to add/change anything that would help with that.

script/import_forecasts.rb
jcoglan: Again this was a trade-off of urgency vs return on investment. The classes that do the actual work here are tested, and the stuff in this script is mostly argument parsing, printing output, and asking for user input, which are all fairly painful to write tests for.

I'd be interested in knowing what looks risky here -- are there specific things we could write tests for, without exhaustively setting up tests that fake out the whole process of running a shell script? Any of this that could be moved into classes that we write unit tests for, without the complexity of the UI layer?

jcoglan:
I'm putting this back into draft because we need to amend how zero values are handled.

app/services/import_planned_disbursements.rb
jcoglan: I've had to refactor this to take in new requirements, and the latest version is now rescuing a specific error to catch this. This class no longer converts the value itself, it lets `PlannedDisbursementHistory` take care of that, so we need to rescue the conversion error to detect this type of failure.

app/services/import_planned_disbursements.rb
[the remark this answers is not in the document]
pezholio: Fair enough - I didn't realise there'd be columns we wouldn't be using either, so getting rid of the ID isn't worth the effort

app/services/import_planned_disbursements.rb
pezholio: 👍

app/services/import_planned_disbursements.rb
pezholio: Yeah, there are a lot of params, which could get messy

script/import_forecasts.rb
pezholio: Yeah, fair enough. It's not actually doing too much, so I think it's not really worth the pain of testing

jcoglan:
This is blocked on #778 being merged, in order to work correctly.

jcoglan:
#778 has been merged so this is ready for review again.

mec (approved):
Super nice PR. All explained beautifully!

Was there a reason to choose a script over a rake task?

jcoglan:
@mec I used a script because using `OptionParser` and `gets` inside of rake tasks has strange behaviour -- `rake` can interpret arguments to your task as arguments to itself, even if they're placed after the `--` word, and `gets` returns values from `ARGV` instead of reading from stdin.

I considered using the rake way of passing arguments but you run into problems with the way the shell and rake parse arguments with spaces in them, as in `rake forecasts:import['Royal Academy']`. I really wanted to use normal command-line switch-style arguments so that quoting works properly and the arguments have names, rather than having to reemember the argument order when this task takes five inputs.

PR #777, opened 2020-12-02
(1264) Validate countries from complete country list

pezholio:
Before, we were validating country codes from the recipient_country and intended_benificiaries codelists, which have the countries that are _currently_ valid. As we are importing historic data, we want to validate against **all** country codes, not just countries that are currently allowed to be added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -406,6 +406,7 @@
 - Update importer date format to be MM/DD/YYYY
 - Separate intended beneficiaries with a pipe
 - Add a field to record the UK delivery partner named contact for project-level activities
+- Validate country codes against all valid-country codes in the importer
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-23...HEAD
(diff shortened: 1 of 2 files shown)


pezholio:
[Importer validates against a list of all countries](https://trello.com/c/YSbJSm8d/1264-importer-validates-against-a-list-of-all-countries)

CHANGELOG.md
```diff
@@ -406,6 +406,7 @@
 - Update importer date format to be MM/DD/YYYY
 - Separate intended beneficiaries with a pipe
 - Add a field to record the UK delivery partner named contact for project-level activities
+- Validate country codes against all valid-country codes
```
mec, line 409: Could we just be clear it is only the importer that has changed?

mec (approved):
This is the right thing to do now! 👍 

We know this whole geography/beneficiaries issue needs to be resolved and is on the list for straight after Q3 reporting - right now we have to import mostly historic data that may contain countires that are no longer eligibile and this allows that to happen. 

🚢

CHANGELOG.md
pezholio: Yeah, makes sense. Have updated this and the commit message to be clearer

PR #778, opened 2020-12-02
Handle zero values in planned disbursement history

jcoglan:
## Changes in this PR

These changes are required in order to unblock #773, and they also improve how forecast entry works for end users. The commits go into more detail on the implementation, but the gist is:

When we see a value of `0` in an imported CSV, or when the user enters this value on the site, we should interpret this as saying there is no spending planned for that quarter. In general, we don't want to delete history, and so we handle this by storing the `0` so that it supersedes previous historical values. We avoid displaying zero-valued forecasts on the site, so from the user's point of view they're effectively deleted.

There are some exceptions to this behaviour:

- If there are no existing forecast records stored for that activity and quarter, then we ignore the input rather than storing a record with `value = 0`.
- If the activity is level C/D and therefore stores report-based history, and there is only one forecast record in the history for the quarter, and that record belongs to the current report, we delete it.

In all other situations, the `0` is stored as a new revision, so that it overrides previous values without deleting them.

### Before

Suppose we have a few forecasts stored for an activity:

Screenshot 2020-12-02 at 15 10 10

We go to edit one and set its value to zero:

Screenshot 2020-12-02 at 15 10 30

This is currently a validation error; planned disbursements are not allowed to have a zero value.

Screenshot 2020-12-02 at 15 10 44

### After

This is now not an error, and it acts to effectively delete the forecast, without destroying history. Whereas before, if there were any zero values in the database, they would be displayed:

Screenshot 2020-12-02 at 15 11 29

Now, they no longer appear -- setting a value of zero means you're asserting that no spending is planned in that quarter, and so we stop displaying it.

Screenshot 2020-12-02 at 15 11 48

However, the previous value is not destroyed; we represent the "deletion" by storing `0` as the latest value, superseding the previous one.

Screenshot 2020-12-02 at 15 12 46

## Next steps

### DIFF
--- a/app/models/planned_disbursement.rb
+++ b/app/models/planned_disbursement.rb
@@ -21,5 +21,5 @@ class PlannedDisbursement < ApplicationRecord
     :providing_organisation_reference,
     :financial_quarter,
     :financial_year
-  validates :value, inclusion: {in: 0.01..99_999_999_999.00}
+  validates :value, inclusion: {in: 0..99_999_999_999.00}
 end
--- a/app/services/planned_disbursement_history.rb
+++ b/app/services/planned_disbursement_history.rb
@@ -86,6 +86,8 @@ class PlannedDisbursementHistory
   end
 
   def create_original_entry(value, report = nil)
+    return if value == 0
+
     attributes = series_attributes.merge(required_attributes).merge(
       planned_disbursement_type: :original,
       value: value,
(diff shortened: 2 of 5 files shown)


jcoglan:
[Decide how to handle someone entering 0 on the forecast form](https://trello.com/c/1EZT7IZ7/1254-decide-how-to-handle-someone-entering-0-on-the-forecast-form)

mec (approved):
This looks good – the behaviour is exactly what I was expecting to see! 👍 

I have a few quieries around how all this plays out for a user in the applicaiton, but I don't think they relate directly to this work and I feel like it is a sound approach to get all the 'mechanics' around forecasts in place and then come back and review things in the application - I am wondering how best to caprture them, or if we even need to?

PR #779, opened 2020-12-03
(1261) Importer can import policy markers

pezholio:
This adds the policy marker fields for import. We want to treat blank fields as “not_assessed”, but the way the convertor works at the moment, we skip any blank fields. 

I’ve got round this by adding a constant called `ALLOWED_BLANK_FIELDS`. Rather than just checking a field is present before converting it, we additionally check is a field is in the `ALLOWED_BLANK_FIELDS` constant too. This way we can convert any fields that are left blank to read as `not_assessed`

### DIFF
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -168,11 +168,30 @@ module Activities
         actual_end_date: "Actual end date",
         sector: "Sector",
         collaboration_type: "Collaboration type (Bi/Multi Marker)",
+        policy_marker_gender: "DFID policy marker - Gender",
+        policy_marker_climate_change_adaptation: "DFID policy marker - Climate Change - Adaptation",
+        policy_marker_climate_change_mitigation: "DFID policy marker - Climate Change - Mitigation",
+        policy_marker_biodiversity: "DFID policy marker - Biodiversity",
+        policy_marker_desertification: "DFID policy marker - Desertification",
+        policy_marker_disability: "DFID policy marker - Disability",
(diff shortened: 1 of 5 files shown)


pezholio:
[Importer can import policy markers](https://trello.com/c/jYTzpuYz/1261-importer-can-import-policy-markers)

app/services/activities/import_from_csv.rb
```diff
@@ -238,6 +261,20 @@ def convert_gdi(gdi)
...
+        policy_markers_iati_codes_to_enum(policy_marker)
+      end
+      alias convert_policy_marker_gender convert_policy_marker
+      alias convert_policy_marker_climate_change_adaptation convert_policy_marker
+      alias convert_policy_marker_climate_change_mitigation convert_policy_marker
+      alias convert_policy_marker_biodiversity convert_policy_marker
+      alias convert_policy_marker_desertification convert_policy_marker
+      alias convert_policy_marker_disability convert_policy_marker
+      alias convert_policy_marker_disaster_risk_reduction convert_policy_marker
+      alias convert_policy_marker_nutrition convert_policy_marker
```
jcoglan, line 276: I like how this is organised -- avoiding quite a lot of boilerplate for defining the same conversion process for all these fields but using the allowed list to make it safe.

jcoglan (approved):

app/services/activities/import_from_csv.rb
pezholio: Ta 👍 I did consider using `method_missing`, but this seemed clearer

PR #781, opened 2020-12-03
(1257) Report Channel of delivery code

CristinaRO:
## Changes in this PR

- Allow Channel of delivery code to be imported from CSV and reported in the CSV

## Screenshots of UI changes

N/A

## Next steps


CristinaRO:
[I can report Channel of delivery code](https://trello.com/c/FVozLelR/1257-i-can-report-channel-of-delivery-code)

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -162,53 +164,41 @@
...
-        },
+        invalid_activity_attributes,
       ]
 
       expect { subject.import(activities) }.to_not change { existing_activity }
 
       expect(subject.created.count).to eq(0)
       expect(subject.updated.count).to eq(0)
-
-      expect(subject.errors.count).to eq(1)
```
leeky, line 180: As the `it` block starts with "has an error", we should probably leave the expectation to check that there is an error in, even if we're no longer checking for an exact error message.

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -162,53 +164,41 @@
...
```
leeky, line 206: Same here.

leeky (approved):
Two comments about restoring some expectations that were removed, but otherwise LGTM! :shipit:

leeky (approved):

PR #782, opened 2020-12-07
Ignore Ruby files when running Shellcheck

pezholio:
Currently, import_forecasts.rb is causing Shellcheck to fail when run via `script/test`. I've updated the Rake task to ignore any Ruby files that are in the `script` directory.


leeky (approved):
LGTM! :shipit:

PR #783, opened 2020-12-08
(1268) Import can create an activity implementing organisation

leeky:
Add the implementing organisation fields to `Activities::ImportFromCsv`:

 - Implementing organisation name
 - Implementing organisation reference
 - Implementing organisation sector (aka organisation type)

If all of the above are blank, we don't create an associated `ImplementingOrganisation`. However, if either "Implementing organisation name" or "Implementing organisation sector" are provided, these BOTH need to be present and valid to import the row correctly.

**NB: If an activity has more than one associated `ImplementingOrganisation`, these are all deleted and replaced with a single one when imported, which is by design. Obviously, this will need to work differently once end-users are able to use the importer directly.**

#  Next steps

### DIFF
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -85,9 +85,9 @@ module Activities
       def update
         return unless @activity && @errors.empty?
 
-        attributes = @converter.to_h
+        @activity.assign_attributes(@converter.to_h)
 
-        return if @activity.update(attributes)
+        return if @activity.save
 
         @activity.errors.each do |attr_name, message|
           @errors[attr_name] ||= [@converter.raw(attr_name), message]
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -6,7 +6,7 @@ module Activities
       end
 
       def csv_column
-        Converter::FIELDS[column] || column.to_s
+        Converter::FIELDS[column] || Converter::IMPLEMENTING_ORGANISATION_FIELDS[column.to_sym] || column.to_s
       end
     }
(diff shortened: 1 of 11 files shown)


app/services/activities/import_from_csv.rb
```diff
@@ -87,11 +88,27 @@ def update
...
+
+        @activity.implementing_organisations = [associated_implementing_organisation]
+
         return if @activity.save
 
         @activity.errors.each do |attr_name, message|
           @errors[attr_name] ||= [@converter.raw(attr_name), message]
         end
+
+        @errors.delete(:implementing_organisations)
```
pezholio: What does this line do?

app/services/activities/import_from_csv.rb
```diff
@@ -87,11 +88,27 @@ def update
 
         @activity.assign_attributes(@converter.to_h)
 
+        associated_implementing_organisation = ImplementingOrganisation.find_or_initialize_by(
+          activity_id: @activity.id
+        ) { |implementing_organisation|
+          implementing_organisation.name = row["Implementing organisation name"]
+          implementing_organisation.reference = row["Implementing organisation reference"]
+          implementing_organisation.organisation_type = row["Implementing organisation sector"]
+        }
```
pezholio: I feel like this (and the lines in the creator) are similar enough to have as a class of their own? There's a fair bit of repetition in the error lines too, so was wondering if that stuff could be encapsulated in the class too?

pezholio (approved):
Had a couple of issues around repetition, but noticed they were fixed in a subsequent commit! Punch it! :shipit:

PR #784, opened 2020-12-08
(1250) Add new fields to the activity importer

leeky:
Adds the following fields that were missing from the activity importer:

- BEIS ID (`beis_id`)
- UK DP Named Contact (NF) (`uk_dp_named_contact`)
- NF Partner Country DP (`country_delivery_partners`)

## Next steps


mec (approved):
The change here is so small, I was expecting more, that can only be a good thing! 😄 🚢

leeky:
@mec Yeah, I've double-checked the other fields in the checklist, and they're all included in the import, so 🥳

PR #785, opened 2020-12-08
(1305) Change Programme Status field to an enum

pezholio:
This updates the Programme Status field to be an enum, rather than a string. I've also tweaked the codelists to be in the BEIS folders, as these codes are actually BEIS, not IATI.

I've dispensed with my usual mantra of making sure all the tests pass for each commit, as there are quite a few changes here, and I wanted to make sure I explained everything I'd done, instead of lumping everything in as one commit, which would make it tricky to explain everything in a neat and tidy narrative.

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -3,7 +3,7 @@ class Staff::ActivityFormsController < Staff::BaseController
   include DateHelper
   include ActivityHelper
 
-  DEFAULT_PROGRAMME_STATUS_FOR_FUNDS = "07"
+  DEFAULT_PROGRAMME_STATUS_FOR_FUNDS = "spend_in_progress"
 
   FORM_STEPS = [
     :blank,
--- a/db/migrate/20201208093803_change_activity_programme_status_to_enum.rb
+++ b/db/migrate/20201208093803_change_activity_programme_status_to_enum.rb
@@ -1,0 +1,27 @@
+class ChangeActivityProgrammeStatusToEnum < ActiveRecord::Migration[6.0]
+  def up
+    add_column :activities, :programme_status_integer, :integer
+
+    Activity.find_each do |activity|
+      programme_status = Integer(activity.programme_status_before_type_cast, exception: false)
+      activity.update_column(:programme_status_integer, programme_status)
+    end
+
+    remove_column :activities, :programme_status, :string
+    rename_column :activities, :programme_status_integer, :programme_status
+  end
+
+  def down
+    add_column :activities, :programme_status_string, :string
+
+    Activity.find_each do |activity|
+      next if activity.programme_status.nil?
+
+      programme_status = "%02d" % activity.programme_status_before_type_cast
+      activity.update_column(:programme_status_string, programme_status)
+    end
+
+    remove_column :activities, :programme_status, :string
+    rename_column :activities, :programme_status_string, :programme_status
+  end
+end
--- a/vendor/data/codelists/BEIS/programme_status.yml
+++ b/vendor/data/codelists/BEIS/programme_status.yml
@@ -1,46 +1,42 @@
 ---
-attributes:
-  category-codelist:
-  name: ProgrammeStatus
-  complete: '1'
 data:
-  - code: '01'
+  - code: 1
     name: Delivery
     description: Activities related to delivery of ODA activities only
-  - code: '02'
+  - code: 2
     name: Planned
     description: Scoping negotiations and/or workshops have begun but no formal commitment has been made between the UK Delivery Partner. Scoping activity. Financial allocation has been made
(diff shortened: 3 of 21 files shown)


pezholio:
[Change Programme Status field to an enum](https://trello.com/c/ZYg0ovgt/1305-change-programme-status-field-to-an-enum)

db/migrate/20201208093803_change_activity_programme_status_to_enum.rb
```diff
@@ -0,0 +1,25 @@
+class ChangeActivityProgrammeStatusToEnum < ActiveRecord::Migration[6.0]
+  def up
+    add_column :activities, :programme_status_integer, :integer
+
+    Activity.all.each do |activity|
```
leeky, line 5: This should probably be `find_each` which avoids loading every Activity into memory at once.

db/migrate/20201208093803_change_activity_programme_status_to_enum.rb
```diff
@@ -0,0 +1,25 @@
...
+      programme_status = activity.programme_status.to_i
+      activity.update_attribute(:programme_status_integer, programme_status)
```
leeky, line 7: Would `update_column` be a better match here?

vendor/data/codelists/BEIS/programme_status.yml
```diff
@@ -1,46 +1,42 @@
 ---
-attributes:
-  category-codelist:
-  name: ProgrammeStatus
-  complete: '1'
 data:
-  - code: '01'
+  - code: 1
```
leeky, line 3: This codelist has been agreed by BEIS, so I'm wary about changing the `code`s from strings to integers here, without checking first.

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -3,7 +3,7 @@ class Staff::ActivityFormsController < Staff::BaseController
   include DateHelper
   include ActivityHelper
 
-  DEFAULT_PROGRAMME_STATUS_FOR_FUNDS = "07"
+  DEFAULT_PROGRAMME_STATUS_FOR_FUNDS = "spend_in_progress"
```
leeky, line 6: Might not matter, but this should probably be a symbol `:spend_in_progress`.

db/migrate/20201208093803_change_activity_programme_status_to_enum.rb
```diff
@@ -0,0 +1,25 @@
+class ChangeActivityProgrammeStatusToEnum < ActiveRecord::Migration[6.0]
+  def up
+    add_column :activities, :programme_status_integer, :integer
+
+    Activity.all.each do |activity|
+      programme_status = activity.programme_status.to_i
```
leeky, line 6: On production, we have 131 activities where `programme_status` is `nil` and 5 where it's `Replace me`. Are there any risks with these all being converted to `0`..?

db/migrate/20201208093803_change_activity_programme_status_to_enum.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    remove_column :activities, :programme_status, :string
+    rename_column :activities, :programme_status_integer, :programme_status
+  end
+
+  def down
+    add_column :activities, :programme_status_string, :string
+
+    Activity.all.each do |activity|
+      programme_status = "%02d" % activity.programme_status
```
leeky, line 18: As per above, should those current `0` programme statuses be changed back to `nil` here?

leeky (approved):
Just have some suggestions around the migration mainly, otherwise LGTM :shipit: 

It does seem that Activity is getting overloaded with `enums` which add lots of magic Rails methods, not sure if this impacts performance-wise though.

db/migrate/20201208093803_change_activity_programme_status_to_enum.rb
pezholio: Good shout. I've made this safer by using `Integer` with `exception: false`, so if it encounters anything that's not a number, it returns `nil`. I've also used `programme_status_before_type_cast`, rather than just `programme_status`, so we can migrate and revert safely with the enums in place.

vendor/data/codelists/BEIS/programme_status.yml
mec: @leeky I can confirm, all BEIS want is that the status are from their list - it is our decision to code them and how the coding is defined - generally speaking, I like the idea of coding the values BEIS define in the same way IATI do and that is one of the motivation for doing this work.

db/migrate/20201208093803_change_activity_programme_status_to_enum.rb
leeky: Didn't know about `Integer(value, exception: false)` 👍

PR #789, opened 2020-12-09
(1305) Fix programme status enum migration

leeky:
When this migration ran, some of the programme statuses were ending up being `nil` unexpectedly.

This is because calling `Integer` with a zero-prefixed number is treated as octal, so "08" and "09" are invalid, and `exception: false` was causing them to be set to `nil`


CristinaRO (approved):
I would like to understand more about various conversion things, such as why couldn't we use `to_i` (which assumes decimal), but as a one-time fix to have the correct data it's OK.

PR #792, opened 2020-12-09
(1200) Add missing fields to the CSV export

leeky:
Change the order of the CSV report file to match the data migration template and add the following columns which were previously missing:

- Free Standing Technical Cooperation (`fstc_applies`)
- SDG 1 (`sdg_1`)
- SDG 2 (`sdg_2`)
- SDG 3 (`sdg_3`)
- GCRF Challenge Area (`gcrf_challenge_area`)

## Next steps

### DIFF
--- a/app/services/export_activity_to_csv.rb
+++ b/app/services/export_activity_to_csv.rb
@@ -9,110 +9,11 @@ class ExportActivityToCsv
   end
 
   def call
-    [
-      activity_presenter.funding_organisation_name,
-      activity_presenter.transparency_identifier,
-      activity_presenter.delivery_partner_identifier,
-      activity_presenter.roda_identifier,
-      activity_presenter.beis_id,
-      activity_presenter.level,
-      activity_presenter.title,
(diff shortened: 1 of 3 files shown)


app/services/export_activity_to_csv.rb
```diff
@@ -127,11 +28,76 @@ def next_four_quarter_forecasts
     report_presenter.next_four_financial_quarters.map { |quarter, year| "Q#{quarter} #{year}" }
   end
 
+  private def columns
```
MariluzHerrera, line 31: Great idea to convert this into a hash to ensure we keep the order for headers and values!

MariluzHerrera (approved):
Thanks for this PR Leeky! 
Everything looks good. At first, I got a bit confused with the placeholder comments for the missing fields and I was unsure about whether they should be there. I had a chat with Meyric and everything looks clear now. 

I'm going to approve and merge this PR since you're not back for a few more days if you don't mind :-)

PR #793, opened 2020-12-10
(1300) Simplify transactions form

MariluzHerrera:
Trello: https://trello.com/c/cPiODeqP

## Changes in this PR

The current flow for reporting a transaction asks users to complete lots of fields, not all of which are needed. We can save users time and effort by streamlining the form.

Our proposed approach is to remove a number of fields that can be autopopulated by RODA in case they are needed. This PR removes the fields that are not necessary in both the `Add a transaction` form and the transaction details page, but retaining them in the data model. 

I have also modified the specs to reflect these changes.

### Before

Screenshot 2020-12-10 at 15 46 03

Screenshot 2020-12-10 at 15 47 14

### After

Screenshot 2020-12-09 at 16 27 22

Screenshot 2020-12-09 at 16 28 18

## Next steps

### DIFF
--- a/config/locales/models/transaction.en.yml
+++ b/config/locales/models/transaction.en.yml
@@ -30,7 +30,7 @@ en:
         value: Transaction amount
     legend:
       transaction:
-        date: Date transaction was made
+        date: Date of transaction
         providing_organisation: Providing organisation
         receiving_organisation: Receiving organisation
     hint:
--- a/spec/features/staff/users_can_create_a_transaction_spec.rb
+++ b/spec/features/staff/users_can_create_a_transaction_spec.rb
@@ -11,6 +11,21 @@ RSpec.feature "Users can create a transaction" do
     before { authenticate!(user: user) }
     let(:user) { create(:beis_user) }
 
+    scenario "the form only shows relevant fields" do
+      activity = create(:programme_activity, :with_report, organisation: user.organisation)
+
+      visit activities_path
+
+      click_on(activity.title)
+
+      click_on(t("page_content.transactions.button.create"))
+
+      expect(page).to have_content t("page_title.transaction.new")
+      expect(page).to have_content t("form.label.transaction.value")
+      expect(page).to have_content t("form.legend.transaction.date")
+      expect(page).to have_content t("form.legend.transaction.receiving_organisation")
+    end
+
     scenario "successfully creates a transaction on an activity" do
       activity = create(:programme_activity, :with_report, organisation: user.organisation)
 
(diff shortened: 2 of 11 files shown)


config/locales/models/transaction.en.yml
```diff
@@ -30,7 +30,7 @@ en:
         value: Transaction amount
     legend:
       transaction:
-        date: Date transaction was made
+        date: Date of transaction
         providing_organisation: Providing organisation
```
CristinaRO, line 34: I think the translation strings for the removed fields are not being used anywhere else, so we could remove them. (But if we're not sure, it's also ok to leave them in for now.)

CristinaRO (changes requested):
Needs rebasing because of conflicts in the changelog.

app/views/staff/shared/transactions/_table.html.haml
```diff
@@ -2,31 +2,19 @@
   %table.govuk-table.transactions
     %thead.govuk-table__head
       %tr.govuk-table__row
-        %th.govuk-table__header
```
CristinaRO, line 5: Ah, we've ended up working in parallel on the same thing and duplicating some work.

This card https://trello.com/c/O13SejbU/1299-the-actuals-transactions-summary-is-consistent-with-the-forecasts-summary is specifically about removing these fields from the view and replacing the date with quarter, so my PR being merged will create even more conflicts for you. :/

Let me know if you have any trouble with the conflicts, I'll be available to help.

config/locales/models/transaction.en.yml
MariluzHerrera: Yeah, I had a similar thought but I decided to leave them just in case

spec/features/staff/users_can_create_a_transaction_spec.rb
```diff
@@ -11,6 +11,26 @@
...
+
+      click_on(activity.title)
+
+      click_on(t("page_content.transactions.button.create"))
+
+      expect(page).to have_content t("page_title.transaction.new")
+      expect(page).to have_content t("form.label.transaction.value")
+      expect(page).to have_content t("form.legend.transaction.date")
+      expect(page).to have_content t("form.legend.transaction.receiving_organisation")
+      expect(page).to_not have_content t("form.label.transaction.description")
```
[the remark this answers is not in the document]
MariluzHerrera: I wrote this test to allow me to drive the changes on my code, and to make sure I wasn't leaving any non-relevant field in the form, but I see your point. I'll remove the `to_not have_content` expectations since it will definitely look odd in the future if I leave them there.

app/views/staff/shared/transactions/_table.html.haml
MariluzHerrera: I see! I'll have a look and give you a shout if I need help. Thanks Cristina!! :-)

MariluzHerrera: Okay, I've rebased and I think everything looks the way it should, but please double check and let me know if that's not the case.

CristinaRO (approved):

PR #796, opened 2020-12-11
Add local development user seeds

mec:
## Changes in this PR

A common scenario is for a developer to take a copy of production data
to run it locally.

In this situation rather than running all the database seeds it is
helpful to add the least amount of entities to the data to get the local
environment working.

The only data we really need is the users from the local development
Auth0 tenant.

This change allows a developer to add those users by running the
following command on the Rails console after duplicating the production
database locally:

`load File.join(Rails.root, "db", "seeds", "local_users.rb")`


mec:
D'oh! we already have `development_users.rb`!

PR #797, opened 2020-12-11
Update planned disbursement usage

jcoglan:
The intent of this PR is to amend all places in the application that display planned disbursements so that they only display the latest value for each quarter, rather than all the records linked to an activity. Some of those records will represent old historical data that has been superseded by newer entries.

I have amended the instances of this that I have found but I would like to know if there are others I've missed. I have removed the `Activity#planned_disbursements` association to discourage people from accidentally rendering all the historical records -- we should use `Activity#latest_planned_disbursements` instead.

### DIFF
--- a/app/controllers/staff/activities_controller.rb
+++ b/app/controllers/staff/activities_controller.rb
@@ -18,7 +18,7 @@ class Staff::ActivitiesController < Staff::BaseController
 
     @transactions = policy_scope(Transaction).where(parent_activity: @activity).order("date DESC")
     @budgets = policy_scope(Budget).where(parent_activity: @activity).order("period_start_date DESC")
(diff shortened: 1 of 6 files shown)


CristinaRO (approved):
Thanks for the clear separate commits! This looks thorough to me, but I'm not familiar enough with the application to think of any other places where we might display this info.

Can be merged after rebasing.

jcoglan:
> I'm not familiar enough with the application to think of any other places where we might display this info

My approach here was to review every place in the app where `planned_disbursement` or `PlannedDisbursement` appears, to look for places where sets of them are being loaded. Anywhere that I found we were loading _all_ the forecasts for an activity, I amended to load the _latest_ ones.

PR #798, opened 2020-12-11
Importer handles no implementing organisation

mec:
## Changes in this PR

I'll be honest and say I am struggling to follow how the
ImplementingOrganisationBuilder is working, but I do know that when
trying to import real data to production we have some issues.

I've written the tests to reproduce the issues and get around them,
although I feel like there is probably a better approach than this, with
time in short supply to get these imports done I think this potential
work-around is good enough.

I believe the issues are:

When the importer is creating an activity it assumes the import row will
have the details to make an implementing organisation and this is not
always the case.

When the importer is updating an existing activity it assumes the
activity has an existing implementing organisation or that the import
row has the details to create one, again this is not always the case.

When updating an activity we only add the result of trying to build an
implementing organisation if it is valid, for an implementing
organisation to be valid it will need the activity_id, name and type.

We cannot use this approach for creating new activities because the
activity will not have an id until it is saved.

It looks like we create a new activity and implementing organisation at the
same time and the implementing organisation will only become valid once
the new activity is saved and able to give the implementing organisation
its `activity_id`. Because these two things happen at the point the
activity is saved, it may be we are trying to save an invalid
implementing organisation, there is no way to know until the entire
activity is saved.

The fix here is if  Implementing organisation name or type is present
in the row we have to assume the user is trying to set an implementing
organisation for this row, and if one of the values is missing we need to tell them
with an error.

But if neither is present, we don't want to try and create an implementing
organisation at all.

### DIFF
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -89,7 +89,9 @@ module Activities
         @activity.assign_attributes(@converter.to_h)
 
         implementing_organisation_builder = ImplementingOrganisationBuilder.new(@activity, row)
-        @activity.implementing_organisations = [implementing_organisation_builder.build]
+        implementing_organisation = implementing_organisation_builder.build
+
+        @activity.implementing_organisations = [implementing_organisation] if implementing_organisation.valid?
 
         return if @activity.save
 
@@ -131,7 +133,10 @@ module Activities
         @activity.cache_roda_identifier
 
         implementing_organisation_builder = ImplementingOrganisationBuilder.new(@activity, row)
-        @activity.implementing_organisations = [implementing_organisation_builder.build]
+        if row["Implementing organisation name"].present? || row["Implementing organisation sector"].present?
+          implementing_organisation = implementing_organisation_builder.build
+          @activity.implementing_organisations = [implementing_organisation]
+        end
 
         return if @activity.save(context: Activity::VALIDATION_STEPS)
 
@@ -139,7 +144,7 @@ module Activities
           @errors[attr_name] ||= [@converter.raw(attr_name), message]
         end
 
-        implementing_organisation_builder.add_errors(@errors)
+        implementing_organisation_builder.add_errors(@errors) if @activity.implementing_organisations.any?
       end
 
       def calculate_level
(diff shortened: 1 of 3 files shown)


spec/services/activities/import_from_csv_spec.rb
```diff
@@ -273,8 +299,6 @@
       expect(new_activity.call_close_date).to eq(DateTime.parse(new_activity_attributes["Call close date"]))
       expect(new_activity.call_present).to eq(true)
       expect(new_activity.planned_start_date).to eq(DateTime.parse(new_activity_attributes["Planned start date"]))
-      expect(new_activity.planned_end_date).to eq(DateTime.parse(new_activity_attributes["Planned end date"]))
-      expect(new_activity.actual_start_date).to eq(DateTime.parse(new_activity_attributes["Actual start date"]))
```
MariluzHerrera, line 277: Are these two deleted lines intentional?

MariluzHerrera (approved):
Thanks for this PR @mec!

The commit message has helped me a lot to understand the issues around the ImplementingOrganisationBuilder, as I wasn't very familiar with this part of the project.

I agree with you that given the time constraints this fix gives us the functionality we need right now, so I'm happy to approve.

I have just left a question in regards some deleted expectations from one of the specs that I cannot see the relation to this piece of work, so not sure if they're intentional. They don't block this PR in any case 🚀

spec/services/activities/import_from_csv_spec.rb
mec: No! Thank you! How did I not see those. I'll go and have a look.

mec: Must have been a mec+vim typo!

mec:
@MariluzHerrera I spotted something else that would stop the import happening on this so have updated this PR, could you take another look.

app/services/activities/import_from_csv.rb
```diff
@@ -131,15 +133,18 @@ def create
...
+        end
 
         return if @activity.save(context: Activity::VALIDATION_STEPS)
 
         @activity.errors.each do |attr_name, message|
           @errors[attr_name] ||= [@converter.raw(attr_name), message]
         end
 
-        implementing_organisation_builder.add_errors(@errors)
+        implementing_organisation_builder.add_errors(@errors) if @activity.implementing_organisations.any?
```
MariluzHerrera, line 147: I'm guessing this is the latest change you committed. Well spotted! I hadn't noticed that without that conditional things wouldn't work as expected. 
Happy for this PR to get merged :-)

MariluzHerrera (approved):
I saw the changes. Thanks for letting me know @mec! 🚀

PR #799, opened 2020-12-11
(1318) export next twelve quarters forecasts

jcoglan:
The changes are covered by the commit messages, but to summarise them: we are changing the forecast data included in exported activity CSV.

Before this change, the export includes forecasts for the _four_ quarters following _today's date_.

After this change the export includes forecasts for the _twelve_ quarters following _the date of the report_.

### DIFF
--- a/spec/models/report_spec.rb
+++ b/spec/models/report_spec.rb
@@ -157,16 +157,14 @@ RSpec.describe Report, type: :model do
   end
 
   describe "#next_twelve_financial_quarters" do
-    it "returns an array with the next twelve financial quarters" do
-      travel_to(Date.parse("1 April 2020")) do
-        report = create(:report)
+    it "returns an array with the next twelve financial quarters from the date of the report" do
+      report = travel_to(Date.parse("1 April 2020")) { create(:report) }
 
-        expect(report.next_twelve_financial_quarters).to eq [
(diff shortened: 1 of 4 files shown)


jcoglan:
[Include the next twelve quarters' forecasts in export data](https://trello.com/c/j53fGhCU/1318-include-the-next-twelve-quarters-forecasts-in-export-data)

spec/models/report_spec.rb
```diff
@@ -157,16 +157,14 @@
...
-        report = create(:report)
-
-        expect(report.next_twelve_financial_quarters).to eq [
-          [2, 2020], [3, 2020], [4, 2020], [1, 2021],
-          [2, 2021], [3, 2021], [4, 2021], [1, 2022],
-          [2, 2022], [3, 2022], [4, 2022], [1, 2023],
-        ]
-      end
+    it "returns an array with the next twelve financial quarters from the date of the report" do
+      report = travel_to(Date.parse("1 April 2020")) { create(:report) }
```
MariluzHerrera, line 161: Good point! I saw this test in the previous commit and I didn't notice that the whole test was inside the do block, instead of just the report.

MariluzHerrera (approved):
Looks good to me. Thank you for the explanation on your commit message about wrapping the creation on the Report on the `travel_to` rather than the whole test. I don't think I would've noticed if you hadn't said it! 😅

PR #800, opened 2020-12-11
(1297) Update transactions to reflect actual disbursements

jcoglan:
The definition of actuals/transactions has changed over time and we need
to update production data to reflect this. There are two parts to this
problem:

- Some transactions are included in the data which do not fit our
  definition, and they need to be removed.
- Some transactions are labelled incorrectly, and they need to be
  relabelled to fit our definition.

Any transactions of type 'incoming commitments' (code `11`) and
'outgoing commitments' (code `2`) should be deleted

All transactions in the database should be of type 'disbursement' code
`3`. To achieve this we need to change the type for 'expenditure' (code
`4`) and 'credit guarantee' (code `10`) to code `3`.

### DIFF
--- a/db/data/20201209112603_transactions_reflect_actual_disbursements.rb
+++ b/db/data/20201209112603_transactions_reflect_actual_disbursements.rb
@@ -1,0 +1,10 @@
+class TransactionsReflectActualDisbursements < ActiveRecord::Migration[6.0]
+  def up
+    Transaction.where(transaction_type: [2, 11]).delete_all
+    Transaction.where(transaction_type: [4, 10]).update_all(transaction_type: 3)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end


jcoglan:
[Transactions in production reflect the actual disbursements and nothing else](https://trello.com/c/T8TEjZaa/1297-transactions-in-production-reflect-the-actual-disbursements-and-nothing-else)

MariluzHerrera (approved):
I've read the Trello card and the comments left by Dan, Meyric and you to get some understanding about this issue. I still have a pending question. 

In your commit message you said that all transactions in the DB should be code `3`, and according to the data migration from this PR you are updating all types `4` and `10` to code `3`, which looks good to me. I see you are deleting all `2` and `11` too. 

I was wondering if we have any other transaction types in the DB. There are 13 different types of transactions, according to the [codelist](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/chore/1297-transactions-reflect-actual-disbursements/vendor/data/codelists/IATI/2_03/transaction/transaction_type.yml) so just curious to know how we will handle other types of transactions if they are any. However, this question is not a blocker.

Overall, it looks good to me so I'll approve it. Thanks James!

jcoglan:
@MariluzHerrera Thanks for reminding me to check this before it goes out. Right now, these are all the transaction types that exist in production:

```rb
Transaction.pluck(:transaction_type).uniq
# => ["3", "2", "10", "4", "11"]
```

So I think this migration covers all the cases we need it to.

PR #801, opened 2020-12-11
(1098) CSV template now shows activity title

MariluzHerrera:
Trello: https://trello.com/c/V0k57Ky5

## Changes in this PR

On the CSV report, we were showing the activity description for the column `activity name`. The activity title would be more appropiate in this case. The activity title is only shown here to help guide users, we do not use this column from the template during import.

Specs have been modified accordingly.

## Next steps


CristinaRO (approved):
Can be merged after resolving the conflict in the changelog.

PR #806, opened 2020-12-14
(1320) Update the transaction importer

CristinaRO:
## Changes in this PR

- Transaction importer sets description from attributes
- Transaction importer does not process disbursement channel

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/spec/services/import_transactions_spec.rb
+++ b/spec/services/import_transactions_spec.rb
@@ -29,7 +29,6 @@ RSpec.describe ImportTransactions do
         "Receiving Organisation Type" => "80",
         "Receiving Organisation IATI Reference" => "",
         "Disbursement Channel" => "",
-        "Description" => "Fees for Q3",
       }
     end
 
@@ -50,7 +49,7 @@ RSpec.describe ImportTransactions do
         value: 50.0,
         receiving_organisation_name: "Example University",
         receiving_organisation_type: "80",
-        description: "Fees for Q3",
+        description: "Q4 1999-2000 spend on Example Project",
       )
     end
 
@@ -329,26 +328,11 @@ RSpec.describe ImportTransactions do
         ])
       end
     end
-
-    context "when Description is blank" do
-      let :transaction_row do
-        super().merge("Description" => "")
-      end
-
-      it "imports the transaction" do
-        expect(report.transactions.count).to eq(1)
-      end
-
-      it "generates a default description" do
-        transaction = report.transactions.first
-        expect(transaction.description).to eq("Q4 1999-2000 spend on Example Project")
-      end
-    end
   end
 
   describe "importing multiple transactions" do
     let :sibling_project do
-      create(:project_activity, organisation: project.organisation, parent: project.parent)
+      create(:project_activity, organisation: project.organisation, parent: project.parent, description: "Sibling Project")
     end
 
     let :first_transaction_row do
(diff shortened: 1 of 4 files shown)


CristinaRO:
[Update the transaction/actuals importer](https://trello.com/c/7FZVkV8e/1320-update-the-transaction-actuals-importer)

spec/services/import_transactions_spec.rb
```diff
@@ -398,20 +379,20 @@
...
 
     it "assigns each transaction to the correct report" do
-      expect(report.transactions.map(&:description).sort).to eq([
-        "Christmas Donation",
-        "Fees for Q3",
-        "Rent Payments",
+      expect(report.transactions.map(&:value).sort).to eq([
+        50,
+        150,
+        5000,
```
jcoglan, line 336: I'm wondering if there's some way we could make this test clearer -- previously the `description` functioned as a way of identifying the different transaction rows from the CSV, and it's harder to parse using just the `value` numbers.

CristinaRO: I've wondered that too. Is it worth giving the factoried activities custom names, so we can predictably test descriptions? (Otherwise it will be random Lorem ipsum.)

jcoglan: Yes, that sounds good -- so we could put `meaningful_name.value` in the assertions instead of `50`.

CristinaRO: Apologies it's taken me this long to get to this @jcoglan . Is the new approach better?

spec/services/import_transactions_spec.rb
```diff
@@ -399,19 +331,18 @@
...
+        "Q4 1999-2000 spend on Example Project",
+        "Q4 1999-2000 spend on Sibling Project",
+        "Q4 1999-2000 spend on Sibling Project",
       ])
     end
 
     it "assigns each transaction to the correct activity" do
-      expect(project.transactions.map(&:description)).to eq([
-        "Rent Payments",
+      expect(project.transactions.map(&:description).uniq).to eq([
```
leeky, line 341: This is testing the association between Transaction and Activity gets set up, so we could avoid having to update the strings in the future:

```ruby
expect(project.transactions).to be_present
expect(sibling_project.transactions).to be_present
```

leeky (approved):
Have suggested a couple of changes to the specs, but LGTM! :shipit:

spec/services/import_transactions_spec.rb
CristinaRO: I'm not sure I understand why it would be enough to test that some transactions are present. How are we sure they are, as the spec says "to the correct activity"? This is an area that I haven't worked in at all before, so assume zero knowledge (but more-than-zero ability to understand ;) ).

leeky: We're testing transactions fetched *through* the activity association, so therefore they have to belong to the 'correct activity', surely?

CristinaRO: That's what I don't get - in the hypothetical world where something went wrong, it could be that one of the transactions was assigned to another activity. Isn't that why we were previously testing the descriptions?

For example, the £5000 transaction could end up in the `project` instead of the `sibling_project`, and then we'd have a `"Q4 1999-2000 spend on Sibling Project"` showing up in `project.transactions.map(&:description).uniq`. I don't know how that could happen, but well, the spec was there, so I wanted to modify it to work with the new descriptions while keeping its spirit intact. 😬 😅

CristinaRO:
Thank you @leeky ! Updated and re-pushed.

PR #807, opened 2020-12-15
(1311) Fix bug on `programme_status` translation file

MariluzHerrera:
Trello: https://trello.com/c/qg4g3Xyd

## Changes in this PR

One of the options for programme status (`Stopped`) was not displaying correctly on the activity details page. We were getting an error that the translation was missing from the file. This was caused by a typo on the key for this option.

### Before

Screenshot 2020-12-15 at 10 16 17

### After

Screenshot 2020-12-15 at 10 24 16

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -433,6 +433,8 @@
 
 ## [unreleased]
 
+- Fix bug on option `stopped` for `programme_status`
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-25...HEAD
 [release-25]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-24...release-25
 [release-24]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-23...release-24
--- a/config/locales/codelists/BEIS/activity.en.yml
+++ b/config/locales/codelists/BEIS/activity.en.yml
@@ -17,6 +17,6 @@ en:
       spend_in_progress: Spend in progress
       finalisation: Finalisation
       completed: Completed
-      stoped: Stopped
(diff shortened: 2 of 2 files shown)


jcoglan:
Do you know where the code is that uses this translation? Missing translations are supposed to cause an exception so that we catch it during the build, rather than displaying "missing translation" in page text.

MariluzHerrera:
> Do you know where the code is that uses this translation? Missing translations are supposed to cause an exception so that we catch it during the build, rather than displaying "missing translation" in page text.

The `activity_presenter` [here](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/fix/1311-display-programme-status-on-activity-details-page/app/presenters/activity_presenter.rb#L46) access the translations and use them to display the information on the details page. 
I didn't know an exception should've been raised during the build. In fact, I think I've seen similar errors in the past and we probably still have some missing translations here and there (the translations for the browser tab tends to be one of them)

MariluzHerrera:
This is what I meant by translations for the browser tab:

Screenshot 2020-12-15 at 11 49 42

jcoglan:
It makes sense that this is happening in a presenter, as those need to use `I18n.t` rather than the safer `t` function available in views/helpers. I have an idea to fix that, but in the meantime let's merge this fix.

jcoglan:
Sorry, I didn't mean to imply this is a stop-gap fix that will be replaced later -- the fix to the locale file is definitely needed, I just want to make it easier to detect these bugs without them making their way into rendered pages :)

jcoglan (approved):

MariluzHerrera:
Sounds good to me. I'll merge this PR and we can implement the idea you had separately (probably worth writing a Trello card about it?). Thanks for reviewing this James!

PR #808, opened 2020-12-15
(1315) Activities importer sets hidden values on new activities

leeky:
## Changes in this PR

Update `Activities::ImportFromCsv` to:

- set `form_state` to "complete", so the form wizard will behave itself with imported activities
- set `status` to appropriate value based off `programme_status`
- set the funding and accountable organisations to BEIS, when creating a new activity

I've also added a data migration to correct the existing production data to reflect the above changes.

## Next steps


app/services/activities/import_from_csv.rb
```diff
@@ -288,6 +288,7 @@ def convert_to_attributes
         attributes[:recipient_region] ||= inferred_region
         attributes[:call_present] = (@row["Call open date"] && @row["Call close date"]).present?
         attributes[:sector_category] = get_sector_category(attributes[:sector])
+        attributes[:status] = infer_status(attributes)
```
MariluzHerrera, line 291: Just simple curiosity... Is there a reason why we pass all attributes here as an argument instead of passing just `programme_status`?

db/data/20201215171646_fix_imported_activity_defaults.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  private
+
+  def activities_with_nil_form_state
+    non_fund_activities.where(form_state: nil)
+  end
+
+  def activities_with_programme_status_but_no_iati_status
+    non_fund_activities.where(status: nil).where.not(programme_status: nil)
```
MariluzHerrera, line 46: I didn't know you could chain `where`s this way. Very useful!

MariluzHerrera (approved):
Thanks for this PR @leeky! 

I really like the migration file. It's so well organised, with meaningful method names. It was very easy to follow. 

I left a question about something I was curious about, but not a blocker. I think this PR is good to go! 🚀

app/services/activities/import_from_csv.rb
leeky: No reason, other than consistency with the other `infer_` method!

db/data/20201215171646_fix_imported_activity_defaults.rb
leeky: Yep - you might have used them already via scopes, e.g. `Activity.fund.where(...)` (`fund` is hiding a `where`)

PR #811, opened 2020-12-16
(1093) Planned end date cannot be earlier than planned start date

MariluzHerrera:
Trello: https://trello.com/c/MYdDqivE

## Changes in this PR

The service currently allows users to enter planned end date which is earlier than planned start date. We need to introduce validation to prevent this from happening, since it is more likely that users will accidentally introduce incorrect information when creating or updating an activity. 

The validator in this PR makes sure that `planned_end_date` is after, or at least the same, than `planned_start_date`

### After

Screenshot 2020-12-15 at 16 58 40

## Next steps

### DIFF
--- a/app/validators/end_date_after_start_date_validator.rb
+++ b/app/validators/end_date_after_start_date_validator.rb
@@ -1,0 +1,13 @@
+class EndDateAfterStartDateValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return unless value
+
+    unless end_date_after_start_date(record.planned_start_date, record.planned_end_date)
+      record.errors.add :planned_end_date, :not_before_start_date
+    end
+  end
+
+  private def end_date_after_start_date(start_date, end_date)
+    end_date >= start_date
+  end
+end
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -416,6 +416,16 @@ RSpec.describe Activity, type: :model do
       end
     end
 
+    context "when planned_end_date is not blank" do
+      let(:activity) { build(:activity) }
+
+      it "does not allow planned_end_date to be earlier than planned_start_date" do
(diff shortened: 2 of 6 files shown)


spec/models/activity_spec.rb
```diff
@@ -416,6 +416,16 @@
       end
     end
 
+    context "when planned_end_date is not blank" do
+      let(:activity) { build(:activity) }
+
+      it "does not allow planned_end_date to be earlier than planned_start_date" do
+        activity = build(:activity, planned_start_date: Date.today, planned_end_date: 1.day.ago)
```
leeky, line 423: For consistency with your other specs which use `Date.today` and `Date.tomorrow`, you could use `Date.yesterday` here.

app/validators/end_date_after_start_date_validator.rb
```diff
@@ -0,0 +1,13 @@
+class EndDateAfterStartDateValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return unless value
+
+    unless end_date_after_start_date(record.planned_start_date, record.planned_end_date)
+      record.errors.add :planned_end_date, I18n.t("activerecord.errors.models.activity.attributes.planned_end_date.not_before_start_date")
```
leeky, line 6: You could shorten this, as the validator already knows where to look in the I18n scopes:

```ruby
record.errors.add :planned_end_date, :not_before_start_date
```

leeky (approved):
I've added two tiny suggestions, but LGTM! :shipit:

spec/models/activity_spec.rb
MariluzHerrera: That's true! I didn't notice. Thanks for that!

MariluzHerrera:
Thanks for reviewing my PR @leeky! I really like your suggestions so I'll make the changes and then merge the PR :-)

PR #813, opened 2020-12-17
(1090) Organisation names must be unique

MariluzHerrera:
Trello: https://trello.com/c/gfrNbdqZ

## Changes in this PR

To prevent BEIS users from accidentally duplicate an organisation, we are now introducing validation on the Organisation model to make the name field unique, regardless of case sensitivity.

### After

Screenshot 2020-12-17 at 08 46 18

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -440,6 +440,7 @@
 - Activity importer sets BEIS as the funding and accountable organisation
 - Activity importer infers `status` from `programme_status`
 - Activity importer sets `form_state` to "complete" to ensure correct behaviour
+- Organisation name must be unique
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-25...HEAD
 [release-25]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-24...release-25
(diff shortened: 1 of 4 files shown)


pezholio (changes requested):
Build has failed. Think we might need to add something to the factories to make sure the Organisation names are unique now - maybe using Ffaker?

MariluzHerrera:
Never heard of Ffaker. I've seen we're using Faker::Company.name to generate the organisation names. 
Shall we do this as a separate piece of work?

pezholio:
Ha, sorry, Ffaker is a fork of Faker, I thought we were using that. It's the same deal really, but a bit faster. Not suggesting we move to that. We might have to add some kind of guard in to make sure the company name is unique. Could probably use a sequence. Something like:

```
name { "#{Faker::Company.name} #{n}" }
```

MariluzHerrera:
I'll have a look. Thanks @pezholio!

MariluzHerrera:
@pezholio If you have a moment to have a look at the changes. I read about sequences on Factory Bot and I think I'm using the correct syntax but I'd appreciate if you can check too. Thank you!

pezholio (approved):
LGTM 👍

PR #814, opened 2020-12-17
Move CI to Github Actions

pezholio:
This moves the CI to Github Actions. This should hopefully speed up build times, as well as starting the process of moving away from Travis (and paying them money), as well as keeping the CI closer to the code.

I've made a couple of tiny tweaks to the Docker config, mainly removing volumes from the docker-compose config, and adding a dependency to the test layer of the Dockerfile. This shouldn't have any impact on the project, as I don't think the docker-compose config is being used by anyone locally (shout if it is!).

I've added a couple of secrets to the Github config - namely the Coveralls Token and the AWS config (so we can get the Terraform testing working). 

Next stage is to get the deploy working, but I think it'll be better to do this in a seperate PR, once we're happy with this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,3 +14,5 @@ jobs:
       - uses: actions/checkout@v2
       - name: Run tests
         run: script/cibuild
+      - name: Test terraform scripts
+        run: script/terraform_test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


pezholio:
[Look into moving to GitHub actions](https://trello.com/c/PP6NjrA2/1332-look-into-moving-to-github-actions)

leeky (approved):
LGTM :shipit: 

It's a pity you don't seem to get realtime feedback of test progress any more - is there any way of fixing that? I note that it all shows up once the build is complete.

pezholio:
Yeah, I noticed that. I'll have a quick look and see if there's anything we can do about that

pezholio:
Doesn't seem like we can at first glance. We can always iterate later if we find a way.

PR #815, opened 2020-12-17
Change CI Badge to Github actions

pezholio:
This was missed off from #814, and badges are important 👍


leeky (approved):
It's shows up in the rich diff of README.md, so LGTM! :shipit:

PR #816, opened 2020-12-17
(1048) Show error message when user enters invalid dates

CristinaRO:
## Changes in this PR

- Show error message when the user tries to enter an invalid date on the `dates` step. Covers `planned_start_date`, `planned_end_date`, `actual_start_date`, `actual_end_date`

We are not able to reflect to the user what it was that they typed in. We are already swimming against Rails a bit, so it would take more effort and possibly make a more brittle feature if we tried to do that.

### Before

No feedback was given to the user.

### After
Screenshot 2020-12-17 at 22 35 44

## Next steps

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -141,12 +141,11 @@ class Staff::ActivityFormsController < Staff::BaseController
     when :country_delivery_partners
       @activity.assign_attributes(country_delivery_partners: country_delivery_partners)
     when :dates
-      @activity.assign_attributes(
-        planned_start_date: format_date(planned_start_date),
-        planned_end_date: format_date(planned_end_date),
-        actual_start_date: format_date(actual_start_date),
-        actual_end_date: format_date(actual_end_date),
-      )
+      assign_inputs_on_dates_step
+      if @activity.errors.present?
+        render_step :dates
+        return
+      end
     when :geography
       @activity.assign_attributes(geography: geography, recipient_region: nil, recipient_country: nil)
     when :region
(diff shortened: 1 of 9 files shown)


CristinaRO:
[Users are not given an error message when entering an invalid month](https://trello.com/c/7X5QfHFz/1048-users-are-not-given-an-error-message-when-entering-an-invalid-month)

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -282,24 +281,22 @@ def country_delivery_partners
...
-    {day: actual_start_date["actual_start_date(3i)"], month: actual_start_date["actual_start_date(2i)"], year: actual_start_date["actual_start_date(1i)"]}
+  def extract_date_parts_for(date_attr)
+    date_params = params.require(:activity).permit(date_attr)
+    {day: date_params["#{date_attr}(3i)"], month: date_params["#{date_attr}(2i)"], year: date_params["#{date_attr}(1i)"]}
   end
 
-  def actual_end_date
-    actual_end_date = params.require(:activity).permit(:actual_end_date)
-    {day: actual_end_date["actual_end_date(3i)"], month: actual_end_date["actual_end_date(2i)"], year: actual_end_date["actual_end_date(1i)"]}
+  def validate_inputs_on_dates_step
```
pezholio, line 289: Can we call this `assign_inputs_on_dates_step` instead to better reflect what this method is doing?

pezholio (approved):
LGTM, just one point about naming, but otherwise looks great!

app/controllers/staff/activity_forms_controller.rb
CristinaRO: Done, thank you for the review and for suggesting a better name!

PR #818, opened 2020-12-18
Fix spec that wasn't running

pezholio:
While I was investigating something else, I noticed that this spec was not named correctly, and actually failing. I've fixed the test to not worry about the order of the array, and also renamed the file, so it runs as part of the main test suite. I think this code is already touched in other specs, so I don't think it'll improve the coverage


MariluzHerrera (approved):
Looks good to me!!

PR #819, opened 2020-12-18
(1248) Codify ODA Eligibility

MariluzHerrera:
Trello: https://trello.com/c/hXMGMMGf

## Changes in this PR

For consistency throughout the service, we are now going to codify the options for `oda_eligibility`, using numbers instead of strings. This field is already an enum on the DB, so the changes on this PR are small modifications of the data in the yaml file and how we use this file to display the different options in the UI.

Likewise, we want to be consistent and ask users to provide codes when we can, instead of strings. Until now, `oda_eligibility` on the CSV import work was taken as a string. We are now collecting the numeric values instead. This carries a few changes on the importer tool, along with amendments to the specs.

I believe a DB migration is not needed since we already did one when we changed `oda_eligibility` column in the DB from a string to an integer (enum). We [migrated the existing data](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/db/migrate/20201102105419_change_oda_eligibility_type.rb) back then, so we should be okay about this, but please let me know if you disagree! :-)

## Next steps

### DIFF
--- a/app/helpers/codelist_helper.rb
+++ b/app/helpers/codelist_helper.rb
@@ -173,6 +173,15 @@ module CodelistHelper
     }.compact.sort_by(&:code)
   end
 
+  def oda_eligibility_radio_options
+    yaml = YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/BEIS/oda_eligibility.yml"))
+
+    Activity.oda_eligibilities.map do |name, code|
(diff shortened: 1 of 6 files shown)


pezholio (approved):
LGTM. This makes the import format more consistent, which I think was the aim of the card 👍

PR #822, opened 2020-12-21
Show error when cell contains invalid characters

pezholio:
At the moment, if a cell contains an invalid (i.e. non-UTF-8) character, we get a 500 error, because the template code attempts to show the value of the cell, and this raises an `Encoding::CompatibilityError`. We can get around this by catching that specific error and forcing the encoding of the cell. We can also display a slightly more friendly error back to the user. The wording is up for debate - it can proably be improved!

## Screenshots of UI changes

image


leeky (approved):
LGTM! :shipit:

CristinaRO:
I merged one of my PRs, so this needs updating again now. 🙈 

I will hold off on merging the other one (there are pending questions anyway, but even if they get solved), so you can be done with this one.

PR #825, opened 2020-12-21
(1335) Handle forecasted spend errors

CristinaRO:
## Changes in this PR

- Show error messages when the user tries to enter invalid values for a forecasted spend. Covers financial quarter being in the past and forecast value being an invalid number.

### Before

Server error

### After
Screenshot 2020-12-21 at 19 30 28
Screenshot 2020-12-21 at 19 30 50

## Next steps

### DIFF
--- a/app/controllers/staff/planned_disbursements_controller.rb
+++ b/app/controllers/staff/planned_disbursements_controller.rb
@@ -15,7 +15,16 @@ class Staff::PlannedDisbursementsController < Staff::BaseController
     authorize @activity
 
     history = history_for_create
-    history.set_value(planned_disbursement_params[:value])
+
+    begin
+      history.set_value(planned_disbursement_params[:value])
+    rescue PlannedDisbursementHistory::SequenceError
+      @planned_disbursement = PlannedDisbursement.new(planned_disbursement_params)
+      @planned_disbursement.parent_activity = @activity
+      @planned_disbursement.errors.add(:financial_quarter, :in_the_past)
+      render :new
+      return
+    end
 
     flash[:notice] = t("action.planned_disbursement.create.success")
     redirect_to organisation_activity_path(@activity.organisation, @activity)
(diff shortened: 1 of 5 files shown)


CristinaRO:
[Handle forecasts in the past error](https://trello.com/c/tGabvSls/1335-handle-forecasts-in-the-past-error)

app/controllers/staff/planned_disbursements_controller.rb
```diff
@@ -15,7 +15,16 @@ def create
...
 
     history = history_for_create
-    history.set_value(planned_disbursement_params[:value])
+
+    begin
+      history.set_value(planned_disbursement_params[:value])
+    rescue PlannedDisbursementHistory::SequenceError => _e
+      @planned_disbursement = PlannedDisbursement.new(planned_disbursement_params)
+      @planned_disbursement.parent_activity = @activity
+      @planned_disbursement.errors.add(:financial_quarter, t("activerecord.errors.models.planned_disbursement.attributes.financial_quarter.in_the_past"))
```
leeky, line 24: Might as well use `errors.add` as Rails intended, and let it do the I18n lookup  here:

```ruby
@planned_disbursement.errors.add(:financial_quarter, :in_the_past)
```

app/controllers/staff/planned_disbursements_controller.rb
```diff
@@ -15,7 +15,16 @@ def create
     authorize @activity
 
     history = history_for_create
-    history.set_value(planned_disbursement_params[:value])
+
+    begin
+      history.set_value(planned_disbursement_params[:value])
+    rescue PlannedDisbursementHistory::SequenceError => _e
```
leeky, line 21: If it's not needed, I'd omit the `=> _e` altogether.

leeky (approved):
LGTM! :shipit:

app/controllers/staff/planned_disbursements_controller.rb
CristinaRO: Symbol, not string! 🤦🏻‍♀️  (I had tried that before, but with `"in_the_past"`.) Thank you!

PR #833, opened 2021-01-04
Fix time sensitive specs

pezholio:
A couple of specs have started failing because we're now in another financial quarter, I've fixed these by using Rails's built in `travel_to` functionality to travel to a date where the tests will pass.

@CristinaRO also pointed out another test that will fail in the [Trello card](https://trello.com/c/rbYJza9l/1342-some-specs-are-written-in-a-time-brittle-way?menu=filter&filter=email), so I've fixed that too.

### DIFF
--- a/spec/features/staff/users_can_create_a_planned_disbursement_spec.rb
+++ b/spec/features/staff/users_can_create_a_planned_disbursement_spec.rb
@@ -93,6 +93,9 @@ RSpec.describe "Users can create a planned disbursement" do
     end
 
     scenario "they receive an error message if the forecast is not in the future" do
+      start_of_third_quarter = Date.parse("2020-10-01")
+      travel_to start_of_third_quarter
+
       project = create(:project_activity, :with_report, organisation: user.organisation)
       visit activities_path
       click_on project.title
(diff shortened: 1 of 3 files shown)


spec/features/staff/users_can_create_a_planned_disbursement_spec.rb
```diff
@@ -108,6 +111,8 @@
       )
 
       expect(page).to have_content t("activerecord.errors.models.planned_disbursement.attributes.financial_quarter.in_the_past")
+
+      travel_back
```
MariluzHerrera, line 115: I don't have much experience using Rails' `travel_to` functionality, so I have a question about this: If we don't use `travel_back`, would we still be in the same time frame for the rest of the tests?

pezholio: In some cases, yeah, so it's always good to get in the habit of travelling back after you've stubbed the time

MariluzHerrera (approved):
I like the way you solved the failing tests by adding the `travel_to` functionality. It looks very clean. Thanks!

When I was having a look at this issue I saw that in the UI, when a user wants to add a forecasted spend (with the financial quarter and year being prefilled) we are now getting the wrong financial year for some reason I couldn't figure out. The pre-filled information is Q4, FY 2021 - 2022 (FY should still be 2020-2021). Would fixing that be part of this PR or shall we make a new card? 
I'm happy with the changes on the specs, so I'll approve this PR anyway. Thanks Stu!

Here's the screenshot:

Screenshot 2021-01-04 at 11 26 08

PR #834, opened 2021-01-04
Fix deploy conditionals

pezholio:
I got the syntax wrong. Looking at the docs, this should work now.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,7 +44,7 @@ jobs:
           TF_VAR_google_tag_manager_environment_preview: ${{ secrets.STAGING_GOOGLE_TAG_MANAGER_ENVIRONMENT_PREVIEW }}
         run: |
           script/deploy-terraform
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           TF_VAR_google_tag_manager_environment_preview: ${{ secrets.STAGING_GOOGLE_TAG_MANAGER_ENVIRONMENT_PREVIEW }}
         run: |
           script/deploy-terraform
-        if: ${{ github.ref == "refs/heads/develop" }}
+        if: github.ref == 'refs/heads/develop'
```
leeky, line 47: I assume it's the single quotes that are significant here (the `${{ .. }}` is optional for conditionals..?

leeky (approved):

.github/workflows/deploy.yml
pezholio: Aye, I just removed the `${{ ... }}` while I was there, as I couldn't find many examples of them in the wild.

PR #839, opened 2021-01-05
(1310) Activity details page shows correct info for SDGs

MariluzHerrera:
Trello: https://trello.com/c/mO5Kbcdx

## Changes in this PR

Up until now, when a user selected the radio option that the Sustainable Development Goals were not applicable to a certain activity, the activity details page would not reflect this. Instead, the section for SDGs would appear blank and we could still see the edit button as `Add`. This could make users think that the app didn't save the value they just entered. In order to avoid this, this PR introduces changes on the activity presenter, so when sdgs_apply is set to false, the user will see `Not applicable` on the activity details page, and the edit button will appear as `Edit`.

Relevant specs have been fixed as part of this task, and I added some new ones to cover the new behaviour of the presenter at this step.

### Before

Screenshot 2021-01-04 at 15 34 08

### After

Screenshot 2021-01-04 at 15 01 55

## Next steps


pezholio (approved):
LGTM 👍

PR #840, opened 2021-01-05
(1313) Refactor away `funding organisation`

pezholio:
This PR removes the `funding_organisation_*` fields, because all Activities are funded by BEIS, and all level A funds are not reported to IATI anyway. Now we just return the Service Owner (which in our case will always be BEIS)

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -209,12 +209,6 @@ class Activity < ApplicationRecord
     organisation.default_currency
   end
 
-  def has_funding_organisation?
-    funding_organisation_reference.present? &&
-      funding_organisation_name.present? &&
-      funding_organisation_type.present?
-  end
-
   def has_accountable_organisation?
     accountable_organisation_reference.present? &&
       accountable_organisation_name.present? &&
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -239,7 +239,16 @@ class Activity < ApplicationRecord
   end
 
   def providing_organisation
-    return organisation if third_party_project? && !organisation.is_government?
+    third_party_project? && !organisation.is_government? ? organisation : service_owner
+  end
+
+  def funding_organisation
+    return nil if fund?
+
+    service_owner
+  end
+
+  def service_owner
     Organisation.find_by(service_owner: true)
   end
 
--- a/db/migrate/20210105131930_remove_funding_organisation_fields.rb
+++ b/db/migrate/20210105131930_remove_funding_organisation_fields.rb
@@ -1,0 +1,13 @@
+class RemoveFundingOrganisationFields < ActiveRecord::Migration[6.0]
+  def up
+    remove_column :activities, :funding_organisation_name, :string
+    remove_column :activities, :funding_organisation_reference, :string
+    remove_column :activities, :funding_organisation_type, :string
+  end
+
+  def down
+    add_column :activities, :funding_organisation_name, :string, default: "Department for Business, Energy and Industrial Strategy"
+    add_column :activities, :funding_organisation_reference, :string, default: "GB-GOV-13"
+    add_column :activities, :funding_organisation_type, :string, default: "10"
+  end
+end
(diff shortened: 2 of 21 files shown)


app/models/activity.rb
```diff
@@ -239,7 +239,14 @@ def associated_fund
   end
 
   def providing_organisation
-    return organisation if third_party_project? && !organisation.is_government?
+    third_party_project? && !organisation.is_government? ? organisation : service_owner
+  end
+
+  def funding_organisation
+    fund? ? nil : service_owner
```
leeky, line 240: I'm never keen on ternary operators, particularly when combined with methods that end with `?`.

My preference would be:

```ruby
def funding_organisation
  return nil if fund?

  service_owner
end
```

But feel free to ignore 😄

spec/support/shared_examples/behaves_like_valid_activity_xml.rb
```diff
@@ -20,9 +20,13 @@
 
   it "contains the funding organisation XML" do
     visit organisation_activity_path(organisation, activity, format: :xml)
-    expect(xml.at("iati-activity/participating-org[@role = '1']/@ref").text).to eq(activity.funding_organisation_reference)
-    expect(xml.at("iati-activity/participating-org[@role = '1']/@type").text).to eq(activity.funding_organisation_type)
-    expect(xml.at("iati-activity/participating-org[@role = '1']/narrative").text).to eq(activity.funding_organisation_name)
+    if activity.funding_organisation.present?
```
leeky, line 23: Not sure how else you'd do this with shared examples, but we now have a test "contains the funding organisation XML" which in the case of a fund, expects that it doesn't.

spec/features/staff/users_can_create_a_fund_level_activity_spec.rb
```diff
@@ -51,20 +51,6 @@
       expect(page.find("option[@selected = 'selected']").text).to eq activity_presenter.flow
     end
 
-    scenario "the activity has the appropriate funding organisation defaults" do
-      identifier = "a-fund-has-a-funding-organisation"
-
-      visit activities_path
-      click_on(t("page_content.organisation.button.create_activity"))
-
-      fill_in_activity_form(delivery_partner_identifier: identifier, level: "fund")
```
leeky, line 60: Any removal of `fill_in_activity_form` is 🎉

spec/models/activity_spec.rb
```diff
@@ -736,14 +736,6 @@
       it { should validate_presence_of(:extending_organisation_id).on(:update_extending_organisation) }
     end
 
-    context "when the form state is blank" do
-      it "allows updates to be made to other fields set on creation" do
-        blank_activity = create(:activity, funding_organisation_name: "old", form_state: :blank)
```
leeky, line 741: Was originally wondering if test should be kept in, and another field substituted for `funding_organisation_name` in this example, but since the `:blank` form state is only here for legacy reasons, it seems fine to remove this 👍

db/migrate/20210105131930_remove_funding_organisation_fields.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveFundingOrganisationFields < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :activities, :funding_organisation_name, :string
```
leeky, line 3: Might be completely unnecessary, but rolling this back should probably fill in those fields with BEIS details rather than leaving them blank.

spec/models/activity_spec.rb
```diff
@@ -987,6 +987,20 @@
       end
     end
 
+    describe "#funding_organisation" do
+      let!(:beis) { create(:beis_organisation) }
+
+      it "returns BEIS if the activity is a project" do
```
leeky, line 971: Worth adding cases for programme and third-party project to be on the safe side?

leeky (approved):
Have left some really minor comments, but LGTM! :shipit:

PR #841, opened 2021-01-05
(1314) Remove form steps label from child activities view

MariluzHerrera:
Trello: https://trello.com/c/h6QNiHF6

## Changes in this PR

This is an old feature that shows if a user has completed all the form steps of an activity or not, which used to be an indicator of an activity being valid. This is no longer the case and this information has no value for the users anymore, so we are now removing this column from the child activities view. This applies to programmes, projects and third-party projects.

I have also removed the specs related to this feature.

### Before

Screenshot 2021-01-05 at 11 58 08

### After

Screenshot 2021-01-05 at 15 07 26

Screenshot 2021-01-05 at 15 10 19

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -457,6 +457,7 @@
 - SDGs on activity details page now shows `Not applicable` when the user selects this option on the form
 - Refactor away `funding organisation` field
 - Forecasted spend form always includes the current financial year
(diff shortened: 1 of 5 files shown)


leeky (approved):
LGTM! :shipit:

PR #842, opened 2021-01-05
(1312) Refactor away the status field for activities

CristinaRO:
## Changes in this PR

- IATI status is calculated on the fly from the programme status

## Screenshots of UI changes

N/A

## Next steps


app/models/activity.rb
```diff
@@ -364,4 +364,8 @@ def is_newton_funded?
   def requires_country_delivery_partners?
     is_newton_funded? && programme?
   end
+
+  def status
```
leeky, line 368: As this is only used in the IATI XML view, is `ActivityPresenter` a better place for it - maybe renamed to `iati_status`?

spec/models/activity_spec.rb
```diff
@@ -1402,4 +1402,34 @@
       expect(fund.is_newton_funded?).to be_falsey
     end
   end
+
+  describe "#status" do
```
leeky, line 1406: Aren't these already covered by `ProgrammeToIatiStatus` tests?

CristinaRO: Yep, but this was when I wasn't sure we'd be keeping `ProgrammeToIatiStatus`, pending my questions to Meyric. As it is, I think I'll wait until #843 is merged, then use the new `load_yaml` method to do that, possibly in the presenter like you suggested, and then remove the service object after all.

app/models/activity.rb
CristinaRO: I've renamed it, but it turns out that we don't use the presenter to show the activity in XML format, so it's still in the model.

leeky (approved):
LGTM! :shipit:

spec/models/activity_spec.rb
leeky: Good call! The YAML files are the right place to store these BEIS-approved mappings 🎉

PR #843, opened 2021-01-05
Refactor helper to use `load_yaml` for BEIS codes

CristinaRO:
## Changes in this PR

Refactor the codelist helper to standardise the way we load BEIS codes. IATI is still the default source.

Code lists referred originally only to IATI codes.

However, practice has shown that there are a lot of BEIS-specific codes that we are attempting to group and load in a consistent way. Having a recognised, common pattern to load BEIS codes would make it easier to add new ones in the future by decreasing the cognitive load of figuring it out from first principles.

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/app/helpers/codelist_helper.rb
+++ b/app/helpers/codelist_helper.rb
@@ -2,6 +2,7 @@
 
 module CodelistHelper
   class UnreadableCodelist < StandardError; end
+  class UnrecognisedSource < StandardError; end
 
   DEVELOPING_COUNTRIES_CODE = "998"
   ALLOWED_AID_TYPE_CODES = [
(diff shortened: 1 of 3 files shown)


app/helpers/codelist_helper.rb
```diff
@@ -144,46 +145,52 @@ def policy_markers_select_options
...
-  def load_yaml(entity:, type:)
-    yaml = YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/IATI/#{IATI_VERSION}/#{entity}/#{type}.yml"))
+  def load_yaml(entity:, type:, source: "iati")
+    raise UnrecognisedSource unless source.in?(%w[iati beis])
+
+    if source == "iati"
+      yaml = YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/IATI/#{IATI_VERSION}/#{entity}/#{type}.yml"))
+    else
+      yaml = YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/BEIS/#{type}.yml"))
+    end
```
pezholio, line 193: I think what Standard wants here is:

```suggestion
yaml = if source == "iati"
  YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/IATI/#{IATI_VERSION}/#{entity}/#{type}.yml"))
else
  YAML.safe_load(File.read("#{Rails.root}/vendor/data/codelists/BEIS/#{type}.yml"))
end
```

pezholio (approved):
I really like this, thanks! Good to go, assuming you get the build fixed 👍

app/helpers/codelist_helper.rb
CristinaRO: Thanks for decoding that for me, Stu! 😅 It's one of those where I don't really agree with Standard that it makes for more readable code, but I let it go because that's why we have Standard. ;)

PR #844, opened 2021-01-06
(1359) Importer stops activities being associated to an incomplete parent

leeky:
On production, we have several activities in an incomplete state, which meant that `Activity.by_roda_identifier` would return a result, instead of `nil`, so the expected `parent_not_found` was not raised. This would lead to these activities being imported and incorrectly parented to a random activity.

## Next steps


pezholio (approved):
LGTM - Good catch! 🚢

PR #845, opened 2021-01-06
(1087) Prevent admnistrators from changing an email address

pezholio:
When carrying out user testing, we found out that if an administrator created an account with the wrong email address and the email address was subsequently changed, the new user would not recieve a confirmation email for them to activate their account. 

We did initially consider sending a new confirmation email if an email is changed, but on further investigation / discussion, we discovered that Auth0 does not allow a user's email to be changed, and in any case, we should treat the email of a user as the primary identifier. 

With this in mind, I've changed the system to not allow emails to be changed, both via a validation, and by disabling the email field on the edit page. If a user has been created with the wrong email address, they should be deactivated. We don't currently allow deletion of users, though if the system gets clogged up with loads of erroneously created deactivated users, we can reconsider and prioritise appropriately.

### Before

image

### After

image

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -4,6 +4,7 @@ class User < ApplicationRecord
   belongs_to :organisation
   validates_presence_of :name, :email
   validates :email, format: {with: URI::MailTo::EMAIL_REGEXP}
+  validates :email, with: :email_cannot_be_changed_after_create, on: :update
 
   enum role: {
     administrator: "administrator",
@@ -25,5 +26,13 @@ class User < ApplicationRecord
 
   def delivery_partner?
     !service_owner?
+  end
+
+  private
+
+  def email_cannot_be_changed_after_create
+    if email_changed?
+      errors.add(:email, :cannot_be_changed)
+    end
   end
 end
--- a/spec/features/staff/beis_users_can_edit_a_user_spec.rb
+++ b/spec/features/staff/beis_users_can_edit_a_user_spec.rb
@@ -14,12 +14,11 @@ RSpec.feature "BEIS users can editing other users" do
     target_user = create(:administrator, name: "Old Name", email: "old@example.com")
 
     updated_name = "New Name"
-    updated_email = "new@example.com"
 
     stub_auth0_update_user_request(
       auth0_identifier: target_user.identifier,
-      email: updated_email,
-      name: updated_name
+      name: updated_name,
+      email: target_user.email,
     )
 
     # Navigate from the landing page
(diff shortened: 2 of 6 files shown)


spec/models/user_spec.rb
```diff
@@ -4,6 +4,15 @@
   describe "validations" do
     it { should validate_presence_of(:name) }
     it { should validate_presence_of(:email) }
+
+    it "should not allow an email to be changed" do
+      user = create(:administrator)
+
+      user.email = "new@example.com"
+
+      expect(user.valid?).to eq(false)
```
leeky, line 13: You could also write: `expect(user).to_not be_valid` or even `expect(user).to be_invalid`

spec/features/staff/beis_users_can_edit_a_user_spec.rb
```diff
@@ -7,19 +7,33 @@
...
+
+    target_user = create(:administrator, name: "Old Name", email: "old@example.com")
+
+    visit organisation_path(user.organisation)
+    click_on(t("page_title.users.index"))
+
+    find("tr", text: target_user.name).click_link("Edit")
+
+    email_field = find("input[name='user[email]']")
+    expect(email_field["disabled"]).to eq("disabled")
```
leeky, line 22: Capybara provides a `disabled?` method on returned elements, so you can do:

`expect(email_field).to be_disabled`

spec/features/staff/beis_users_can_edit_a_user_spec.rb
```diff
@@ -7,19 +7,33 @@
...
 
     updated_name = "New Name"
-    updated_email = "new@example.com"
 
     stub_auth0_update_user_request(
       auth0_identifier: target_user.identifier,
-      email: updated_email,
-      name: updated_name
+      name: updated_name,
+      email: target_user.email,
```
leeky, line 36: 👍

spec/models/user_spec.rb
```diff
@@ -4,6 +4,15 @@
   describe "validations" do
     it { should validate_presence_of(:name) }
     it { should validate_presence_of(:email) }
+
+    it "should not allow an email to be changed" do
+      user = create(:administrator)
```
leeky, line 9: Is it worth explicitly doing `user = create(:administrator, email: "old@example.com")` ?

app/models/user.rb
```diff
@@ -26,4 +27,12 @@ def service_owner?
   def delivery_partner?
     !service_owner?
   end
+
+  private
+
+  def email_cannot_be_changed_after_create
+    if email_changed?
+      errors.add(:email, "cannot be changed once a user has been created")
```
leeky, line 35: This should probably be I18n'd (NB: Use `errors.add(:email, :blah)` here rather than calling I18n.t directly)

leeky (approved):
Have left some minor suggestions, but otherwise LGTM! :shipit:

app/models/user.rb
pezholio: Good call 👍

PR #848, opened 2021-01-08
(1379) Fix: Users can change the extending organisation

mec:
## Changes in this PR
Whilst looking at if this was even possible I noticed it was broken.

The reason is that the activity model is not valid by default without
all required attributes. We often see activities being completed over
time and so cannot always rely on the object being valid.

It turns out we even have a validation context that was used for these
reasons to test the validity but it was no called on the `save` so the
save never completed.

This was compounded by the fact the UI gives the impression that the
save was completed when it had not.

The fix is to simply validation in the `update_extending_organisation`
context which will only validate the extending organisation attribute
and not the entire Activity object.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -465,6 +465,7 @@
 - Infer the value of FSTC applies from aid type, where possible
 - Require UK Delivery partner named contact for all projects and third-party projects
 - Fix a brittle spec
+- Fix users can change the extending organisation on a level B (programme) activity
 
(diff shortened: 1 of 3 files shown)


pezholio (approved):
Yikes, good catch! Let's ship it 👍

CristinaRO:
@mec Needs rebasing and fixing the conflict in the changelog, when you've got a bit of time.

PR #849, opened 2021-01-08
(1140) Automatically fill in FSTC question where applicable

CristinaRO:
## Changes in this PR

- Infer the answer to the "FSTC applies" question from the value of the aid type, where possible
- Do not link to edit FSTC if it is inferred from aid type

## Next steps

### DIFF
--- a/app/views/staff/shared/activities/_activity.html.haml
+++ b/app/views/staff/shared/activities/_activity.html.haml
@@ -320,7 +320,7 @@
       - unless activity_presenter.fstc_applies.to_s.blank?
         = t("summary.label.activity.fstc_applies.#{activity_presenter.fstc_applies}")
     %dd.govuk-summary-list__actions
-      - if policy(activity_presenter).update? && step_is_complete_or_next?(activity: activity_presenter, step: :fstc_applies)
+      - if policy(activity_presenter).update? && step_is_complete_or_next?(activity: activity_presenter, step: :fstc_applies) && !can_infer_fstc?(@activity.aid_type)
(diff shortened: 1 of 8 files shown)


spec/helpers/codelist_helper_spec.rb
```diff
@@ -229,6 +229,46 @@
...
+      end
+
+      it "returns true for aid types 'D02', 'E01', 'G01'" do
+        %w[D02 E01 G01].each do |at|
+          expect(helper.can_infer_fstc?(at)).to eql true
+        end
+      end
+
+      it "returns false for any other aid type" do
+        (CodelistHelper::ALLOWED_AID_TYPE_CODES - %w[D02 E01 G01]).each do |at|
```
pezholio, line 267: 👍

app/views/staff/shared/activities/_activity.html.haml
```diff
@@ -320,7 +320,7 @@
       - unless activity_presenter.fstc_applies.to_s.blank?
         = t("summary.label.activity.fstc_applies.#{activity_presenter.fstc_applies}")
     %dd.govuk-summary-list__actions
-      - if policy(activity_presenter).update? && step_is_complete_or_next?(activity: activity_presenter, step: :fstc_applies)
+      - if policy(activity_presenter).update? && step_is_complete_or_next?(activity: activity_presenter, step: :fstc_applies) && !can_infer_fstc?(@activity.aid_type)
```
pezholio, line 323: I wonder if it might be better to have an small method in the helpers here? Something like `ftsc_not_inferred?`, just to make the intention of this line a bit clearer?

pezholio (approved):
LGTM - Just one teeny thing, but no biggie if you choose not to implement 👍

PR #850, opened 2021-01-11
(1366) Users can delete a transaction added by mistake

pezholio:
## Changes in this PR

This allows BEIS and Delivery Partner users to delete transactions, which may have been added by mistake. It also adds a "Cancel" button to allow users to go back if they have accidentally tried to add a transaction.

### Before

image

image

### After

image

image

### DIFF
--- a/app/controllers/staff/transactions_controller.rb
+++ b/app/controllers/staff/transactions_controller.rb
@@ -52,6 +52,18 @@ class Staff::TransactionsController < Staff::BaseController
     end
   end
 
+  def destroy
+    @transaction = Transaction.find(id)
+    authorize @transaction
+
+    @transaction.create_activity key: "transaction.destroy", owner: current_user, parameters: {activity_id: activity.id}
+    @transaction.destroy
+
+    flash[:notice] = t("action.transaction.destroy.success")
+
+    redirect_to organisation_activity_path(activity.organisation, activity)
+  end
+
   private
 
   def transaction_params
--- a/app/views/staff/transactions/_form.html.haml
+++ b/app/views/staff/transactions/_form.html.haml
@@ -19,4 +19,5 @@
 
 - if action_name == "new"
   = link_to t("form.link.activity.back"), organisation_activity_path(@activity.organisation, @activity), class: "govuk-button govuk-button--secondary", "data-module": "govuk-button", role: "button"
-
+- if action_name == "edit"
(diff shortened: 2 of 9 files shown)


app/controllers/staff/transactions_controller.rb
```diff
@@ -52,6 +52,18 @@ def update
     end
   end
 
+  def destroy
+    @transaction = Transaction.find(id)
+    authorize @transaction
+
+    @transaction.create_activity key: "transaction.destroy", owner: current_user
```
jcoglan, line 59: What's the effect of creating public_activity records for something you then delete? Do the logs remain present, and if so how does one view them?

pezholio: I don't think we can view `public_activity` records for anything can we? This is more so we have an audit trail that we can get at right? Or am I missing something? (I frequently do!)

pezholio: I could potentially add parameters to the Activity, to make it clearer what has been deleted?

app/views/staff/transactions/_form.html.haml
```diff
@@ -16,3 +16,7 @@
   = f.govuk_text_field :receiving_organisation_reference
 
 = f.govuk_submit t("default.button.submit")
+
+- if action_name == "new"
+  = link_to t("form.link.activity.back"), organisation_activity_path(@activity.organisation, @activity), class: "govuk-button govuk-button--secondary", "data-module": "govuk-button", role: "button"
```
leeky, line 21: I know I've already approved your other PR that also adds this button(!), but I think this should be a [back link](https://design-system.service.gov.uk/components/back-link/) instead.

spec/features/staff/users_can_create_a_transaction_spec.rb
```diff
@@ -257,6 +257,20 @@
         expect(page).to have_content t("activerecord.errors.models.transaction.attributes.date.blank")
       end
     end
+
+    scenario "they can cancel their transaction" do
```
leeky, line 261: How about something like "they can choose not to submit a transaction"? This reads like 'cancel' is a specific action.

leeky (approved):
LGTM (with a couple of tiny suggestions) :shipit:

app/views/staff/transactions/_form.html.haml
pezholio: See, I did wonder about that, but the ticket seemed to suggest something that felt more like an action, rather than navigating

PR #854, opened 2021-01-12
(1363) Validate `form_state` to accept only expected values

MariluzHerrera:
Trello: https://trello.com/c/9ljDJ4D6

## Changes in this PR

We recently had an issue where after a data migration we accidentally set the value of `form_state` to `completed` instead of `complete` for some activities on production. There was no error raised because this field lacked of any validation. To avoid similar errors in the future we are now introducing validation to only accept one of the expected values. These values are all the FORM_STEPS plus some custom ones. We also allow nil.

As part of this piece of work I have refactored the `activity_forms_controller`. FORM_STEPS used to live there, but now that the Activity model needs access to this list makes more sense to have the list in the model and have the controller accessing it rather than the other way around. 

## Next steps


pezholio (approved):
LGTM 👍

PR #855, opened 2021-01-12
Users can delete a forecast added by mistake

pezholio:
## Changes in this PR

This allows users to delete forecasts that may have been added by mistake. After a bit of back and forth and discussion, we decided that when a user wants to "delete" a forecast, it'll be because they've erroneously created one, either for the wrong reporting period, or the wrong activity, rather than they've just entered the wrong amount. With this in mind, I decided that the expected behaviour would be to delete all the Planned Disbursements for that activity / period, and remove all history. We did discuss just setting the latest value to zero, which would behave as if the forecast was deleted, but this would still keep the erroneously created value as the "original", which may not be the intent.

### Before

image

image

### After

image

image

### DIFF
--- a/app/services/planned_disbursement_history.rb
+++ b/app/services/planned_disbursement_history.rb
@@ -24,6 +24,10 @@ class PlannedDisbursementHistory
     end
   end
 
+  def clear!
+    entries.destroy_all
+  end
+
   def all_entries
     entries.to_a.reverse
   end
--- a/app/services/planned_disbursement_history.rb
+++ b/app/services/planned_disbursement_history.rb
@@ -25,7 +25,10 @@ class PlannedDisbursementHistory
   end
 
   def clear!
-    entries.destroy_all
+    entries.each do |entry|
+      entry.create_activity(key: "planned_disbursement.destroy", owner: @user, parameters: {associated_activity_id: @activity.id})
+      entry.destroy
+    end
   end
 
   def all_entries
--- a/spec/services/planned_disbursement_history_spec.rb
+++ b/spec/services/planned_disbursement_history_spec.rb
@@ -94,6 +94,35 @@ RSpec.describe PlannedDisbursementHistory do
         ["original", nil, nil, 10],
       ])
     end
+
+    context "when deleting a report" do
+      it "deletes the original planned disbursement when there is only one entry" do
+        history.set_value(10)
+
+        history.clear!
+
+        expect(history_entries).to eq([])
+      end
+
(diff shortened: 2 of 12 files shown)


app/services/planned_disbursement_history.rb
```diff
@@ -24,6 +24,10 @@ def set_value(value)
     end
   end
 
+  def clear
```
leeky, line 27: Since this is destructive, perhaps it should be called `clear!` ?

app/services/planned_disbursement_history.rb
```diff
@@ -25,7 +25,10 @@ def set_value(value)
   end
 
   def clear
-    entries.destroy_all
+    entries.each do |entry|
+      entry.create_activity(key: "planned_disbursement.destroy", owner: @user, parameters: {associated_activity_id: @activity.id})
```
leeky, line 29: 👍  

Storing the `associated_activity_id` here seems sensible. I'm wondering if this warrants documenting (ADR or something else?) so that we use the same parameter when doing this sort of thing on other models.

spec/services/planned_disbursement_history_spec.rb
```diff
@@ -122,6 +122,22 @@ def history_entries
...
+          history.set_value(10)
+          history.set_value(5)
+
+          old_entries = history.all_entries
+
+          history.clear
+
+          old_entries.each do |entry|
+            auditable_event = PublicActivity::Activity.find_by(trackable_id: entry.id, parameters: {associated_activity_id: activity.id})
+            expect(auditable_event).to_not be_nil
```
leeky, line 137: Since `find_by` will blow up if there's no matching entry, I'm wondering if it's better to check that two `PublicActivity::Activity` records were created with the right `key` and `parameters` set?

leeky (approved):
Left some small comments, but LGTM! :shipit:

spec/services/planned_disbursement_history_spec.rb
pezholio: Ah yeah, good catch. I did think about doing that, as that's how other tests for Public Activity work.

app/services/planned_disbursement_history.rb
pezholio: Nice, done 👍

app/services/planned_disbursement_history.rb
pezholio: Yeah, I'm not sure exactly where to document this. An ADR seems like it'd be too much.

PR #858, opened 2021-01-14
(1403) Refactor ActivityFormsController

pezholio:
The ActivityFormsController has a lot of repetition and makes the controller quite large. I've factored all the update actions into a lean(er) class that accepts the activity and the specific step, then works out which parameters to fetch and apply, as well as any extra steps that need to be performed. I haven't bothered with any unit tests for this class, which does make me a little anxious, but I think it's fine, as this is all covered by feature tests.

### DIFF
--- a/app/controllers/staff/activity_forms_controller.rb
+++ b/app/controllers/staff/activity_forms_controller.rb
@@ -1,10 +1,6 @@
 class Staff::ActivityFormsController < Staff::BaseController
   include Wicked::Wizard
-  include DateHelper
-  include ActivityHelper
   include CodelistHelper
-
-  DEFAULT_PROGRAMME_STATUS_FOR_FUNDS = "spend_in_progress"
 
   steps(*Activity::FORM_STEPS)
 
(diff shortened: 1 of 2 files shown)


app/services/activity/updater.rb
```diff
@@ -0,0 +1,180 @@
...
+      if respond_to?(updater, :include_private)
+        __send__(updater)
+      else
+        set_for_step(step)
+      end
+    end
+
+    private
+
+    def set_for_step(step)
```
leeky, line 25: To be consistent, should this be called `set_step` ? Although I'm confused what this is used for.

Also, would it be useful to have a test that checks that `ActivityUpdater` has a `set_` method for every step? I note that `set_programme_status` is missing, for example.

I'm wondering if the `else` is a mistake here, as it's swallowing `respond_to?` failures.

app/controllers/staff/activity_forms_controller.rb
```diff
@@ -70,105 +68,12 @@ def update
       @activity.assign_attributes(programme_status: DEFAULT_PROGRAMME_STATUS_FOR_FUNDS)
     end
```
leeky, line 69: Should this get moved into `ActivityUpdater` too?

app/services/activity/updater.rb
leeky: I've also wondered why, since all users who can manage an activity can access all steps, we can't globally do...

`activity.assign_attributes(activity_params)`

...rather than carefully making sure the form only contains the fields from the step you're currently on, which should always be the case, unless the user hacks the form, which would be pointless.

leeky (changes requested):

app/services/activity/updater.rb
pezholio: `set_for_step` is a generic catch-all method for any steps that don't require any special actions, so we can just set the relevant params and move on. That said, we could probably just set the whole params and move on, as you say. I'll have a crack at that and see if anything else falls out

app/controllers/staff/activity_forms_controller.rb
pezholio: Good call 👍

app/services/activity/updater.rb
pezholio: Oh, also, just running `activity.assign_attributes(activity_params)` gives an `ActiveModel::ForbiddenAttributesError` in some cases, presumably because there's some stuff submitted in forms that aren't present in the model.

leeky: Maybe keep the individual `assign_attributes` for now so it's a straight refactor.  

I've realised I completely misread `step => params_for(step)` as the equivalent of `"step" => params_for("step")` instead of `step` being a variable for some reason or another 🤯  (!)

I still think having a distinct method name to keep it separate from the dynamic `step_*` methods might be useful: `fallback_set_for_step`

leeky (approved):

app/services/activity/updater.rb
pezholio: Cool, I've gone for `assign_attributes_for_step` and put it seperate from the `step_*` methods to make it clearer

PR #861, opened 2021-01-18
Fix display of previous actuals in the report CSV

CristinaRO:
## Changes in this PR

During Q3 reporting we realised that the report CSV was showing the wrong values for the previous quarter’s actuals. It was showing the same values as for the current quarter. (The values were not overwritten, it was only a computation/display bug.)

The reason was that the service object was memoising all the actuals values for the current report, so when the code attempted to calculate the actuals for the previous report, it was using the already computed values.

The reason the spec didn’t catch this is that it was not set up with a full real-life scenario.

We have now enriched the spec’s setup to replicate the conditions under which the bug occurs.

None of the other methods which memoise values are being called for a different report, so we didn’t see a reason to preemptively change them.

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -470,6 +470,7 @@
 ## [unreleased]
 
 - Allow users to delete their transactions
+- Fix display of previous actuals in the report CSV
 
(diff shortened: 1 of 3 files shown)


pezholio (approved):
LGTM. Good sleuthing! 👍

PR #863, opened 2021-01-19
(1370) Separate BEIS from other organisations

MariluzHerrera:
Trello: https://trello.com/c/iEz6socI

## Changes in this PR

When a user wants to add a new user, they have to select what organisation this user belongs to. There is a radio button list for this, with organisations sorted alphabetically. However, there was no distinction between BEIS and the Delivery Partners on this list. BEIS could appear in the middle of the list, which made easier for users to accidentally assign a new user to BEIS when they don't belong to this organisation. This would give that user higher privileges than they should have. To avoid this, we have now separated BEIS org from the rest of organisations on display, adding a divider in the radio button list.

### Before

Screenshot 2021-01-12 at 14 39 42

### After

Screenshot 2021-01-19 at 16 20 12

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -474,6 +474,8 @@
 
 ## [unreleased]
 
+- On the user administration page, BEIS now appears as a separate organisation to avoid users being assigned to this org by accident
+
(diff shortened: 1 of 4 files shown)


pezholio (approved):
LGTM 👍

PR #864, opened 2021-01-19
(1388) Users can bulk upload activity data

CristinaRO:
## Changes in this PR

- Users can upload activities in bulk from a CSV

## Next steps


app/views/staff/reports/_actions.html.haml
```diff
@@ -1,5 +1,6 @@
 .govuk-grid-row
   .govuk-grid-column-full.page-actions
+    = link_to t("action.activity.upload.link"), new_report_activity_upload_path(@report_presenter), class: "govuk-button govuk-button--secondary govuk-!-margin-left-4"
```
pezholio, line 3: Should this be on the Activities page?

pezholio (approved):
Looks good on the whole - I'm just a bit unsure about the placement of the Upload link. Shouldn't this be on the Activities page?

app/views/staff/reports/_actions.html.haml
CristinaRO: After timeboxing this with James over half an hour, the conclusion is: probably, it would make the most sense, but trying to take the upload out of the report scope creates a cascade of changes, and running against Rails, that would delay/block this feature for too long.

PR #865, opened 2021-01-20
(1416) Parent step only shows complete parent activities

pezholio:
Previously when prompting for a parent activity, all parent activities were shown on the form. This resulted in a bunch of incomplete or erroneously created activities showing up. This adds a filter to the `scoped_parent_activities` method that makes sure only activities with a `form_state` of `complete` are shown.

### Before

image

### After

image

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -46,7 +46,7 @@ module FormHelper
   end
 
   def scoped_parent_activities(activity:, user:)
-    case activity.level.to_sym
+    activities = case activity.level.to_sym
     when :fund
       Activity.none
     when :programme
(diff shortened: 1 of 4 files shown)


app/helpers/form_helper.rb
```diff
@@ -59,6 +59,8 @@ def scoped_parent_activities(activity:, user:)
       FindProjectActivities.new(organisation: activity.organisation, user: user)
         .call(eager_load_parent: false)
```
MariluzHerrera, line 60: Out of curiosity... Is this `eager_load_parent: false` to avoid errors from the Bullet gem or is there any other reason?

MariluzHerrera (approved):
It looks good to me! I left a question but it's related to some code that was already there, and just because I was curious about it :-)

PR #867, opened 2021-01-21
(1426)  Load codelists into memory in production

pezholio:
This adds an initializer that loads all the codelists we have into memory when the application boots, rather than loading them on demand.


jcoglan:
Would it be possible to avoid caching this data in development, and instead reload it on each request? With the current implementation, we don't need to restart the app to see changes to these files, whereas we would if they're cached at boot time.

pezholio:
Good shout. This has inspired me to do some further refactoring which should help increase the visibility of codelists across the application.

spec/helpers/codelist_helper_spec.rb
```diff
@@ -4,14 +4,14 @@
...
-        expect { helper.yaml_to_objects(entity: "generic", type: "favourite_colours") }
+    describe "#codelist_to_objects" do
+      it "raises an error when the codelist is missing or incorrect" do
+        expect { helper.codelist_to_objects(type: "favourite_colours") }
           .to raise_error "CodelistHelper::UnreadableCodelist"
       end
 
-      it "formats the data in a yaml file to an array of objects for use in govuk form builder" do
-        expect(helper.yaml_to_objects(entity: "generic", type: "default_currency"))
+      it "formats the data from a codelinst to an array of objects for use in govuk form builder" do
```
CristinaRO, line 13: Small typo in the description, if you want to fix it: "codelinst" -> "codelist".

CristinaRO (commented):
Perhaps I'm missing something, but it looks like the codelists are still being cached in development. I tried locally and a change in a codelist wasn't reflected until I restarted the server.

PR #874, opened 2021-01-26
(1423) Add a transfer model

leeky:
Add a `Transfer` model for storing the transfers of money between activities.

I've added a foreign key constraint on `source_id` and `destination_id`, that will prevent those activities being deleted whilst transfers exist that reference them. 

NB: I didn't name the associations `source_activity` and `destination_activity` both to save typing and in case we need to have a source that isn't an activity in the future (e.g. external funding).

## Next steps


pezholio (approved):
LGTM. Nice work on the custom matcher too - really cool! 👍

PR #879, opened 2021-01-27
Document all the types of identifiers we use for activities

jcoglan:
This imports the [document](https://gist.github.com/jcoglan/e525c9059e38f0c812069f6171c37652) I wrote a while ago about the different types of activity identifier we use in RODA, as I need to keep referring people to it. I believe we've added one more type since this was written so it may need updating?


leeky (approved):
LGTM! :shipit:

PR #881, opened 2021-01-27
(1422) Add a Type attribute to the Budget model

pezholio:
This adds a `type` attribute to the Budget model, allowing a user to specify if the funds come from the Newton or GCRF Fund. This can be extended to other fund types in the future. 

This PR (should have been two PRs TBH) has also removed the `period_start_date` and `period_end_date` columns from the UI in favour of a `financial_year` column. We then work out the Start and End dates based on the start and end of a given financial year. 

I've added a data migration to set the financial year for existing Budgets based on their start date. We'll need to remove the `period_start_date` and `period_end_date` columns from the DB once the data migrations have been run in staging and prod (I know in theory the data and DB migrations run in parallel, but I think it's safer to wait until everything is out in the wild)

### Before

image

### After

image


db/schema.rb
```diff
@@ -50,17 +50,16 @@
     t.boolean "call_present"
     t.date "call_open_date"
     t.date "call_close_date"
-    t.string "intended_beneficiaries", array: true
     t.string "roda_identifier_fragment"
     t.string "roda_identifier_compound"
+    t.string "intended_beneficiaries", array: true
```
leeky, line 55: Did you mean to commit these unrelated `db/schema.rb` changes?

spec/features/staff/users_can_create_a_budget_spec.rb
```diff
@@ -150,10 +151,12 @@
     end
   end
 
-  def fill_in_and_submit_budget_form
+  def fill_in_and_submit_budget_form(financial_year: "#{Date.current.year}-#{Date.current.next_year.year}")
```
leeky, line 147: `financial_year` doesn't get overridden anywhere, so is this needed?

app/models/financial_year.rb
```diff
@@ -0,0 +1,38 @@
+class FinancialYear
```
leeky, line 1: Looks like there might be some behaviour shared between this and `FinancialPeriod` (defined in `lib/financial_period.rb`).

app/models/financial_year.rb
```diff
@@ -0,0 +1,38 @@
...
+      if [1, 2, 3].include?(date.month)
+        new(date.year - 1)
+      else
+        new(date.year)
+      end
+    end
+  end
+
+  def start_date
+    Date.parse("01-04-#{start_year}")
```
leeky, line 24: Would `Date.new(start_year, 4, 1)` make more sense here..?

spec/models/budget_spec.rb
```diff
@@ -113,10 +115,16 @@
   context "when the period start and period end dates are invalid" do
     it "performs budget date validation" do
       budget = build(:budget, period_start_date: Date.today, period_end_date: Date.today + 366.days)
+  it "returns an instance of FinancialYear for the financial year" do
```
leeky, line 79: Might be worth using `travel_to` and locking this into an exact date and known values for the expectations?

db/data/20210127153525_set_financial_year_based_on_start_date.rb
```diff
@@ -0,0 +1,12 @@
+class SetFinancialYearBasedOnStartDate < ActiveRecord::Migration[6.0]
+  def up
+    Budget.all.each do |budget|
```
leeky, line 3: Probably worth doing `Budget.all.find_each` here instead.

leeky: Do we want to drop the `period_start_date` and `period_end_date` columns too?

leeky (approved):
Have made some comments, but :shipit: when ready!

db/schema.rb
pezholio: Whoops, no!

app/models/financial_year.rb
pezholio: Yeah, probably faster too!

app/models/financial_year.rb
pezholio: Ah, good call. I've done a bit of refactoring to get rid of that module altogether, and introduce a new FinancialQuarter class. Would appreciate a look before merging if that's cool?

db/data/20210127153525_set_financial_year_based_on_start_date.rb
pezholio: I deliberately skipped that. I don't 100% trust the Data Migrate gem, and would rather wait until the data migrations are done before dropping the column. Will make a card so we don't forget 👍

pezholio:
@leeky Thanks for the review! ⭐ I know you've approved it, but I'd appreciate another quick look over https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/881/commits/edd18bdeafa6f7ab04804475f89337c6e093828d, as I've introduced a bit of a refactor 👍

leeky (approved):
Recent refactors look good - have just added three small suggestions!

PR #882, opened 2021-01-27
(1418) Users can see Current and Historic activities separately

MariluzHerrera:
Trello: https://trello.com/c/NyH0BprF

## Changes in this PR

This new feature separates the activities view page into two different tabs. One containing the `Current` activities, which are more relevant for users, and the other containing the `Historic` activities, which are the ones where users are unlikely to add any financials, information, etc to them. These `historic activities` are the ones in which `programme_status` is set to either `completed`, `cancelled` or `stopped`. These activities cannot have any financial data added or modified and the users are unlikely to have to use them in the future. However, we still want to keep them in RODA.

The activities view will display two tabs `Current` and `Historic`, and the user will be able to select which ones they want to see.

### Before

Screenshot 2021-01-27 at 17 17 54

### After

Screenshot 2021-01-27 at 17 13 25

Screenshot 2021-01-27 at 17 13 40

## Next steps

### DIFF
--- a/app/controllers/staff/activities_controller.rb
+++ b/app/controllers/staff/activities_controller.rb
@@ -2,11 +2,12 @@
 
 class Staff::ActivitiesController < Staff::BaseController
   include Secured
+  after_action :verify_authorized, except: [:index, :historic]
+  after_action :verify_policy_scoped, only: [:index, :historic]
 
   def index
     @organisation_id = organisation_id
-    @activities = policy_scope(Activity.where(organisation: organisation_id))
-    authorize @activities
+    @activities = policy_scope(Activity.where(organisation: organisation_id)).current
     @activity_presenters = @activities.includes(:organisation).order("created_at ASC").map { |activity| ActivityPresenter.new(activity) }
   end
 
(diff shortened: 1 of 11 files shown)


app/controllers/staff/activities_controller.rb
```diff
@@ -5,7 +5,8 @@ class Staff::ActivitiesController < Staff::BaseController
 
   def index
     @organisation_id = organisation_id
-    @activities = policy_scope(Activity.where(organisation: organisation_id))
+    @activities = policy_scope(Activity.where(organisation: organisation_id).where.not(programme_status: ["completed", "stopped", "cancelled"])
+      .or(Activity.where(organisation: organisation_id, programme_status: nil)))
```
jcoglan, line 9: This feels like something it would be worth refactoring into a scope on `Activity`, to make the intent clearer, for example:

```rb
class Activity
  scope :historic, -> {
    where.not(programme_status: ["completed", "stopped", "cancelled"]).or(where(programme_status: nil))
  }
end
```

We could then write in the controller:

```rb
Activity.historic.where(organisation: organisation_id)
```

app/controllers/staff/activities_controller.rb
```diff
@@ -40,6 +41,13 @@ def create
     redirect_to activity_step_path(@activity.id, :blank)
   end
 
+  def historic
+    @organisation_id = organisation_id
+    @historic_activities = policy_scope(Activity.where(organisation: organisation_id, programme_status: ["completed", "stopped", "cancelled"]))
```
jcoglan, line 46: Similarly here we could define an `Activity.historic` scope to tidy this up.

app/controllers/staff/activities_controller.rb
MariluzHerrera: Nice idea! I like how your example reads. I'll change my code. Thanks for the suggestion!

jcoglan (approved):

PR #883, opened 2021-01-27
(1448) Group activities by hierarchy on single report view

CristinaRO:
## Changes in this PR

- Group activities by hierarchy on the view of a single report page

### Before
Screenshot 2021-01-27 at 18 23 04

### After
Screenshot 2021-01-27 at 18 22 26

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -417,4 +417,17 @@ class Activity < ApplicationRecord
 
     iati_status_from_programme_status(programme_status)
   end
+
+  def self.hierarchically_grouped_projects
+    activities = all.to_a
+    projects = activities.select(&:project?).sort_by(&:roda_identifier_fragment)
+    third_party_projects = activities.select(&:third_party_project?).group_by(&:parent_id)
+
+    grouped_projects = []
+    projects.each do |project|
+      grouped_projects << project
+      grouped_projects += third_party_projects.fetch(project.id, []).sort_by(&:roda_identifier_fragment)
+    end
+    grouped_projects
+  end
 end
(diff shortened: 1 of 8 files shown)


app/models/activity.rb
```diff
@@ -409,4 +409,16 @@ def iati_status
 
     iati_status_from_programme_status(programme_status)
   end
+
+  def self.hierarchically_grouped_projects
+    projects = select(&:project?).sort_by(&:roda_identifier_fragment)
```
leeky, line 414: Since this gets chained with methods that return an `Activity::ActiveRecord_Relation` couldn't you do:

`projects = project.order(:roda_identifier_fragment)`

...and let the filtering/sorting take place in the database..?

leeky (approved):
I've suggested doing filtering/sorting in the database rather than in Ruby, but LGTM :shipit:

app/models/activity.rb
CristinaRO: @jcoglan can correct this if it's wrong, but it was my understanding that we wanted to only hit the database once for the whole load of C and D level projects.

That's how we started, anyway, this method had a few iterations before it landed in the model. It could be that we're still hitting the db more times than intended now? Not sure.

leeky: Probably better to leave this alone and get it merged. Just needs a rebase, @CristinaRO.

jcoglan: Yes, the purpose here is to load whatever scope `hierarchically_grouped_projects` is called on _once_, and that query will return both level C and D activities, which then need sorting as this method describes. The sorting by RODA identifier is really part of the process of topologically sorting these records in the application -- it needs to happen after the level D activities are grouped by parent, so putting an `ORDER BY` into the database query would mean the sorting happens too early in the process.

jcoglan: I found out that our implementation actually runs two queries -- both `select` calls execute the query. I've modified the method to load all the records once, and then do all the sorting/grouping in memory.

PR #892, opened 2021-02-01
(chore) Update docs for copying production data to local

mec:
## Changes in this PR

Over time we've developed a quicker, simpler and more robust way to do
this.


MariluzHerrera (approved):
Looks great! Thanks for updating this docs 🚀

PR #894, opened 2021-02-02
(1428) Use 'high_voltage' gem for static pages

leeky:
Use `high_voltage` gem (always included from dxw's standard Rails starter template) instead of separate controllers for static content.

I've also fixed some typos and simplified the tests as these pages are *always* publicly accessible.


pezholio (approved):
LGTM 👍

PR #895, opened 2021-02-02
(1368) Users can report Channel of delivery code via form

CristinaRO:
There may be something in this PR that makes the test suite unworkably slow. Thus, a review and/or pairing on investigating this would be much appreciated.

## Changes in this PR

- Users can report Channel of delivery code via form
- The presence of Channel of delivery code is validated from a short list of BEIS-approved codes

#### Select form open
Screenshot 2021-02-01 at 17 15 24

#### Select form with selection
Screenshot 2021-02-01 at 17 15 19

#### Select form with error message
Screenshot 2021-02-01 at 17 15 37

#### Activity details
Screenshot 2021-02-01 at 17 15 07

## Next steps


app/helpers/codelist_helper.rb
```diff
@@ -170,4 +170,20 @@ def organisation_type_options
...
     Codelist.new(type: "language_code").to_objects
   end
+
+  def beis_allowed_channel_of_delivery_codes
+    Codelist.new(type: "channel_of_delivery_code", source: "beis").list
+  end
+
+  def channel_of_delivery_codes
+    iati_data = Codelist.new(type: "channel_of_delivery_code")
+    iati_data.select { |item| item["code"].in?(beis_allowed_channel_of_delivery_codes) }
```
leeky, line 180: This is calling `beis_allowed_channel_of_delivery_code` once for every entry in the `channel_of_delivery_code` codelist. Because it isn't memoised, it's creating a new array 414 times (one of each entry in the IATI list), which means any test which hits the Activity details page is taking 30 seconds to run!

jcoglan: Does `Codelist` memoise the read from disk during tests? Is it reading from disk that's expensive, or allocating the array in memory each time?

pezholio (approved):
Nice, let's ship it 🚢

PR #896, opened 2021-02-03
Memoize codelists in tests

pezholio:
While it's potentially useful to reload codelists in dev, it's less useful in test, and almost doubles the time it takes our tests to run (from c. 3 minutes to c. 6 minutes on my local machine!). I have had to update the Codelist tests to reset the class's instance variable (fact fans: a variable starting with an `@` is still called an instance variable when it's being used inside a class, not a class variable)

### DIFF
--- a/app/models/codelist.rb
+++ b/app/models/codelist.rb
@@ -17,10 +17,10 @@ class Codelist
 
   class << self
     def codelists
-      if Rails.env.production?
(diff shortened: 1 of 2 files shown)


pezholio:
@dependabot rebase

leeky (approved):
LGTM :shipit:

PR #897, opened 2021-02-03
(1456) Add a `source_fund` to an Activity

pezholio:
This adds the concept of a `source_fund` to an Activity. This is stored as a code in the database, and can be returned as a `Fund` class (which may eventually be a database model) with some useful helper methods.

The `source_fund_code` is set at the point of setting the parent, and I've also added a data migration to set source funds on all existing Activities.

With this in place we can now change the `is_gcrf_funded?` and `is_newton_funded?` columns to use the source fund, rather than have to walk up the tree. There's probably some more improvements we can make, but I wanted to get some eyes on this and get it out there first.


app/models/fund.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  class << self
+    def from_activity(activity)
+      raise InvalidActivity unless activity.fund?
+      raise InvalidFund unless valid_fragments.include?(activity.roda_identifier_fragment)
+
+      case activity.roda_identifier_fragment
+      when "NF"
+        new(1)
```
CristinaRO, line 36: I spent some time being confused at why we'd use "magic numbers" when we have them in a codelist. It's much easier to make a mistake when setting or comparing with a number than when using their meaningful names or identifiers.

Perhaps this is part of "future optimisations", in which case ignore me! But if identifiers continue to play a big part in the whole thing, could we store the valid identifiers in the codelist, too?

jcoglan: Yes, it would at least be helpful to pull these into named constants on the `Fund` model for readability, and perhaps map that back into the migration in the previous commit?

spec/factories/activity.rb
```diff
@@ -384,6 +384,7 @@
   trait :level_form_state do
     form_state { "level" }
     level { nil }
+    source_fund_code { nil }
```
jcoglan, line 387: As well as defining named constants for the fund codes, it would be useful to have a trait for creating Newton/GCRF activities that encapsulates how the fund relationship is represented.

jcoglan:
This all looks good to me, nothing very much surprising about it. My main worry as always with denormalising things is maintaining consistency -- what if an activity changes parent, do we need to update `source_fund` on all its children. However I think the intention is that activities do not move around the tree in this way, and certainly not between funds, so this should be safe.

app/models/fund.rb
pezholio: Done and done! 👍

jcoglan:
I don't think it's a blocker in terms of whether the code works or not, but I would really like to see the magic numbers `1` and `2` replaced with named constants throughout for readability.

pezholio:
Oh, I thought I had?

pezholio:
@jcoglan Pretty sure that's done now 👍

CristinaRO (approved):
I think the latest version has addressed all the outstanding suggestions, @jcoglan ?

Great work @pezholio  🙌🏻

PR #899, opened 2021-02-03
(1461) Flow is always "10" ("ODA")

CristinaRO:
## Changes in this PR

- We don't ask the users for the Flow type, because it is always going to be ODA ("10"). We don't import it from CSV. When exporting to CSV and to XML, the flow is added and hardcoded to "10".

## Next steps


pezholio (approved):
LGTM 👍

PR #900, opened 2021-02-04
(1477) Handle nil RODA fragments when sorting

pezholio:
We have some limited situations where a RODA ID is nil, this causes the sorting to break, which is less than ideal. I've fixed this so we now cast all RODA IDs to a string when sorting.


CristinaRO (approved):
Ouch! Thanks for investigating and fixing!

PR #904, opened 2021-02-05
(1475) Encode CSV downloads for Excel

jcoglan:
This PR addresses a problem that happens when people use Excel to open a CSV downloaded from RODA. All the data stored in RODA is UTF-8, and does not contain any _encoding_ errors; that is, all text stored in RODA is a valid UTF-8 byte sequence. The CSVs served by RODA are also valid UTF-8 encoded CSV, but when opened in Excel some of their content is mis-interpreted.

We ran some experiments where we encoded a CSV file containing text from a variety of Unicode blocks. We encoded this file as UTF-8, UTF-16BE and UTF-16LE, and made copies of each of these with and without a byte order mark (BOM) at the start.

We found that Excel is able to read UTF-8 successfully if it's prefixed with a BOM. If this is missing, Excel will mis-read many of the characters in the file, including mis-interpreting the boundaries between rows and cells. It appears to be interpreting the data in some other ASCII-compatible encoding such as Windows-1252, but we're unsure which one.

Excel can read UTF-16LE with a BOM, in the sense that it reads all the codepoints represented in the file correctly. However, it ignores the commas that separate cells and so merges all the data into one column.

We saw mixed results for UTF-16BE on various platforms.

Given this, I've unified the code for all parts of the application that serve downloadable CSV, so we have one implementation for how this works. It encodes the data in UTF-8 and prefixes it with a BOM.

One thing we learned while investigating this is that some programs will strip the BOM off the file, rendering it unreadable. For example, sending the file via Slack results in the BOM being removed from it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -491,6 +491,10 @@
 - Add links to the guidance across the site
 - Users can report Channel of delivery code through the activity form
 
+## [unreleased]
+
+- Serve CSV downloads encoded in UTF-8, prefixed with a byte order mark
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-31...HEAD
 [release-31]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-30...release-31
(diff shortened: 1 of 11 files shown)


pezholio (approved):
Yeah, this is great stuff. Big fan of removing boilerplate too, so it's nice to solve that problem too! 👍

PR #905, opened 2021-02-05
(1439) Add the last 12 actuals to the CSV activity export, if available

pezholio:
This updates the CSV activity exporter to include actuals data for anything up to twelve quarters ago, if the reports are available. If there are no reports, the headers and data are not made shown in the CSV.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -490,6 +490,7 @@
 - Group activities by hierarchy on the view of a single report page
 - Add links to the guidance across the site
 - Users can report Channel of delivery code through the activity form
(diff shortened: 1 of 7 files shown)


mec:
@pezholio apologies for this, ths is not a review - but I did notice that in this work we only inlcude previous actuals if there is a previous report - this is not the behaviour we would like to see.

The report csv should always include the actuals for the same financial quarter as the report and the previous 12, even if that means a whole lot of 0.00.

Happy to chat this through as I know the logic can seem questionable at times - let me know?

mec:
I am going to close this PR for now, the full reasons are in trello and the short version is up in the comments ☝️

PR #911, opened 2021-02-09
(chore) Backing services can be run with Docker

mec:
## Changes in this PR

I prefer to run the two backing services (Postgres and Redis) with
Docker as I find managing version with Homebrew can effect my entire
system and I often find I update the backing services without meaning to
when updating my own tooling, which can cause horrible problems.

For those that prefer it, a simple docker compose file allows one to
start and stop the backing services via docker and not effect their
system at all. This has an added benefit of having the option to pin the
versions of the backing services to that of the ones used in production.

Whilst I am loathe to add alternatives to the one-script-to-rule-them-all,
I much prefer this approach personally and have real technical
reasons to avoid the Homebrew path, therefore I am adding this compose
file and a small section in the readme to cover its use for those that
prefer.

If the team are opposed to this in the readme, could we keep the compose file
so I don't have to have it somewhere else? 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,23 @@ script/server
 script/test
 ```
 
+## Running backing services with Docker compose
+
+If you prefer not to install the backing services (Postgres and Redis) with
+Homebrew via the scripts above, run them in the background with Docker and
+then use standard rails commands to interact with the application (you will need
+Docker installed on your device):
+
+````
+docker-compose -f backing-services-docker-compose.yml up -d
+````
+
+To stop the backing services:
+
+````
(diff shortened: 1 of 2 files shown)


pezholio (approved):
Good idea. I might give this a try myself. This is a nice alternative to running the whole app in Docker locally, which comes with its own problems :shipit:

mec:
> Good idea. I might give this a try myself. This is a nice alternative to running the whole app in Docker locally, which comes with its own problems :shipit:

Yep, I think this strikes the best balance between both worlds! Thanks for the review!

PR #912, opened 2021-02-09
(1494) Fix static pages when logged in

pezholio:
When a user is logged in, the static pages return a 500, this is because the navigation partial contains calls to a `policy` method, which requires Pundit to be included in the controller.

We can fix this by having our own pages controller, which includes all the things we need from HighVoltage as well as our Pundit helper methods


config/initializers/high_voltage.rb
```diff
@@ -0,0 +1,3 @@
+HighVoltage.configure do |config|
+  config.routes = false
```
MariluzHerrera, line 2: Haven't seen `HighVoltage` before. I'm guessing you're removing the default routes here because we are using our own customised routes for the static pages, right? (the ones created on the `routes.rb` below I mean)

MariluzHerrera (approved):
This looks good to me. I wasn't familiar with the `HighVoltage` gem but I've been reading a bit of their documentation and I left a question here. Not a blocker though.
Thanks Stu!

config/initializers/high_voltage.rb
pezholio: Yeah, that's right - we need to change the Routes to point to out custom controller now

PR #915, opened 2021-02-10
(1476) Show textarea content with wrapper HTML on the Activity details page

pezholio:
Now descriptions and objectives that be presented with the line breaks as HTML

### Before

image

### After

image


mec (approved):
Cor, that was so easy it makes me wonder why it was just done like this in the first place!!? `simple_format` looks like exactly the sort of pragmatic and simple approach to a problem I ❤️ Rails for!

PR #916, opened 2021-02-10
(chore) Update readme to include a link to the DNS repo

mec:
Even though the repo is private, I think it is best to document everything we can, imagine being on support and not knowing where the dns is or who manages it?


pezholio (approved):
LGTM 👍

PR #917, opened 2021-02-10
Add cdn-route

Stretch96:
Creates a cdn-route, and maps it to the app using the custom domain

This introduces 2 new tfvars:

```
custom_domain (the apex domain of where the app will be hosted)
custom_hostname (the subdomain, eg. www or staging)
```

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -95,3 +95,13 @@ variable "rollbar_disabled" {
   description = "Flag to turn off rollbar reporting"
   default     = "false"
 }
+
+variable "custom_domain" {
+  type        = string
+  description = "Name of custom domain created in the cf org"
+}
+
+variable "custom_hostname" {
+  type        = string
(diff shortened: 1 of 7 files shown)


terraform/variables.tf
```diff
@@ -95,3 +95,13 @@ variable "rollbar_disabled" {
...
   default     = "false"
 }
+
+variable "custom_domain" {
+  type        = string
+  description = "Name of custom domain created in the cf org"
+}
+
+variable "custom_hostname" {
+  type = string
```
CristinaRO, line 105: @Stretch96 I think the linting wants this to be aligned with the row under it, if I understand its output right.

I can't review the rest, but this might unblock the CI running successfully.

Stretch96: Ah yes, I forgot to run `terraform fmt`, thanks 👍

mec (approved):

CristinaRO (approved):
🤞🏻

PR #918, opened 2021-02-10
Fix bug with `Fund.all`

leeky:
Whilst looking at the recently-added Fund model, I noticed that the `Fund.all` method was always erroring.

This commit fixes that error and adds a couple of unit tests.


pezholio (approved):
Yikes, good catch!

PR #919, opened 2021-02-10
(1457) BEIS users can add a programme-level activity associated with a source fund

leeky:
Allow a BEIS user to create a new programme-level activity from an organisation's page, which will be associated automatically with the relevant `parent` and `source_fund`. 

Screenshot 2021-02-11 at 13 22 10

Clicking on "Add child activity to GCRF" will create a programme-level activity associated to GCRF, and redirect the user to the "Identifier" step in the wizard.

Previously, creating any level of activity would involve choosing the level and then the parent activity separately, which was awkward to use and often lead to 'empty' activities being created (those with neither a `level` or `parent` set). 

**NB: This currently sets `extending_organisation` to BEIS, which whilst incorrect, mirrors the current behaviour. There are a few places where this is set inconsistently, so I'll fix this in a later PR**

## Next steps


app/models/fund.rb
```diff
@@ -31,6 +31,14 @@ def newton?
     id == MAPPINGS["NF"]
   end
 
+  def activity
+    Activity.fund.find_by!(source_fund_code: id)
+  end
+
+  def ==(other)
+    self.class == other.class && id == other.id && name == other.name
```
pezholio, line 39: Nice 👍

pezholio: Do we need to match against the name too? Would the ID alone not be enough?

app/views/staff/organisations/show.html.haml
```diff
@@ -8,6 +8,18 @@
...
+    .govuk-grid-row
+      .govuk-grid-column-full
+        %h2.govuk-heading-m
+          = t("create_programme", scope: "page_content.organisation")
+
+        - @source_funds.each do |source|
+          = link_to t("form.button.activity.new_with_source_fund", name: source.name),
+            organisation_source_fund_children_path(@organisation_presenter, source.id),
+            method: :post, class: "govuk-button"
+          %br/
```
pezholio, line 21: I think these might look better aligned side by side with a `govuk-!-margin-right-4` class added, like this:

image

spec/features/staff/beis_users_can_create_a_programme_level_activity_spec.rb
```diff
@@ -13,6 +13,47 @@
...
+
+          within ".govuk-table__row##{delivery_partner.id}" do
+            click_on t("default.link.show")
+          end
+
+          click_on t("form.button.activity.new_with_source_fund", name: source_fund.name)
+
+          expect(page).to have_content t("form.label.activity.roda_identifier_fragment")
+
+          programme = Activity.programme.first
```
pezholio, line 37: Should this be `.last`? Just in case we have any others created for some reason

pezholio (approved):
Looks good! I really like how this is all coming together. A couple of teeny things, but feel free to ignore!

app/views/staff/organisations/show.html.haml
leeky: Yeah, I think you're right, was just following the mock up too closely 👍

spec/features/staff/beis_users_can_create_a_programme_level_activity_spec.rb
leeky: Well, it'd have to be `Activity.order(:created_at).last`, because of the UUIDs..? The `CreateProgramme` service already tests that the defaults are set more thoroughly, so I think this is probably ok.

PR #923, opened 2021-02-11
Add an action to automatically rebase

pezholio:
Now, adding `/rebase` as a comment rebases a PR, saving us from having to rebase manually when a PR gets out of date.


mec (approved):
Interested to see how this goes what happens if it can't rebase?

pezholio:
I think the action just fails and returns an error, so it's on us to check. I'd like to improve this to post a comment if it can't rebase, but will probably be a Friday thing 👍

PR #924, opened 2021-02-11
(1491) Handle the BOM on uploaded files

pezholio:
This adds a new `CsvFileUpload` class that takes params with an uploaded file, reads the file and optionally removes the Byte Order Mark, which is added by Excel. There are shortcuts to doing this in Ruby, but we felt that it was more sensible to document it as a problem we're aware of in code.


app/models/csv_file_upload.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def to_csv
+    return nil unless file
+
+    CSV.parse(body, headers: true)
+  rescue
+    nil
+  end
+
+  private def body
```
jcoglan, line 18: I think there's potential for this method to confuse the CSV parser about the encoding of the string, or what bytes it contains. As written `BYTE_ORDER_MARK` is a string with encoding `ASCII_8BIT` containing the bytes `EF BB BF`, because the default string encoding of Ruby string literals is UTF-8.

When checking the first bytes of the file, we want to specifically check the _bytes_, not the _codepoints_ it contains, and then we need to hand off a valid UTF-8 encoded string to `CSV.parse`.

This appears to assume `file.read` returns a string with encoding `ASCII_8BIT`. The `String#[]` method operates on "characters" (codepoints) not bytes, but on a "binary" string `contents[0..2]` will effectively return the first 3 bytes, as an `ASCII_8BIT` string. Comparing this to `BYTE_ORDER_MARK` is valid, as is using `delete_prefix!`, although the check isn't necessary -- `delete_prefix!` does nothing if the string does not begin with the given argument, it does not raise an error.

So if `file.read` returns a `ASCII_8BIT` string, which I think is the default for data read from a socket, then the leading BOM is removed, but the string passed to `CSV.parse` is tagged as `ASCII_8BIT` that means the CSV parser will interpret any multi-byte UTF-8 sequences not as UTF-8 characters but as random bytes, potentially giving incorrect output. This could be fixed by calling `body.force_encoding(Encoding::ASCII_8BIT)` before passing it to `CSV.parse`.

If `file.read` returns a `UTF_8` string, which is the default for files read from disk, then the code for deleting the BOM does nothing -- `contents[0..2]` returns the first 3 _codepoints_ from the string, not bytes, so it does not match `BYTE_ORDER_MARK`, and nothing is removed. The string passed to `CSV.parse` has the correct encoding tag, but it will have a BOM at the start.

If the thing we want to implement is: assume the input is UTF-8, and remove a BOM from it if present, I believe this will work:

```rb
  BYTE_ORDER_MARK = "\uFEFF".encode(Encoding::UTF_8)

  private def body
    contents = file.read.force_encoding(Encoding::UTF_8)
    contents.delete_prefix!(BYTE_ORDER_MARK)
    contents
  end
```

There are also ways of doing the "check and remove" at the byte level, if what you're trying to do is _detect_ the file's encoding, but they're more fiddly.

jcoglan: This is one of those things that's very easy for tests to give you misleading information, because the string encoding depends on whether the file is read from a socket or from disk, how the Rails stack handles the file upload in production vs in test, etc. You don't have any control over this in your test data -- once the data is in a file or sent over the wire it's just a sequence of bytes and it's entirely up to the web stack how they get interpreted.

spec/models/csv_file_upload_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  it "returns a parsed csv" do
+    expect(result).to be_a(CSV::Table)
+    expect(result.first.to_h).to eq({"foo" => "1", "bar" => "2", "baz" => "3"})
+  end
+
+  context "when the file has a BOM prefix" do
+    let(:bom) { "\uFEFF".force_encoding("binary") }
+    let(:csv) do
+      bom + <<~CSV
```
jcoglan, line 32: There's an encoding bug here caused by concatenating an `ASCII_8BIT` string with a `UTF_8` one. It happens to work because the CSV data does not contain any non-ASCII codepoints, but if you change `3` to `hëllo`, for example, you'll see the error.

This is fixed by removing `.force_encoding("binary")`, but it exposes the bug I mentioned above, with a confusing failure message:

```
       expected: {"bar"=>"2", "baz"=>"hëllo", "foo"=>"1"}
            got: {"bar"=>"2", "baz"=>"hëllo", "foo"=>"1"}

       (compared using ==)

       Diff:
       @@ -1,4 +1,4 @@
        "bar" => "2",
        "baz" => "hëllo",
       -"foo" => "1",
       +"foo" => "1",
```

(This failure happens if you don't modify `csv` too.) This is because, with the `force_encoding` call present in the definition of `bom`, `csv` is an `ASCII_8BIT` string. If you remove `force_encoding`, then `csv` becomes a `UTF_8` string, and `CsvFileUpload` fails to remove the BOM from it, so the first key in the resulting hash is actually `"<U+FEFF>foo"`, which you can't see because U+FEFF is a non-printing character.

Implementing `CsvFileUpload#body` as above fixes this bug, so I'd recommend including some non-ASCII data in the tests, and making `file.read` return both `ASCII_8BIT` and `UTF_8` strings to make sure the bug is fixed.

pezholio: That's really clear and useful, ta 👍

jcoglan (approved):
This looks good to me, thank you @pezholio for all your work on this :)

pezholio:
Thanks for all your work helping me to understand it! 💯

PR #925, opened 2021-02-11
Fix rebase action

pezholio:
The action can't trigger an action from another action without using a personal access token. I thought the access token we had by default was enough, but apparently not


CristinaRO (approved):
🤞🏻

PR #926, opened 2021-02-15
(1419) BEIS users can download a reports CSV file for all delivery partners

MariluzHerrera:
Trello: https://trello.com/c/PuEeSnuJ

## Changes in this PR

With this new feature, BEIS users can download all relevant reports in just one click. The reports they will get are those which are subject to changes or additions on the financial data (active, submitted, in_review and awaiting_changes). Reports that are inactive or approved will not be included on this CSV file.

The Reports CSV file contains all the activities from the relevant reports. These are sorted by Delivery Partner first, and then by hierarchy level, by recommendation from our user researcher team. They include the same data than the individual reports.

To avoid BEIS users getting incorrect information due to some reports being from a different quarter (and hence having a different set of financial_quarters information) I'm guarding against downloading the report if one or more reports
are from different financial_quarters. In that case, the user will get an error message and the download will not happen.

### Before

Screenshot 2021-02-01 at 14 47 50

### After

Screenshot 2021-02-01 at 15 47 34

Screenshot 2021-02-15 at 15 56 07

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -499,6 +499,8 @@
 
 ## [unreleased]
 
+- BEIS users can download a CSV report for all DPs
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-32...HEAD
 [release-32]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-31...release-32
 [release-31]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-30...release-31
--- a/app/controllers/staff/reports_controller.rb
+++ b/app/controllers/staff/reports_controller.rb
@@ -8,7 +8,20 @@ class Staff::ReportsController < Staff::BaseController
 
   def index
     if current_user.service_owner?
-      reports_for_service_owner
+
+      respond_to do |format|
+        format.html do
+          reports_for_service_owner
+        end
+        format.csv do
+          if reports_have_same_quarter?
+            send_all_reports_csv
+          else
+            flash[:error] = t("action.report.download.failure")
+            redirect_to action: "index"
+          end
+        end
+      end
     else
       reports_for_delivery_partner
     end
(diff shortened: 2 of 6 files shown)


app/controllers/staff/reports_controller.rb
```diff
@@ -8,7 +8,20 @@ class Staff::ReportsController < Staff::BaseController
...
   def index
     if current_user.service_owner?
-      reports_for_service_owner
+
+      respond_to do |format|
+        format.html do
+          reports_for_service_owner
+        end
+        format.csv do
+          if downloadable_reports_for_beis_users.map(&:financial_quarter).uniq.length == 1
```
leeky, line 17: I'm wondering whether this should extracted into its own method (`reports_have_same_quarter?`), since it's a little confusing to understand what's happening here..?

leeky (approved):
LGTM! :shipit: 

I think it might be worth adding a card to remove those two BEIS reports from production, as it'll make it less likely that "downloading all reports" will result in an error.

app/controllers/staff/reports_controller.rb
MariluzHerrera: Good idea! Thanks for the suggestion :-)

MariluzHerrera:
@leeky, I had to redo an important part of the code on the `reports_controller.rb` due to things not working after rebasing. I have incorporated what James did on [his PR](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/904). That's what has driven the most recent changes on my code. If you could have another look it would be great. Let me know if you're still happy for me to merge this.

I'll request James' review as well, just so we have confidence that I'm using the new code in the way it was intended.

CHANGELOG.md
```diff
@@ -487,6 +487,7 @@
 - Work out a Budget's period based on the financial year
 - Users can see Current activities and Historic activities in different tabs
 - Reorder reports in a more intuitive manner
+<<<<<<< HEAD
```
leeky, line 490: This probably shouldn't be here :)

app/controllers/staff/reports_controller.rb
```diff
@@ -158,4 +171,28 @@ def send_csv
...
+  end
+
+  def reports_have_same_quarter?
+    downloadable_reports_for_beis_users.map(&:financial_quarter).uniq.length == 1
+  end
+
+  def send_all_reports_csv
+    report_sample = downloadable_reports_for_beis_users.first
+    headers = ExportActivityToCsv.new(report: report_sample).headers
+    stream_csv_download(filename: "Reports.csv", headers: headers) do |csv|
```
jcoglan, line 190: This looks exactly like what I was expecting to see here -- sorry for the disruption of adding `stream_csv_download` and changing the response types from `ExportActivityToCsv` at the same time you were working on this!

CHANGELOG.md
MariluzHerrera: Oops! Completely missed that one. Thanks!!

app/controllers/staff/reports_controller.rb
MariluzHerrera: No worries!! Glad that I managed to get it right :-)

leeky:
Looks good to me (again) :)

PR #928, opened 2021-02-16
(1496) Set default values for various Transaction attributes

CristinaRO:
## Changes in this PR

- Transaction description is populated from the financial quarter and year and from the activity's title
- The default type for a transaction is Disbursement, set during creation and import
- The providing organisation for a transaction is set from the activity

## Next steps

### DIFF
--- a/app/services/create_transaction.rb
+++ b/app/services/create_transaction.rb
@@ -1,8 +1,6 @@
 class CreateTransaction
   attr_accessor :activity, :report, :transaction
 
-  DEFAULT_TRANSACTION_TYPE = "3"
-
   def initialize(activity:, report: nil)
     self.activity = activity
     self.report = report || Report.editable_for_activity(activity)
@@ -16,7 +14,7 @@ class CreateTransaction
     convert_and_assign_value(transaction, attributes[:value])
 
     transaction.description = default_description if transaction.description.nil?
-    transaction.transaction_type = DEFAULT_TRANSACTION_TYPE if transaction.transaction_type.nil?
+    transaction.transaction_type = Transaction::DEFAULT_TRANSACTION_TYPE if transaction.transaction_type.nil?
 
     unless activity.organisation.service_owner?
       transaction.report = report
--- a/app/services/create_transaction.rb
+++ b/app/services/create_transaction.rb
@@ -1,6 +1,8 @@
 class CreateTransaction
   attr_accessor :activity, :report, :transaction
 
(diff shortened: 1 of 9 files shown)


app/services/create_transaction.rb
```diff
@@ -34,4 +35,10 @@ def convert_and_assign_value(transaction, value)
   rescue ConvertFinancialValue::Error
     transaction.errors.add(:value, I18n.t("activerecord.errors.models.transaction.attributes.value.not_a_number"))
   end
+
+  def default_description
+    return nil unless transaction.date.present?
+
+    "#{transaction.financial_quarter_and_year} spend on #{activity.title}"
```
leeky, line 46: Does it matter that this will go out of sync if `activity.title` is changed in the future?

db/data/20210215165822_set_default_attributes_on_transaction.rb
```diff
@@ -0,0 +1,25 @@
...
+      .or(transaction_scope.where(providing_organisation_reference: nil))
+
+    transactions.each do |transaction|
+      transaction.transaction_type = Transaction::DEFAULT_TRANSACTION_TYPE if transaction.transaction_type.blank?
+
+      transaction.description = "#{transaction.financial_year} spend on #{transaction.parent_activity.title}" if transaction.description.blank?
+
+      transaction.providing_organisation_name = transaction.parent_activity.providing_organisation.name if transaction.providing_organisation_name.blank?
+      transaction.providing_organisation_type = transaction.parent_activity.providing_organisation.organisation_type if transaction.providing_organisation_type.blank?
+      transaction.providing_organisation_reference = transaction.parent_activity.providing_organisation.iati_reference if transaction.providing_organisation_reference.blank?
```
leeky, line 18: Does this need a `transaction.save` to commit the changes?

leeky (approved):
Thank you for the small easily-digestible commits!

LGTM (with a couple of minor comments) :shipit:

app/services/create_transaction.rb
mec: It's a great point, my view is that the title was accurate at the time the transaction was added and that is absolutely fine.

db/data/20210215165822_set_default_attributes_on_transaction.rb
mec: Nice spot, the class name is also incorrect, I'll have a look.

PR #929, opened 2021-02-16
(1497) Add ADR for text and encoding

mec:
We thought it would be valuable to included an ADR arround the issues we faced and the learnings we took from the text encoding work we had to do recently.

This is also in the decision log 

https://docs.google.com/spreadsheets/d/1BW12sa9_5N4lAt-ITODO2KPzAxd9b2umZOrRtdeg2_s/edit#gid=1451037787


pezholio (approved):
LGTM 👍

PR #930, opened 2021-02-16
(1501) Store financial quarter and year on transaction

jcoglan:
This PR is the initial work that adds `transactions.financial_{quarter,year}` to the schema, and begins populating them for existing and newly created records. Once this is deployed, we should migrate the UI to set these fields directly, rather than the date.

I have one question related to this so far, which is how the quarter is inferred. There are two possible sources for this inference: the transaction's date, and the transaction's report's financial quarter. Some background information I found out while reviewing our existing data:

- In production, there are currently 3,116 transactions and all have a date set.

- Some transactions do not belong to a report, since they belong to level A/B activities. 2,989 transactions _do_ belong to a report.

- Most transactions belong to "historic" reports, i.e. those containing IATI ingested data, that do not have a financial quarter. Only 420 transactions belong to reports that have a financial quarter.

- Of those, 10 transactions have a date that falls outside their report's quarter; i.e. `FinancialQuarter.for_date(transaction.date)` returns a quarter different from that on `transaction.report`.

### DIFF
--- a/app/models/transaction.rb
+++ b/app/models/transaction.rb
@@ -17,9 +17,22 @@ class Transaction < ApplicationRecord
   validates :value, numericality: {other_than: 0, less_than_or_equal_to: 99_999_999_999.00}
   validates :date, date_not_in_future: true, date_within_boundaries: true
 
+  before_save :set_financial_quarter_from_date
+
   def financial_quarter_and_year
     return nil if date.blank?
 
     FinancialQuarter.for_date(date).to_s
+  end
+
+  private
+
+  def set_financial_quarter_from_date
+    return if date.blank?
+    return if financial_quarter.present? && financial_year.present?
+
+    financial_quarter = FinancialQuarter.for_date(date)
+    self.financial_quarter = financial_quarter.quarter
+    self.financial_year = financial_quarter.financial_year.start_year
   end
 end
--- a/db/migrate/20210216113434_add_financial_quarter_to_transactions.rb
+++ b/db/migrate/20210216113434_add_financial_quarter_to_transactions.rb
@@ -1,0 +1,6 @@
+class AddFinancialQuarterToTransactions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :transactions, :financial_quarter, :integer
+    add_column :transactions, :financial_year, :integer
+  end
+end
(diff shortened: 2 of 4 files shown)


app/models/transaction.rb
```diff
@@ -13,4 +13,15 @@ class Transaction < ApplicationRecord
     :receiving_organisation_type
   validates :value, numericality: {other_than: 0, less_than_or_equal_to: 99_999_999_999.00}
   validates :date, date_not_in_future: true, date_within_boundaries: true
+
+  before_save :set_financial_quarter_from_date
```
mec, line 17: Did you settle on this hook rather than in the create transaciton service as we expect this all to go away when we collect the financial quarter in the interface?

What is you take on using Rails hooks? Lots of folks tend to avoid them - I am not sure where I land yet?

db/data_schema.rb
```diff
@@ -1,2 +1,2 @@
 # encoding: UTF-8
-DataMigrate::Data.define(version: 20210202104522)
+DataMigrate::Data.define(version: 20210216140931)
```
mec, line 2: I am interested to know how we will approach this need to update the data in code if we do decide to get rid of the data migrate gem?

mec (approved):
This looks very neat and tidy - thank you! I have two general question that don't block this, but I love to learn more 👍

db/data_schema.rb
pezholio: I have some Opinons on that, based on how we did it on Claim

mec: @pezholio I like opinions!

app/models/transaction.rb
jcoglan: My strategy here is that I want the app to start converting all input dates into quarters implicitly, no matter how the transaction is created, so it will affect the web form, the bulk importer, etc.

Then we deploy this change, which gets the app to start adding a quarter to all new transactions, and back-fills all the existing ones.

Once that behaviour is in place, we change the UI so you input a quarter rather than a date, and make the model convert the quarter into date, rather than the other way around.

I tend not to use hooks often, and usually only to ensure the internal consistency of a single record, not to perform other side effects since the control flow rapidly becomes confusing and causes surprises. For example here I'm making sure two fields agree with each other, I'll often use hooks to "clean up" data before validation e.g. removing leading/trailing space, etc. But not to, say, send an email -- that doesn't belong in the data layer.

jcoglan: I also only use hooks when you definitely always want that behaviour to implicitly happen on creating/saving a record. If you ever want to create records in a different state in the app or in tests, then that logic does not belong in a hook -- it's something the caller should be doing explicitly.

PR #931, opened 2021-02-16
(1451) Upload financial data journey

CristinaRO:
## Changes in this PR

- The CSV template download links for actuals and forecasts are more prominent (from button style to link style, and given a medium heading size)
- The new upload page for actuals and forecasts includes a modified version of the upload form, where the label is styled to be visually the same weight as the CSV download link
- The error report page for actuals and forecasts has different copy (label and hint) for the upload form (compared to the initial upload page)

### Before
Screenshot 2021-02-16 at 18 22 42
Screenshot 2021-02-16 at 18 22 50
Screenshot 2021-02-16 at 18 24 28
Screenshot 2021-02-16 at 18 24 38

### After
Screenshot 2021-02-16 at 18 26 51
Screenshot 2021-02-16 at 18 26 59
Screenshot 2021-02-16 at 18 27 11
Screenshot 2021-02-16 at 18 27 18

## Next steps

- [ ] What else can be used from the original design to make the upload form more intuitive/friendly?

### DIFF
--- a/db/seeds/activity.rb
+++ b/db/seeds/activity.rb
@@ -85,3 +85,5 @@ Activity.find_or_create_by(third_party_project_params)
     activity.save!
   end
 end
+
+Report.for_activity(first_project).create!(state: :active)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -513,6 +513,7 @@
 - Clicking a link when signed-out should take you to the right place
 - Accept financial quarters instead of dates when inputting transactions
 - Redirect old domains to canonical one
+- Overhaul the interface for uploading financial data (actuals and forecasts)
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-33...HEAD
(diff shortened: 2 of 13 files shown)


db/seeds/activity.rb
```diff
@@ -85,3 +85,6 @@
     activity.save!
   end
 end
+
+report = Report.where.not(state: :approved).find_or_create_by!(organisation_id: delivery_partner.id, fund_id: gcrf_fund.id)
+report.active!
```
jcoglan, line 90: I'm not sure what the effect of calling `find_or_create_by!` on a `where.not` relation is. It might be more reliable to use the abstractions already in the application and do something like this:

```rb
Report.for_activity(first_project).create!(state: :active)
```

jcoglan (approved):

db/seeds/activity.rb
mec: Happy to go with your suggestion - reset my DB with it in place and all looks good, I suppose I would like to see this in its own `reports` seed file, but I thinnk we can live without.

I'll make that change and then merge this.

PR #932, opened 2021-02-16
(1483) Upload activities on index page

CristinaRO:
## Changes in this PR

- Move the link to initiate bulk **activity** upload to the Activities index page

We have decided to not add this link to the top level navigation.

The simplest change that makes sense is to move the activity upload from the individual report page to the Activities index page. The activity upload and the corresponding template are not directly tied to a specific report, which is why we want to move it away from the report page.

### Before
Screenshot 2021-02-16 at 17 54 49

### After
Screenshot 2021-02-16 at 17 54 19

## Next steps

- [ ] Styling the upload activities link (shouldn't stack vertically)


mec:
I am going to push back on this. I think that although technically adding activties is not tied to a report, there is a wider expectation that no data should be changed in RODA unless there is a report in an editable state.

I expect we have some work to do to reflect that expectation in the application and I will happily accept that I am wrong about all this, but for now, I will close this PR and try reflect the changes we might need to do in the backlog and with our designer.

PR #933, opened 2021-02-16
(1458) Show programmes grouped by fund on organisation show page

leeky:
Allow users to see a list of programmes an organisation is extending from the organisation's show page (which is also the first page delivery partners will see when signing into RODA).

The next step will allow users to create a child activity from this page.

### Before

before

### After

download

## Next steps

### DIFF
--- a/app/views/staff/organisations/_grouped_programmes.html.haml
+++ b/app/views/staff/organisations/_grouped_programmes.html.haml
@@ -1,0 +1,28 @@
+- if grouped_programmes.any?
+  .govuk-grid-row
+    .govuk-grid-column-full
+      %h2.govuk-heading-m
+        = t("page_content.organisation.programmes")
+
+      - grouped_programmes.each do |fund, activities|
+        - if activities.any?
+          %table{id: fund.id, class: "govuk-table"}
+            %caption{class: "govuk-table__caption govuk-table__caption--s"}
+              = fund.title
+            %thead{class: "govuk-table__head"}
+              %tr{class: "govuk-table__row"}
+                %th{scope: "col", class: "govuk-table__header"}
+                  = Activity.human_attribute_name(:title)
+                %th{scope: "col", class: "govuk-table__header"}
+                  = Activity.human_attribute_name(:roda_identifier_compound)
+                %th{scope: "col", class: "govuk-table__header"}
+
+            %tbody{class: "govuk-table__body"}
+              - activities.each do |activity|
+                %tr{class: "govuk-table__row", id: activity.id}
+                  %td{scope: "row", class: "govuk-table__cell"}
+                    = activity.title
+                  %td{class: "govuk-table__cell"}
+                    = activity.roda_identifier_compound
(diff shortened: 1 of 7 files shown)


app/views/staff/organisations/_grouped_programmes.html.haml
```diff
@@ -0,0 +1,28 @@
...
+
+            %tbody{class: "govuk-table__body"}
+              - activities.each do |activity|
+                %tr{class: "govuk-table__row", id: activity.id}
+                  %td{scope: "row", class: "govuk-table__cell"}
+                    = activity.title
+                  %td{class: "govuk-table__cell"}
+                    = activity.roda_identifier_compound
+                  %td{class: "govuk-table__cell"}
+                    = link_to(t("default.link.view"), organisation_activity_details_path(activity.extending_organisation, activity), class: "govuk-link")
```
pezholio, line 28: Should we maybe get this in a partial?

app/views/staff/organisations/_grouped_programmes.html.haml
```diff
@@ -0,0 +1,28 @@
+- if grouped_programmes.any?
+  .govuk-grid-row
+    .govuk-grid-column-full
+      %h2.govuk-heading-m
+        = t("page_content.organisation.programmes")
+
+      - grouped_programmes.each do |fund, activities|
+        - if activities.any?
+          %table{id: fund.id, class: "govuk-table"}
+            %caption{class: "govuk-table__caption govuk-table__caption--s"}
```
pezholio, line 10: How does `govuk-table__caption--m` look with this? First instinct when looking at the screenshot was the text was a smidge too small

pezholio (approved):
LGTM - just a couple of questions on the view, feel free to ignore 👍

app/views/staff/organisations/_grouped_programmes.html.haml
leeky: I was going by the design and it's one step down from the medium-sized `<h2>`.

app/views/staff/organisations/_grouped_programmes.html.haml
leeky: I'm not sure - I generally split a partial further if those parts could get re-used elsewhere. I think this is a one-off however.

app/views/staff/organisations/_grouped_programmes.html.haml
leeky: And it turns out that `govuk-table__caption--m` doesn't exist in the outdated version of `govuk-frontend` we've been using for a while... 🐰 🕳️

leeky: Have amended the caption, adding some spacing and fixed the width of the "title" column so the fund tables both line up:

Screenshot 2021-02-17 at 12 33 39

I've also changed "Identifier" to "RODA identifier" to be consistent with the rest of the site.

PR #935, opened 2021-02-17
Revert "Bump data_migrate from 6.6.0 to 6.6.1"

pezholio:
6.6.1 has introduced a new bug, documented here https://github.com/ilyakatz/data-migrate/issues/170


MariluzHerrera (approved):
Thanks for spotting this and fix it!! 🚀

PR #938, opened 2021-02-18
Add RODA ID column to the activity import template

pezholio:
This allows users to update activities, if they want / need to. I've also added an extra test with a minimal CSV to make sure we can update Activities via the upload


leeky (approved):
LGTM! :shipit: 

I wonder if [this conditional](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/1473-show-the-roda-id-column-in-activity-import-template/app/services/activities/import_from_csv.rb#L39) should strip out whitespace from`RODA ID` (since even a single space might cause it to try an `update` rather than `create`)

pezholio:
We're good, because `" ".present?` still returns `false`. Good thought though! 👍

PR #939, opened 2021-02-18
(1490) No search engine indexing on environments other than production

pezholio:
This adds a new environment variable to the application: `ROBOT_NOINDEX`. If it is set to true, the application renders two `meta` tags in the header, which tell search engines to not crawl the site.


mec (approved):
You've got a double 'staging' typo in the second commit but otherwise all good 👍

PR #943, opened 2021-02-19
(1466) Remove ingested from activities

mec:
Some history:

The `ingested` flag was part of the old ingesting of historic activity
data from published activities. This was our first attempt to ingest
historic data and it was difficult for many reasons.

At the time we were also developing the application and that included
adding fields that were entirely new i.e. the fileds would not be in the
ingested data set.

This meant an `ingested` activity would be missing the new fields but
still needed to be considered valid in certain contexts.

You can see how the ingested flag was used to force certain validation
rules.

Why are we removing the field?

We no longer need the ingested flag as as new delivery partners are on
boarded to the service ALL of their historic is prepared by BEIS and
imported into the application replacing everything that would have been
ingested previously - therefore the ingested flag serves no purpose and
should have not impact on an activity's validity.

We performed a lot of checks against production data to be sure (which
are here https://trello.com/c/60NRg2BP) but the short version is:

Of the 949 activities, 674 are invalid before AND after removing the
column so we can see how it has no effect on validity.

We do have an issue of 'what does being valid actual mean' but that is
not part of this work.

We also still have an ingested flad on other entities, but they do not
change their validity so are not a concern for the moment.

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -396,7 +396,7 @@ class Activity < ApplicationRecord
   end
 
   def requires_call_dates?
-    !ingested? && is_project?
+    is_project?
   end
 
   def comment_for_report(report_id:)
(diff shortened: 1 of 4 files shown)


pezholio (approved):
LGTM 👍

PR #944, opened 2021-02-19
(1501) Store financial quarter and year on transaction

jcoglan:
## Changes in this PR

We have already deployed a change that adds `financial_quarter` and `financial_year` to `Transaction`, and fills those in implicitly from the `date` value. This PR changes the input mechanisms to accept quarters instead of dates.

I would particularly like the validation logic checked, as I'm uncertain about how report quarters work. My understanding is that, for example, the _current time_ is that we're in Q4, and I would therefore be filling out the Q4 report, and therefore entering transactions whose quarter is Q4 2020. These would have a date of 31 March 2021 assigned, which is in the future, and the current validation rules disallow that. Is this correct?

## Screenshots of UI changes

The web form for entering a single transaction now uses the same quarter inputs as we do on forecasts.

Screenshot 2021-02-22 at 11 32 32

The bulk upload template now includes financial quarters instead of a date column for each transaction. If any CSV is uploaded with the old schema (using a `Date` column) it will fail.

Screenshot 2021-02-22 at 11 33 55

Both of these input mechanisms default the transaction's quarter to match that of the current report for the relevant activity.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -511,6 +511,7 @@
 ## [unreleased]
 
 - Clicking a link when signed-out should take you to the right place
+- Accept financial quarters instead of dates when inputting transactions
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-33...HEAD
 [release-33]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-32...release-33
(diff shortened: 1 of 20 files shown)


pezholio (approved):
Looks fine code-wise, but, to your question about dates - do we even need the dates now, or can we remove this field now? The only reason I left the column in originally was because I didn't want to deploy a change that both used a field in a data migration and removed the field. Might be one for @mec to confirm though?

jcoglan:
@pezholio I asked about this on Trello and Meyric's advice was to retain the `date` column for now.

jcoglan:
Either way, the data migration to back-fill `financial_{quarter,year}` from `date` has already shipped -- I made sure to deploy that before further work for exactly the reason you've spotted.

pezholio:
Cool, lets crack on in that case :shipit:

PR #946, opened 2021-02-22
Fix flickering test

pezholio:
This test was sometimes failing because the order of the new activities was not guaranteed.

### DIFF
--- a/spec/features/staff/users_can_upload_activities_spec.rb
+++ b/spec/features/staff/users_can_upload_activities_spec.rb
@@ -95,8 +95,7 @@ RSpec.feature "users can upload activities" do
 
       expect(new_activites.count).to eq(2)
 
-      expect(new_activites[0].transparency_identifier).to eq("1234")
-      expect(new_activites[1].transparency_identifier).to eq("1235")
+      expect(new_activites.pluck(:transparency_identifier)).to match_array(["1234", "1235"])
     end
   end
 


spec/features/staff/users_can_upload_activities_spec.rb
```diff
@@ -91,7 +91,7 @@
       expect(page).to have_text(t("action.activity.upload.success"))
       expect(page).not_to have_xpath("//tbody/tr")
 
-      new_activites = Activity.where(created_at: DateTime.now)
+      new_activites = Activity.where(created_at: DateTime.now).order("created_at")
```
jcoglan, line 94: How does ordering the activities by `created_at` when all of them will have the same `created_at` value make this test more stable?

jcoglan: Sorry that sounds like a rhetorical question like I think this won't work -- that's not what I'm saying, I'd like to know why it _does_ work :)

pezholio: D'oh, of course! There's a `freeze_time` block there isn't there?

jcoglan: I didn't notice the `freeze_time` block, I was going purely off the fact that `Activity.where(created_at: t)` will return a set of records with the same `created_at` value of `t`, so ordering by that field won't give a predictable ordering.

pezholio: I've tweaked this now to just match against the `transparency_identifier`s of the newly created activities

pezholio: If we didn't have the `freeze_time` block, the `created_at`s would be fractions of a second different, so the ordering by created_at would work. As it stands we need the `freeze_time` to make sure the `created_at` values are exactly the same

jcoglan: Cool, if the test doesn't mind what order the records appear in this looks good 👍

jcoglan (approved):

PR #948, opened 2021-02-23
(1512) Set correct domain in the worker container

pezholio:
This had me scratching my head for a while, but I discovered that the worker container (which I didn't think we were using!) is used to send emails via Notify, and still has the old domain set, so any emails sent will have the old domain in URLs.


mec (approved):
I can imagine how tricky this was to spot! Nice work! 👍

PR #950, opened 2021-02-24
Release 34

pezholio:
- Clicking a link when signed-out should take you to the right place
- Accept financial quarters instead of dates when inputting transactions
- Redirect old domains to canonical one
- Overhaul the interface for uploading financial data (actuals and forecasts)


mec (approved):
🚢

PR #951, opened 2021-02-24
(1507) Jettison Data Migrate gem into the sun

pezholio:
image

This removes the Data Migrate gem and adds a simple generator for us to create our data migrations. This means we have to run our migrations manually, but this encourages us to be careful an deliberate with our data migrations, so I think this is fine.


db/schema.rb
```diff
@@ -48,17 +48,16 @@
     t.boolean "call_present"
     t.date "call_open_date"
     t.date "call_close_date"
-    t.string "intended_beneficiaries", array: true
```
CristinaRO, line 51: What makes these fields jump around, and could we separate this change from the others? I'm not saying do it or not, I'm trying to understand why sometimes we get in this cycle of different machines reordering existing fields in the schema. 😬

CristinaRO (approved):
One non-blocking question.

Thank you for writing up the ADR and the reasoning - I was/am ever so slightly worried about running migrations manually, but it makes sense given the alternative, and I see the point about being deliberate.

db/schema.rb
pezholio: Yeah, I don't know why it does that, it's annoying as hell. Usually when this happens, I just commit the relevant lines, but I was running on autopilot and forgot, so have edited the commit 👍

CristinaRO (approved):

tahb:
Hello, I've ended up down a rabbit hole! After being involved in adding the `data_migrate` gem _in_ with the best intentions, it has been absolutely excellent to be able to come back and follow the story to understand why it didn't work out. Thank you for this ADR @pezholio.

PR #953, opened 2021-02-24
Feature/1439 add historic actuals transactions to the export csv

jcoglan:
## Next steps


mec:
I'll read through and comment as I go - hope that is okay.

First comment is based on:

>  having long lists of "0.00" values everywhere.

This is exactly what we expect to happen so we should no be trying to avoid it. Have a look at an 'svt', this is the file BEIS use to do most of the their work currently, here is a link you want the 'GCRF MT' tab:

https://docs.google.com/spreadsheets/d/1qi3ZhzdKw8HNVAqSzfae3MgmGgni2GAI/edit#gid=965001144

This should give you a sense of what we are trying to get to, notice how the actuals columns are mostly 0.

This file is from the Q2 report, so you'll notice how the actuals go from Q2 backwards and the forecasts start at Q3 and go forwards, all chronologically by financial quarter.

mec:
Point one: Summerised as taking a financial quarter not a report.

100% agree, this is exactly what it looks like to me. Even when generating the report csv, the report is really only used to set the 'current period' (for want of better language) or the mid-point between the acutals and forecasts, it doesn't matter if there was a prior report or not, they still want the previous actual data and future forecasts in the report.

Sticking with actuals for the moment, this means that the report csv is given a financial quarter and it then gets all prior financial quarter actuals and sums them per financial quarter going back n number of quarters.

Looking at the SVT in the previous comment might help.

mec:
Point two: summerised as perf!

Yep, 100% agree again! This is why we have to refactor this in my view, the performance here is horrible and it is mostly because we rely on a Report to get what we need, when actually all we want is all transactions sumed per financial quarter.

In my simple world this looks something like a group query on each activity:

`act.transactions.group(:financial_quarter, :financial_year).sum(:value)`

Which gets us all the figures we want, I would then map those against all the financial quarters we know we want, leaving 0.00 where there is no match.

I am sure this can be optimised further, but I do belive even this would improve things.

mec:
Point three: summerised as which columns!? 

This is similar to my first point which is we want all the columns from the 'current period' going back n number of financial quarters, a column for each financial quarter, if there is no value, we want to show the 0 - there is no need to try and do anything more complex.

When we concatenate reports together this is the syptom of the design and not the actual need, this need is really something like

> I need all the report files combined for the same financial quarter

So again here, becasue we take a report we stuggle, which report? Well it's all of them, that doesn't work?!? 😞 

But what we actual want is simply all the data for a given financial quarter, which from the user perspective is the same thing but from our quite different!

Again this is another reason to be refactoring all this, I am not saying it is simple though and will propably need to be introduced over time.

mec:
The other parts:

Transaction without a financial quarter and year - this is just wrong, when we introduce financial quarter and financial year to transaction we need to ensure they all have one set, based on the date, as you demonstrate, all Transactions have a date to derive a financial quarter and year from.

The trouble with report: I think all this screams out how 'report' as a concept is just making life very difficult, but as you say, for forecasts it is important to know which 'version' of the forecast you want. I believe this is still really a financial quater and year, but it is encapsulated in our idea of 'report' so 

> show me the value of this activitys Q4 2020-2021 forecast when it was 'reported' in Q1 2020-2021?

Here we can see that we want the version that relates to the report, but the report (in this context) is just anohter way of saying 'show me the version from Q1 2020-2021'.

Reports with transactions from previous financial quarters?!!? Yes indeed, a report could be thought of as 'all the changes to the data set given in a specified period' - we fully expect users to forget to add a transaction in a reporting period, ideally this would not get through BEIS assurance as, in theory, this would show variance against the forecast that BEIS would pick up on, but I suspect life is not that perfect. And once again, this goes into the reasons we have to refactor the exportToCsv class and related summing methods, now not just for performance but as they simply do not do what we need any longer.

mec:
The final part: I hope that all makes some sense, happy to talk more, here, on a call or in Slack - whatever works for you.

I think you instinct is spot on, the code that sums the actuals can simply take a financial quater and year and gather up the n number of totals from there back.

Forecasts are, as ever, more complex as we have the 'which version' issue to contend with. I thought that internally they still rely on what is essentailly a financial quater and year - but I haven't looked at them for a while and so the detail has been replaced in my head. Conceptually I thought they work in a way where I can ask for the version of the forecast for a given financial quarter and year and I get back the version from that date or the last version before that date if it was not altered on that date (report) (goodness! even reading that back is complicated!) - we need to change the number of forecasts in the report csv (captured here: https://trello.com/c/kejDCbwK so I suspect sitting down and going over all of this and coming up with a strategy would be wise?

I fully expected these stories to result in refactoring of underlying code, I generally work to the principal that is expected and anyone working on these who is concerned can raise it and break out more detailed cards to manage any transition, happy to talk about that as well if that is not working for us or if that needs to be made more explicit? 

And as ever I would welcome anyones contributuion to writing the cards up so the expectation is set on them as early as we can.

I think we may want to talk over a good strategy to handle all this and propose defining what that looks like as our next step, what do you think?

jcoglan:
This is probably ready for review now. A detailed explanation of all the changes is in the commit messages, I think the most critical is `ed8e5fb0 (Introduce TransactionOverview for loading per-quarter transaction totals, 2021-03-02)` which explains how I'm loading past quarters' transaction data and how it relates to report versioning. Probably best to read the commits in order and check they make sense though.

jcoglan:
This is blocked on #962, which contains changes that conflict with this branch. I'd rather merge that PR first and then resolve the conflicts in this PR.

CristinaRO:
> This is blocked on #962, which contains changes that conflict with this branch. I'd rather merge that PR first and then resolve the conflicts in this PR.

Update: #962 has been superseded by #973 (still blocking it though).

PR #956, opened 2021-02-25
(1515) Make the transaction upload CSV show activities hierarchically

pezholio:
This is a two-step process, the first commit removes any activities that aren't reportable (i.e. not current) from the `reportable_activites` scope. The second commit uses the `hierarchically_grouped_projects` method we already use on the report variance page to organise the activities hierarchically. We hope this will make is easier for delivery partners to find the relevant activity, particularly when there are a lot of activities to parse in a big CSV.

# Excel Screenshots

## Before

image

# After

image


leeky (approved):
LGTM! :shipit:

PR #959, opened 2021-03-01
(1460) Delivery partner can add child activities to a level B parent

leeky:
## Changes in this PR

Add the new "Add child activity" link on the organisation show page. This link is only visible to delivery partners. 

Clicking on this link will create a new activity which is already has its `source_fund_code`, `parent_id` and `level` set, among other attributes. The user will then continue creating the activity as usual, only starting from the 'delivery partner identifier' step of the form.

This PR introduces the `Activity.new_child` method, which initialises a new Activity with sensible defaults set (based on the parent activity and delivery partner organisation passed in). This, along with the `ActivityDefaults` service object, can likely DRY up other parts of the app.

NB: This doesn't yet remove the old path to adding activities (manually choosing a level and parent), but that'll be done in a separate PR once adding child activities to level C has been done, as removing it now breaks a *lot* of tests.

### Before

Screenshot 2021-03-01 at 12 32 22

### After

Screenshot 2021-03-01 at 12 33 13

## Next steps


app/models/activity.rb
```diff
@@ -228,6 +228,15 @@ class Activity < ApplicationRecord
...
   }
 
+  def self.new_child(parent_activity:, delivery_partner_organisation:, &block)
+    attributes = ActivityDefaults.new(
+      parent_activity: parent_activity,
+      delivery_partner_organisation: delivery_partner_organisation
+    ).call
+
+    new(attributes, &block)
+  end
```
pezholio, line 238: ❤️ it!

pezholio:
I'll have a proper look tomorrow, but this looks great from an initial drive-by

app/views/staff/organisations/_grouped_programmes.html.haml
```diff
@@ -25,4 +25,6 @@
                   %td{class: "govuk-table__cell"}
                     = activity.roda_identifier_compound
                   %td{class: "govuk-table__cell"}
-                    = link_to(t("default.link.view"), organisation_activity_details_path(activity.extending_organisation, activity), class: "govuk-link")
+                    - if policy(:project).create?
+                      = link_to(t("action.activity.add_child"), organisation_activity_children_path(activity.extending_organisation, activity), method: :post, class: "govuk-link govuk-!-margin-right-4")
```
pezholio, line 29: Just wondering if this should be a `button_to` as it's an action that results in an activity being created? The `view` link could then be linked to the activity title

pezholio (approved):
Looks really good. Just one teensy thing about the layout, but happy to let it slide and we can iterate later 👍

app/views/staff/organisations/_grouped_programmes.html.haml
leeky: I did wonder about that - there are several places where we have "View" links and others where we make the title clickable. The only problem with the latter is how you render empty titles.

I'm never sure about having rows of "primary action" style buttons on the same page. Might see if I the design system will let me style a button *as* a link.

leeky: Think I'm going to leave this alone for now, even though it's not ideal.

PR #961, opened 2021-03-01
(1517) Make file names descriptive, part 1: Add BEIS org reference to organisations

CristinaRO:
## Changes in this PR

- New attribute on `Organisation`, `beis_organisation_reference`, currently set via data migration only
- Data migration to populate the new attribute

## Next steps

- [ ] Use the new attribute to generate better filenames for downloaded files
- [ ] Implement validations for the new attribute
- [ ] Expose `beis_organisation_reference` in create and edit forms


pezholio (approved):
LGTM 👍

PR #962, opened 2021-03-01
(1517) Make file names descriptive, part 2: generate filenames

CristinaRO:
## Changes in this PR

- Downloaded files have more descriptive filenames, using financial quarter, source fund, and organisation short name

## Next steps

- [ ] Implement validations for `beis_organisation_reference`
- [ ] Expose `beis_organisation_reference` when creating and editing an organisation


app/controllers/staff/activity_uploads_controller.rb
```diff
@@ -13,7 +13,10 @@ def new
   end
 
   def show
-    stream_csv_download(filename: "activities.csv", headers: csv_headers)
+    @report_presenter = ReportPresenter.new(@report)
+    filename = "#{ERB::Util.url_encode(@report_presenter.financial_quarter_and_year)}-#{ERB::Util.url_encode(@report_presenter.fund.title)}-#{ERB::Util.url_encode(@report.organisation.beis_organisation_reference)}_activities_upload.csv"
+
+    stream_csv_download(filename: filename, headers: csv_headers)
```
jcoglan, line 19: I'm wondering if I should have made `stream_csv_download` perform the URL-encoding of the filename -- it's really the whole thing that needs to be encoded for the context it's being put into, not individual pieces of it, and the above could be written

```rb
filename = ERB::Util.url_encode("#{@report_presenter.financial_quarter_and_year}-#{@report_presenter.fund.title}-#{@report.organisation.beis_organisation_reference}_activities_upload.csv")
```

app/presenters/report_presenter.rb
```diff
@@ -13,6 +13,6 @@ def deadline
 
   def financial_quarter_and_year
     return nil if financial_quarter.nil? || financial_year.nil?
-    "Q#{financial_quarter} #{financial_year}-#{financial_year + 1}"
+    "FQ#{financial_quarter} #{financial_year}-#{financial_year + 1}"
```
jcoglan, line 16: My branch `feature/1439-add-historic-actuals-transactions-to-the-export-csv` removes this method and all other implementations of `financial_quarter_and_year` into a single module -- does this method need to be different for reports, or should all instances of `financial_quarter_and_year` have an `F` at the start?

jcoglan:
We should probably put a commit at the start of this that makes `stream_csv_download` deal with URL-encoding the filename -- it would clear a lot of the following changes up as well as being safer -- callers of this method should not need to know how the filename is encoded in the response.

There's a few conflicts with my open PR where it affects `ExportActivityToCsv` but I can deal with those if I rebase my branch after this is merged.

app/presenters/report_presenter.rb
CristinaRO: I don't know, this is one of those cases where I don't want to build a bikeshed for us to paint, but I need to find out if this "Q" versus "FQ" is really important for the users. (It seems *to me* that having it consistent everywhere would be preferable, but there might be business logic reasons I'm missing.)

app/controllers/staff/activity_uploads_controller.rb
CristinaRO: I did aggregate all of this together in one of the later commits, where I'm making the report presenter responsible for generating the filenames, and then it's encoding the whole string at the end.

I didn't think of messing with `stream_csv_download`, I can try that out.

CristinaRO:
> We should probably put a commit at the start of this that makes `stream_csv_download` deal with URL-encoding the filename -- it would clear a lot of the following changes up as well as being safer -- callers of this method should not need to know how the filename is encoded in the response.
> 
> There's a few conflicts with my open PR where it affects `ExportActivityToCsv` but I can deal with those if I rebase my branch after this is merged.

I am also OK to deal with conflicts if your PR gets merged before this one!

jcoglan:
No no, it would add a relatively large amount of noise to this PR vs a relatively small amount to my PR, or at least that's how it looks given my current familiarity with each.

jcoglan:
I could sort out the URL encoding stuff on this branch, if you like @CristinaRO ?

CristinaRO:
> I could sort out the URL encoding stuff on this branch, if you like @CristinaRO ?

Yes, please. :)

CristinaRO:
Superseded by https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/973

PR #965, opened 2021-03-02
Release 35

leeky:
- Remove Data Migrate gem and run data migrations manually
- Activities can recursively total the transactions for all of their children
- Add an attribute to `Organisation` to record the short name, `beis_organisation_reference`,
- Delivery partners can add level C activities, automatically parented to an existing level B


pezholio (approved):
LGTM 👍

PR #967, opened 2021-03-02
(1459) Users can add a child activity to an existing project

leeky:
On the Activity "child activities" tab, change the "Add activity" button so that it points to the new `ActivityChildren#create` action.

Because the new activity is automatically associated with its parent activity and the `form_state` is set to beyond the 'level' and 'parent' steps, the `fill_in_activity_form` test helper breaks.

Therefore, I've temporarily added a `skip_level_and_parent_steps` option to the helper method, which will be removed once the old activity creation code becomes obsolete.

To prevent breaking the audit-trail tests, I've added a `create_activity` call in `ActivityChildrenController#create` which mirrors the existing behaviour in `ActivitiesController#create`.

## Next steps


jcoglan:
Why is it not possible to create a level D activity unless a relevant report exists?

app/policies/third_party_project_policy.rb
```diff
@@ -1,2 +1,14 @@
 class ThirdPartyProjectPolicy < ProjectPolicy
+  def create?
+    !beis_user? && editable_report?
+  end
+
+  private
+
+  def editable_report?
+    # FIXME: Remove once 'level' form_step removed
+    return false if record.is_a?(Symbol)
```
CristinaRO, line 10: This is probably just my confusion: in which situations can the record still be a symbol here? And how does it relate to removing the `form_step`?

leeky: You can call Pundit two ways:

- with a symbol, e.g. `policy(:third_party_project).create?` which checks to see if the current user can perform an action on *any* third-party project activity.
- with a model instance, e.g. policy(@activity).create?` you can check if the current user can perform an action on that specific third-party project activity .

The `level` form step in the wizard calls a view helper `create_activity_level_options` which is the only other place the `create?` is called on this policy.  I thought it was better to add a guard clause and comment rather than fixing tests that will I'm going to remove in my next PR.

CristinaRO (approved):

PR #968, opened 2021-03-02
(1541) Fix inconsistencies with activity tabs and their ARIA states

leeky:
There were a few places where the ARIA selected hint didn't match the currently selected tab, so I'm using a bit of logic to tie the CSS classes and ARIA attributes together.

## Next steps


app/views/staff/shared/activities/_tab_nav_item.html.haml
```diff
@@ -0,0 +1,4 @@
+- selected = controller_name.ends_with?(tab)
+%li{class: ["govuk-tabs__list-item", ("govuk-tabs__list-item--selected" if selected)] }
```
CristinaRO, line 2: Tiny comment: would it be neater to add a `.compact` at the end of the array, or does the `nil` class not produce any cruft in the output at all?

CristinaRO (approved):
I like the views having a Railsier structure! I was always struggling to guess their place in the file tree. And good catch on the ARIA stuff.

app/views/staff/shared/activities/_tab_nav_item.html.haml
leeky: It actually swallows any `nil`s automatically :)

PR #972, opened 2021-03-04
(1420) Data migration to remove all legacy data

mec:
__** Requires data migration to be run manually once deployed **__

This data migration will remove all data for known  Delivery Partner
organisations.

The goal of this work is to remove all the data we have imported via the
legacy IATI data for:

- Royal Academy of Engineering
- British Academy

We are doing this because BEIS would like to change the RODA identifier
for all the activities which work 'top down' i.e. any changes to the ID
at the top level need to propagate down so effectively we would need to
change all the IDs. We decided that as the data is to be replaced
entirely (see below) the simplest approach is to remove the data in its
entirety.

It is safe to carry out this action as neither organisation has used RODA
to report as yet and the new data migration process will
replace all of the data with a cleaned and verified data set prepared
for this purpose.

This code has been run against a copy of production data and had the
results verified, we decided, on balance that the change here is at such
a high level that any automated test would not increase our confidence
in this code doing the right thing beyond that of running it locally.

The expectation, as with all data migrations, is that this will be
deleted once it has been deployed and run.

### DIFF
--- a/db/data/20210303102617_remove_royal_accademy_of_engineering_and_british_accademy_legacy_data.rb
+++ b/db/data/20210303102617_remove_royal_accademy_of_engineering_and_british_accademy_legacy_data.rb
@@ -1,0 +1,30 @@
+# Run me with `rails runner db/data/20210303102617_remove_royal_accademy_of_engineering_and_british_accademy_legacy_data.rb`
+#
+# Put your Ruby code here
+def remove_all_data_set_for_organisation_id(organisation_id)
+  # collect all activities
+  activities = Activity.where(organisation_id: organisation_id).or(Activity.where(extending_organisation_id: organisation_id))
+  activity_ids = activities.pluck(:id)
+
+  # destroy all transactions
+  transactions = Transaction.where(parent_activity_id: activity_ids)
+  transactions.destroy_all
+
+  # destroy all forecasts
+  forecasts = PlannedDisbursement.where(parent_activity_id: activity_ids)
+  forecasts.destroy_all
+
+  # destroy all budgets
+  budgets = Budget.where(parent_activity_id: activity_ids)
+  budgets.destroy_all
+
+  # destroy all activites
+  activities.destroy_all
+end
+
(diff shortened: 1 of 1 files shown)


db/data/20210303102617_remove_royal_accademy_of_engineering_and_british_accademy_legacy_data.rb
```diff
@@ -0,0 +1,29 @@
...
+  # destroy all budgets
+  budgets = Budget.where(parent_activity_id: activity_ids)
+  budgets.destroy_all
+
+  # destroy all activites
+  activities.destroy_all
+end
+
+royal_accademy_of_engineering_id = Organisation.find_by(name: "Royal Academy of Engineering", iati_reference: "GB-CHC-293074")&.id
+british_accademy_id = Organisation.find_by(name: "British Academy", iati_reference: "GB-COH-RC000053")&.id
```
jcoglan, line 26: Should these raise an error if the given records are not found? As written, this script could silently do nothing if the expected organisation records are not loaded.

mec: It's a fair point, I originally wrote it with the IDs of the organisation in the data migraton as this is explicitally to be run in production only, where the IDs are known. 

Then I changed it as testing with known IDs became too complicated, I should have changed it back at the point we decided that a migration like this one, that removes a bunch of data, probably doesn't need a test to give us confidence, or rather writting test for it feels like too much effort for the confidence it would bring, we felt running it against a copy of the production database would deliver more confidence.

If I switch back to the known IDs I can guarantee the code will execute, how does that sit with you?

jcoglan: I think using the names rather than IDs makes the code clearer, but we could make the script error out if the required records are not found?

pezholio (approved):
LGTM 👍

PR #973, opened 2021-03-04
(1517) Generate descriptive filenames for downloadable CSVs

CristinaRO:
## Changes in this PR

- Downloaded files have more descriptive filenames, using financial quarter, source fund short name, and organisation short name
- The financial quarter display name is now in the format `FQ1 2020-2021`

## Next steps


app/controllers/staff/reports_controller.rb
```diff
@@ -162,7 +162,7 @@ def approved_reports
   end
 
   def send_csv
-    filename = "#{@report_presenter.fund.title}-#{@report_presenter.financial_quarter_and_year}-#{@report.description}.csv"
+    filename = "#{@report_presenter.financial_quarter_and_year}-#{@report_presenter.fund.roda_identifier_fragment}-#{@report.organisation.beis_organisation_reference}-#{@report.description}.csv"
```
mec: I would be tempted to even drop the user specified `description` here and just set it to `_report`, my reasoning being that so far the descriptions BEIS have used don't really bring anything informational and so the file names end up looking very duplicated, here is one from production:

`FQ4 2020-2021-GCRF-UKSA-Q4 Report - FY 2020_21.csv`

And the user can always modify the filename at the download point or later on their device.

mec (approved):
These file names are such an improvement!

Lovely set of commits that I could follow the story of, thank you.

I left one comment about not including the user specified report description, but don't let it block this - if you think it brings value, I am happy to go with this 🚢

PR #990, opened 2021-03-15
Add Coveralls GitHub action

mec:
Rather than have this hidden away in out internal scripts, we can use a
nice simple GitHub action.

## Next steps


mec:
Just a look at coveralls github action.

PR #991, opened 2021-03-15
(1487) Implementation of RODA email notifications

pezholio:
This adds emails to be sent to users when a report is activated, submitted, awaiting changes, or approved. These are usually sent to all delivery partners in the report's organisation, apart from when a report is submitted, when an email is sent to all service owners and all delivery partners in the report's organisation. We use `mail-notify`'s `view_mail` functionality, which makes it easier for us to keep the email templates tightly coupled to our code and versioned.

# Screenshots

## Activated

image

## Submitted (delivery partner version)

image

## Submitted (service owner version)

image

## Awaiting Changes

image

## Approved

image


app/services/report/send_state_change_emails.rb
```diff
@@ -0,0 +1,54 @@
...
+      send_mail_to_users(:awaiting_changes)
+    end
+
+    def send_approved
+      send_mail_to_users(:approved)
+    end
+
+    def send_mail_to_users(action, users = delivery_partners)
+      users.each do |user|
+        ReportMailer.with(report: report, user: user).send(action).deliver_later
```
jcoglan, line 42: Does the use of `deliver_later` mean we're now depending on ActiveJob or Sidekiq or some other background task runner?

pezholio: Yeah, we've have Sidekiq running for a while now and delivering use invitiation emails, see: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/7dea2b943cc2a686a310e7c5bd00437dde135e43/app/services/send_welcome_email.rb#L9

PR #995, opened 2021-03-16
(1571) Remove accountable_organisation fields

leeky:
These fields can be inferred from `extending_organisation`, so they don't need to be stored separately.

## Next steps


app/models/activity.rb
```diff
@@ -346,6 +340,36 @@ def funding_organisation
     service_owner
   end
 
+  def accountable_organisation
+    return service_owner if fund? || programme?
+
+    extending_organisation.is_government? ? service_owner : extending_organisation
+  end
+
+  def accountable_organisation_name
```
CristinaRO, line 349: These getters aren't being tested right now, is that ok?

CristinaRO (commented):
One question re: the code. Also, did you intentionally not describe this change in the changelog?

leeky:
@CristinaRO I usually only add a changelog entry for features that impact the end user, as per the instructions on the checklist itself.

app/models/activity.rb
leeky: Good point! I've pushed an update which adds a spec for each of the getters

CristinaRO:
> @CristinaRO I usually only add a changelog entry for features that impact the end user, as per the instructions on the checklist itself.

Eh, I find that guideline to be highly subjective. I tend to err on the left side of the equation ("an entry should always be made"), because what is "a small tweak which has no impact outside the development team" is fuzzy to me.

CristinaRO (approved):
Thanks for the clear commit messages and making the tweaks!

PR #997, opened 2021-03-16
Update terraform docs

jcoglan:
## Changes in this PR

This updates `terraform/README.md` to accurately reflect what we did to deploy to `pentest` most recently, correcting some mistakes and filling in some missing details that might not be obvious to people who've not done this process before.

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -45,40 +45,21 @@ to do one.
   in the item "terraform state S3 bucket credentials". The CF credentials should
   be your username and password for GPaaS. Once set, load these variables into
   your shell by running `source deploy-credentials.sh`
+- download a copy of the Terraform variables from the RODA 1Password vault;
+  they're in the item "pentest.tfvars" -- save them to a file named
+  `pentest.tfvars` in the current directory
 - build and push a new docker image if needed `cf app` can tell you what image
   is currently being used [DockerHub has a list of existing docker images you
   can
   use](https://hub.docker.com/repository/docker/thedxw/beis-report-official-development-assistance/tags?page=1),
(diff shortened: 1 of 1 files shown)


terraform/README.md
```diff
@@ -1,62 +1,76 @@
...
+environment.
 
 We store our state in an s3 bucket on the PaaS in the terraform space.
 
 Deploys will happen automatically from Github Actions
 
 ## Manual Deployment
 
-We should avoid manual deployments but the following instructions will allow you to do one.
+We should avoid manual deployments but the following instructions will allow you
```
mec, line 15: Could we change this to account for the fact we have to manually deploy the training/pentest environment? Perhaps somehting like:

> We avoid manual deployments on production and staging, but require them on our training environment (pentest).

terraform/README.md
```diff
@@ -1,62 +1,76 @@
...
+- set environment variables for your AWS and CF credentials by creating a file
+  named `deploy-credentials.sh` containing the following:
+
+  ```
+  export AWS_ACCESS_KEY='...'
+  export AWS_SECRET_ACCESS_KEY='...'
+  export CF_USER='...'
+  export CF_PASSWORD='...'
+  ```
+
```
mec, line 42: Could we add somehitng here like 'Never commit sensitive information like keys, secrets or passwords'?

mec (changes requested):
This is so much better. Thank you for taking the time to do this. I left two comments I think we should address.

jcoglan:
@mec I've made the changes you requested now.

mec (approved):
🚢

PR #999, opened 2021-03-17
Fix organisation seeds

robbiepaul:
## Changes in this PR

When setting up the repo for the first time I received an error as part
of the database seeding process.

```
NoMethodError: undefined method `organisation' for nil:NilClass
```

This is was caused by the organisation records not being created due to
being considered invalid, meaning that no users were created.

There was a new validation added in a95d2f92c2c62721c4a80318ceed0c22da139426
to the Organisation model that checks for the validity of the
`beis_organisation_reference`, this field wasn't part of our sample
data for our seeds.

I've added the `short_name` property to the sample data YAML file and
populated it using the corresponding values from the
`vendor/data/beis_organisation_references.csv` file.

I've also added a bang to the the organisation update action so that
if something similar happens again, the error will be thrown much closer
to where the issue is.

## Next steps

### DIFF
--- a/db/seeds/organisations.rb
+++ b/db/seeds/organisations.rb
@@ -4,10 +4,11 @@ organisations = YAML.safe_load(File.read(File.join(Rails.root, "db", "seeds", "o
 organisations.each do |organisation|
   organisation_params = {
     name: organisation["name"],
+    beis_organisation_reference: organisation["short_name"],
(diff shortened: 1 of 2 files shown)


jcoglan (approved):

jcoglan:
Thanks for fixing this @robbiepaul :)

PR #1000, opened 2021-03-18
(1579) Remove ExtendingOrganisationController

leeky:
Changing an activity's `extending_organisation` without affecting the `organisation` attribute, and those of its children will likely cause lots of issues, so let's remove the ability to do this for now. 

## Next steps


pezholio (approved):
I'm all for deleting code, LGTM! 👍

PR #1002, opened 2021-03-18
(1555) Implement the GOV.UK cookie component banner

pezholio:
This adds the [GOV.UK Cookie Banner](https://design-system.service.gov.uk/components/cookie-banner/) to the site, as well as a [Cookie settings](https://design-system.service.gov.uk/patterns/cookies-page/#letting-users-accept-or-reject-cookies-on-the-cookie-page) form to the Cookies page, using the [GOV.UK Cookie Consent](https://github.com/dxw/govuk-cookie-consent) JavaScript library, which I cobbled together based on a similar one for the NHS. 

This works by initially setting all cookie-setting JavaScript to a content-type of `text/plain`, so it doesn't get loaded until express permission has been given. At that point the JS changes the content-type to `application/javascript`, so it then gets executed.

We've taken the decision to make this functionality JS-only, and assume that non-JS users have not consented to tracking, which means that we've removed the non-JS fallback for Google Tag Manager.

## Screenshots

image

image

image

image


robbiepaul (approved):
LGTM 👍

PR #1004, opened 2021-03-19
(1582) Only create activities from a known parent

leeky:
Remove the remaining places where activities can be added without a parent, and link any existing "Add activity" buttons to the new `ActivityChildrenController#create` action.

## Next steps


app/views/staff/organisations/show.html.haml
```diff
@@ -15,9 +15,8 @@
           = t("create_programme", scope: "page_content.organisation")
 
         - @funds.each do |fund|
-          = link_to t("form.button.activity.new_child", name: fund.title),
-            organisation_activity_children_path(@organisation_presenter, fund.id),
-            method: :post, class: "govuk-button govuk-!-margin-right-4"
+          = form_tag organisation_activity_children_path(@organisation_presenter, fund), method: :post do
+            = submit_tag t("form.button.activity.new_child", name: fund.title), class: "govuk-button govuk-!-margin-right-4"
```
pezholio, line 19: You could use [`button_to`](https://apidock.com/rails/ActionView/Helpers/UrlHelper/button_to) here

app/services/activity/updater.rb
```diff
@@ -40,7 +40,7 @@ def set_parent
       case activity.level.to_sym
       when :programme then UpdateActivityAsProgramme.new(activity: activity, parent_id: parent_id).call
       when :project then UpdateActivityAsProject.new(activity: activity, parent_id: parent_id).call
-      when :third_party_project then UpdateActivityAsThirdPartyProject.new(activity: activity, parent_id: parent_id).call
+        # when :third_party_project then UpdateActivityAsThirdPartyProject.new(activity: activity, parent_id: parent_id).call
```
pezholio: Can this line just be removed, given we've deleted the service?

app/views/staff/organisations/show.html.haml
leeky: I always forget about `button to` ! 🤦

app/views/staff/shared/activities/_projects.html.haml
```diff
@@ -3,6 +3,6 @@
     %h2.govuk-heading-m
       = t("page_content.activity.projects")
     - if policy(:project).create?
-      = form_tag(organisation_activities_path(current_user.organisation), method: "post") do
+      = form_tag(organisation_activity_children_path(activity.extending_organisation, activity), method: "post") do
```
CristinaRO, line 6: Could you add something about this change to the commit message, so we understand the reason for the change?

app/services/activity/updater.rb
```diff
@@ -39,7 +39,7 @@ def set_parent
 
       case activity.level.to_sym
       when :programme then UpdateActivityAsProgramme.new(activity: activity, parent_id: parent_id).call
-      when :project then UpdateActivityAsProject.new(activity: activity, parent_id: parent_id).call
+        # when :project then UpdateActivityAsProject.new(activity: activity, parent_id: parent_id).call
```
CristinaRO: This service is accumulating commented lines. Could you remove them at the point where they become deprecated?

app/helpers/form_helper.rb
```diff
@@ -69,7 +69,7 @@ def create_activity_level_options(user:)
       policy.create? || policy.update?
     }
 
-    authorised_levels = authorised_levels.reject { |level| level == "fund" || level == "third_party_project" }
+    authorised_levels = authorised_levels.reject { |level| level == "fund" || level == "third_party_project" || level == "project" }
```
CristinaRO, line 72: Could you add more detail about this change to the commit message, please?

db/data/20210319102418_make_organisations_consistent.rb
```diff
@@ -0,0 +1,23 @@
...
+
+# Put your Ruby code here
+
+service_owner = Organisation.find_by(service_owner: true)
+
+def extending_organisation_for_activity(activity)
+  if activity.level == "programme"
+    activity.children ? activity.children.first.organisation : activity.organisation
+  else
+    activity.parent.organisation
```
CristinaRO, line 11: I'm getting mired in organisations and extending organisations, parents and children, so I might have it wrong here, but:

For a *project* activity, the extending organisation should be the delivery partner org, if we know it. If the ext org for a project is nil, we have to get the info elsewhere.

But if we get the organisation of the parent activity, which might be (must be?) a *programme*, the org of the programme is going to be BEIS, the `service_owner`, so we end up with the extending org of the project being BEIS, when it should be the delivery partner (*if* we have this info somewhere in the programme).

Is this what is intended by this code?

CristinaRO (changes requested):
I have a few requests and a few questions. I am quite confused by what is intended, and the Miro board didn't entirely help compensate for the lack of narrative in the original Trello card. 😬

app/services/activity/updater.rb
leeky: Good point - think I made the mistake of commenting them out whilst I was figuring this all out, but didn't go back and removing them entirely 👍

app/views/staff/organisations/show.html.haml
CristinaRO: Was this meant as an agreement to use `button_to` now, or in the future (but not changing the code now)?

leeky: @CristinaRO I'm going to change to using `button_to` now.

db/data/20210319102418_make_organisations_consistent.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+end
+
+Activity.where(extending_organisation: nil).each do |activity|
```
leeky, line 15: On production, there are no activities with an empty `extending_organisation` whose form is complete, so not how useful this will be.

I'm wondering whether this needs to be in a separate branch which also includes fixes to the policy files to enforce based on this change?

pezholio: Yeah, I did wonder how useful this would be on reflection. Happy to cherry pick it into another branch.

pezholio: I've cherry picked this commit into a temp branch, so feel free to delete it

leeky: @pezholio Does the above (second paragraph) seem reasonable..?

pezholio: @leeky "fixes to the policy files to enforce based on this change?" - Is there a story for this?

leeky: @pezholio Not yet!

db/data/20210319102418_make_organisations_consistent.rb
leeky: We've removed this commit from the PR, because the data migration wasn't fixing all places where the data was inconsistent.

I've also realised that the activity policy files should get updated at the same time, as they make assumptions based the way `extending_organisation` was used quite a while ago.

app/views/staff/shared/activities/_projects.html.haml
CristinaRO: Having discussed this over a Slack call, I interpret my confusion as being caused by this commit doing two things at once:
- removing the ability to choose Project (Level C), like the commit message says
- fixing the bug? misunderstanding? where we were sometimes using the wrong organisation to create the project, which is not mentioned in the commit message (`current_user.organisation` -> `activity.extending_organisation` in here, various other changes in specs)

Is this correct @leeky ?


It's this bugfix? misunderstanding-fix? that I would like to see documented in the commit message, because it might be the only institutional memory we have for this. (Think about you-six-months-from-now, and whether you'd be able to trace your own steps if you had to read this cold during a support shift. ;) )

CristinaRO (changes requested):

app/views/staff/shared/activities/_projects.html.haml
leeky: I'll update this tomorrow - I get your point, I think I was avoiding explaining the same changes four times - one for each level!

leeky: Mostly this is about creating a child activity based on its parent, rather than assuming that `current_user.organisation` will be the same thing.

CristinaRO: > I'll update this tomorrow - I get your point, I think I was avoiding explaining the same changes four times - one for each level!

Thanks! Replacing the org with extending_org isn't explained at all, much less four times 😅 I know it probably seems "obvious" to you, after spending all that time in the code, that it's symmetrical to the changes done in the other paths, but it's not obvious to me!

CristinaRO (approved):
Thank you for patiently walking me through the changes and for the commit messages! :)

PR #1005, opened 2021-03-19
(1583) Remove obsolete activity creation code

leeky:
General clean up of code that is no longer needed, including:

- `blank`, `level` and `parent` form steps (and remove any activities in these states to prevent errors)
- Remove special cases in policies for handling activities without a `level` or `parent`.
- `create_activity_level_options` and `scoped_parent_activities` form helper methods

## Next steps


db/migrate/20210326085937_remove_unused_activity_form_states.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveUnusedActivityFormStates < ActiveRecord::Migration[6.0]
+  def up
+    Activity.where(form_state: [nil, :blank, :level, :parent]).destroy_all
```
CristinaRO, line 3: Is this safe to do in production?

CristinaRO (commented):
I have a question about deleting activities - is there a risk of deleting an activity that someone is in the process of creating? Or any that should be turned into valid activities rather than deleted?

db/migrate/20210326085937_remove_unused_activity_form_states.rb
leeky: It would be impossible to know what needs to be added to these activities to make them valid, as none have a `parent_id` yet and most don't even know if they're a fund, programme, project or third-party project.

They're very likely to have been created in error - i.e. clicking "Add activity" and then abandoning the form.

Since, https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1004 was merged, there is no way for an end user to create an activity without a `level` or `parent` being set, so this won't affect activities being currently created.

pezholio: I think this is fine, but I'm always a bit iffy about seeing model code in a DB migration. Would this be better as a data migration?

leeky: If your development database contains any of the above form_states, you'll get weird errors when editing them (because the validations have changed to always expect `form_state` to be set).

Also, wouldn't the next line `change_column_null(:activities, :form_state, false)`  break if any of the existing records contained `form_state == nil` ?

I don't see any harm in doing this here. The only activities that will ever be affected on in obsolete states (i.e. likely most of the ones that show up as `Untitled`).

pezholio: OK, cool. That sounds fine 👍

CristinaRO (approved):
Thanks, that makes sense!

PR #1007, opened 2021-03-19
Display RODA Identifiers consistently everywhere

jcoglan:
There are a few places left in the application where we display the
Delivery Partner Identifier in a column labelled "RODA Identifier" or
just "Identifier". This is probably left over from when we first
introduced RODA Identifiers, and it creates confusion as activities are
shown with different IDs across pages.

This changes most of the places in the app where we show a single ID for
an activity, to display the RODA Identifier. We still display Delivery
Partner Identifiers in places where they're explicitly called out as
such, like the activity details page and XML export.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -548,6 +548,8 @@
 
 ## [unreleased]
 
+- Display the RODA Identifer anywhere we have an activity table with an "Identifier" column
+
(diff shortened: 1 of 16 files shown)


pezholio (approved):
This is great. It tripped me up a couple of times when going through the training environment, so good to be consistent.

PR #1009, opened 2021-03-22
Write a script to automate the database restore process

pezholio:
This adds a script that:

- dumps the data from a live database (either staging, prod,  or pentest) to a known location (in `/tmp`) WITHOUT the `users` table
- drops the existing development database
- loads the data into the dev databse
- removes any extraneous Postgres extensions (added by GOV.UK PaaS),
- makes sure the migrations are updated
- seeds the local users

There is also a trap function that runs no matter what happens, to clean up our backup file. I've also updated the documentation, so users know to just run this script.

### DIFF
--- a/bin/db-restore
+++ b/bin/db-restore
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+# bin/restore: Restore the local database from staging/production.
+
+set -e
+
+ENVIRONMENT_NAME="${1:-staging}"
+environments="pentest prod staging"
+
+# Check `ENVIRONMENT_NAME` is one we expect
+if ! echo "$environments"| grep -w "$ENVIRONMENT_NAME" > /dev/null; then
+    echo "The environment name must be one of the following: $environments"
+    exit 1
+fi
+
+# Make sure we're in the root of the project
+cd "$(dirname "$0")/.."
+
+# Check prerequisites are installed
+if ! command -v cf > /dev/null; then
+    echo "The Cloud Foundry (cf) tool is not installed. See https://docs.cloud.service.gov.uk/get_started.html#set-up-the-cloud-foundry-command-line for how to get started"
+    exit 1
+fi
+
+if ! cf plugins | grep -q "conduit"; then
+    echo "The cf conduit plugin is not installed. See https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/database-backup-and-restore.md#installing-the-cf-conduit-plugin for installation instructions."
+    exit 1
+fi
+
+# exit with this by default, if it is not set later
+exit_code=0
+
+# Create a unique filename, so we can delete it later
+timestamp=$(date +"%Y%m%d%I%M%S")
+filename="/tmp/dump-$timestamp.sql"
+
+# Cleanup on exit, no matter what happens
+cleanup () {
+    echo "==> Cleaning up..."
+    rm "$filename" &> /dev/null
+    exit "$exit_code"
+}
+
+# register the cleanup function for all these signal types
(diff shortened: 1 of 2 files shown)


bin/db-restore
```diff
@@ -0,0 +1,88 @@
...
+
+# bin/restore: Restore the local database from staging/production.
+
+set -e
+
+ENVIRONMENT_NAME="${1:-staging}"
+environments=(pentest prod staging)
+
+# Check `ENVIRONMENT_NAME` is one we expect
+if ! [[ ${environments[*]} =~ $ENVIRONMENT_NAME ]]; then
```
jcoglan, line 11: This condition might have unintended behaviour since it checks whether the string `"pentest prod staging"` is a regex match for the string in `ENVIRONMENT_NAME`. So for example running `db-restore ent` or `db-restore '.*'` both trigger this condition.

bin/db-restore
```diff
@@ -0,0 +1,88 @@
...
+
+# Create a unique filename, so we can delete it later
+timestamp=$(date +"%Y%m%d%I%M%S")
+filename="/tmp/dump-$timestamp.sql"
+
+# Cleanup on exit, no matter what happens
+cleanup () {
+  echo "==> Cleaning up..."
+  rm "$filename"
+  exit $exit_code
```
jcoglan, line 42: I am possibly being paranoid here, but I would quote `"$exit_code"` although `shellcheck` seems fine with this.

bin/db-restore
```diff
@@ -0,0 +1,88 @@
...
+
+# register the cleanup function for all these signal types
+trap cleanup EXIT ERR INT TERM
+
+echo "==> Logging into GOV.UK PaaS..."
+
+cf login -s "$ENVIRONMENT_NAME"
+
+# Set the Postgres service name, depending on our environment
+service="beis-roda-$ENVIRONMENT_NAME-postgres"
```
jcoglan, line 53: Small style thing here -- might be worth bracing `${ENVIRONMENT_NAME}` to make it clear that `ENVIRONMENT_NAME-postgres` is not a parameter name.

bin/db-restore
```diff
@@ -0,0 +1,88 @@
...
+        --exclude-table="users"
+
+echo "==> Destroying the existing local database"
+
+# Check the user wants to continue before blowing away the development data
+echo "THIS WILL DELETE ALL DATA IN YOUR DEVELOPMENT ENVIRONMENT."
+echo "Are you sure you want to continue? (n to exit)"
+read -p "" -r delete_data
+
+if [ "$delete_data" != "n" ]; then
```
jcoglan, line 71: I'd prefer we check for a specific input before doing something destructive, rather than accepting any input other than `n`.

bin/db-restore
```diff
@@ -0,0 +1,88 @@
...
+    echo "==> Dropping and recreating the roda-development database..."
+    psql -c "\set AUTOCOMMIT on\n DROP DATABASE \"roda-development\"; CREATE DATABASE \"roda-development\";" -d postgres
```
jcoglan, line 73: You can probably use single quotes around this to avoid escaping the internal double-quotes -- I assume Postgres requires double-quotes here. Might actually be worth using single quotes for shell strings throughout unless you _need_ double quotes for parameter substitution.

pezholio: Yeah, that's habit from using double quotes all the time in Ruby

pezholio:
@jcoglan I've addressed all those comments, so should be good for another look now

jcoglan (approved):
Thanks for adding this @pezholio, will save a lot of effort and prevent us making mistakes while loading data :)

PR #1010, opened 2021-03-22
(1578) Receiving organisation is optional for transactions

pezholio:
This updates the transaction validations to make the Organisation optional. However, if one of the Organisation fields are present, then at least the Name and Type should be present. I've also updated the form to make it clearer that Organisation is now optional. The XML should now not include receiving organisation data, and the receiving organisation shows as `N/A` if an organisation is not provided too.

## Screenshots

image

image

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -553,6 +553,7 @@
 - Budgets do not collect IATI fields or currency as they are set by default
 - Budgets tables do not show IATI fields and only show the financial year
 - Budgets funding type must be the same as the parent activity
+- Make receiving organisation optional
 
 ## [unreleased]
 
--- a/app/models/transaction.rb
+++ b/app/models/transaction.rb
@@ -10,11 +10,9 @@ class Transaction < ApplicationRecord
   belongs_to :parent_activity, class_name: "Activity"
   belongs_to :report, optional: true
 
+  validates_with TransactionOrganisationValidator
   validates_presence_of :report, unless: -> { parent_activity&.organisation&.service_owner? }
-  validates_presence_of :value,
-    :financial_year,
-    :receiving_organisation_name,
-    :receiving_organisation_type
+  validates_presence_of :value, :financial_year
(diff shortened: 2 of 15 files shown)


spec/features/staff/users_can_view_transactions_on_an_activity_page_spec.rb
```diff
@@ -32,7 +32,20 @@
 
         expect(page).to have_content(transaction_presenter.financial_quarter_and_year)
         expect(page).to have_content(transaction_presenter.value)
-        expect(page).to have_content(transaction_presenter.receiving_organisation_name)
+        expect(page).to have_content(transaction.receiving_organisation_name)
```
CristinaRO, line 35: What's the value in changing this expectation to check for the field on the transaction rather than the result of the transaction presenter?

The new presenter method isn't tested in the presenter spec; would there be value in testing it there?

CristinaRO (approved):
One non-blocking question. (I'd like to know the answer, though, testing is one of the areas where I'm constantly trying to understand more and learn!)

CHANGELOG.md
```diff
@@ -553,6 +553,7 @@
 - Budgets do not collect IATI fields or currency as they are set by default
 - Budgets tables do not show IATI fields and only show the financial year
 - Budgets funding type must be the same as the parent activity
+- Make recieving organisation optional
```
CristinaRO, line 556: If you're amending / fixing conflicts, would you mind fixing the typo here too? "recieving" --> "receiving")

spec/features/staff/users_can_view_transactions_on_an_activity_page_spec.rb
pezholio: Ah, well spotted. I've removed this tests and added a test to the presenter spec, which I agree, makes more sense. I wasn't 100% comfortable testing for the presence of `N/A` anywhere in the page anyway, because that could be introduced somewhere else

CHANGELOG.md
pezholio: Done!

PR #1016, opened 2021-03-24
(1523) Add the budget types to the application

mec:
## Changes in this PR
I won't lie, I am not super sure how this work will go down with you dear reviewer! :) I am super happy to do a pair/sync review if it will help, please let me know!?

The main goal is to use the `budget_type` database column to actually store what we now consider the 'budget type' and introduce more 'budget types' so that the application can support more of the 'funding arragngements' which is the main user need (I can report actvities with the 'external income' funding arrangement ).

Along the way we've dropped some other columns, deleting data and come up with a way to load constants from code lists that I am looking forward to feedback on (I suspect there is a better/nicer way to do this!?)

I've leant heavily on the fact that the new BEIS defined budget types have the same first two values as the funding_types codes that we already have.

Budgets are far from complete at this point so more work will follow so this is very much another small step or many small steps to come…

## Screenshots of UI changes

Screenshot_2021-03-24 Create budget — Report your Official Development Assistance

Screenshot_2021-03-24 Activity Global Challenges Research Fund (GCRF) — Financials — Report your Official Development Assis

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -579,6 +579,7 @@
 - Add supporting hint text about the soft limits to title and description
 - Change channel of delivery code to radio buttons
 - Collect transferred and external budget types
+- Show the type of budget in the application
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-42...HEAD
 [release-42]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-41...release-42
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -578,6 +578,7 @@
 - Data dictionary link opens in a new window/tab by default
 - Add supporting hint text about the soft limits to title and description
 - Change channel of delivery code to radio buttons
+- Collect transferred and external budget types
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-42...HEAD
 [release-42]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-41...release-42
--- a/app/models/budget.rb
+++ b/app/models/budget.rb
@@ -1,20 +1,21 @@
 class Budget < ApplicationRecord
   include PublicActivity::Common
 
+  IATI_TYPES = Codelist.new(type: "budget_type", source: "iati").hash_of_coded_names
+  IATI_STATUSES = Codelist.new(type: "budget_status", source: "iati").hash_of_coded_names
+  BUDGET_TYPES = Codelist.new(type: "budget_type", source: "beis").hash_of_coded_names
+  DIRECT_BUDGET_TYPES = [BUDGET_TYPES["direct_newton_fund"], BUDGET_TYPES["direct_global_challenges_research_fund"]]
+
   belongs_to :parent_activity, class_name: "Activity"
   belongs_to :report, optional: true
 
   validates_presence_of :report, unless: -> { parent_activity&.organisation&.service_owner? }
   validates_presence_of :value,
     :currency,
-    :funding_type,
     :financial_year
   validates :value, numericality: {other_than: 0, less_than_or_equal_to: 99_999_999_999.00}
-  validates :funding_type, inclusion: {in: ->(_) { valid_funding_type_codes }}
-  validate :funding_type_must_match_source_fund, unless: -> { parent_activity&.source_fund_code.blank? }
-
-  IATI_TYPES = Codelist.new(type: "budget_type", source: "iati").hash_of_coded_names
-  IATI_STATUSES = Codelist.new(type: "budget_status", source: "iati").hash_of_coded_names
+  validates :funding_type, inclusion: {in: BUDGET_TYPES.values}
+  validate :direct_budget_type_must_match_source_fund, if: -> { DIRECT_BUDGET_TYPES.include?(funding_type) }
 
   def financial_year
     return nil if self[:financial_year].nil?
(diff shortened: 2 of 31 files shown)


app/models/codelist.rb
```diff
@@ -38,6 +38,22 @@ def initialize_codelists
     end
   end
 
+  def hash_of_coded_names
+    coded_values = {}
+    list.each do |code|
+      coded_values[ActiveSupport::Inflector.parameterize(code["name"], separator: "_").to_sym] = code["code"]
```
jcoglan, line 44: I would avoid using `to_sym` here and leave these values as strings -- they're data, not names of things internal to the program like method names or column names. This also makes this consistent with the type of `hash_of_named_codes` which has string keys.

app/models/budget.rb
```diff
@@ -32,6 +30,14 @@ def period_end_date
     financial_year&.end_date
   end
 
+  def iati_type
+    IATI_TYPES[:original]
```
jcoglan, line 36: If you use `IATI_TYPES.fetch(:original)` here, we'll get an error if that key doesn't exist. If the application expects certain pieces of data to exist it should raise an error as soon as it finds out the data is missing, rather than returning `nil` and causing a bug somewhere else.

jcoglan:
I'm not sure if the amount of work required to fix the following is worth it but half the commits on this branch don't build, because of the order in which the changes are done. The commit `Migration to make funding_type into  budget_type` changes the schema before any of the requires changes to the code have been made that make it compatible with that schema change.

A possible way to re-order things would be to break the migration into smaller pieces, for example:

- One migration just removes `budgets.status`, and in the same commit we remove all uses of that column from the code, so the build passes
- We then make functional changes to the codebase so that we stop referencing `budget_type`, and `funding_type` behaves in the way we ultimately want
- Finally, a migration to effectively rename `funding_type` to `budget_type`, either by actually renaming the column (if you want to retain data) or by dropping and replacing it

The alternative is to squash some commits together so that the existing migration doesn't break the build.

CHANGELOG.md
```diff
@@ -556,6 +556,7 @@
 - Users can report the strategic area under which the GCRF allocation was made
 
 ## [unreleased]
+- collect additional budget types in the applicaiton
```
CristinaRO, line 559: When fixing the conflicts in the changelog, would you mind capitalising the sentence and fixing the typo? ("applicaiton" -> "application")

mec:
@jcoglan I've reworked this based on your comments, there are a few differneces to keep it green.

I am a bit nervous about the loading constants from codelists, I feel like this is a valuable thing to do as a concept but I am not sure about this implementation, it feels like there is a better approach to this that I am not seeing - I would appreciate your view (or anyone else reading along!)

What I am trying to achieve is:

- define the values that can be stored in an object from the code list
- define the human readable parts of the values (names,descriptions, etc) from the code list
- remove the duplication of the human readable values from the locales
- allow BEIS to publish the codelist and for users to know what values the applicaiton accepts i.e. the mapping between code and 'real word' meaning is not in code itself and can be easily understood by BEIS and other users, for example when preparing their spreadsheets for uploading.

These sort of feel like class things, but then there are sort of about a specific instance as well. I've read about singletons and that sort of feels right but then there is lots of 'singletons are evil' stuff as well.

I would like to try and set the pattern we use here so we can follow for other attrbutes in the application, we'll be looking at the policy markers very soon with a similar story.

jcoglan:
@mec this looks really great, it's a nice step-by-step transition to the new behaviour now :)

The main concern I have about constants/singletons is that, when you write something like:

```rb
  IATI_TYPES = Codelist.new(type: "budget_type", source: "iati").hash_of_coded_names
```

The constant `IATI_TYPES` is assigned once, when the process starts up. That's great for values that are not going to change for the lifetime of the process, but for things that can change while the process is running, that's a problem.

`Codelist.new` already memoises loading the data from disk in production, so we're not gaining anything performance-related by storing its result in a constant. And by storing it in this way, it _might_ be the case that the constant does not get updating if you change a codelist file during development, which is undesirable.

I'd have to check this, but the way Rails reloads code in development means this might not be a problem -- try changing one of the codelist files and see if your changes show up in on refresh. If they don't, one solution is to replace the constant with a class method that does the same thing:

```rb
  def self.iati_types
    Codelist.new(type: "budget_type", source: "iati").hash_of_coded_names
  end
```

This is still cheap in production, because `Codelist.new` caches codelist data in production. And because `iati_types` is not a method call rather than a constant reference, its body gets executed _every_ time you need it, so it will update properly in development too.

But like I way, Rails code reloading might mean this isn't actually a problem, and using constants for this is fine.

jcoglan:
By the way, one thing I try to stick to which you are doing here but I thought I'd draw attention to, is only referring to constants from inside the classes they're defined in. Code outside the class should only interact with it by calling methods. That is, if something is defined as:

```rb
class Budget
  IATI_TYPES = Codelist.new(...)
end
```

Then if you have code all over the codebase using the expression `Budget::IATI_TYPES` you have to go and change all those if you decide to replace the constant with something dynamic. If everything is instead using a method call, `Budget.iati_types`, you can change how that method works without needing to change any of its callers.

Think of constants as "implementation details" and only refer to them directly from within the same class.

mec:
Thanks @jcoglan that all makes lots of sense! 

I think, in production at least, what you describe is the behaviour we want, the code lists are code that is loaded at boot and should never change, they are the values that are accepted by the application for as long as it runs in that 'version'. If we had a way to 'upload' codelist, things would be different, but I feel that is so far distance as to not be an issue.

I remember seeing something about how things work in devlopment when I was looking at the `Codelist` model: 

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/e55a0200857ff82b7a097ebbad8b4778f40542fa/app/models/codelist.rb#L19 it looks like we only mememoise the `codelist` in production, in development it is loaded every call - I feel like that at the model level, that is what we want.

mec:
So the only place I can see I am using `Budget::IATI_TYPES` is in the specs to check the values in validation - this feels okay to me.

mec:
> Just to double-check: if you change the codelists in development, do those changes take effect without restarting the app? The only thing I can see a potential problem with here is that storing these values as constants might mean they don't get reloaded correctly, so I wanted to check that.

Yeah, they do 👍

jcoglan (approved):

PR #1018, opened 2021-03-24
(1581) Backfill forecasts data into historic reports for new partners

jcoglan:
## Changes in this PR

This revives a script we used in the previous quarter to back-fill some forecast data into the previous quarter's approved report, before opening the next reporting round.

Since this was first introduced and then removed, the forecast import service has been exposed to end users and so has gained some authorization logic and other checks that meant this functionality had to be reworked beyond just reverting the commits that removed it. Details in the commit messages.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -558,6 +558,7 @@
 ## [unreleased]
 
 - Users can report the strategic area under which the GCRF allocation was made
+- Bring back script for importing forecast data into an in-review/approved report
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-39...HEAD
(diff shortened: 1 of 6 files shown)


jcoglan:
[Allow Q4 forecasts to be imported into Q3 reports](https://trello.com/c/H6MF69Fg/1581-allow-q4-forecasts-to-be-imported-into-q3-reports)

leeky (approved):
LGTM, just I wondered if there were any report states which we should always bail out on?

jcoglan:
That's a good question -- this is probably only supposed to hit approved or in-review reports, but I'm not 100% sure this should be a hard restriction. This is why I printed the report's state before confirmation, in case it's something the person doing the import cares about.

PR #1020, opened 2021-03-24
(1603) Remove `legacy_iati_xml` column

CristinaRO:
## Changes in this PR

This column was added in 71ba7adcf04cc700fba6014e6fbbe0652838dad3 in order to preserve the IATI data used to generate ingested activities.

This historical data hasn’t been needed since, and is adding visual noise when inspecting activities via a console.

We think it’s safe to drop this column. We will check with the responsible person from BEIS before doing so in production.

## Next steps


CristinaRO:
[Remove `legacy_iati_xml` column from Activity](https://trello.com/c/eV6E5PZ0/1603-remove-legacyiatixml-column-from-activity)

leeky (approved):
LGTM! :shipit:

PR #1021, opened 2021-03-25
(1608) "bin/db-restore" script fixes

leeky:
The command wasn't working correctly because it was attempting to seed the `users` table that didn't exist as it was explicitly excluded from the database dump.

Changing `--exclude-table` to `--exclude-table-data` means that the schema for the users table is dumped, but its contents isn't.

I've also excluded the internal `ar_internal_metadata` and`spatial_ref_sys` tables from the backup as well.

Finally calling `rails db:environment:set` let's Rails know we're running a development environment (despite the copy of the database looking like a production one).

NB: I'm entirely sure why this is needed when we're excluding `ar_internal_metadata` from the dump, but I got warnings without it.

Using `pgsql -f dump.sql` instead of `pgsql ... < dump.sql` seemed to resolve a lot of the random `Invalid command \n` errors.

## Next steps

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -28,3 +28,4 @@ spec/examples.txt
 # SQL files
 **/*.sql
 
+bin/db-restore
--- a/bin/db-restore
+++ b/bin/db-restore
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# bin/restore: Restore the local database from staging/production.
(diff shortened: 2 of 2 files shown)


pezholio (approved):
Changes all look good, and can confirm it all works for me 👍

PR #1022, opened 2021-03-25
(1609) Refactor handling of implementing organisations

pezholio:
In certain circumstances, when trying to update an activity, we hit an error in the ImplemeningOrganisationBuilder, where the builder tried to add an error to the Errors array, but because there were no implementing organisations available, we got an `undefined method `errors' for nil:NilClass` error.

This reworks how the `ImplemeningOrganisationBuilder` works to memoize the organisation that gets built, rather than fetching the organisation's last implementing organisation, which may or may not be the correct one, or even exist at all.

I've also updated the updater, so it only attempts to build an implementing organisation if the relevant fields are present, in a similar way to how the creator works.

### DIFF
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -96,18 +96,22 @@ module Activities
 
         @activity.assign_attributes(@converter.to_h)
 
-        implementing_organisation_builder = ImplementingOrganisationBuilder.new(@activity, row)
-        implementing_organisation = implementing_organisation_builder.build
+        if row["Implementing organisation name"].present? || row["Implementing organisation sector"].present?
+          implementing_organisation_builder = ImplementingOrganisationBuilder.new(@activity, row)
+          implementing_organisation = implementing_organisation_builder.organisation
 
-        @activity.implementing_organisations = [implementing_organisation] if implementing_organisation.valid?
+          if implementing_organisation.valid?
+            @activity.implementing_organisations = [implementing_organisation]
+          else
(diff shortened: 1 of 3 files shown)


pezholio:
[Refactor handling of implementing organisations](https://trello.com/c/WXwcBgoI/1609-refactor-handling-of-implementing-organisations)

app/services/activities/import_from_csv.rb
```diff
@@ -181,20 +185,22 @@ def initialize(activity, row)
...
+            implementing_organisation.organisation_type = row["Implementing organisation sector"]
+          }
+        end
       end
 
       def add_errors(errors)
         errors.delete(:implementing_organisations)
 
-        implementing_organisation_errors = @activity.implementing_organisations.first.errors.messages
+        implementing_organisation_errors = organisation.errors.messages
```
leeky, line 203: Won't `organisation.errors` always be empty (`organisation.valid?` isn't called)?

I think there's a test case missing for when the activity's implementing_organisation is invalid 🤔

leeky (changes requested):

app/services/activities/import_from_csv.rb
pezholio: Ah, good call. I'll tweak this

pezholio: @leeky Think this is ready for another look

leeky (approved):
LGTM! :shipit:

PR #1028, opened 2021-03-26
(1567) Activity importer can update an activity without requiring the sector code.

pezholio:
When we're updating an activity, we don't always want to have to set every single variable. As it stands, when we update a record, there are some variables we're expected to set every single time (e.g. sector code).

This updates the Converter to behave differently when updating to when creating a new activity. If we're updating, we grab the row, and only run the converter on fields that are expressly filled out. We also don't do any of the inferrence when updating, as this can overwrite data that is already set on the Activity.

This also has the side effect of any policy marker fields not being set to `not_assessed` by default. This is fine for now, and we're going to write a seperate card to write a migration that sets all Activity policy marker fields to default to `not_assessed` at the database level.

### DIFF
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -87,7 +87,7 @@ module Activities
         @activity = find_activity_by_roda_id(row["RODA ID"])
         @organisation = organisation
         @row = row
-        @converter = Converter.new(row)
+        @converter = Converter.new(row, :update)
         @errors.update(@converter.errors)
       end
 
@@ -269,9 +269,10 @@ module Activities
         "NF Partner Country DP",
       ]
 
-      def initialize(row)
+      def initialize(row, method = :create)
         @row = row
         @errors = {}
+        @method = method
         @attributes = convert_to_attributes
       end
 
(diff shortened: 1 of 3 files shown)


pezholio:
[Activity importer can update an activity without requiring the sector code.](https://trello.com/c/Z5i2pmBf/1567-activity-importer-can-update-an-activity-without-requiring-the-sector-code)

app/services/activities/import_from_csv.rb
```diff
@@ -284,19 +285,30 @@ def to_h
...
+        attributes = fields.each_with_object({}) { |(attr_name, column_name), attrs|
           attrs[attr_name] = convert_to_attribute(attr_name, @row[column_name]) if field_should_be_converted?(column_name)
         }
 
-        attributes[:geography] = infer_geography(attributes)
-        attributes[:requires_additional_benefitting_countries] = (@row["Recipient Country"] && @row["Intended Beneficiaries"]).present?
-        attributes[:recipient_region] ||= inferred_region
-        attributes[:call_present] = (@row["Call open date"] && @row["Call close date"]).present?
-        attributes[:sector_category] = get_sector_category(attributes[:sector])
+        if @method == :create
```
CristinaRO, line 292: What happens if there is a change in the attributes that determine geography during update? Isn't the existing data going to be mismatched with the new data from the upload?

app/services/activities/import_from_csv.rb
```diff
@@ -284,19 +285,30 @@ def to_h
...
+      def fields
+        return FIELDS if @method == :create
+
+        columns_to_update = @row.to_h.reject { |_k, v| v.blank? }.keys
+        converter_keys = FIELDS.select { |_k, v| columns_to_update.include?(v) }.keys
+        FIELDS.slice(*converter_keys)
+      end
+
       def field_should_be_converted?(column_name)
         ALLOWED_BLANK_FIELDS.include?(column_name) || @row[column_name].present?
```
CristinaRO, line 313: I think the original purpose of this method was to allow the policy markers to be set to `not_assessed`. If we're not doing that anymore, is there any point to allowing some fields to be converted if they're blank?

CristinaRO (commented):
A couple of questions, let's reconvene on Monday and perhaps discuss synchronously.

app/services/activities/import_from_csv.rb
pezholio: Good shout 👍

app/services/activities/import_from_csv.rb
pezholio: Also, a good shout. This will also make the logic a lot clearer. Thanks for noticing 👍

pezholio: @CristinaRO Thinking about this more, I think this is definitely a thing we should do, but won't be able to do until we set a default for the policy markers, which will come in a seperate feature.

CristinaRO (approved):
Thanks for making the changes! Needs conflict solving.

pezholio:
I'm on it! 👍

pezholio:
Thanks for the review!

PR #1032, opened 2021-03-29
Fix puzzling test result ordering bug

leeky:
The setup for the test in question generates a CSV containing two result rows. The order of these was consistently reversed on two developers' machines (including mine), whereas it was fine for another two.

Rather than spend to long puzzling over this strange issue, I've updated the test to check that the expected hash is *included* in `rows` rather than expecting it to be the first result.

We suspect something to do with Postgres versions (the two it was affecting are running older versions), but can't be certain.

## Next steps


pezholio (approved):
Seems legit 👍

jcoglan:
This has made me realise the test as I wrote it doesn't actually do what I intended, which is to assert that all forecast cells are empty -- it should check all the rows present, not just the first one. I'll amend it.

PR #1033, opened 2021-03-29
(1630) Set all activity policy marker fields to `not_assessed` by default

pezholio:
After some discussion about how we handle certain blank fields when updating / creating Activities via CSV upload for https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1028, we decided that all Activities should default to `not_assessed` by default, as that is how all activities start out. This is also the default selected option when creating / editing an activity via the web form.

I've added a migration to handle this, as well as updating the CSV import test.


pezholio:
[Set all Activity Policy marker fields to `not_assessed` by default](https://trello.com/c/QIc0Bmqw/1630-set-all-activity-policy-marker-fields-to-notassessed-by-default)

robbiepaul (approved):
LGTM 👍

PR #1035, opened 2021-03-29
(1620) Data dictionary link opens in a new tab

CristinaRO:
## Changes in this PR
- Data dictionary link opens in a new window/tab by default
- Existing links that open in new windows now use the same helper instead of being manually coded
 
## Next steps

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -19,4 +19,8 @@ module ApplicationHelper
       link_to("#{text} #{raw(span)}".html_safe, href, class: "govuk-link")
     end
   end
+
+  def link_to_new_window(text, href)
+    link_to("#{text} (opens in new window)", href, class: "govuk-link", target: "_blank")
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -20,7 +20,7 @@ module ApplicationHelper
     end
   end
 
-  def link_to_new_window(text, href)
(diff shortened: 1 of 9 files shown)


CristinaRO:
[Clicking a data dictionary link should open a new tab](https://trello.com/c/9on7ClxP/1620-clicking-a-data-dictionary-link-should-open-a-new-tab)

app/helpers/application_helper.rb
```diff
@@ -19,4 +19,8 @@ def a11y_action_link(text, href, context = "")
       link_to("#{text} #{raw(span)}".html_safe, href, class: "govuk-link")
     end
   end
+
+  def link_to_new_window(text, href)
+    link_to("#{text} (opens in new window)", href, class: "govuk-link", target: "_blank")
```
pezholio, line 24: I think we'd probably be better saying "opens in new tab" here, as the default behaviour for `_blank` is to open a new tab nowadays. Also [GOV.UK Design system](https://design-system.service.gov.uk/styles/typography/#opening-links-in-a-new-tab) says to: 

> "Include `rel="noreferrer noopener"` along with `target="_blank"` to reduce the risk of [reverse tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing)."

pezholio (approved):
LGTM 👍

app/helpers/application_helper.rb
CristinaRO: Thank you for catching this @pezholio !

PR #1040, opened 2021-03-30
(1617) Update policy markers and use radio buttons with descriptions

pezholio:
This updates the policy markers to use codelists, rather than enums, and changes the select boxes to radio buttons. It makes for a longer page, but should aid usability.

## Before

image

## After

image


pezholio:
[Update Policy Markers and use radio buttons with descriptions](https://trello.com/c/dWGFzb5i/1617-update-policy-markers-and-use-radio-buttons-with-descriptions)

app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -37,22 +37,22 @@
...
-    %policy-marker{"vocabulary" => "1", "code" => "7", "significance" => Activity.policy_marker_climate_change_adaptations[activity.policy_marker_climate_change_adaptation]}/
-  - unless activity.desertification_not_assessed?
-    %policy-marker{"vocabulary" => "1", "code" => "8", "significance" => Activity.policy_marker_desertifications[activity.policy_marker_desertification]}/
-  - unless activity.disaster_risk_reduction_not_assessed?
-    %policy-marker{"vocabulary" => "1", "code" => "10", "significance" => Activity.policy_marker_disaster_risk_reductions[activity.policy_marker_disaster_risk_reduction]}/
-  - unless activity.disability_not_assessed?
-    %policy-marker{"vocabulary" => "1", "code" => "11", "significance" => Activity.policy_marker_disabilities[activity.policy_marker_disability]}/
-  - unless activity.nutrition_not_assessed?
-    %policy-marker{"vocabulary" => "1", "code" => "12", "significance" => Activity.policy_marker_nutritions[activity.policy_marker_nutrition]}/
+  - unless activity.policy_marker_gender == 1000
```
jcoglan, line 40: Is there a way we could use a meaningful name instead of the numeral `1000` in these checks? I'm assuming it refers to the "not assessed" enum value.

spec/features/staff/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -133,14 +133,14 @@
...
             delivery_partner_identifier: "IND-ENT-IFIER",
-            policy_marker_gender: "not_targeted",
-            policy_marker_biodiversity: "significant_objective",
-            policy_marker_disability: "principal_objective",
-            policy_marker_desertification: "principal_objective_and_in_support",
-            policy_marker_nutrition: "not_assessed",
-            policy_marker_climate_change_adaptation: "not_assessed",
-            policy_marker_climate_change_mitigation: "not_assessed",
-            policy_marker_disaster_risk_reduction: "not_assessed")
+            policy_marker_gender: 0,
```
jcoglan, line 136: It would likewise be useful to use meaningful names rather than magic numbers in these definitions and in tests.

pezholio:
@jcoglan I thought about what you said, and it does actually seem like keeping enums as a thing is a useful thing to have, so rather than do away with enums completely, I'm now using the `hash_of_coded_names` approach that @mec introduced recently to load the enums when the application boots. This means we keep all the knowledge of what the codes are in the codelist, while not sacrificing the utility of enums (which, as you pointed out, are useful to avoid magic numbers all over the place)

jcoglan:
@pezholio That's a nice approach, and much clearer, thanks for making the changes :)

jcoglan (approved):

mec:
@pezholio oooo, loading enums witht the code list data sounds like exactly what I was hoping to get to - we get all the magic™ of enums but loaded from the code lists! I need to come and check up on how this went?

PR #1041, opened 2021-03-30
(1639) Introduce "create_child?" method to ActivityPolicy

leeky:
Currently the only way to determine whether an activity can be created at a particular level is to check using `ActivityPolicy.create?`. This doesn't really work when you're trying to run this check against an activity that doesn't yet exist. This means we're forced to call the `create?` check without passing in a given activity, which means we're unable to check for an editable report, for example.

Instead of doing this, we introduce a new method `ActivityPolicy.create_child?`. Asking the parent activity whether a child can be created under it means we have enough information to give an accurate response.

## Next steps

### DIFF
--- a/app/policies/activity_policy.rb
+++ b/app/policies/activity_policy.rb
@@ -10,7 +10,7 @@ class ActivityPolicy < ApplicationPolicy
     if beis_user?
       return true if record.fund? || record.programme?
     end
-    return false if editable_report_for_organisation_and_fund.nil?
+    return false unless editable_report?
     record.organisation == user.organisation
   end
 
@@ -24,7 +24,7 @@ class ActivityPolicy < ApplicationPolicy
     if delivery_partner_user?
       return false if record.organisation != user.organisation
       return false if record.fund? || record.programme?
-      return false if editable_report_for_organisation_and_fund.nil?
+      return false unless editable_report?
       return true
     end
     false
@@ -47,8 +47,9 @@ class ActivityPolicy < ApplicationPolicy
     end
   end
 
-  private def editable_report_for_organisation_and_fund
-    fund = record.associated_fund
-    Report.editable.find_by(organisation: record.organisation, fund: fund)
+  private
+
+  def editable_report?
+    Report.editable.for_activity(record).exists?
   end
 end
--- a/app/policies/activity_policy.rb
+++ b/app/policies/activity_policy.rb
@@ -10,7 +10,7 @@ class ActivityPolicy < ApplicationPolicy
     if beis_user?
       return true if record.fund? || record.programme?
     end
-    return false if editable_report_for_organisation.nil?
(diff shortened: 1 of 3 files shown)


app/policies/activity_policy.rb
```diff
@@ -47,12 +47,8 @@ def resolve
...
   end
 
-  private def editable_report_for_organisation
-    Report.editable.find_by(organisation: record.organisation)
-  end
-
   private def editable_report_for_organisation_and_fund
-    fund = record.parent.associated_fund
+    fund = record.associated_fund
     Report.editable.find_by(organisation: record.organisation, fund: fund)
```
jcoglan: There is a scope on `Report` that implements this so you can write `Report.editable.for_activity(record).first` here.

app/policies/activity_policy.rb
```diff
@@ -47,8 +47,10 @@ def resolve
     end
   end
 
-  private def editable_report_for_organisation_and_fund
+  private
+
+  def editable_report?
     fund = record.associated_fund
-    Report.editable.find_by(organisation: record.organisation, fund: fund)
+    Report.editable.exists?(organisation: record.organisation, fund: fund)
```
jcoglan: Or, to replace my previous comment, `Report.editable.for_activity(record).exists?`

app/policies/activity_policy.rb
```diff
@@ -14,6 +14,14 @@ def create?
     record.organisation == user.organisation
   end
 
+  def create_child?
+    return record.fund? if beis_user?
+    return false if record.third_party_project?
+    return false unless editable_report?
```
jcoglan, line 20: Why does there need to be an editable report in existence in order for a new activity to be created?

leeky: Well, it's matching the [existing behaviour](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1041/files/a5e852050964c43ef778fa9f2a32bd81d69f7faa..56710ca14a2bc3c506ee8abf40b9d00f44decca5#diff-d9caded5b030bd11eaf9e5847ec15f8fbc14f2f357a0aa27da212846c6a0a736L13) in the `create?` method. I will check however!

jcoglan: Did we find an answer to this question?

leeky: We're going to take this out for now, although there is a user need to prevent anything being created or edited whilst reports are being approved. This will be added after some more thought has been put into it.

app/policies/activity_policy.rb
leeky: I couldn't do this in the end, as I needed to do: 

`Report.editable.where(..., organisation: [record.extending_organisation, record.organisation]).exists?`

leeky:
@jcoglan Should be good to go now 🤞

jcoglan (approved):

PR #1044, opened 2021-03-31
(Chore) Document the process for importing new data

mec:
When we on-board new delivery partners we need to bring in all their
existing activities and financial data at the time they are on-boarded.

This is a one time only operation run by developers with data supplied
by BEIS.

We want to document the process in the code as the most useful place to
do so and expect it to be removed once the service completes on-boarding
all delivery partners in Q1 2021-2022.

### DIFF
--- a/doc/importing-new-delivery-partner-data.md
+++ b/doc/importing-new-delivery-partner-data.md
@@ -1,0 +1,323 @@
+# Importing delivery partner data to the application
+
+## General
+We have enough confidence in our code to run this as a pair on production.
+
+If you want to gain confidence about any aspect and would get that by running
+the import locally first, feel free to do so. Running the import locally lets
+you get the data into perfect shape before you have to copy it over to
+production.
+
+All imports are run in a transaction, if any row fails the whole import is
+skipped.
+
+Once all the data is imported, zip up the actual files used and attach them to
+the Trello card (see below), this keeps a record of what might have been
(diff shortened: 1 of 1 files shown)


robbiepaul (approved):
LGTM

doc/importing-new-delivery-partner-data.md
```diff
@@ -0,0 +1,309 @@
...
+```
+Organisation.all.pluck(:name, :id)
+```
+
+- quit the console
+
+- run the import 
+
+```
+cd /app && /usr/local/bin/ruby bin/rails activities:import CSV=FILENAME ORGANISATION_ID=ORGANISATION_ID
```
CristinaRO, line 101: We have now added the requirement to provide the email of the uploader to the task: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1058/commits/6af9d88a21bd4234ca894b5b3fe7a06c834f990c.

The uploader is then passed to the `ImportFromCsv` service and used to determine the permissions of the uploader via the organisation that they belong to, so you might want to make it clearer in the instructions that the uploader will have to change their account to belong to the correct organisation for the task they're running.

CristinaRO (changes requested):
We've made some changes to the activity import process that need to be reflected in the documentation.

CristinaRO (approved):

PR #1046, opened 2021-04-01
Fix brittle transfer tests

pezholio:
Now we've moved into a new financial year, the tests for the transfers were breaking, as the financial years and quarters were hardcoded to the previous financial year. To fix this, I've set the parameters to always default to the current financial year and quarter.

### DIFF
--- a/spec/features/staff/beis_users_can_create_a_transfer_spec.rb
+++ b/spec/features/staff/beis_users_can_create_a_transfer_spec.rb
@@ -11,14 +11,19 @@ RSpec.feature "BEIS users can create a transfer" do
   end
 
   scenario "successfully creates a transfer" do
-    transfer = fill_in_transfer_form
+    quarter = FinancialQuarter.for_date(Date.today)
+
+    transfer = fill_in_transfer_form(
+      financial_quarter: quarter.to_i,
+      financial_year: quarter.financial_year.to_i
+    )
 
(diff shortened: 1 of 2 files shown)


CristinaRO (approved):
Good catch, prompt action! :D

spec/features/staff/beis_users_can_create_a_transfer_spec.rb
```diff
@@ -11,14 +11,19 @@
   end
 
   scenario "successfully creates a transfer" do
-    transfer = fill_in_transfer_form
+    quarter = FinancialQuarter.for_date(Date.today)
+
+    transfer = fill_in_transfer_form(
```
leeky, line 16: Why are you overriding the quarter and year here, when your new defaults in `fill_in_transfer_form` will do the same thing..?

leeky (approved):
Added a comment about duplication, but otherwise LGTM :shipit:

spec/features/staff/beis_users_can_create_a_transfer_spec.rb
pezholio: Just to be a bit clearer when we're testing the outcome. Before, we were assuming that the Financial Year was Q1 2020, without explicity setting it. This way, we have access to an instance of the class too

PR #1048, opened 2021-04-01
Add script for converting import data files

jcoglan:
To perform the initial import of new delivery partners' records into
RODA, we need to take the spreadsheets they've provided and convert them
into the schema supported by the transaction and forecast importers.

To run the script:

    $ ruby script/convert_import.rb \
        --level C \
        --input path/to/input_file.csv \
        --output path/to/output

This reads the data from `path/to/input_file.csv` and writes two new
files: `path/to/output_transactions.csv` and
`path/to/output_forecasts.csv`, which will be compatible with our
existing importers. It does not write anything to the database or
interact with the application in any other way, it purely converts one
CSV file into two others.

The `--level` option must be either `C` or `D` to indicate the level of
the activities being imported.

The main job of this converter is to recognise the various ways that
identifiers, transactions and forecasts are represented in the files
we've been provided, and canonicalise them.

Identifiers are presented as two columns: `Parent RODA ID` and `RODA ID
Fragment`, as is usual for creating new activities. To build the RODA ID
of the new activity we combine these fields with either `-` at level C
or the empty string at level D.

Transactions are presented column-wise, whereas our importer expects
them row-wise, with these standard columns:

    Activity RODA Identifier
    Financial Year
    Financial Quarter
    Value

So we convert a single input row into multiple possible output rows by
matching column headings that represent transactions. We recognise the
following formats that we have observed in files provided to us:

    Act <yyyy>/<yy> FY Q<q> (<month>, <month>, <month>)
    Act <yyyy>/<yy> Q<q>
    Q<q> <yyyy>-<yyyy> actuals

We extract the year and quarter from these and emit one row of output
for each column we find in the input.

Forecasts are presented column-wise and are also consumed column-wise by
our importer, but there are many different heading formats in the input
we've been given. They include:

    FC Q<q> <yyyy>-<yy>
    FC Q<q> <yyyy>
    FC <yyyy>/<yy> FY Q<q> (<month>, <month>, <month>)
    Q<q> <yyyy>-<yyyy> forecast

We recognise all these and emit a CSV with headings in the form
recognised by our importer:

    FC <yyyy>/<yy> FY Q<q>

Since this set of output columns is dynamic and we're emitting a single
CSV, we need to make sure all rows are emitted with the same set of
quarters in the same order. To achieve this, we use the first row of the
input to extract a set of `forecast_mappings` -- a list of the columns
we plan to emit along with their source column names in the input. We
pass these mappings in to convert each data row.

## Next steps

### DIFF
--- a/app/services/import_converter.rb
+++ b/app/services/import_converter.rb
@@ -1,0 +1,94 @@
+class ImportConverter
+  TRANSACTION_HEADER_PATTERNS = [
+    /^Act +(?<year>\d{4})\/\d{2} +FY +Q(?<quarter>[1-4]) +\(.*\)$/,
+    /^Act +(?<year>\d{4})\/\d{2} +Q(?<quarter>[1-4])$/,
+    /^Q(?<quarter>[1-4]) +(?<year>\d{4})-\d{4} actuals$/,
+  ]
+
+  FORECAST_HEADER_PATTERNS = [
+    /^FC +Q(?<quarter>[1-4]) +(?<year>\d{4})-\d{2}$/,
+    /^FC +Q(?<quarter>[1-4]) +(?<year>\d{4})$/,
+    /^FC +(?<year>\d{4})\/\d{2} +FY +Q(?<quarter>[1-4]) +\(.*\)$/,
+    /^Q(?<quarter>[1-4]) +(?<year>\d{4})-\d{4} forecast$/,
+  ]
+
+  IDENTIFIER_HEADERS = ["Activity RODA Identifier"]
+  TRANSACTION_HEADERS = ["Financial Year", "Financial Quarter", "Value"].freeze
+
+  LEVEL_C = "C"
+  LEVEL_D = "D"
+
+  def initialize(row, level:, forecast_mappings: nil)
+    @row = row
+    @level = level
+    @forecast_mappings = forecast_mappings
+  end
+
+  def transaction_headers
+    IDENTIFIER_HEADERS + TRANSACTION_HEADERS
+  end
+
+  def transaction_tuples
+    tuples = []
+
+    @row.each do |header, value|
+      year, quarter = match_quarter(header, TRANSACTION_HEADER_PATTERNS)
+      if year && quarter && value.strip != "0"
+        tuples << identifier_tuple + [year, quarter, value]
+      end
+    end
+
+    tuples
+  end
+
+  def forecast_headers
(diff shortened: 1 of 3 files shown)


app/services/import_converter.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  def forecast_tuples
+    tuple = forecast_mappings.map { |_, row_header| @row.fetch(row_header) }
+    [identifier_tuple + tuple]
+  end
+
+  def forecast_mappings
+    return @forecast_mappings if @forecast_mappings
+    @forecast_mappings = []
```
pezholio, line 55: Couldn't this be wrapped in a `@forecast_mappings ||= begin ... end` block?

script/convert_import.rb
```diff
@@ -0,0 +1,40 @@
+require "csv"
```
pezholio, line 1: No a huge deal, but I'd prefer to see ocassionally used scripts in `bin/`, rather than `script/`, as the latter seems to be more for commonly used scripts to rule them all type things

pezholio (approved):
Looks largely fine, a couple of things, but feel free to ignore (especially now I've just noticed that there is already a Ruby script in `scripts/` anyhoo)

app/services/import_converter.rb
jcoglan: Sure, I'll change it.

jcoglan: I'm not 100% sure I prefer this b/c it still needs an internal mutable variable which might make this more confusing. What do you think?

```rb

  def forecast_mappings
    @forecast_mappings ||= begin
      mappings = []

      @row.each do |row_header, _|
        year, quarter = match_quarter(row_header, FORECAST_HEADER_PATTERNS)
        next unless year && quarter

        output_header = forecast_header(year, quarter)
        mappings << [output_header, row_header]
      end

      mappings.sort_by(&:first)
    end
  end
```

script/convert_import.rb
jcoglan: I thought `bin` was for executables that come from gems?

app/services/import_converter.rb
pezholio: Yeah, good point. Let's leave it as is

script/convert_import.rb
pezholio: I'm pretty sure it can be for any executable, but there's already a Ruby executable in `script/` already, so let's just leave it as is 👍

PR #1052, opened 2021-04-06
(1649) Change GCRF stategic area checkboxes to integers

pezholio:
This was  also causing issues where the `gcrf_strategic_area_options` were returning the code as an integer, but the codes were stored as a string. I looked into tweaking how the options were presented, or even changing the underlying YAML to made the codes into strings, but I finished up taking the approach that we should trust the codelist and if the codes are integers, so the data should be stored as an integer too.

### DIFF
--- a/app/presenters/activity_presenter.rb
+++ b/app/presenters/activity_presenter.rb
@@ -171,7 +171,7 @@ class ActivityPresenter < SimpleDelegator
 
   def gcrf_strategic_area
     return if super.blank?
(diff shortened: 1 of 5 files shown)


pezholio:
[GCRF stategic area checkboxes do not reflect the state when editing](https://trello.com/c/ss74c7f8/1649-gcrf-stategic-area-checkboxes-do-not-reflect-the-state-when-editing)

CristinaRO (approved):

PR #1054, opened 2021-04-06
Remove byte order mark from the start of import files

jcoglan:
Our script for extracting transactions and forecasts from import data
files was failing to recognise the first heading in CSV files because of
the presence of a byte order mark. We remove this before parsing.

We also make the converter throw an error if the ID columns it requires
are not present; `values_at` returns nil instead of throwing on
unrecognised keys.

### DIFF
--- a/script/convert_import.rb
+++ b/script/convert_import.rb
@@ -3,6 +3,8 @@ require "optparse"
 require_relative "../config/environment"
 require_relative "../app/services/import_converter"
 
+BYTE_ORDER_MARK = "\uFEFF".encode(Encoding::UTF_8)
+
 parser = OptionParser.new { |args|
   args.on "-i", "--input FILE"
   args.on "-o", "--output FILE"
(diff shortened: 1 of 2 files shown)


jcoglan:
Could I request that others not merge this PR as it's a work in progress -- we're ironing out bugs in the converter script as we go through the import files and I'd like to treat that as a unit of work and decide when to merge all of it.

CristinaRO:
> Could I request that others not merge this PR as it's a work in progress -- we're ironing out bugs in the converter script as we go through the import files and I'd like to treat that as a unit of work and decide when to merge all of it.

You can make it a Draft PR and then it *can't* be merged. (And it also sends a clearer message.) :)

script/convert_import.rb
```diff
@@ -12,7 +15,9 @@
 parser.parse!(into: options)
 
 level = options.fetch(:level).upcase
-rows = CSV.read(options.fetch(:input), headers: true)
+input_data = File.read(options.fetch(:input))
+input_data.delete_prefix!(BYTE_ORDER_MARK)
+rows = CSV.read(input_data, headers: true)
```
mec, line 20: Do we want `parse` here as `input_data` is a string now?

mec (approved):
I'd like the commit message to speak to why we had to add the Rails environment and the change to add fetch, other than that, looks good. 🚢

jcoglan:
@mec Thanks, I'll actually split these out into their own commits, this ended up being a grab-bag of random fixes while I was working on it.

PR #1056, opened 2021-04-07
(1648) Users cannot create or edit activites whilst there is no active report

pezholio:
This is already controlled by the ActivityPolicy, which means that a delivery partner can't create or edit an activity when there is no editable report. This PR just hides the "Add Child" buttons based on the policy - currently clicking on "Add child" will result in a "Not Authorised" error.

I've also added a test to make sure a delivery partner can't see "Add" or "Edit" links on any part of an Activity's details when there is no editable report. This is already the case, but gives us more confidence and documents the expected behaviour.

### DIFF
--- a/app/views/staff/organisations/_grouped_programmes.html.haml
+++ b/app/views/staff/organisations/_grouped_programmes.html.haml
@@ -25,6 +25,6 @@
                   %td{class: "govuk-table__cell"}
                     = activity.roda_identifier_compound
                   %td{class: "govuk-table__cell"}
-                    - if policy(:project).create?
+                    - if policy(activity).create_child?
                       = link_to(t("action.activity.add_child"), organisation_activity_children_path(activity.extending_organisation, activity), method: :post, class: "govuk-link govuk-!-margin-right-4")
                     = link_to(t("default.link.view"), organisation_activity_details_path(activity.extending_organisation, activity), class: "govuk-link govuk-!-margin-right-4")
--- a/app/views/staff/shared/activities/_projects.html.haml
+++ b/app/views/staff/shared/activities/_projects.html.haml
@@ -2,7 +2,7 @@
   .govuk-grid-column-full
     %h2.govuk-heading-m
       = t("page_content.activity.projects")
-    - if policy(:project).create?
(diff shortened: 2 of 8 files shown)


pezholio:
[Users cannot create or edit activites whilst there is no active report](https://trello.com/c/vblqkCHu/1648-users-cannot-create-or-edit-activites-whilst-there-is-no-active-report)

CristinaRO (commented):
Flagging up that this work intersects or overlaps with https://trello.com/c/b7wCEUkE/1644-use-createchild-when-deciding-whether-to-show-add-activity-buttons.

Up to you if the overlap is such that you're willing to remove the old `create` style policies as well like the card suggests, or if that would seem to expand the scope too much!

app/views/staff/organisations/_grouped_programmes.html.haml
```diff
@@ -25,6 +25,6 @@
                   %td{class: "govuk-table__cell"}
                     = activity.roda_identifier_compound
                   %td{class: "govuk-table__cell"}
-                    - if policy(:project).create?
+                    - if policy(:project).create? && policy(activity).create_child?
```
CristinaRO, line 28: I think, although I'm not 100%, but based on the scope of this card https://trello.com/c/b7wCEUkE/1644-use-createchild-when-deciding-whether-to-show-add-activity-buttons , that the `policy(:project).create?` check is now redundant and can be removed.

app/views/staff/shared/activities/_projects.html.haml
```diff
@@ -2,7 +2,7 @@
   .govuk-grid-column-full
     %h2.govuk-heading-m
       = t("page_content.activity.projects")
-    - if policy(:project).create?
+    - if policy(:project).create? && policy(activity).create_child?
```
CristinaRO, line 5: Same as above.

CristinaRO (approved):

app/views/staff/organisations/_grouped_programmes.html.haml
pezholio: Yeah, I think you're right. I'll see what happens if I get rid

PR #1058, opened 2021-04-08
(1566) Activity importer respects the policies

CristinaRO:
## Changes in this PR

- Activity importer checks permissions

## Next steps


CristinaRO:
[Activity import respects the policies](https://trello.com/c/W7orkOeN/1566-activity-import-respects-the-policies)

app/services/activities/import_from_csv.rb
```diff
@@ -67,7 +69,7 @@ def update_activity(row, index)
       elsif row["Parent RODA ID"].present?
         add_error(index, :parent_id, row["Parent RODA ID"], I18n.t("importer.errors.activity.cannot_update.parent_present")) && return
       else
-        updater = ActivityUpdater.new(row, @organisation)
+        updater = ActivityUpdater.new(row: row, user: @uploader, delivery_partner_organisation: @delivery_partner_organisation)
```
leeky, line 72: For consistency, it's probably worth reordering the parameters here to match the `ActivityCreator.new` call above.

app/services/activities/import_from_csv.rb
```diff
@@ -82,12 +84,20 @@ def add_error(row_number, column, value, message)
     class ActivityUpdater
       attr_reader :errors, :activity, :row
 
-      def initialize(row, organisation)
+      def initialize(row:, user:, delivery_partner_organisation:)
         @errors = {}
         @activity = find_activity_by_roda_id(row["RODA ID"])
-        @organisation = organisation
+        @user = user
+        @organisation = user.organisation
```
leeky, line 91: Is `@organisation` still used anywhere. I couldn't spot it.

app/services/activities/import_from_csv.rb
```diff
@@ -82,12 +84,20 @@ def add_error(row_number, column, value, message)
     class ActivityUpdater
       attr_reader :errors, :activity, :row
 
-      def initialize(row, organisation)
+      def initialize(row:, user:, delivery_partner_organisation:)
```
leeky, line 87: How about renaming `user` to `uploader` here as well. Handy to differentiate in cases (BEIS uploading?) where the uploader might not belong to the delivery partner organisation..?

app/services/activities/import_from_csv.rb
```diff
@@ -125,12 +135,20 @@ def find_activity_by_roda_id(roda_id)
...
+      def initialize(user:, row:, delivery_partner_organisation:)
+        @user = user
+        @organisation = user.organisation
+        @delivery_partner_organisation = delivery_partner_organisation
         @errors = {}
         @row = row
         @converter = Converter.new(row)
         @parent_activity = fetch_parent(@row["Parent RODA ID"])
+
+        if @parent_activity && !ActivityPolicy.new(@user, @parent_activity).create_child?
```
leeky, line 147: I wonder if `ActivityPolicy.create_child?` would benefit from `return false if record.nil?`, to remove the need to check it's not nil here (`if @parent_activity && ...`).

lib/tasks/import_activities.rake
```diff
@@ -28,8 +31,8 @@ namespace :activities do
     end
   rescue Errno::ENOENT
     abort "Cannot find the file at #{path}"
-  rescue ActiveRecord::RecordNotFound
-    abort "Can't find an organisation with the ID '#{organisation_id}'"
+  rescue ActiveRecord::RecordNotFound => error
```
leeky, line 34: 👍

leeky (approved):
Had some minor comments about naming things, but otherwise LGTM :shipit: 

I note that Coveralls is currently causing CI to fail because coverage is now below 98%.

app/services/activities/import_from_csv.rb
robbiepaul: I don't think we're able to do this, if the parent activity is nil we don't want to add an additional error to say the uploader is not authorised. But I agree it could be tidied up, not sure how best to though.

CristinaRO:
Thank you @leeky for the thorough review!

PR #1059, opened 2021-04-09
(1643) Link each question on the activity form to the guidance

pezholio:
This adds a new YAML file to the vendor directory, containing the IDs of the guidance pages contained in Zendesk, there is also a class that checks if guiance exists for a specific model and field. If it does, we return the full URL, otherwise we return an empty string. 

I've then extended the GOVUK form builder to check there is guidance available before rendering the input, if there is, we render a link to the guidance, which opens in a new tab.

# Screenshots

image

image

image


pezholio:
[Link each question on the activity form to the guidance](https://trello.com/c/QCtaQ6zB/1643-link-each-question-on-the-activity-form-to-the-guidance)

config/initializers/form_builder.rb
```diff
@@ -45,6 +47,11 @@ def has_errors?
 end
 
 module GOVUKDesignSystemFormBuilder
+  class FormBuilder < ActionView::Helpers::FormBuilder
+    delegate :content_tag, :tag, :safe_join, :link_to, :capture, to: :@template
+
+    include Roda::FormBuilder::Builder
```
leeky, line 53: Could you also create a custom form builder that inherits from `GOVUKDesignSystemFormBuilder::FormBuilder` and then specifying it on line 1 (https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1059/files#diff-bf1117bdafe92b2011b9ed317d701a0973101a9893f47509641c2d47f656df47L1), rather than injecting custom code into their class?

DFE Digital have been doing a reorganising of their code lately, and an internal change to this class has the potential to bite us in the future.

lib/roda/form_builder/builder.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      def with_guidance(attribute_name, html)
+        return if html.nil?
+
+        guidance_url = GuidanceUrl.new(@object_name, attribute_name).to_s
+
+        if guidance_url.present?
+          doc = Nokogiri::HTML.fragment(html)
+
+          doc.xpath("div").first.add_child(content_tag(:p, class: "guidance") {
```
leeky, line 15: I think this is good for now. Wondering if we should add ARIA rules to make sure that the guidance hint is associated correctly with the form. Maybe overriding the block (see [injecting content](https://govuk-form-builder.netlify.app/building-blocks/injecting-content/))?

lib/roda/form_builder/builder.rb
```diff
@@ -0,0 +1,47 @@
...
+        :govuk_url_field,
+        :govuk_number_field,
+        :govuk_text_area,
+        :govuk_collection_select,
+        :govuk_collection_radio_buttons,
+        :govuk_collection_check_boxes,
+        :govuk_date_field,
+      ].each do |method|
+        define_method(method) do |attribute_name, *args, &block|
+          with_guidance(attribute_name, super(attribute_name, *args))
```
leeky, line 42: Should `&block` be passed along too?

leeky (approved):
This works for me, just had a couple of concerns about accessibility and the way we inject into the form builder.

config/initializers/form_builder.rb
pezholio: Yeah, good call. It actually has bitten us here already as it goes - https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1066

lib/roda/form_builder/builder.rb
pezholio: Good call

lib/roda/form_builder/builder.rb
pezholio: I had a crack at this, but the way `&block` works injects content before the field, not after, which means we'll have to dig into quite a bit of the code to make this work. Might be nice as an improvement for the upstream gem though

PR #1062, opened 2021-04-12
(1645) Make organisation and extending_organisation consistent

leeky:
This data migration ensures that existing activities have `organisation` and `extending_organisation` set consistently. This mirrors the values that `ActivityDefaults` assigns for new activities depending on their level:

- Fund: organisation / extending_organisation are always both BEIS
- Programme: organisation is BEIS and extending_organisation is set to a delivery partner
- Project: organisation / extending_organisation are the same DP
- Third-party Project: organisation / extending_organisation are the same DP

In the past `extending_organisation_id` has been an optional field, so we copy this value from `organisation_id` for projects and third-party projects.

I output a warning for any programmes whose `extending_organisation` is incorrectly set to BEIS, as these will need to be manually updated, to allow the relevant delivery partner to create child activities (projects) under them.

## Next steps

### DIFF
--- a/db/data/20210412134546_fix_extending_organisation_inconsistencies.rb
+++ b/db/data/20210412134546_fix_extending_organisation_inconsistencies.rb
@@ -1,0 +1,18 @@
+# Run me with `rails runner db/data/20210412134546_fix_extending_organisation_inconsistencies.rb`
+
+beis = Organisation.find_by(service_owner: true)
+
+Activity.fund.update_all(organisation_id: beis.id, extending_organisation_id: beis.id)
+
+Activity.programme.where.not(extending_organisation_id: beis.id).update_all(organisation_id: beis.id)
+
+Activity.project.update_all("extending_organisation_id = organisation_id")
+
+Activity.third_party_project.update_all("extending_organisation_id = organisation_id")
+
+bad_programmes = Activity.programme.where(extending_organisation_id: beis.id)
+
+if bad_programmes.any?
+  puts "Found #{bad_programmes.size} programme(s) whose extending_organisation is incorrectly set to BEIS and should be manually corrected:"
+  puts bad_programmes.ids.join("\n")
+end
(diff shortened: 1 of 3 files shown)


spec/models/activity_spec.rb
```diff
@@ -77,8 +77,8 @@
       end
     end
 
-    describe ".publishable_to_iati?" do
-      it "only returns activities with a `form_state` of :complete and/or `publish_to_iati` of true" do
+    describe ".publishable_to_iati" do
+      it "only returns activities with a 'complete' and `publish_to_iati` is true" do
```
CristinaRO, line 81: While you're in there making things better (🎉 ), could you rephrase "with a 'complete'" to something else?

db/data/20210412134546_fix_extending_organisation_inconsistencies.rb
```diff
@@ -0,0 +1,18 @@
...
+
+Activity.project.update_all("extending_organisation_id = organisation_id")
+
+Activity.third_party_project.update_all("extending_organisation_id = organisation_id")
+
+bad_programmes = Activity.programme.where(extending_organisation_id: beis.id)
+
+if bad_programmes.any?
+  puts "Found #{bad_programmes.size} programme(s) whose extending_organisation is incorrectly set to BEIS and should be manually corrected:"
+  puts bad_programmes.ids.join("\n")
```
CristinaRO, line 17: How would you feel about generating the list of "bad" programmes as a list of production URLs? ("Nope" is a valid answer.)

CristinaRO (approved):
Approved in principle, a couple of comments.

db/data/20210412134546_fix_extending_organisation_inconsistencies.rb
leeky: Nice idea, but the Activity details page is scoped on organisation, and this warning is flagging places where it is unreliable.

PR #1067, opened 2021-04-14
(1644) Use ActivityPolicy#create_child? for more "create activity" buttons

leeky:
Update Organisation#show and the third-party projects shared partial to use the `ActivityPolicy#create_child?` check.

NB: There are still a few places where we're using the old level-specific `create?` checks, but it proved pretty difficult to remove without breaking many tests, so let's leave it for now.

## Next steps


pezholio (approved):
LGTM 👍

PR #1068, opened 2021-04-14
(1641) Start documenting view helpers

leeky:
Create a guide to the various view helpers that developers have available to them. Starting with:

- `a11y_action_link`
- `link_to_new_tab`

I did wonder about implementing a proper pattern library, especially as we have a number of shared partials that might be useful to document, but this like a good start.

---
Screenshot 2021-04-14 at 16 18 29

## Next steps


pezholio (approved):
LGTM 👍

PR #1070, opened 2021-04-15
(1673) Upgrade Rails and Ruby versions

pezholio:
This upgrades our Rails verson to 6.1.3, as there were some dependencies that were blocking this. I've also taken the opportunity to put us on the latest 2.x version of Ruby (2.7.3) too


pezholio:
[Upgrade Rails and Ruby versions](https://trello.com/c/t6Bbd446/1673-upgrade-rails-and-ruby-versions)

CristinaRO:
Could we try separating updating the ruby version from the Rails version? It might make it clearer which step is causing the CI build to fail. 🤷🏻

spec/views/staff/shared/message_spec.rb
```diff
@@ -9,11 +9,16 @@
 
   context "when passed a flash message with a hash" do
     it "renders a flash message with a hash that has title and errors keys" do
```
CristinaRO, line 11: Could you update the spec description as well, please, since this seems to be checking for the error messages and not the error keys?

pezholio: Oooh, good catch

Gemfile
```diff
@@ -23,7 +23,7 @@ gem "public_activity", "~> 1.5"
 gem "pundit"
 gem "rollbar"
 gem "rails", "~> 6.0.3"
-gem "redis", "~> 4.2"
+gem "redis", "< 4.2"
```
CristinaRO, line 26: The commit message says "the Redis version, which was pinned to `> 4.2` in Sidekiq 5.2.9". It should say "< 4.2", right?

pezholio: Whoops! Fixed

CristinaRO (approved):

PR #1071, opened 2021-04-15
(1664) Export spending and refund data

jcoglan:
## Changes in this PR

This introduces a new report export CSV with a spending breakdown showing spending, refunds, and net values as distinct amounts for recent quarters. Specifically for each activity in the report, it includes:

- RODA, BEIS and DP identifiers
- title and description
- programme status and ODA eligibility
- for the most recent 7 quarters, the total spend, total refund and net
- for 13 quarters before those, the total net spend
- for 5 quarters after the report date, the forecast spend
The final commit is a refactoring of some code I encountered that I thought I would need to complete this feature. It ended up not being relevant but I thought I'd include it here in case people want it included. I'm happy to remove it if not.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -601,6 +601,7 @@
 - Bump Ruby and Rails versions
 - Update Rails configuration files with v6.1 settings
 - Add tests to guard against missing Activity field translations
+- Provide a breakdown of spend, refund and net values as a report CSV download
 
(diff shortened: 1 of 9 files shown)


jcoglan:
[Export spending and refund data](https://trello.com/c/9mtRGz43/1664-export-spending-and-refund-data)

pezholio (approved):
LGTM - I think that last commit is fine to leave in FWIW. Always nice to make things better as we go

PR #1072, opened 2021-04-15
(1522) Budgets have providing organisations

CristinaRO:
## Changes in this PR

- Direct budgets automatically set the providing organisation to BEIS (the service owner)
- Transferred budgets require the user to select a delivery partner organisation
- External budgets require the user to enter the name and type of the providing organisation (and optionally its IATI reference)

With JavaScript enabled, the users will only be offered the relevant fields, depending on the selected budget type:
- Direct budgets: hide all providing org fields
- Transferred budgets: show only the radio buttons select
- External budgets: show only the name, type, and reference fieldset

### After
Screenshot 2021-04-15 at 17 57 45
Screenshot 2021-04-15 at 17 58 06
Screenshot 2021-04-15 at 17 58 34

## Next steps


CristinaRO:
[Budgets record the 'providing organisation'](https://trello.com/c/xMfYxd2t/1522-budgets-record-the-providing-organisation)

app/models/budget.rb
```diff
@@ -16,13 +16,17 @@ class Budget < ApplicationRecord
...
+  before_validation :infer_and_assign_providing_org_attrs
+
   validates_presence_of :report, unless: -> { parent_activity&.organisation&.service_owner? }
   validates_presence_of :value,
     :currency,
     :financial_year
   validates :value, numericality: {other_than: 0, less_than_or_equal_to: 99_999_999_999.00}
   validates :budget_type, inclusion: {in: BUDGET_TYPES.values}
   validate :direct_budget_type_must_match_source_fund, if: -> { DIRECT_BUDGET_TYPES.include?(budget_type) }
+  validates_presence_of :providing_organisation_id, if: -> { DIRECT_BUDGET_TYPES.include?(budget_type) }
```
leeky: I'm wondering if it's worth [grouping these validations](https://guides.rubyonrails.org/active_record_validations.html#grouping-conditional-validations) together?

```ruby
with_options if: -> { DIRECT_BUDGET_TYPES.include?(budget_type) } do |direct_budget|
  direct_budget.validate :direct_budget_type_must_match_source_fund
  direct_budget.validate :direct_budget_providing_org_must_be_beis
  direct_budget.validates :providing_organisation_id, presence: true
end
```

app/models/budget.rb
```diff
@@ -52,4 +56,14 @@ def iati_status
       errors.add(:budget_type, I18n.t("activerecord.errors.models.budget.attributes.funding_type.source_fund.#{parent_activity.source_fund_code}"))
     end
   end
+
+  private def direct_budget_providing_org_must_be_beis
+    errors.add(:providing_organisation_name, "Providing organisation for direct funding must be BEIS!") unless providing_organisation_id == Organisation.service_owner.id
```
leeky, line 63: Could you `service_owner?` on the association instead?

`unless providing_organisation.service_owner?`

app/models/organisation.rb
```diff
@@ -32,4 +32,8 @@ def ensure_beis_organisation_reference_is_uppercase
   def is_government?
     %w[10 11].include?(organisation_type)
   end
+
+  def self.service_owner
```
leeky, line 36: 🎉

spec/models/budget_spec.rb
```diff
@@ -40,6 +42,20 @@
       end
     end
 
+    describe "providing organisation" do
+      context "when the budget_type is a direct type" do
```
leeky, line 46: Do you need a test for the opposite case - i.e. that it **doesn't** get set for non-direct types?

jcoglan:
Just mentioning https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1082#issuecomment-823933367 as something we wanted to decide as part of this work -- do we stick to IE11-compatible JS, or write modern JS and possibly transpile it for older browsers?

My personal preference would be for the latter, as it's likely to be more effort for devs to remember which syntax they can and can't use, especially anyone that started learning JS after around 2016.

app/presenters/budget_presenter.rb
```diff
@@ -36,4 +36,10 @@ def currency
     return if super.blank?
     I18n.t("generic.default_currency.#{super.downcase}")
   end
+
+  def providing_organisation_name
```
leeky, line 40: I'm wondering if this should exist in the `Budget` model itself..?

spec/features/staff/users_can_create_a_budget_spec.rb
```diff
@@ -177,6 +179,39 @@
...
+        click_on(project_activity.title)
+        click_on(t("page_content.budgets.button.create"))
+
+        choose(user.organisation.name)
+        select "#{Date.current.year}-#{Date.current.next_year.year}", from: "budget[financial_year]"
+        fill_in "budget[value]", with: "1000.00"
+        click_button t("default.button.submit")
+
+        expect(page).to have_content(t("action.budget.create.success"))
+        expect(project_activity.budgets.last.providing_organisation.service_owner).to eql(true)
```
leeky: Would something like this be more readable:

`expect(project_activity.budgets.last.providing_organisation).to eql(beis)`

NB: `beis` isn't currently being set.

config/locales/models/budget.en.yml
```diff
@@ -14,6 +14,8 @@ en:
     legend:
       budget:
         budget_type: Type
+        providing_organisation: Providing organisation
```
leeky, line 20: Glad to see the automatic I18n lookups actually being used for once!

leeky (approved):
Have left a few small comments, but this otherwise looks good :shipit: 

I keep wondering if `Organisation` should be expanded to include external organisations, so we're not forcing users to enter the name and type endlessly for every budget line, but that's a discussion for the future!

app/models/budget.rb
CristinaRO: Thank you for the suggestion! Upcoming with the next push.

spec/models/budget_spec.rb
CristinaRO: Good catch! Incoming updates will include tests for the discarded inputs as well.

CristinaRO:
Thank you for the detailed review @leeky !

The only outstanding question about this PR is @jcoglan 's question about JavaScript: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1072#issuecomment-823949913

CristinaRO:
Robbie and I discussed this and concluded that we are ok to merge as is, given that even Microsoft pushes users towards using Edge, and it's a free OS upgrade.

We will also try to find out of we have any IE users (Google Tag Manager might have stats).

PR #1076, opened 2021-04-16
Make sure activity fields have all translations

pezholio:
This adds a shared example that covers all presenter fields that have a translation. It runs through all the codes in a given codelist and checks that the output does not match the `/translation missing/` regex. I did consider raising an error, but I was concerned that might cause issues with legacy data. This at least makes sure any data that we add with the codelists we have now (and any codelists that change going forward) are covered by the translations.

### DIFF
--- a/app/presenters/activity_presenter.rb
+++ b/app/presenters/activity_presenter.rb
@@ -6,7 +6,7 @@ class ActivityPresenter < SimpleDelegator
 
   def aid_type
     return if super.blank?
-    I18n.t("activity.aid_type.#{super.downcase}")
+    translate("activity.aid_type.#{super.downcase}")
   end
 
   def aid_type_with_code
@@ -15,12 +15,12 @@ class ActivityPresenter < SimpleDelegator
   end
 
   def covid19_related
-    I18n.t("activity.covid19_related.#{super}")
+    translate("activity.covid19_related.#{super}")
   end
 
   def sector
     return if super.blank?
-    I18n.t("activity.sector.#{super}")
+    translate("activity.sector.#{super}")
   end
 
   def sector_with_code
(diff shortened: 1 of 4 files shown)


pezholio:
[Translation missing for sector code "11260": Lower secondary education](https://trello.com/c/r9fCRN3a/1666-translation-missing-for-sector-code-11260-lower-secondary-education)

app/presenters/activity_presenter.rb
```diff
@@ -178,12 +178,12 @@ def gcrf_strategic_area
 
   def gcrf_challenge_area
     return if super.blank?
-    I18n.t(super, scope: "form.label.activity.gcrf_challenge_area_options")
+    I18n.t!(super, scope: "form.label.activity.gcrf_challenge_area_options")
```
CristinaRO, line 181: An errant `t!` asking to be replaced. ;)

spec/presenters/activity_presenter_spec.rb
```diff
@@ -3,7 +3,29 @@
 require "rails_helper"
 
 RSpec.describe ActivityPresenter do
+  RSpec.shared_examples "a code translator" do |field, args, field_type|
+    let(:field) { field }
+    let(:args) { args.reverse_merge(source: "iati") }
+    let(:cast) do
```
CristinaRO, line 9: Thanks Stu for explaining the layers of meta-programming! We agreed to replace this part with a less exciting but easier to follow `if` statement.

CristinaRO (approved):
Approved with the two agreed suggested changes.

Thanks for working out a way to reliably detect when the translations get out of date! This is awesome and will save us time in the long run.

pezholio:
Ta for the review and chat @CristinaRO! 👍

PR #1077, opened 2021-04-16
(1674) Activity search

jcoglan:
## Changes in this PR

This adds a search function to the site, capable of searching for activities on the following fields:

- RODA identifier (both the complete ID and the activity-specific fragment)
- BEIS identifier
- Delivery partner identifier
- Previous identifier
- Title

All the identifier fields are case-insensitive complete matches -- you cannot search for a substring of an activity's ID. The title is also case-insensitive but supports partial matches, so you can find an activity without remember its complete title.

We could extend this partial matching to IDs, and perform word splitting on the query, and so on, but I'd rather put this in front of users and test it before adding such complexity.

## Screenshots of UI changes

This shows the search box on the home page:

Screenshot 2021-04-16 at 18 01 03

Results for searching for an activity's RODA ID fragment:

Screenshot 2021-04-16 at 18 01 27

Results for searching for a substring of some activities' titles:

Screenshot 2021-04-16 at 18 04 10

## Next steps

### DIFF
--- a/app/views/staff/searches/_form.html.haml
+++ b/app/views/staff/searches/_form.html.haml
@@ -1,0 +1,12 @@
+.govuk-grid-row
+  .govuk-grid-column-full
+    %h2.govuk-heading-m
+      = t("form.activity_search.heading")
+
+    = form_with model: ActivitySearch.new, url: search_path, method: "get" do |f|
+      = f.govuk_text_field :query,
+        name: :query,
+        label: { text: t("form.activity_search.query") },
+        hint: { text: t("form.activity_search.hint") },
+        width: 20
+      = f.govuk_submit t("form.activity_search.submit")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -608,6 +608,7 @@
 ## [unreleased]
 
 - Link to the Service performance Zendesk article in the footer
+- Add functionality to search for activities by identifiers and title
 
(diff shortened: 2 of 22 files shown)


app/views/staff/searches/_form.html.haml
```diff
@@ -0,0 +1,8 @@
+.govuk-grid-row
+  .govuk-grid-column-full
+    %h2.govuk-heading-m
+      = t("form.activity_search.heading")
+
+    = form_with model: ActivitySearch.new, url: search_path, method: "get" do |f|
+      = f.govuk_text_field :query, name: :query, width: 20
```
pezholio, line 7: Could we add a more useful label here? Something like `Enter your search query`? We could then have a hint - something like `You can search by RODA, Delivery Partner, or BEIS identifier, or by the activity's title`

pezholio (changes requested):
Just one potential change here, just to tighten up some of the wording to make the search more approachable. Looks great other than that!

farrahfa:
[I can search for activities in RODA](https://trello.com/c/XsphJOMd/1674-i-can-search-for-activities-in-roda)

app/views/staff/searches/_form.html.haml
jcoglan: Thanks, I've added this copy and rebased my commits.

pezholio: Sweet 👍

pezholio (approved):
LGTM! 👍

PR #1084, opened 2021-04-20
Link to the Service performance article in footer

CristinaRO:
## Changes in this PR
- Link to the Service performance Zendesk article in the footer

I have chosen to link to the generic article that will collect all service performance reports for each quarter, rather than to the specific link for the current quarter.

My thinking was that we don’t want to have to update the link during every quarter, but anyone disagrees this can be discussed.

### Before
Screenshot 2021-04-20 at 17 34 35

### After
Screenshot 2021-04-20 at 17 34 23

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -607,6 +607,8 @@
 
 ## [unreleased]
 
+- Link to the Service performance Zendesk article in the footer
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-46...HEAD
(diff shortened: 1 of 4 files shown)


CristinaRO:
[Link to "Service performance" page in Zendesk in the RODA footer](https://trello.com/c/npjj3nvv/1556-link-to-service-performance-page-in-zendesk-in-the-roda-footer)

pezholio (approved):
I'm happy with this. I think it'll be a pain to have to remember to change the link each quarter, so would be better to have something generic.

PR #1086, opened 2021-04-22
(1704) Fix JavaScript deployment

jcoglan:
## Changes in this PR

Deployment is currently failing with the following error:

```
Uglifier::Error: Unexpected token: operator (>). To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).
```

This is because some JS we wrote recently uses an arrow function `=>`.

The recommended solution is to put this in our production config:

```
  config.assets.js_compressor = Uglifier.new(harmony: true)
```

However this does not work, because `Uglifier` does not understand some more recent JS syntax:

```
Uglifier::Error: Unexpected token: punc (.)
```

This is caused by a use of the `?.` operator, which is a more recent language edition that this tool does not understand.

Alongside this, we have a need to support IE11, per the [GOV.UK browser support docs](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).

We can fix both these problems by, either:

1. Avoiding using any modern JS syntax, sticking to ES5 or earlier features
2. Switch to modern tooling that understands new JS syntax and can convert and minify it for our supported browsers

I would prefer option 2, because option 1 requires developers to remember which JS features are supported by which tools. In particular, avoiding ES6 features will be more challenging for anyone that learned JS since 2016 and will only get more difficult over time.

I think it will impose less cognitive overhead if we allow modern JS syntax and have our deployment tools convert it as necessary.

The first 2 commits in this PR fix the problem without changing our toolchain, by removing the problematic JS code and allowing `assets:precompile` to work. The commits after that migrate the app to use [Webpacker](https://github.com/rails/webpacker), which is now the default method for compiling JS in Rails apps. Further details in the commit messages.

The main change to be aware of is that, instead of adding new JavaScript in `app/assets/javascripts`, you need to add files in `app/javascript/src` and import/execute your new code from `app/javascript/packs/application.js`.

## Next steps


jcoglan:
[Fix JavaScript deployment](https://trello.com/c/zYSQRLRx/1704-fix-javascript-deployment)

pezholio:
Looks like we need to add `yarn` as a dependency to the Dockerfile too

pezholio (approved):
LGTM. Great that we can now use modern JS with some confidence. Would be great to get this into the Rails template so other folks can benefit 👍

PR #1087, opened 2021-04-22
(1654) Show activities in a collapsable/expandable tree view table

robbiepaul:
## Changes in this PR

**Note: This replaces #1082 -- we're now using webpacker to generate assets so we can use more modern JavaScript. Thanks to @jcoglan for updating this code to work with webpacker**

Grouping the activities by parent allows users to more easily find the activity they are looking for in a list.

Grouping by programme means that BEIS users would lose out on the ability to view "fund" level activities. As a compromise, I decided to make the fund heading a link that will link to the fund details page.

The table tree view component is based on the GOV.UK Accordion component and has all the accessibility features associated with it.

For non-JS users all the content will be available on the page.

## Screenshots of UI changes

Screenshot 2021-04-20 at 15 10 23

## Next steps


pezholio (approved):
This is ace. Much snappier, and a really big UI improvement that I think users will love 👍

PR #1089, opened 2021-04-23
Use `pg_restore` to restore the dump

pezholio:
There has recently been a row introduced into the database, which had a `%` literal in the title, which has caused issues with the database restore. To get around this, and give us a more reliable restore process, I've set the format to be a `.tar` file, which can be restored via `pg_restore`, rather than a plain text SQL dump.


leeky (approved):
Worked for me! :shipit:

PR #1091, opened 2021-04-26
(1715) Remove Reporting Organisation from activities

pezholio:
This is the first PR in a series - working towards the aim of https://trello.com/c/kJG0qMT3/1647-replace-organisation-and-extendingorganisation-with-a-deliverypartner-attribute. 

As the reporting organisation is always going to be BEIS, this PR removes the `reporting_organisation` `belongs_to` association from the Activity model and instead sets the first part of the RODA identifier to be the value of the `SERVICE_OWNER_IATI_REFERENCE`, which is hardcoded to be `GB-GOV-13` (The IATI refefence of BEIS). The only place that `reporting_organisation` was actually returned was in the IATI XML, so I've tweaked that, so the IATI XML always returns BEIS.


pezholio:
[Remove Reporting Organisation association from the Activity](https://trello.com/c/kjHiSpzy/1715-remove-reporting-organisation-association-from-the-activity)

app/models/organisation.rb
```diff
@@ -34,6 +36,6 @@ def is_government?
   end
 
   def self.service_owner
-    Organisation.find_by(service_owner: true)
+    find_by(service_owner: true, iati_reference: SERVICE_OWNER_IATI_REFERENCE)
```
leeky, line 39: Since there's a database index is on `iati_reference`, I'm wondering if it'd be quicker to use that column alone here?

Perhaps the Organisation factory could also use this constant?

leeky (approved):
Have left a couple of minor comments, but otherwise LGTM! :shipit:

PR #1092, opened 2021-04-26
Database restore script passes organisation flag

leeky:
To prevent a GPaaS user who has access to multiple tenants from restoring the database for the wrong project, pass the `-o` switch to `cf login`.

Additionally, I've added a check which calls `cf target`. If the user is already logged in and has access to the BEIS tenant (i.e. they recently accessed the console in production), they won't be asked to reauthenticate.

### DIFF
--- a/bin/db-restore
+++ b/bin/db-restore
@@ -4,6 +4,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


pezholio (approved):
LGTM 👍

PR #1093, opened 2021-04-26
(1694) Feature include implementing organisation in the csv export

mec:
## Changes in this PR
Users want to see the implementing organisations in the report csv. As
there can be multiple implementing organisations for a given activity we
need a separator when outputting into a single csv cell. As with the
rest of the application we use a pipe "|" symbol to achieve this.

As a general pattern we use a presenter to change the way data is
presented in certain contexts, here the csv presenter does the job.

## Next steps

### DIFF
--- a/app/presenters/activity_csv_presenter.rb
+++ b/app/presenters/activity_csv_presenter.rb
@@ -12,4 +12,9 @@ class ActivityCsvPresenter < ActivityPresenter
     return if super.blank?
     super.join(" | ")
   end
+
+  def implementing_organisations
+    return if super.empty?
+    super.pluck(:name).join("|")
+  end
 end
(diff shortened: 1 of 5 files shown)


mec:
https://trello.com/c/JDw66wtV

app/presenters/activity_csv_presenter.rb
```diff
@@ -12,4 +12,10 @@ def country_delivery_partners
     return if super.blank?
     super.join(" | ")
   end
+
+  def implementing_organisations
+    return if super.empty?
+    implementing_organisations = super.collect { |implementing_organisation| implementing_organisation.name }
```
pezholio, line 18: You can probably achieve the same with a one-liner:

```suggestion
    super.pluck(:name).join("|")
```

pezholio (approved):
LGTM - just one teeny suggestion. Feel free to implement or not!

app/presenters/activity_csv_presenter.rb
mec: Gah! I knew there would be a one-liner, I just couldn't get to it! Thank you! I will use this 👍

PR #1094, opened 2021-04-26
Allow reports without financial year to be accessed

CristinaRO:
## Changes in this PR

Historical “ingested” reports don’t have any financial year and quarter.

Many of the features built around a report rely on the report having a financial quarter and year, and break if the report doesn’t.

In order to allow the historical reports to be accessed and downloaded, we need to add a lot of guards against the lack of financial year and quarter.

(Any reports being created now enforce the presence of a FY/FQ by assigning default values at creation, and the fields are readonly at database level.)

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -614,10 +614,10 @@
 - Show activities in an collapsable and expandable tree view table
 
 ## [unreleased]
-- implementing organisations are shown in the report csv file.
-
+- Implementing organisations are shown in the report csv file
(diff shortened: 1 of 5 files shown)


CristinaRO:
[Allow old ingest reports to be accessed](https://trello.com/c/6aSGAaxz/1695-allow-old-ingest-reports-to-be-accessed)

pezholio (approved):
LGTM - thanks for doing this! 👍

PR #1095, opened 2021-04-26
(1677) Sum programme and child activity transactions by financial quarter in IATI XML

robbiepaul:
## Changes in this PR

When reporting programme level activities to IATI we want to include transactions belonging to all child activities as well. We don't want an entry for each transaction rather we want to group them by financial quarter.

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -277,6 +277,20 @@ class Activity < ApplicationRecord
     Transaction.where(parent_activity_id: activity_ids)
   end
 
+  def reportable_transactions_for_level
+    if programme?
+      spend_by_financial_quarter(own_and_descendants_transactions.order("date DESC"))
+    else
+      transactions.order("date DESC")
+    end
+  end
+
+  private def spend_by_financial_quarter(reportable_transactions)
+    reportable_transactions.group_by(&:own_financial_quarter).map do |financial_quarter, actuals|
+      Transaction.new(date: financial_quarter.end_date, value: actuals.sum(&:value), transaction_type: Transaction::DEFAULT_TRANSACTION_TYPE)
+    end
+  end
+
   def valid?(context = nil)
     context = VALIDATION_STEPS if context.nil? && form_steps_completed?
     super(context)
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -250,9 +250,7 @@ class Activity < ApplicationRecord
   end
 
   def total_spend(financial_quarter = nil)
-    activity_ids = descendants.pluck(:id).append(id)
-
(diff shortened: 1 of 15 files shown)


robbiepaul:
['Rolled up' actual spend is calculated and included at level B in the IATI xml](https://trello.com/c/OSqf6t4D/1677-rolled-up-actual-spend-is-calculated-and-included-at-level-b-in-the-iati-xml)

app/models/activity.rb
```diff
@@ -277,6 +277,24 @@ def own_and_descendants_transactions
...
+  def reportable_transactions_for_level
+    reportable_transactions = if programme?
+      own_and_descendants_transactions
+    else
+      transactions
+    end.order("date DESC")
+
+    return reportable_transactions unless programme?
+
+    spend_by_financial_quarter(reportable_transactions)
```
pezholio, line 289: This tripped me up a bit - would it be clearer to return the `spend_by_financial_quarter` on the `own_and_descendants_transactions` if the activity is a programme, otherwise return the transactions? I may be missing something here though, so let me know if I am 👍 

```suggestion
    if programme?
      spend_by_financial_quarter(own_and_descendants_transactions.order("date DESC"))
    else
      transactions.order("date DESC")
    end
```

pezholio: Also, what happens if the activity is a fund?

app/models/activity.rb
```diff
@@ -277,6 +277,24 @@ def own_and_descendants_transactions
...
+  end
+
+  private def spend_by_financial_quarter(reportable_transactions)
+    reportable_transactions.group_by(&:own_financial_quarter).map do |financial_quarter, actuals|
+      Transaction.new(date: financial_quarter.end_date, value: actuals.sum(&:value), transaction_type: Transaction::DEFAULT_TRANSACTION_TYPE)
```
pezholio, line 294: Could we add a description to the Transaction to explain what activity the transaction is for?

app/models/activity.rb
robbiepaul: Nice suggestion, that's much clearer.

I don't think it's possible to report on a whole fund.

app/models/activity.rb
robbiepaul: Yeah, we discussed this with Meyric when we were planning out how to tackle it. We weren't sure what to use as a description for the grouped activities/child activities by financial quarter. The description isn't required so we decided to leave it off.

pezholio: Ah, fair enough 👍

pezholio (approved):
LGTM 👍

PR #1096, opened 2021-04-27
(1717) Validate that activities have the correct organisations

pezholio:
This adds a validation to make sure activities are owned by the correct organisation with the following rules:

* Funds should be owned by BEIS
* Programmes should be owned by BEIS
* Projects should be owned by a delivery partner
* Third Party Projects should be owned by a delivery partner

We have business logic in the app which forces this to happen anyway, but we have A LOT of tests that don't reflect this reality, and mean making changes to how we handle organisations is quite brittle.

I've also renamed the generic `activity` factory to `__activity`, to reflect the fact that its purpose is to set up defaults, and shouldn't be used on its own - in the app an activity is either a Fund, a Programme, a Project, or a Third-party Project, so we should be using these.

### DIFF
--- a/spec/factories/activity.rb
+++ b/spec/factories/activity.rb
@@ -1,5 +1,5 @@
 FactoryBot.define do
-  factory :activity do
+  factory :__activity do
     title { Faker::Lorem.sentence }
     delivery_partner_identifier { "GCRF-#{Faker::Alphanumeric.alpha(number: 5).upcase!}" }
     roda_identifier_fragment { Faker::Alphanumeric.alpha(number: 5) }
--- a/spec/features/staff/users_can_add_benefitting_countries_spec.rb
+++ b/spec/features/staff/users_can_add_benefitting_countries_spec.rb
@@ -1,8 +1,9 @@
 RSpec.feature "users can add benefitting countries as intended beneficiaries" do
-  context "when the user is signed as a BEIS user" do
-    let(:user) { create(:beis_user) }
+  context "when the user is signed as a delivery partner user" do
+    let(:user) { create(:delivery_partner_user) }
     before { authenticate!(user: user) }
     let(:activity) { create(:project_activity, :at_geography_step, organisation: user.organisation, intended_beneficiaries: nil) }
+    let!(:report) { create(:report, :active, fund: activity.associated_fund, organisation: user.organisation) }
 
     context "after choosing either country or region on the geography step" do
       scenario "the user will be asked if there are other benefitting countries to be added" do
--- a/spec/features/staff/users_can_add_benefitting_countries_spec.rb
+++ b/spec/features/staff/users_can_add_benefitting_countries_spec.rb
@@ -2,7 +2,7 @@ RSpec.feature "users can add benefitting countries as intended beneficiaries" do
   context "when the user is signed as a BEIS user" do
     let(:user) { create(:beis_user) }
     before { authenticate!(user: user) }
-    let(:activity) { create(:activity, :at_geography_step, organisation: user.organisation, intended_beneficiaries: nil) }
+    let(:activity) { create(:project_activity, :at_geography_step, organisation: user.organisation, intended_beneficiaries: nil) }
 
     context "after choosing either country or region on the geography step" do
       scenario "the user will be asked if there are other benefitting countries to be added" do
--- a/spec/features/staff/users_can_manage_activity_geography_spec.rb
+++ b/spec/features/staff/users_can_manage_activity_geography_spec.rb
@@ -1,8 +1,8 @@
 RSpec.feature "Users can provide the geography for an activity" do
-  context "when the user belongs to BEIS" do
-    let(:user) { create(:beis_user) }
+  context "when the user is a delivery partner" do
+    let(:user) { create(:delivery_partner_user) }
     before { authenticate!(user: user) }
-    let(:activity) { create(:project_activity, :at_geography_step, organisation: user.organisation) }
+    let(:activity) { create(:project_activity, :with_report, :at_geography_step, organisation: user.organisation) }
 
     scenario "they are asked to choose the geography" do
       visit activity_step_path(activity, :geography)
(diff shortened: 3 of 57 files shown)


pezholio:
[Validate that a fund is owned by the service owner](https://trello.com/c/iwLNdtWf/1717-validate-that-a-fund-is-owned-by-the-service-owner)

spec/policies/fund_policy_spec.rb
```diff
@@ -3,8 +3,8 @@
 RSpec.describe FundPolicy do
   subject { described_class.new(user, activity) }
 
-  let(:organisation) { create(:organisation) }
-  let(:activity) { create(:fund_activity, organisation: organisation) }
+  let(:beis_organisation) { create(:beis_organisation) }
+  let(:activity) { create(:fund_activity, organisation: beis_organisation) }
```
leeky, line 7: Can we remove `organisation` here (and therefore `beis_organisation` too?)

spec/factories/activity.rb
```diff
@@ -1,5 +1,5 @@
 FactoryBot.define do
-  factory :activity do
+  factory :__activity do
```
leeky, line 2: I'm surprised you didn't have to declare `class: "Activity"` here, so it knows what to create (unless there's some magic happening under the hood).

spec/features/staff/users_can_view_fund_level_activities_spec.rb
```diff
@@ -12,7 +12,7 @@
     before { authenticate!(user: user) }
 
     scenario "can view a fund level activity" do
-      fund_activity = create(:activity, level: :fund, organisation: user.organisation)
+      fund_activity = create(:fund_activity, organisation: user.organisation)
```
leeky, line 15: Remove `organisation` as in line 51?

spec/features/staff/users_can_add_benefitting_countries_spec.rb
```diff
@@ -1,8 +1,9 @@
 RSpec.feature "users can add benefitting countries as intended beneficiaries" do
-  context "when the user is signed as a BEIS user" do
-    let(:user) { create(:beis_user) }
+  context "when the user is signed as a delivery partner user" do
```
leeky, line 2: Good spot!

spec/features/staff/users_can_manage_activity_geography_spec.rb
```diff
@@ -1,8 +1,8 @@
 RSpec.feature "Users can provide the geography for an activity" do
-  context "when the user belongs to BEIS" do
-    let(:user) { create(:beis_user) }
+  context "when the user is a delivery partner" do
```
leeky, line 2: Again - well spotted!!

leeky (changes requested):
Great work - I'm very glad to be finally rid of those misleading tests!

There were a few places I spotted issue, mostly around overriding `user` in the policy specs.

spec/features/staff/users_can_add_benefitting_countries_spec.rb
pezholio: Yeah, I think there's a couple of tests like this, where it was just easier to be a BEIS user (no added complication around reports etc), but this feels much closer to reality.

spec/policies/transaction_policy_spec.rb
```diff
@@ -56,10 +56,9 @@
     let(:user) { create(:delivery_partner_user) }
 
     context "when the activity is a fund" do
-      let(:activity) { create(:fund_activity, organisation: user.organisation) }
-
-      it { is_expected.to permit_action(:show) }
+      let(:activity) { create(:fund_activity) }
 
+      it { is_expected.to forbid_action(:show) }
```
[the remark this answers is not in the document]
pezholio: Yeah, I spent a bit of time scratching my head over this, then I realised that it was doing the wrong thing!

pezholio:
Thanks for the thorough review - sometimes it's difficult to see the wood for the trees when you're doing tasks like this!

spec/features/staff/users_can_view_an_organisation_as_xml_spec.rb
```diff
@@ -8,7 +8,7 @@
     context "when the user is viewing the BEIS organisation show page" do
       scenario "they cannot download the organisation's projects as XML" do
         beis = user.organisation
-        _project = create(:project_activity, organisation: beis)
+        _project = create(:programme_activity, organisation: beis)
```
leeky, line 11: This changed from `create(:project_activity)` to `create(:programme_activity)`, so the variable should be updated accordingly.

spec/features/staff/users_can_view_an_organisation_as_xml_spec.rb
```diff
@@ -140,8 +140,9 @@
       end
 
       scenario "the XML file does not contain fund or programme activities in the organisation" do
-        _fund = create(:fund_activity, organisation: organisation)
-        _programme = create(:programme_activity, extending_organisation: organisation)
+        beis = create(:beis_organisation)
+        _fund = create(:fund_activity, organisation: beis)
+        _programme = create(:programme_activity, extending_organisation: beis)
```
leeky, line 145: Shouldn't this have it's ending organisation set to `organisation` i.e. a delivery partner (as defined in the `let!` at the top of the file)..?

spec/models/activity_spec.rb
```diff
@@ -129,36 +129,100 @@
 
     context "overall activity state" do
       context "when the activity form is final" do
-        subject { build(:activity, :at_identifier_step, form_state: "complete") }
+        subject { build(:project_activity, :at_identifier_step, form_state: "complete") }
+        it { should be_invalid }
+      end
+    end
+
+    context "when activity is a fund" do
```
leeky, line 137: I'm wondering if these belong in a `spec/validator` test instead 🤔

leeky (approved):
Have added some minor comments, but otherwise LGTM! :shipit:

spec/models/activity_spec.rb
pezholio: Yeah, probably. That spec is getting pretty big too! 😬

PR #1098, opened 2021-04-27
(1682) Add BEIS contact info to the IATI XML

robbiepaul:
- Add BEIS contact info to the IATI XML

## Next steps

### DIFF
--- a/app/views/staff/shared/xml/_activity.xml.haml
+++ b/app/views/staff/shared/xml/_activity.xml.haml
@@ -30,6 +30,15 @@
   %activity-date{"iso-date" => activity.planned_end_date, type: "3"}/
   - if activity.actual_end_date?
     %activity-date{"iso-date" => activity.actual_end_date, type: "4"}/
+  %contact-info{"type" => "1"}
+    %organisation
+      %narrative= I18n.t("contact_info.organisation")
+    %department
+      %narrative= I18n.t("contact_info.department")
+    %email= I18n.t("contact_info.email")
+    %website= I18n.t("contact_info.website")
+    %mailing-address
+      %narrative= I18n.t("contact_info.mailing_address")
   - if activity.recipient_country?
     %recipient-country{"code" => activity.recipient_country}
       %narrative= country_name_from_code(activity.recipient_country)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -623,6 +623,8 @@
 
 ## [unreleased]
 
+- Add BEIS contact info to the IATI XML
+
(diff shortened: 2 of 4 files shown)


robbiepaul:
[The IATI xml includes the `contact-info` element](https://trello.com/c/xqu3jRe3/1682-the-iati-xml-includes-the-contact-info-element)

app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -30,6 +30,15 @@
...
     %activity-date{"iso-date" => activity.actual_end_date, type: "4"}/
+  %contact-info{"type" => "1"}
+    %organisation
+      %narrative Department of Business Energy and Industrial Strategy
+    %department
+      %narrative General enquiries
+    %email enquiries@odamanagement.org
+    %website https://www.gov.uk/government/publications/beis-official-development-assistance-research-and-innovation
+    %mailing-address
+      %narrative Department of Business, Energy and Industrial Strategy, 4th Floor, 1 Victoria Street, SW1H 0ET
```
pezholio, line 41: Having this info hidden away makes me a bit nervous, becuase it smells like a constant. If there is any organisational change and BEIS changes their name, or there's a change in responsiblities, this could get missed. I wonder about having a singleton class that returns the info we have about the service owner, as well as the additional information like Email, website etc etc. That way, it's more visible, and we can use it in other places too. Something like this:

```ruby
class ServiceOwner < Organisation
  include Singleton

  def department
    "General enquiries"
  end

  def email
    "enquiries@odamanagement.org"
  end

  def address
    "Department of Business, Energy and Industrial Strategy, 4th Floor, 1 Victoria Street, SW1H 0ET"
  end

  def website
    "https://www.gov.uk/government/publications/beis-official-development-assistance-research-and-innovation"
  end

  private

  def initialize(attrs)
    super(organisation.attributes)
  end

  def organisation
    Organisation.find_by(iati_reference: Organisation::SERVICE_OWNER_IATI_REFERENCE)
  end
end
```

Whaddya reckon?

robbiepaul: I feel like a singleton class might be a premature optimisation. I agree though, the details could do with being more visible. I've extracted them out to a locale file, which seems to be the convention on this project. Hopefully, that'll mean that they won't get missed in any future updates.

pezholio: Ah, yeah, a locale file is probs a better idea 😃

pezholio (approved):

PR #1099, opened 2021-04-27
Deleting a forecast is non-destructive

CristinaRO:
## Changes in this PR

Deleting a forecast is equivalent to setting the latest revision to 0.

Rather than plainly deleting all revisions, different things need to happen depending on whether the forecast is pinned to a report, like in the case of level C and D activities, or not pinned, like in the case of levels A and B.

The forecast history already takes care of all these cases when setting a value to 0, whether through the interface or via bulk upload.

Deletion is a nicer way for users to express the intent to remove that forecast, while the history mechanism remains invisible from the outside.

Co-authored-by: jcoglan <james.coglan@dxw.com>

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -625,6 +625,7 @@
 
 - Add BEIS contact info to the IATI XML
 - Validate that activities have the correct organisation
+- Deleting a forecast doesn't delete data from approved reports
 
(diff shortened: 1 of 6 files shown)


CristinaRO:
[Safely delete planned disbursements](https://trello.com/c/O5fMcGRp/1706-safely-delete-planned-disbursements)

mec (approved):
🚢

PR #1100, opened 2021-04-27
Handle forecasts in the past error when importing

CristinaRO:
## Changes in this PR

Attempting to create a forecast for a past financial quarter raises an error.

We now rescue from this error and report it in the same way as any other errors in the uploaded data.

## Next steps


CristinaRO:
[Handle sequence error](https://trello.com/c/jT5K1doK/1719-handle-sequence-error)

PR #1109, opened 2021-05-10
Trigger a rebase on all auto merge PRs on every merge to Develop

pezholio:
As well as triggering a rebase by setting a PR to automerge or commenting `/rebase`, this PR adds the functionality to check over all PRs that are set to automerge and rebase them once a PR has been merged to `develop` or `master`. This is particularly helpful when we have a big backlog of PRs (i.e. from Dependabot), so we don't have to constantly manually rebase every time a new PR is merged.

### DIFF
--- a/.github/workflows/trigger_rebase.yml
+++ b/.github/workflows/trigger_rebase.yml
@@ -1,0 +1,42 @@
+name: Trigger Rebase
+
+on:
+  push:
+    branches:
+      - develop
+env:
+  AUTO_REBASE_PERSONAL_ACCESS_TOKEN: ${{ secrets.AUTO_REBASE_PERSONAL_ACCESS_TOKEN }}
+jobs:
+  fetch_all_prs_awaiting_automerge:
+    runs-on: ubuntu-latest
+    outputs:
+      pr_numbers: ${{ steps.get-all-awaiting-pr-numbers.outputs.pr_numbers }}
+    steps:
+      - name: Get all awaiting PR numbers
+        id: get-all-awaiting-pr-numbers
+        run: |
+          PR_NUMBERS=$(curl https://api.github.com/repos/UKGovernmentBEIS/beis-report-official-development-assistance/pulls\?state\=open | jq -c '.[] | select(.auto_merge!=null) | .number | [.]')
+          echo "::set-output name=pr_numbers::$PR_NUMBERS"
+  trigger_rebase_action:
+    if: ${{ needs.fetch_all_prs_awaiting_automerge.outputs.pr_numbers != '' }}
+    strategy:
+      matrix:
+        pr_number: ${{ fromJson(needs.fetch_all_prs_awaiting_automerge.outputs.pr_numbers) }}
+    needs: fetch_all_prs_awaiting_automerge
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get the latest ref for ${{ matrix.pr_number }}
+        run: |
+          PR_DATA=$(curl https://api.github.com/repos/UKGovernmentBEIS/beis-report-official-development-assistance/pulls/${{ matrix.pr_number }})
+          printf 'ref=%q\n' "$(echo "$PR_DATA" | jq -r .head.ref)" >> "$GITHUB_ENV"
+      - name: Checkout the latest code for ${{ matrix.pr_number }}
+        uses: actions/checkout@v2
+        with:
+          token: ${{ env.AUTO_REBASE_PERSONAL_ACCESS_TOKEN }}
+          fetch-depth: 0
+          ref: ${{ env.ref }}
+      - name: Automatic Rebase
+        uses: dxw/rebase@master
+        env:
+          GITHUB_TOKEN: ${{ env.AUTO_REBASE_PERSONAL_ACCESS_TOKEN }}
+          PR_NUMBER: ${{ matrix.pr_number }}


pezholio:
[Trigger a rebase on all auto merge PRs on every merge to Develop](https://trello.com/c/OcYXmfQC/1754-trigger-a-rebase-on-all-auto-merge-prs-on-every-merge-to-develop)

.github/workflows/trigger_rebase.yml
```diff
@@ -0,0 +1,42 @@
+name: Trigger Rebase
+
+on:
+  push:
```
jcoglan, line 4: Does the `push` event trigger when a branch head is moved by merging a PR via the GitHub site? The `develop` branch does not accept external pushes so I wasn't sure what this event covers.

pezholio: Yep, the CI gets run on develop when we merge PRs. This reponse in the Github Community confirms the same too https://github.community/t/does-a-pull-request-merge-always-trigger-the-push-event/17032/2?u=pezholio

.github/workflows/trigger_rebase.yml
```diff
@@ -0,0 +1,42 @@
...
+    strategy:
+      matrix:
+        pr_number: ${{ fromJson(needs.fetch_all_prs_awaiting_automerge.outputs.pr_numbers) }}
+    needs: fetch_all_prs_awaiting_automerge
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get the latest ref for ${{ matrix.pr_number }}
+        run: |
+          PR_DATA=$(curl https://api.github.com/repos/UKGovernmentBEIS/beis-report-official-development-assistance/pulls/${{ matrix.pr_number }})
+          echo "ref=$(echo $PR_DATA | jq ".head.ref") >> $GITHUB_ENV"
```
jcoglan, line 31: I'm not sure about the quoting of this line, but I might be getting confused as I always do about the quoting rules in yaml and bash and how they play together. If this line is sent verbatim to the shell, then it runs:

```sh
echo "ref=$(echo $PR_DATA | jq ".head.ref") >> $GITHUB_ENV"
```

Which would print, for example:

```
ref="chore/1579-remove-extending-organisation-controller" >> /path/to/github.env
```

Rather than appending the line `ref="chore/1579-remove-extending-organisation-controller"` to the file `/path/to/github.env`. Is that what's intended.

I'm also not sure if you're deliberately relying on the fact that `jq` puts `"` quotes around its output unless you use the `-r` switch, in order to make sure the appended line quotes the value of `ref`. A possibly safer version would do:

```sh
ref='$(echo $PR_DATA | jq -r ".head.ref")'
```

Since that avoids the possibility of a refname containing symbols with special meaning in bash that then get interpreted when inside `"` quotes.

jcoglan: The way I think this should be quoted, given the above, is:

```
echo "ref='$(echo "$PR_DATA" | jq -r .head.ref)'" >> "$GITHUB_ENV"
```

.github/workflows/trigger_rebase.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  fetch_all_prs_awaiting_automerge:
+    runs-on: ubuntu-latest
+    outputs:
+      pr_numbers: ${{ steps.get-all-awaiting-pr-numbers.outputs.pr_numbers }}
+    steps:
+      - name: Get all awaiting PR numbers
+        id: get-all-awaiting-pr-numbers
+        run: |
+          PR_NUMBERS=$(curl https://api.github.com/repos/UKGovernmentBEIS/beis-report-official-development-assistance/pulls\?state\=open | jq -c '.[] | select(.auto_merge!=null) | .number | [.]')
```
jcoglan, line 18: Are `curl` and `jq` installed by default on GitHub's Ubuntu runners?

.github/workflows/trigger_rebase.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Get the latest ref for ${{ matrix.pr_number }}
+        run: |
+          PR_DATA=$(curl https://api.github.com/repos/UKGovernmentBEIS/beis-report-official-development-assistance/pulls/${{ matrix.pr_number }})
+          echo "ref=$(echo $PR_DATA | jq ".head.ref") >> $GITHUB_ENV"
+      - name: Checkout the latest code for ${{ matrix.pr_number }}
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.AUTO_REBASE_PERSONAL_ACCESS_TOKEN }}
+          fetch-depth: 0
+          ref: ${{ ENV.ref }}
```
jcoglan, line 37: Should this be `env.ref` rather than `ENV.ref`? The [docs](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable) use the lowercase version but I don't know if that's important.

.github/workflows/trigger_rebase.yml
pezholio: Ah yeah, that looks better. Too many quotes in a YAML document tend to make my eyes bleed

.github/workflows/trigger_rebase.yml
pezholio: Yeah, I think you're right. Both work, but let's go with the documented version.

.github/workflows/trigger_rebase.yml
pezholio: Yep - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

.github/workflows/trigger_rebase.yml
jcoglan: An even safer way that makes sure the variable is correctly quoted in the environment file would be:

```sh
printf 'ref=%q\n' "$(echo "$PR_DATA" | jq -r .head.ref)" >> "$GITHUB_ENV"
```

pezholio: Oh nice, that's even better

.github/workflows/trigger_rebase.yml
```diff
@@ -0,0 +1,42 @@
+name: Trigger Rebase
+
+on:
+  push:
+    branches:
+      - develop
+env:
+  AUTO_REBASE_PERSONAL_ACCESS_TOKEN: ${{ secrets.AUTO_REBASE_PERSONAL_ACCESS_TOKEN }}
```
CristinaRO, line 8: Where does the "top level" env var get used, if anywhere?

The job steps get the token again from `secrets`, so they don't appear to use the top level var.

CristinaRO (approved):
James explained enough of the stuff herein for me to be confident to approve it.

Would be good to understand why we need to provide the token to the action's "global" environment, when it's already available in `secrets`, but it doesn't seem to expose it to any other processes that we don't already trust.

.github/workflows/trigger_rebase.yml
pezholio: Oooh, well spotted! 👍

jcoglan:
@pezholio Thanks for pointing me at the docs that answer my questions :) This has been a good prompt for me to actually learn GH Actions!

PR #1111, opened 2021-05-10
Infer the value of `sdgs_apply` when importing

CristinaRO:
## Changes in this PR

By default `sdgs_apply` is `false`.

Set it to `true` if at least one Sustainable Development Goal is populated during import.

NB: We don’t have a way to remove attributes during import, which is why we can’t currently detect if the goals are entirely removed in order to set it to `false`.

### Before

The goals themselves were set, but not correctly reflected in the interface. The interface said "Not applicable".

### After

The goals are set and they are reflected in the interface.

## Next steps


pezholio (approved):

pezholio (approved):
LGTM 👍

PR #1115, opened 2021-05-12
Total forecast respects versioning

jcoglan:
## Changes in this PR

This PR does two main things. In its first commit it fixes a bug in `Activity#total_forecasted` caused by this method ignoring that forecast data is versioned, and therefore double-counting multiple versions of the same forecast.

The commits after this are an attempt to address the underlying problem that it's not clear how to use `PlannedDisbursement` correctly. This model should not be used directly, but instead the `PlannedDisbursementHistory` and `PlannedDisbursementOverview` services should be used. The commits here make the codebase conform to this design and then block direct access to the model. Attempting to use the model directly will give an exception with the following message:

```
The PlannedDisbursement model should not accessed directly; please use the PlannedDisbursementHistory and PlannedDisbursementOverview services to create forecasts. See doc/forecasts-and-versioning.md for more information.
```

The document it mentions explains how to access forecast data, and explains how the underlying data model works.

## Next steps


jcoglan:
[Make activity total forecasts work with versioning](https://trello.com/c/fUL2IWr1/1764-make-activity-total-forecasts-work-with-versioning)

jcoglan:
I should add that I'd be fine with the `default_scope` trick not getting merged. Putting this in did help me find all the code that needed correcting, and I wanted some way to signal to other devs how this code should be used, and to make it harder to use it incorrectly by accident. That said it might be too much of a weird trick, and having this code by documented is a sufficient solution to the problem.

pezholio:
Not going to directly approve this, because I think we still need the rest of the team to look over it, but I feel a bit more confident in my understanding of all this. I think there are improvements we can make to make this stuff simpler, but I feel more confident that I could do this myself instead of waiting for `$PERSON_WHO_UNDERSTANDS_THIS_STUFF` to do it

CristinaRO:
Hashtag "saved you a click":

`spec/services/planned_disbursement_overview_spec.rb:344:9: Style/TrailingCommaInArrayLiteral: Put a comma after the last item of a multiline array.`

app/models/planned_disbursement.rb
```diff
@@ -2,6 +2,14 @@ class PlannedDisbursement < ApplicationRecord
   include PublicActivity::Common
   include HasFinancialQuarter
 
+  DIRECT_ACCESS_WARNING = "The PlannedDisbursement model should not be accessed \
+directly; please use the PlannedDisbursementHistory and PlannedDisbursementOverview \
+services to create forecasts. See doc/forecasts-and-versioning.md for more information."
+
+  default_scope do
```
mec, line 9: I never thought I would see an safe use for `default_scope` but this definately fits the bill!

pezholio (approved):

mec (approved):
This all looks good to me and helps cement my own philosophy of approaching PlannedDisbursements very carefully!

The doc is great and will really pay for itself in the future, thank you! I think it is worth a line in `readme.md` to call out the docs directory generally so it is not overlooked, there is some high value things in there and I know it is easily done!

I like the scoping error, if we are going to have something unconvential I think it is better to raise an unconvential error to force folks to think, coming to a Rails project and not interacting with the models directly would be considered pretty unusal by most people I think and this helps them spot that.

🚢

CristinaRO (approved):
Great work, I like the `default_scope` approach for future-proofing.

We're only holding off on merging because of the in-progress release 50 (and Github Actions shenanigans).

PR #1117, opened 2021-05-13
(1727) Extend Organisation model

pezholio:
This PR extends the Organisation model to allow a new type of organisation - Matched Effort Providers. I've also added a tabbed navigation to filter by organisation type and also a button to create each type of organisation.

Matched effort providers can optionally have a `beis_organisation_reference`, and it doesn't have the same kind of constraints on length as with a delivery partner organisation. They can also be set to "inactive", which means they are only kept for reference, and won't show up in the UI when we come to adding them to activities.

## Screenshots

image

image

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -9,27 +9,28 @@ module FormHelper
   end
 
   def list_of_user_roles
-    @list_of_user_roles ||= begin
-      User.roles.map { |id, name| OpenStruct.new(id: id, name: t("activerecord.attributes.user.roles.#{name}")) }
-    end
+    @list_of_user_roles ||= User.roles.map { |id, name| OpenStruct.new(id: id, name: t("activerecord.attributes.user.roles.#{name}")) }
   end
 
   def list_of_financial_quarters
-    @list_of_financial_quarters ||= begin
-      FinancialQuarter::QUARTERS.map { |id| OpenStruct.new(id: id, name: "Q#{id}") }
-    end
+    @list_of_financial_quarters ||= FinancialQuarter::QUARTERS.map { |id| OpenStruct.new(id: id, name: "Q#{id}") }
   end
 
   def list_of_financial_years(years = FinancialYear.next_ten)
-    @list_of_financial_years ||= begin
-      years.map { |year| OpenStruct.new(id: year.to_i, name: year.to_s) }
-    end
+    @list_of_financial_years ||= years.map { |year| OpenStruct.new(id: year.to_i, name: year.to_s) }
   end
 
   def user_active_options
     [
       OpenStruct.new(id: "true", name: t("form.user.active.active")),
       OpenStruct.new(id: "false", name: t("form.user.active.inactive")),
+    ]
+  end
+
+  def organisation_active_options
+    [
+      OpenStruct.new(id: "true", name: t("form.label.organisation.active.true")),
+      OpenStruct.new(id: "false", name: t("form.label.organisation.active.false")),
     ]
   end
 end
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -20,6 +20,7 @@ class Organisation < ApplicationRecord
   validates :iati_reference,
     uniqueness: {case_sensitive: false},
     presence: true,
+    unless: proc { |organisation| organisation.matched_effort_provider? },
     format: {with: /\A[a-zA-Z]{2,}-[a-zA-Z]{3}-.+\z/, message: I18n.t("activerecord.errors.models.organisation.attributes.iati_reference.format")}
   validates :beis_organisation_reference, uniqueness: {case_sensitive: false}
   validates :beis_organisation_reference,
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -7,6 +7,11 @@ class Organisation < ApplicationRecord
   has_many :users
   has_many :funds
 
+  enum role: {
+    delivery_partner: 0,
+    service_owner: 99,
+  }
+
   validates_presence_of :organisation_type, :language_code, :default_currency
   validates :name,
     presence: true,
@@ -21,7 +26,7 @@ class Organisation < ApplicationRecord
     format: {with: /\A[A-Z]{2,10}\z/, message: I18n.t("activerecord.errors.models.organisation.attributes.beis_organisation_reference.format")}
 
   scope :sorted_by_name, -> { order(name: :asc) }
-  scope :delivery_partners, -> { sorted_by_name.where(service_owner: false) }
+  scope :delivery_partners, -> { sorted_by_name.where(role: "delivery_partner") }
 
   before_validation :ensure_beis_organisation_reference_is_uppercase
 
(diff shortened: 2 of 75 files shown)


pezholio:
[Extend Organisation model](https://trello.com/c/kFlaEN56/1727-extend-organisation-model)

spec/models/organisation_spec.rb
```diff
@@ -77,11 +77,11 @@
       end
     end
 
-    context "when an organisation is not deliberately flagged as BEIS" do
-      it "should default to false" do
-        new_organisation = create(:delivery_partner_organisation)
+    context "when an organisation a matched effort provider" do
```
mec, line 94: Typo: "when an organisation is a matched effort provider"

and a sneaky space on line 81.

config/routes.rb
```diff
@@ -22,7 +22,13 @@
         get "historic" => "activities#historic"
       end
     end
-    resources :organisations, except: [:destroy] do
+    resources :organisations, except: [:destroy, :index] do
+      constraints role: /delivery_partners|matched_effort_providers/o do
+        collection do
+          get "(:role)", to: "organisations#index", defaults: {role: "delivery_partners"}, as: ""
```
mec: The commit message mentions the `nulldb` but I am not seeing and I thought we got rid of that - am I missing a it?

app/views/staff/organisations/_table.html.haml
```diff
@@ -1,20 +1,35 @@
+.govuk-grid-row
+  .govuk-grid-column-full
+    .govuk-tabs
+      %h2.govuk-tabs__title
+        Activities
```
mec, line 5: This probably should be 'Organisations'?

spec/features/staff/beis_users_can_view_organisations_spec.rb
```diff
@@ -24,33 +58,28 @@
       context "the role is 'delivery_partners'" do
         let(:role) { "delivery_partners" }
 
-        scenario "it lists delivery partner organisations" do
-          expect(page).to have_content(t("page_title.organisation.index"))
+        include_examples "lists delivery partner organisations"
 
-          delivery_partner_organisations.each do |organisation|
-            expect(page).to have_content(organisation.name)
+        context "it allows toggling to view matched effort providers" do
```
mec, line 63: This is super picky and feel free to disagree/ignore, but I think this would read better as something like:

> when viewing the matched effort providers tab…

config/brakeman.ignore
```diff
@@ -1,13 +1,33 @@
 {
   "ignored_warnings": [
+    {
```
mec, line 3: Are we saying that the word 'role' is what is triggering this?

app/models/organisation.rb
```diff
@@ -21,10 +21,11 @@ class Organisation < ApplicationRecord
     uniqueness: {case_sensitive: false},
     presence: true,
     format: {with: /\A[a-zA-Z]{2,}-[a-zA-Z]{3}-.+\z/, message: I18n.t("activerecord.errors.models.organisation.attributes.iati_reference.format")}
+  validates :beis_organisation_reference, uniqueness: {case_sensitive: false}
   validates :beis_organisation_reference,
     presence: true,
-    uniqueness: {case_sensitive: false},
```
mec, line 20: Have we lost uniqueness or am I missing something?

mec (changes requested):
Nice work! 

I love the role enum, everything reads so nice now (and we now know we have `external income provider` coming into the organisations mix!).

The changes you put in for the organisation factory make a big difference for me in the readability of the specs - thank you for taking the time to do that!

Most of my comments are questions/clarifications, I left one thing about making the IATI reference becoming optional for matched effort providers (which probably should have been on the card - so probs my bad!) as part of this so I've asked for changes - let me know if you think that is achievable or not?

app/models/organisation.rb
```diff
@@ -7,6 +7,11 @@ class Organisation < ApplicationRecord
   has_many :users
   has_many :funds
 
+  enum role: {
+    delivery_partner: 0,
+    service_owner: 99,
```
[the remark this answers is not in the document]
pezholio: I guess my thinking was that we're only likely to have one service owner org, so it's an exception rather than a rule

spec/factories/organisation.rb
```diff
@@ -1,5 +1,5 @@
 FactoryBot.define do
-  factory :organisation do
+  factory :__organisation do
```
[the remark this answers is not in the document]
pezholio: Ah yeah, we did a similar thing to `Activity`s, but I can update the commit message

app/services/planned_disbursement_history.rb
```diff
@@ -96,7 +96,7 @@ def create_original_entry(value, report = nil)
     attributes = series_attributes.merge(required_attributes).merge(
       planned_disbursement_type: :original,
       value: value,
-      report: report,
```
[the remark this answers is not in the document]
pezholio: Hmmm, I'm not sure. I have standardrb running in my IDE, so I think this happened automatically on save

config/routes.rb
pezholio: Nope, I just forgot the edit the commit message. Have binned it.

spec/features/staff/beis_users_can_view_organisations_spec.rb
pezholio: Good point, it's a context after all!

config/brakeman.ignore
pezholio: Yeah, that's right 👍

app/models/organisation.rb
pezholio: No, we validate by uniqueness on line 24. This validation only validates the format now

app/helpers/form_helper.rb
```diff
@@ -9,21 +9,15 @@ def list_of_delivery_partners
   end
 
   def list_of_user_roles
-    @list_of_user_roles ||= begin
```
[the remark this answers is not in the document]
pezholio: I think this is Standardrb / Prettier automatically doing this on save

app/models/organisation.rb
```diff
@@ -21,10 +21,11 @@ class Organisation < ApplicationRecord
     uniqueness: {case_sensitive: false},
     presence: true,
```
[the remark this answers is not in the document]
pezholio: That's not a problem - do we even want the `iati_reference` to appear on the form?

mec (commented):
Just want to know what the `o` on the new route does, then all good?!

config/routes.rb
```diff
@@ -22,7 +22,13 @@
         get "historic" => "activities#historic"
       end
     end
-    resources :organisations, except: [:destroy] do
+    resources :organisations, except: [:destroy, :index] do
+      constraints role: /delivery_partners|matched_effort_providers/o do
```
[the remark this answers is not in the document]
pezholio: Again, this is a StandardRB thing. I think it chucks it in when you have interpolation in a regex. Now we don't have string interpolation, we don't need it, so I can bin it.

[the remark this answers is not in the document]
mec: oh! That is a regex! I didn't even realise that!

mec (approved):

PR #1128, opened 2021-05-19
(1756) Update the request an account link

pezholio:
Also move the Matched Effort Provider entry to unreleased, as it's not yet released


pezholio:
[Request account link on start page is broken](https://trello.com/c/ONOSuR6f/1756-request-account-link-on-start-page-is-broken)

CristinaRO (approved):
When you fix the conflict in the changelog, would you mind moving the line "- Stop forecast totals double-counting multiple versions of the same forecast" under unreleased as well?

PR #1129, opened 2021-05-19
(1805) Transfers must be associated to the report

CristinaRO:
## Changes in this PR
- Allow DP users to create transfers from project and 3rd party project activities, and associate them with the current report
- Allow BEIS users to create transfers from fund and programme activities only

### After
Screenshot 2021-05-19 at 17 41 53

## Next steps

### DIFF
--- a/app/controllers/staff/transfers_controller.rb
+++ b/app/controllers/staff/transfers_controller.rb
@@ -5,10 +5,11 @@ class Staff::TransfersController < Staff::BaseController
 
   def new
     @source_activity = source_activity
+
+    authorize @source_activity, :create_transfer?
+
     @transfer = Transfer.new
     @transfer.source = @source_activity
-
-    authorize @transfer
   end
 
   def edit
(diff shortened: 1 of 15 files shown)


CristinaRO:
[Transfers must be associated to the report in which they are added](https://trello.com/c/iXq86nn5/1805-transfers-must-be-associated-to-the-report-in-which-they-are-added)

app/controllers/staff/transfers_controller.rb
```diff
@@ -5,10 +5,14 @@ class Staff::TransfersController < Staff::BaseController
...
   def new
     @source_activity = source_activity
+
+    authorize @source_activity, :create_transfer?
+
     @transfer = Transfer.new
     @transfer.source = @source_activity
-
-    authorize @transfer
+    if @source_activity.project? || @source_activity.third_party_project?
```
jcoglan, line 13: This probably doesn't need to happen here, because `@transfer.report` won't be used to authorize the action and won't be used in the form displayed on the page. The important thing is that this association is done in the `create` action.

jcoglan (approved):
We made the one change I asked for ourselves, so I'm merging this.

PR #1131, opened 2021-05-20
(1825) Fix roles in database seeds

leeky:
Update the database seed files to reflect the recent changes to the organisation model.

Also, I've moved the confirmation message for destroying the local development database earlier, so the remote database dump only happens when the developer responds positively.


mec (approved):
🚢

PR #1133, opened 2021-05-21
Add some useful stats to our healthcheck endpoint

pezholio:
This sets the current Git sha and the time the image was built as env vars in our Docker file, which we then expose via the healthcheck endpoint, which will be useful for checking deploy throughput, as well as letting us know exactly what version of the code is deployed.


lozette (approved):
🚀

PR #1135, opened 2021-05-24
(1824) Address some policy inconsistencies

leeky:
- Remove roles from users. All users were 'administrators', with their access determined by the organisation they belong to
- Update `delivery_partner?` to use the organisation role instead of assuming "NOT a service owner" means the same thing

## Next steps

### DIFF
--- a/app/policies/organisation_policy.rb
+++ b/app/policies/organisation_policy.rb
@@ -27,14 +27,4 @@ class OrganisationPolicy < ApplicationPolicy
   private def associated_user?
     user.organisation.eql?(record)
   end
-
-  class Scope < Scope
-    def resolve
-      if user.administrator?
-        scope.all
-      else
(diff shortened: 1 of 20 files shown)


app/policies/organisation_policy.rb
```diff
@@ -30,11 +30,7 @@ def download?
 
   class Scope < Scope
     def resolve
-      if user.administrator?
-        scope.all
-      else
-        scope.where(id: user.organisation_id)
-      end
+      scope.all
```
CristinaRO, line 33: Given that this class isn't doing any significant filtering, is this still necessary to keep pundit magic functioning?

CristinaRO (approved):
Great tidy up and good catch on the `delivery_partner` =/= `!beis_user` situation! One non-blocking question.

PR #1137, opened 2021-05-24
Use consistent naming for China

CristinaRO:
## Changes in this PR
The recipient country codelist is lifted from IATI codelists, and is the source for the select boxes used on the activity form to select the recipient countries.

However, in this case, we don’t want to create a BEIS-specific codelist file, and accept that if we ever replace this file with a new codelist release from IATI, we might have to do the change again.

The IATI locale file was compiled by us based on official IATI codelists, and is used to display country names in both recipient country and intended beneficiaries sections on the activity details page.

### Before
Screenshot 2021-05-24 at 14 23 21
Screenshot 2021-05-24 at 14 24 08

### After
Screenshot 2021-05-24 at 14 28 23
Screenshot 2021-05-24 at 14 28 14

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -651,6 +651,7 @@
 - Sum programme and child activity forecasts by financial quarter in IATI XML
 - Remove unused 'role' from User model
 - Open external links in the footer in new tabs
+- Use consistent naming for China (People's Republic of) when selecting and showing recipient country and intended beneficiaries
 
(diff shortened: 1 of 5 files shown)


CristinaRO:
[China's name should consistently be "China (People's Republic of)" inline with DAC designation](https://trello.com/c/VxMUwIVr/1736-chinas-name-should-consistently-be-china-peoples-republic-of-inline-with-dac-designation)

mec (approved):
:ship

PR #1139, opened 2021-05-25
(1726) Introduce MatchedEffort model

pezholio:
This adds a new Matched Effort model to be used when a project has any matched funding applied to it, whether that's in kind or through direct monetary contributions. 

# Screenshots

image

image

image

image

### DIFF
--- a/app/controllers/staff/matched_efforts_controller.rb
+++ b/app/controllers/staff/matched_efforts_controller.rb
@@ -6,6 +6,13 @@ class Staff::MatchedEffortsController < Staff::BaseController
     @matched_effort = MatchedEffort.new
   end
 
+  def edit
+    @activity = Activity.find(params[:activity_id])
+    authorize @activity
+
+    @matched_effort = MatchedEffort.find(params[:id])
+  end
+
   def create
     @activity = Activity.find(params[:activity_id])
     authorize @activity
@@ -18,6 +25,22 @@ class Staff::MatchedEffortsController < Staff::BaseController
       redirect_to organisation_activity_other_funding_path(@activity.organisation, @activity)
     else
       render :new
+    end
+  end
+
+  def update
+    @activity = Activity.find(params[:activity_id])
+    authorize @activity
+
+    @matched_effort = MatchedEffort.find(params[:id])
+    @matched_effort.assign_attributes(matched_effort_params)
+
+    if @matched_effort.valid?
+      @matched_effort.save
+      flash[:notice] = t("action.matched_effort.update.success")
+      redirect_to organisation_activity_other_funding_path(@activity.organisation, @activity)
+    else
+      render :edit
     end
   end
 
--- a/app/controllers/staff/matched_efforts_controller.rb
+++ b/app/controllers/staff/matched_efforts_controller.rb
@@ -21,6 +21,8 @@ class Staff::MatchedEffortsController < Staff::BaseController
 
     if @matched_effort.valid?
       @matched_effort.save
+      @matched_effort.create_activity key: "matched_effort.create", owner: current_user
+
       flash[:notice] = t("action.matched_effort.create.success")
       redirect_to organisation_activity_other_funding_path(@activity.organisation, @activity)
     else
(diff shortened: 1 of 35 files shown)


pezholio:
[Introduce MatchedEffort model](https://trello.com/c/OOg8bsn7/1726-introduce-matchedeffort-model)

spec/models/matched_effort_spec.rb
```diff
@@ -32,5 +31,48 @@
...
+      subject { build(:matched_effort, funding_type: funding_type, category: category) }
+      let(:category) { nil }
+
+      context "when the funding type is in kind" do
+        let(:funding_type) { "in_kind" }
+        let(:category) { "staff_time" }
+
+        it { should validate_presence_of(:category) }
+
+        context "and the category is not applicable to the in kind funding type" do
```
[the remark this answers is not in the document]
jcoglan: Similarly it would be helpful to have this in the other context below to spell out which combinations of category and funding type are valid.

app/models/matched_effort.rb
```diff
@@ -13,4 +13,44 @@ class MatchedEffort < ApplicationRecord
 
   validates_with MatchedEffortOrganisationValidator, if: -> { organisation_id.present? }
   validates_with MatchedEffortCategoryValidator
+
```
jcoglan, line 18: This is a really nice change, it's much nicer for future maintainability to access this functionality via method calls than by accessing constants and hashtables from another class directly.

app/controllers/staff/matched_efforts_controller.rb
```diff
@@ -0,0 +1,39 @@
+class Staff::MatchedEffortsController < Staff::BaseController
+  def new
+    @activity = Activity.find(params[:activity_id])
+    authorize @activity
```
jcoglan, line 6: These `authorize` calls are going to invoke `ActivityPolicy#new?` and `ActivityPolicy#create?` -- is that what we want? This is about attaching new records to an activity, not creating an activity itself.

app/controllers/staff/matched_efforts_controller.rb
```diff
@@ -6,6 +6,13 @@ def new
     @matched_effort = MatchedEffort.new
   end
 
+  def edit
+    @activity = Activity.find(params[:activity_id])
+    authorize @activity
```
jcoglan: Likewise this will invoke `ActivityPolicy#edit?` and `ActivityPolicy#update?`, which might be what we need but I wanted to double check.

app/controllers/staff/matched_efforts_controller.rb
```diff
@@ -1,24 +1,24 @@
...
   end
 
   def edit
     @activity = Activity.find(params[:activity_id])
-    authorize @activity
-
     @matched_effort = MatchedEffort.find(params[:id])
+
+    authorize @matched_effort
```
jcoglan, line 13: Ah, you sorted the authorization out in a later commit :)

spec/policies/matched_effort_policy_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      it { is_expected.to forbid_action(:create) }
+      it { is_expected.to forbid_action(:update) }
+      it { is_expected.to forbid_action(:destroy) }
+    end
+  end
+
+  context "as a Delivery partner user" do
+    let(:user) { build_stubbed(:delivery_partner_user) }
+
+    context "when the matched effort belongs to an activity owned by the user" do
```
jcoglan, line 32: Would it be worth adding a context here for when the `MatchedEffort` belongs to an `Activity` the DP user does not own?

jcoglan:
Wow that was a mammoth PR :) It mostly looked quite straightforward and I like that it's broken into lots of small steps. I would help to flesh the commit messages out a bit, especially where there's just a subject line, to explain the modelling being added in each commit and how it represents the real-world domain.

pezholio:
@jcoglan Cool - thanks for having a look - sorry it's so big, but hopefully breaking it down into smaller commits helped. Will have a look at your comments now 👍

app/models/matched_effort.rb
pezholio: Thanks! I agnoised over it for a bit, because I wasn't sure if I was overengineering things, but it also helps with building out the form, so useful in two places

spec/models/matched_effort_spec.rb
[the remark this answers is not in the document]
pezholio: Cool, done!

jcoglan (approved):
Nice work, thanks for adding those extra tests :)

PR #1143, opened 2021-05-25
(1772) Set autocomplete to `off` for activity search

pezholio:
When usig the search in Firefox (v89) as the text field is not set to `autocomplete=off` the browser show historic
suggestions. these suggestion take a second to appear and  by that point the user is click the search button directly under the autocomplete list thereby selecting a possibly incorrect serach value and submitting the form.

This sets `autocomplete` to `off`, so the browser does not show the historic suggestions.


pezholio:
[Search textfield should not autocomplete in browser](https://trello.com/c/slNdk1bu/1772-search-textfield-should-not-autocomplete-in-browser)

jcoglan (approved):

PR #1144, opened 2021-05-25
(1818) BEIS users receive report approved notification

CristinaRO:
## Changes in this PR
- Send approved report email notifications to BEIS users

## Next steps

### DIFF
--- a/app/mailers/report_mailer.rb
+++ b/app/mailers/report_mailer.rb
@@ -34,9 +34,19 @@ class ReportMailer < ApplicationMailer
     @user = params[:user]
     raise_unless_active_user
 
-    view_mail(ENV["NOTIFY_VIEW_TEMPLATE"],
-      to: @user.email,
-      subject: t("mailer.report.approved.subject", application_name: t("app.title")))
+    @role = if @user.organisation == @report_presenter.organisation
+      :delivery_partner
+    elsif @user.organisation.service_owner?
+      :service_owner
+    end
+
+    if @role.present?
+      view_mail(ENV["NOTIFY_VIEW_TEMPLATE"],
+        to: @user.email,
+        subject: t("mailer.report.approved.#{@role}.subject", application_name: t("app.title")))
+    else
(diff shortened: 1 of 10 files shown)


CristinaRO:
[BEIS users receive the "report approved" notification email](https://trello.com/c/TUVYJDhn/1818-beis-users-receive-the-report-approved-notification-email)

app/mailers/report_mailer.rb
```diff
@@ -34,9 +34,19 @@ def approved
...
-    view_mail(ENV["NOTIFY_VIEW_TEMPLATE"],
-      to: @user.email,
-      subject: t("mailer.report.approved.subject", application_name: t("app.title")))
+    @role = if @user.organisation == @report_presenter.organisation
+      :delivery_partner
+    elsif @user.organisation.service_owner?
+      :service_owner
+    end
+
+    if @role.present?
```
jcoglan, line 43: I'm not sure how this condition could fail, since the statement above it will always assign `@role` to either `:delivery_partner` or `:service_owner`. What situation are we trying to guard against here?

CristinaRO: We're not always assigning it, the second branch is an `elsif`, for the rare case that a user from a different organisation could make it all the way to the mailer.

CristinaRO: I did question myself if it was possible to end up with a situation where _both_ `@user.organisation` and `@report_presenter.organisation` are `nil`, in which case equality would pass for the wrong reasons, but if the report organisation were somehow nil, email sending would fail when executing `@report.organisation.users.active`.

jcoglan: It would make no sense to have a `Report` with no `organisation` present. I'd rather enforce that in the model or database than guard against it here.

CristinaRO: Oh yeah, I wasn't going to add a guard against that here.

jcoglan (approved):

PR #1146, opened 2021-05-26
(1737) Add sdgs_apply to Activity export

pezholio:
We don't currently export the SDG applicability in the CSV export, just the 3 SDG value columns. This adds a presenter method that returns `Yes` if sdgs_apply is true, or `No` if it is false. We then use this presenter method to add a new column to the CSV export.


pezholio:
[Add "sdgs_apply" to CSV export](https://trello.com/c/Ays7IVrf/1737-add-sdgsapply-to-csv-export)

robbiepaul (approved):
LGTM 🚀

PR #1149, opened 2021-05-27
(17993) Move Aid type question

mec:
## Changes in this PR

We will soon be adding logic for questions who's responses can be
calculated based on the answer to this question.

One of those dependant questions is Collaboration type, so the aid type
question must come before it.

As a first step to enabling that work, we move the Aid type question up
the order to before the Collaboration type question.


pezholio (approved):

pezholio (approved):
LGTM 👍

farrahfa:
[Move Aid type question](https://trello.com/c/xedq7lDc/1793-move-aid-type-question)

PR #1151, opened 2021-05-27
(1840) Improvements to single report view

mec:
## Changes in this PR
Previously, we used the report description to derive the title shown on a single report view.

So far we have only ever seen a single report per quarter and the description is designed to be used when more than one report occurs in the same quarter.

The descriptions of the reports we have seen have generally been `[financial quarter] [fund] report` which made the title repetative and unclear.

The report description was also required which meant a user had to fill it in before being able to activate a report. This PR introduces two changes to ease this:

- the single report page now creates the title from two known values so it can be rendered without the user providing a descripiton formalising the report page title to `[financial quarter] [fund name] report` and showing the description in a summary (see below)
- the report description has been made optional, which reduces the number of steps to activate a report.

When viewing a single report some context was lost as users could no longer see some important information, we were showing the report state in a summary, so this work extends that to include other useful information:

- report deadline
- report organisation (for BEIS users)
- report description, if one is available
- whether the report can be changed

This last value has been added to aid new users who may not understand the link between the report states and when the contents of the report can be changed, we felt this was a simple, unobtrusive way to achieve this.

More work is planned to further improve the report experience for users.

## Screenshots of UI changes

Delivery partner:

Screenshot 2021-05-27 at 13-55-28 FQ1 2021-2022 A longish description like this so we can see how this looks variance — Rep

BEIS user:

Screenshot 2021-05-27 at 13-55-58 FQ1 2021-2022 A longish description like this so we can see how this looks variance — Rep

### DIFF
--- a/spec/features/staff/users_can_view_reports_spec.rb
+++ b/spec/features/staff/users_can_view_reports_spec.rb
@@ -63,6 +63,31 @@ RSpec.feature "Users can view reports" do
       expect(page).to have_content report.fund.source_fund.name
     end
 
+    scenario "the report includes a summary" do
+      report = create(:report, :active, organisation: build(:delivery_partner_organisation))
+
+      visit reports_path
+
+      within "##{report.id}" do
+        click_on t("default.link.show")
+      end
+
+      expect(page).to have_content report.description
+      expect(page).to have_content l(report.deadline)
+      expect(page).not_to have_content t("page_content.report.summary.editable.#{report.editable?}")
+      expect(page).to have_content report.organisation.name
+    end
+
+    context "when there is no report descripiton" do
+      scenario "the summary does not include the empty value" do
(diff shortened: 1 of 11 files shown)


spec/features/staff/users_can_view_reports_spec.rb
```diff
@@ -59,7 +59,33 @@
...
+        click_on t("default.link.show")
+      end
+
       expect(page).to have_content report.description
+      expect(page).to have_content l(report.deadline)
+      expect(page).not_to have_content t("page_content.report.summary.editable.#{report.editable?}")
+      expect(page).to have_content report.organisation.name
+    end
+
+    context "when there is no reporti descripiton" do
```
pezholio, line 81:
```suggestion
     context "when there is no report descripiton" do
```

pezholio (approved):
LGTM - just one typo, then it's good to go!

PR #1153, opened 2021-05-28
(1767) Verify and improve fstc_applies question logic and experience

pezholio:
This updates the logic in the `fstc_applies` question step to set it automatically in most cases based on the the Aid Type. Before we were only setting the value of `fstc_applies` in one particular case, now we can set it in virtually all cases, except in a couple where we need to ask.

We do this by creating a new BEIS Aid Type codelist which has a `ftsc_applies` value in the YAML that tells us whether to set it to true or false. If the value is not true or false, we ask.

There is also a data migration that we'll need to run that will set historical activities to the value we expect. BEIS are aware and will tell DPs this is happening. We'll also need to tell BEIS when this is done.

# Screenshots

image

### DIFF
--- a/vendor/data/codelists/BEIS/aid_type.yml
+++ b/vendor/data/codelists/BEIS/aid_type.yml
@@ -1,0 +1,96 @@
+---
+attributes:
+  category-codelist: AidType-category
+  name: AidType
+  complete: "1"
+data:
+  - category: B
+    code: B02
+    name: Core contributions to multilateral institutions
+    description:
+      These funds are classified as multilateral ODA (all other categories
+      fall under bilateral ODA). The recipient multilateral institution pools contributions
+      so that they lose their identity and become an integral part of its financial
+      assets. See Annex 2 of the DAC Directives for a comprehensive list of agencies
+      core contributions to which may be reported under B02 (Section I. Multilateral
+      institutions).
+    ftsc_applies: false
+  - category: B
+    code: B03
+    name:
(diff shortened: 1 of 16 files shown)


pezholio:
[Automate the RODA identifier](https://trello.com/c/JwntB6Fz/1724-automate-the-roda-identifier)

pezholio:
[Verify and improve FSTC applies question logic and experience](https://trello.com/c/xJZATSqb/1767-verify-and-improve-fstc-applies-question-logic-and-experience)

vendor/data/codelists/BEIS/aid_type.yml
```diff
@@ -0,0 +1,96 @@
+---
```
jcoglan, line 1: Does this file duplicate content in `vendor/data/codelists/IATI/2_03/activity/aid_type.yml`? If so, how will future developers know how to keep these two copies up to date? Or is there some way we can share content between them?

pezholio: The way we usually work when adapting codelists from IATI is copying the IATI codes and making them BEIS codelists. Part of the aim of this piece of work was to better codify some of the rules, rather than having to override certain behaviours with constants etc. I think any change that is made to any codelists in future will be done slowly and deliberately, so there's minimal risk of this getting missed / confused.

jcoglan: Okay, that makes sense 👍

jcoglan (approved):

PR #1155, opened 2021-05-28
CSV export of all actual spend

jcoglan:
## Changes in this PR

This adds the ability for BEIS users to download all the transactions for each delivery partner as a CSV file that lists all the partner's activities with their total spending by quarter.

The main points of note here are that we probably want to iterate the design here -- do we want this split up by delivery partner, or should we have one big download? Do we need it more granular and split up by fund, etc.

The permissions here are very coarse, I've just limited this to only BEIS users. I'd expect this to change as iterating the design would probably mean changing the routes/resources/controllers we have here so I didn't want to get too into modelling permissions over those until we've a better idea of what we really want.

## Screenshots of UI changes

All these changes are only visible to BEIS users. A new item is added to the top-level navigation:

Screenshot 2021-05-28 at 16 26 15

This links to a list of all delivery partner organisations:

Screenshot 2021-05-28 at 16 26 39

Each organisation links through to a list of exports for that organisation:

Screenshot 2021-05-28 at 16 26 58

Clicking "Download" provides the user with a CSV file containing all the transaction totals per quarter for all that organisation's activities. All activities are included even if they have no spend in any of the listed quarters. The set of quarter columns covers every quarter from the earliest to the most recent transaction in the set.

Screenshot 2021-05-28 at 16 27 24

## Next steps

### DIFF
--- a/script/test
+++ b/script/test
@@ -28,6 +28,9 @@ else
   echo "==> Linting our shellscripts"
   bundle exec rake shellcheck
 
+  echo "==> Running Brakeman"
+  bundle exec brakeman
+
   echo "==> Running the tests..."
 
   if [ -n "$CI" ]; then
@@ -35,7 +38,4 @@ else
   else
     bundle exec rspec
   fi
-
-  echo "==> Running Brakeman"
-  bundle exec brakeman
 fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -670,6 +670,7 @@
 ## [unreleased]
 
 - add spacing between view and edit links on Organisation index page
+- export complete transaction history per delivery partner for BEIS users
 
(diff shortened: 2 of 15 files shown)


jcoglan:
[BEIS users can download a CSV of 'all' actual spend](https://trello.com/c/M595TPrR/1802-beis-users-can-download-a-csv-of-all-actual-spend)

script/test
```diff
@@ -28,14 +28,14 @@ else
   echo "==> Linting our shellscripts"
   bundle exec rake shellcheck
 
+  echo "==> Running Brakeman"
+  bundle exec brakeman
```
pezholio, line 32: 👍

pezholio (approved):
LGTM 👍

PR #1169, opened 2021-06-03
(1790) Collaboration type logic

jcoglan:
## Changes in this PR

If `aid_type` is set to `B02` or `B03` then we infer the value of `collaboration_type` and do not let the user change it.

I based this on an earlier pattern that @CristinaRO implemented for inferring the FSTC value, but this has recently been refactored in `CodelistHelper`. I have got this working on top of those recent changes but wonder if I should also refactor what I've done here to match the new pattern done by @pezholio.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -684,6 +684,7 @@
 
 - Users can no longer update the delivery partner identifier
 - Rename `Transfer` to `OutgoingTransfer`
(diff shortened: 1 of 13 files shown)


jcoglan:
[Collaboration type logic](https://trello.com/c/bID1KzFm/1790-collaboration-type-logic)

CristinaRO (approved):

PR #1170, opened 2021-06-04
Rename planned disbursements to forecasts

CristinaRO:
## Changes in this PR
- Rename `planned_disbursements` to `forecasts` in application code (not a user-facing change)

The duplicate terminology is creating confusion. “Planned disbursement” originates from IATI, but the users (mainly BEIS and delivery partners) think of them as “forecasts”.

A lot of the initial development of RODA was based on IATI reporting needs, which is how we ended up with IATI terminology.

We want to prioritise the users’ domain language, and we value consistency to make it easier for the delivery team to discuss these concepts, so we’ve decided to use “forecasts” everywhere except within the XML files generated for IATI reporting.

We as the delivery team still need to be aware of the parallel terminology, but we're hoping to decrease confusion during onboarding as well as during day to day work.

## Next steps


CristinaRO:
[Rename `planned_disbursements` to `forecasts`](https://trello.com/c/71pbI2HR/1881-rename-planneddisbursements-to-forecasts)

mec (dismissed):
@CristinaRO Thank you for taking this on! 👍 

I am happy to dig into this PR but if we are going to the effort I think we should change it all, a search in the code for `disbursement` shows a number of places where we are talking about 'planned disbursments' still and probably should not be?

-  in the specs, some `describe` and `it` blocks
- in the locale the is still and `edit_noun` for the accesible links
- and my favorite: the `disbursement_plannable` concern in the routes, which I assume is now `forecastable`?

CristinaRO:
> @CristinaRO Thank you for taking this on! 👍
> 
> I am happy to dig into this PR but if we are going to the effort I think we should change it all, a search in the code for `disbursement` shows a number of places where we are talking about 'planned disbursments' still and probably should not be?
> 
> * in the specs, some `describe` and `it` blocks
> * in the locale the is still and `edit_noun` for the accesible links
> * and my favorite: the `disbursement_plannable` concern in the routes, which I assume is now `forecastable`?

Good catch! Absolutely worth taking the time.

mec:
I am on on holiday until Monday now so someone else will need to pick this review up please, sorry and thank you! 🙏

app/views/staff/shared/xml/_forecast.xml.haml
```diff
@@ -0,0 +1,7 @@
+%planned-disbursement{"type" => forecast.forecast_type}
+  %period-start{"iso-date" => forecast.period_start_date}
+  - if forecast.period_end_date.present?
+    %period-end{"iso-date" => forecast.period_end_date}
+  %value{"currency" => forecast.currency, "value-date" => forecast.period_start_date}= forecast.value
+  %provider-org{"provider-activity-id" => "", "type" => forecast.providing_organisation_type, "ref" => forecast.providing_organisation_reference}
+    %narrative=forecast.providing_organisation_name
```
pezholio, line 7: I have no massive opinions either way, but I wonder if this partial _should_ be called `planned_disbursement`, as that's what we're returning in the XML

pezholio (approved):
Thanks so much for doing this. Just one question, but I'm happy for this to go in as is

app/views/staff/shared/xml/_forecast.xml.haml
CristinaRO: Possibly? :scratches head:

PR #1176, opened 2021-06-09
(1887) Unify dependent fields logic

jcoglan:
## Changes in this PR

This PR introduces a new service named `FieldInference` whose job is to encode and execute rules about dependent field values. I decided to write this after observing the following two stories:

- Story 1: setting `aid_type` to certain values lets us infer the value of `collaboration_type`
- Story 2: setting `collaboration_type` to certain values lets us infer `channel_of_delivery_code`

Plus we already had functionality where `fstc_applies` is determined from `aid_type`.

I picked up story 1 and implemented it. This required changing the following:

- `Activity::Updater` to possibly update `collaboration_type` when `aid_type` is changed
- `Staff::ActivityFormsController` to skip the `collaboration_type` step if it's not editable
- Hide the link to the `collaboration_type` step

I then went to start story 2 and immediately saw a problem. If `channel_of_delivery_code` can be changed by setting `collaboration_type`, then it can also be changed by setting `aid_type`. But someone reading just story 2 would have no idea that `aid_type` is involved unless they have a complete understanding of the inference rules. Scattering the logic for these rules across the controller will make it very easy for us to introduce bugs.

`FieldInference` aims to address this problem by having a single component that holds all the inference rules and ensures that:

- When a field is updated, _all_ resulting changes are applied; if you change `aid_type` then `collaboration_type` and `channel_of_delivery_code` are also updated.
- We can reliably check whether any field is editable by the user, based on the values of all its other properties and the complete set of rules.
- If multiple rules produce different values for the same field, or result in a field having no available allowed values, an exception is thrown so the change does not get stored.

The individual commits give examples of using this service. The way I picture this being used is that when a new requirement comes in:

- We update the codelists to store the dependent values, as we're already doing with `fstc_applies` and `collaboration_type` in the `aid_type` codelist
- We update `Activity::Inference` to load that codelist data into `FieldInference` rules
- Any field that could cause dependent fields to change is updated using this service's `assign()` method
- Steps are skipped and edit links are hidden using the `editable?` method
- Multiple-choice options are determined using the `allowed_values` method

This still requires us to use these methods anywhere they might be applicable, but it avoids the problem above where you need to know all the fields that might directly or indirectly affect the field you're working on. You know that if `assign()` is being used then all applicable changes are being made and any conflicts detected. You know that if `editable?` is being used then we're checking the whole state of the object and you don't need to remember individual inference rules in views and helpers.

I have a couple of open questions about this system that I've not implemented in this PR:

- What should happen when we have a rule like `on(:x, 1).restrict(:y, values)` and the current value of `y` is not in `values`? Should we set `y` to `nil`, throw an error, something else?
- If we have rule like `on(:x, 1).fix(:y, value)` then setting `x` to `1` will set `y` to `value`. If we then change `x` to something else, `y` will remain set to `value`. What should happen to dependent fields when the upstream fields is reassigned?

These don't need answering immediately but are worth thinking about as we continue to implement these rules. We might also want to think about whether inferred values are stored in the database, or are instead stored as `nil` and their inferred values returned by application logic.

## Next steps


jcoglan:
[Unify "dependent fields" logic](https://trello.com/c/XNkesM6Y/1887-unify-dependent-fields-logic)

pezholio (approved):
This is great stuff, really well explained and I can (mainly) follow the logic. This will really stand us in good stead for some of the other changes that are coming down the pipe. Thanks for spending the time to do this 👍

jcoglan:
Thanks @pezholio :)

PR #1178, opened 2021-06-10
(1889) Create a script to prepare the dev environment for a data import

pezholio:
As part of the import process for newly onboarded DPs, we need to test the data in development. 

As part of this process, to make sure our local  environment mirrors that of  the production environment, we need to take  a cut of the production data  and import it to our local environment. 

As well as this, we need to make sure  that we have a user for the DP we are  testing with, as well as an active  report. As we have to do this a number of  times for each DP, this needs to be done a lot, so I've added a script that  allows us to automate this, to make the testing process less painful.

### DIFF
--- a/script/prepare_for_import.rb
+++ b/script/prepare_for_import.rb
@@ -1,0 +1,37 @@
+require "optparse"
+require_relative "../config/environment"
+
+parser = OptionParser.new { |args|
+  args.on "-f", "--fund FUND"
+  args.on "-o", "--organisation ORGANISATION"
+  args.on "-u", "--user EMAIL"
+}
+
+options = {}
+parser.parse!(into: options)
+
+user_email = options.fetch(:user, "roda+dp@dxw.com")
+
+fund = Activity.fund.by_roda_identifier(options.fetch(:fund))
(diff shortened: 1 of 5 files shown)


pezholio:
[Create a script to prepare the dev environment for a data import](https://trello.com/c/h2JGotli/1889-create-a-script-to-prepare-the-dev-environment-for-a-data-import)

script/prepare_for_import.rb
```diff
@@ -0,0 +1,37 @@
...
+
+unless user
+  warn "Could not find organisation with email '#{user_email}'"
+  exit 1
+end
+
+user.organisation = organisation
+user.save
+
+Report.create(fund: fund, organisation: organisation, state: "active")
```
CristinaRO, line 37: It might be worth making sure that this works if invoking the preparation script repeatedly without the full db restore. (But it might not, if the script is only ever intended to be run after a db-restore.)

OTOH, it's the last line, so if it errors because the report already exists, it's not a big deal.

CristinaRO (approved):
Really good to have this automated, to reduce the pain of testing!

Like you said, it's worth documenting the process, even if we won't use it beyond the finite number of DPs to be onboarded, but it's not a blocker.

PR #1179, opened 2021-06-10
(1789) Channel of delivery logic

jcoglan:
## Changes in this PR

This PR implements the following rules:

- If `collaboration_type` is `1` then `channel_of_delivery_code` can be any of the values allowed by BEIS
- If `collaboration_type` is `2` then `channel_of_delivery_code` must be `40000`
- If `collaboration_type` is `3` then `channel_of_delivery_code` may be either `20000` or `51000`

## Screenshots of UI changes

Say the channel of delivery (COD) code is initially `11000`:

Screenshot 2021-06-10 at 16 13 44

If we set collaboration type to `3`:

Screenshot 2021-06-10 at 16 14 03

Then the existing COD code is removed, because the current value of `11000` is not one of the allowed values of `20000` or `51000`:

Screenshot 2021-06-10 at 16 14 45

Visiting the COD step presents only the valid answers as options based on the current collaboration type:

Screenshot 2021-06-10 at 16 15 00

If the collaboration type is set to `2`:

Screenshot 2021-06-10 at 16 15 35

Then the COD code is set to `40000` as this is the only allowed value in that case.

Screenshot 2021-06-10 at 16 16 01

This also deals with the cascading effect of setting the aid type. If we reset everything and make the COD code `11001`:

Screenshot 2021-06-10 at 16 17 09

And then set the aid type to `B02`:

Screenshot 2021-06-10 at 16 17 56

This causes the collaboration type to be set to `2`:

Screenshot 2021-06-10 at 16 18 41

And that in turn locks the COD code to the `40000` value.

Screenshot 2021-06-10 at 16 18 51

Neither of these fields is editable, so the 'edit' links are hidden and the steps are skipped if anyone follows a link to them.

## Next steps


jcoglan:
[Channel of delivery logic](https://trello.com/c/3bysWOpn/1789-channel-of-delivery-logic)

CristinaRO (approved):
The inference service is brilliant!

PR #1180, opened 2021-06-14
(1528) Activities tab

CristinaRO:
## Changes in this PR
- Show a dedicated tab for activities under a report

#### Activities tab
Screenshot 2021-06-14 at 15 03 57

#### Post-upload with errors
Screenshot 2021-06-14 at 15 04 35

#### Post-upload with success
Screenshot 2021-06-14 at 15 44 43

## Next steps


CristinaRO:
[2nd Iteration: Improvements to the activity upload](https://trello.com/c/NcvoRNwi/1528-2nd-iteration-improvements-to-the-activity-upload)

app/views/staff/reports/activities.html.haml
```diff
@@ -23,4 +23,11 @@
             For guidance on adding or updating activities, see the
             = link_to_new_tab "guidance in the help centre", "https://beisodahelp.zendesk.com/hc/en-gb/articles/1500005510061-Understanding-the-Bulk-Upload-Functionality-to-Report-your-Data"
 
+          - if policy(@report_presenter).upload?
+            %p.govuk-body
+              Large numbers of activities can be added or updated via the activities upload.
+
+            %p.govuk-body
+              = link_to t("action.activity.upload.link"), new_report_activity_upload_path(@report_presenter), class: "govuk-link"
```
pezholio, line 31: I think it might look cleaner to make the link to the upload part of the body text:

```suggestion
             Large numbers of activities can be added or updated via the #{link_to("activities upload", new_report_activity_upload_path(@report_presenter), class: "govuk-link")}.
```

pezholio (approved):
LGTM - just one teeny suggestion, but given we're going to be iterating on this, probably not neccessary

app/views/staff/reports/activities.html.haml
CristinaRO: I do agree, but my implementation already deviates from the suggestion to keep the button styling. (I think Robbie mentioned that if we follow the design standard, we shouldn't be showing links as buttons.)

Some of the proposed designs have both Download and Upload buttons on this tab, so yeah, I think we'll be iterating a bit to find what works best!

PR #1181, opened 2021-06-14
(1835) Transfers store historic BEIS identifier

CristinaRO:
## Changes in this PR
- Incoming and outgoing transfers can store the historic BEIS identifier (tracker row ID)

### After
Screenshot 2021-06-14 at 17 28 01
Screenshot 2021-06-14 at 18 08 07

## Next steps

### DIFF
--- a/app/views/staff/shared/transfers/_index.html.haml
+++ b/app/views/staff/shared/transfers/_index.html.haml
@@ -28,6 +28,8 @@
             %th.govuk-table__header{scope: "col"}
               =t("fields.#{transfer_type}.transfer_amount")
             %th.govuk-table__header{scope: "col"}
+              =t("fields.#{transfer_type}.beis_identifier")
+            %th.govuk-table__header{scope: "col"}
         %tbody.govuk-table__body
           - transfers.each do |transfer|
             %tr.govuk-table__row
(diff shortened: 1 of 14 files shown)


CristinaRO:
[Incoming and outgoing transfers store the historic BEIS Identifier](https://trello.com/c/jI0srEm3/1835-incoming-and-outgoing-transfers-store-the-historic-beis-identifier)

app/views/staff/shared/transfers/_index.html.haml
```diff
@@ -39,6 +41,7 @@
                 %td.govuk-table__cell= transfer.source.organisation.name
                 %td.govuk-table__cell= transfer.source.roda_identifier
               %td.govuk-table__cell= transfer.value
+              %td.govuk-table__cell= transfer.beis_identifier
```
pezholio, line 44: I'm not sure of the value of showing this in the table, given it's an optional field, and only useful for historic transfers. I may be wrong though (I frequently am!)

CristinaRO: The reason I added the historic identifier to the table is that there is no individual transfer show page, so the only other way a user could remind themself what they entered would be to go into edit mode, which might not be intuitive to them.

pezholio: Oh course!

pezholio (approved):
LGTM 👍

PR #1187, opened 2021-06-16
(1895) Hide non-DP orgs on the Activities dropdown

pezholio:
As only Delivery Partners have activities assoiciated with them, we don't want to see any non-Delivery Partners to be able to be selected when filtering Activites by organisation.

### Before

image

### After

image

### DIFF
--- a/app/views/staff/shared/activities/_filter.html.haml
+++ b/app/views/staff/shared/activities/_filter.html.haml
@@ -6,6 +6,6 @@
       .govuk-form-group
         %label{ class: "govuk-label", for: :organistaion_id }
           = t("filters.activity.organisation")
-        = select_tag :organisation_id, options_from_collection_for_select(Organisation.all, :id, :name, organisation_id), { class: "govuk-select" }
+        = select_tag :organisation_id, options_from_collection_for_select(list_of_delivery_partners, :id, :name, organisation_id), { class: "govuk-select" }
       .govuk-form-group
         = submit_tag t("filters.activity.submit"), { class: "govuk-button", data: { module: "govuk-button" } }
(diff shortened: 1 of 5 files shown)


pezholio:
[Activity listing page should only have Delivery Partner organisations in the drop downs](https://trello.com/c/Mjpgp4zh/1895-activity-listing-page-should-only-have-delivery-partner-organisations-in-the-drop-downs)

app/views/staff/shared/activities/_filter.html.haml
```diff
@@ -6,6 +6,6 @@
       .govuk-form-group
         %label{ class: "govuk-label", for: :organistaion_id }
           = t("filters.activity.organisation")
-        = select_tag :organisation_id, options_from_collection_for_select(Organisation.all, :id, :name, organisation_id), { class: "govuk-select" }
+        = select_tag :organisation_id, options_from_collection_for_select(Organisation.delivery_partners.order(:name), :id, :name, organisation_id), { class: "govuk-select" }
```
CristinaRO, line 9: Do we have access to the form helper, to use `list_of_delivery_partners` here? If not, is it worth consolidating the list of delivery partners to the application helper, so we don't scatter the definition around? Are there contexts where we *don't* want it sorted by name?

CristinaRO (approved):
It can go in as-is, but I'd like us to consider the reusability of `list_of_delivery_partners`. :)

app/views/staff/shared/activities/_filter.html.haml
pezholio: Good call - I think ordering it by name aids findability in the other case too, so I'll add the order filter in the helper too

PR #1188, opened 2021-06-16
Update import documentation

pezholio:
We now require Reports to be present when importing Activity data for levels C and D, so I've moved the Reports documentation higher up, as well as tweaking some of the wording to be more generic and applicable to any time we do onboarding, not just Q3 2020


mec (approved):
🚢

PR #1189, opened 2021-06-16
Feature/1529 improvements to the reporting forecast journey

edavey:
## Trello
https://trello.com/c/4Km0NgOB/1529-improvements-to-the-reporting-forecast-journey

## Changes in this PR

This PR adds a new 'Forecasts' tab to the 'Report' view. On the new tab:

- the "Upload forecasts" button is on the new tab rather than in the 'actions' partial which is common to all the tabs
- we show a table of forecasts associated with the report, grouped by activity

This also adds to the forecast upload journey, so that the "upload success" page:

- uses the same table of forecasts to list the successfully uploaded forecasts. 
- includes the total amount forecast
- includes a button linking back to the new "report forecasts" path

### Before

before_heading_no_tab

### After

after_heading_with_tab

test_fixtures

Upload success page now uses the same table of forecasts grouped by activity:

successfull_uploads

 
## To do / verify

From Trello: 

- ~the total of all forecasts in the table is shown~
- ~successfully uploading forecasts shows a list of the forecasts added in that upload~ 
- ~the success page has a link back to the report~

## Next steps

### DIFF
--- a/app/controllers/staff/report_forecasts_controller.rb
+++ b/app/controllers/staff/report_forecasts_controller.rb
@@ -9,11 +9,23 @@ class Staff::ReportForecastsController < Staff::BaseController
 
     @report_presenter = ReportPresenter.new(@report)
     @report_activities = @report.reportable_activities
-    @grouped_forecasts = ForecastOverview
-      .new(@report_activities.map(&:id))
-      .latest_values
+    @total_forecast = formatted(forecasts.sum(&:value))
+    @grouped_forecasts = forecasts
+      .map { |forecast| ForecastPresenter.new(forecast) }
       .group_by { |forecast| forecast.parent_activity_id }
 
     render "staff/reports/forecasts"
+  end
+
+  private
+
+  def forecasts
+    ForecastOverview
+      .new(@report_activities.map(&:id))
+      .latest_values
+  end
+
+  def formatted(bigdecimal)
+    ActionController::Base.helpers.number_to_currency(bigdecimal, unit: "£")
   end
 end
--- a/app/controllers/staff/report_forecasts_controller.rb
+++ b/app/controllers/staff/report_forecasts_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class Staff::ReportForecastsController < Staff::BaseController
+  include Secured
+
+  def show
+    @report = Report.find(params["report_id"])
+    authorize @report
+
+    @report_presenter = ReportPresenter.new(@report)
+
+    render "staff/reports/forecasts"
+  end
+end
--- a/app/controllers/staff/report_forecasts_controller.rb
+++ b/app/controllers/staff/report_forecasts_controller.rb
@@ -11,8 +11,9 @@ class Staff::ReportForecastsController < Staff::BaseController
     @report_activities = @report.reportable_activities
     @total_forecast = formatted(forecasts.sum(&:value))
     @grouped_forecasts = forecasts
+      .includes([:parent_activity])
       .map { |forecast| ForecastPresenter.new(forecast) }
-      .group_by { |forecast| forecast.parent_activity_id }
+      .group_by { |forecast| ActivityPresenter.new(forecast.parent_activity) }
 
     render "staff/reports/forecasts"
   end
--- a/app/controllers/staff/report_forecasts_controller.rb
+++ b/app/controllers/staff/report_forecasts_controller.rb
@@ -9,7 +9,7 @@ class Staff::ReportForecastsController < Staff::BaseController
 
     @report_presenter = ReportPresenter.new(@report)
     @report_activities = @report.reportable_activities
-    @total_forecast = formatted(forecasts.sum(&:value))
+    @total_forecast = TotalPresenter.new(forecasts.sum(&:value)).value
     @grouped_forecasts = forecasts
       .includes([:parent_activity])
       .map { |forecast| ForecastPresenter.new(forecast) }
@@ -24,9 +24,5 @@ class Staff::ReportForecastsController < Staff::BaseController
     ForecastOverview
       .new(@report_activities.map(&:id))
       .latest_values
-  end
-
-  def formatted(bigdecimal)
-    ActionController::Base.helpers.number_to_currency(bigdecimal, unit: "£")
   end
 end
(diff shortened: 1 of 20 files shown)


edavey:
[Improvements to the reporting forecast journey](https://trello.com/c/4Km0NgOB/1529-improvements-to-the-reporting-forecast-journey)

CristinaRO:
[Editing Policy markers doesn't show existing selections](https://trello.com/c/TEjkn11h/1900-editing-policy-markers-doesnt-show-existing-selections)

config/locales/models/report.en.yml
```diff
@@ -31,11 +31,16 @@ en:
...
+      forecasts: Forecasts
       variance: Variance
       budgets: Budgets
+
   page_content:
     reports:
       title: Reports
+    tab_content:
+      forecasts:
+        heading: Forecast in this report
```
pezholio, line 43:
```suggestion
        heading: Forecasts in this report
```

spec/features/staff/users_can_view_forecasts_within_report_spec.rb
```diff
@@ -7,8 +7,27 @@
...
     scenario "the report contains a _forecasts_ tab" do
-      report = create(:report, :active, organisation: organisation, description: nil)
+      programme = create(:programme_activity)
+
+      report = create(:report, :active, organisation: organisation, fund: programme.parent)
+      project = create(:project_activity, organisation: organisation, parent: programme)
+
+      activities = 2.times.map {
+        create(:third_party_project_activity, organisation: organisation, parent: project)
+      }
```
pezholio: You can use `create_list` for this, i.e: 

```suggestion
      activities = create_list(:third_party_project_activity, 2, organisation: organisation, parent: project)
```

app/views/staff/reports/forecasts.html.haml
```diff
@@ -21,3 +21,22 @@
...
+          %table.govuk-table
+            %thead.govuk-table__head
+              %tr.govuk-table__row
+                %th.govuk-table__header{scope: "col"} Activity
+                %th.govuk-table__header{scope: "col"} RODA Identifier
+                %th.govuk-table__header{scope: "col"} Financial Quarter
+                %th.govuk-table__header{scope: "col"} Amount
+            %tbody.govuk-table__body
+              - @report_activities.each do |activity|
+                - activity_presenter = ActivityPresenter.new(activity)
```
pezholio: Personally, I'd prefer to set up the presenters in the controller here, to avoid too much logic in the view

spec/features/staff/users_can_view_forecasts_within_report_spec.rb
```diff
@@ -7,8 +7,27 @@
       authenticate!(user: user)
     end
 
+    def expect_to_see_a_table_of_forecasts_grouped_by_activity(activities)
```
pezholio, line 10: 👍

app/views/staff/reports/forecasts.html.haml
```diff
@@ -32,11 +32,21 @@
...
                 - activity_presenter = ActivityPresenter.new(activity)
+
                 %tr.govuk-table__row[activity]
                   %td.govuk-table__cell= activity_presenter.title
                   %td.govuk-table__cell= activity_presenter.roda_identifier
-                  %td.govuk-table__cell
-                  %td.govuk-table__cell
+                  %td.govuk-table__cell{ colspan: 2 }
+                    %table.govuk-table.forecasts
+                      - forecasts.map {|f| ForecastPresenter.new(f) }.each do |forecast|
```
pezholio: Again, if there's a way we can set up these presenters in the controller (or in a helper) to keep the view logic cleaneer, that would be great

app/controllers/staff/report_forecasts_controller.rb
```diff
@@ -9,11 +9,23 @@ def show
...
+
+  def forecasts
+    ForecastOverview
+      .new(@report_activities.map(&:id))
+      .latest_values
+  end
+
+  def formatted(bigdecimal)
+    ActionController::Base.helpers.number_to_currency(bigdecimal, unit: "£")
+  end
```
pezholio, line 31: This feels like it might belong in a view helper, rather than the controller

pezholio (changes requested):
Looks good overall, just one typo and a couple of minor style things. Feel free to push back though! 👍

spec/features/staff/users_can_view_forecasts_within_report_spec.rb
edavey: I'm not convinced that using `create_list` is an improvements:

app/views/staff/reports/forecasts.html.haml
edavey: 👍 thks for this

config/locales/models/report.en.yml
edavey: thanks, I think this makes sense, even though it deviates from Meyric's mockup.

We now show:

- tab name: "Forecasts", consistent with "Activities" and "Budgets" also plural
- page title: "Forecasts in this report" i.e. here is a list of the forecasts associated with this report
- heading of table: "Forecast spend per activity" - forecast is an adjective here, qualifying the noun "spend"

app/views/staff/reports/forecasts.html.haml
edavey: 👍

app/controllers/staff/report_forecasts_controller.rb
edavey: Hmm, I'm not sure about this -- as at present the formatting of numbers is done in Presenters instantiated in Controllers,  rather than in Views via Helpers.

The problem is that there is no presenter for a "total" so we sum up the `Forecast#value` amounts before the forecasts are decorated by presenters and then we need to format the total. Maybe the better answer is a `TotalPresenter`?

edavey:
@pezholio I think this is ready for another look, thks

config/locales/models/report.en.yml
pezholio: Cool - that sounds fine 👍

spec/features/staff/users_can_view_forecasts_within_report_spec.rb
pezholio: Ah, gotcha. I made this comment before all the extra bits were added. I think this is fine in that case 👍

app/controllers/staff/report_forecasts_controller.rb
pezholio: Oooh, yes. Very nice 👍

pezholio (approved):
LGTM - thanks for making those changes 👍

PR #1193, opened 2021-06-21
(1900) Editing policy markers shows and preserves existing selections

CristinaRO:
## Changes in this PR
- Bugfix: existing policy marker values are correctly shown when editing an activity
- Invalid policy marker values are highlighted when importing an activity from a CSV
- Policy marker add/edit links go directly to the relevant form section

### After
Screenshot 2021-06-21 at 15 33 09

## Next steps


CristinaRO:
[Editing Policy markers doesn't show existing selections](https://trello.com/c/TEjkn11h/1900-editing-policy-markers-doesnt-show-existing-selections)

spec/features/staff/users_can_edit_an_activity_spec.rb
```diff
@@ -413,6 +413,42 @@
...
+          expect(page).to have_current_path(
+            activity_step_path(activity, :policy_markers)
+          )
+        end
+
+        choose "activity-policy-marker-gender-significant-objective-field"
+        click_button t("form.button.activity.submit")
+
+        activity.reload
+        expect(activity.policy_marker_gender).to eql("significant_objective")
```
edavey, line 435: These feature specs seem to highlight the problem behaviour perfectly and should prevent regressions.

Just a small suggestion here: as this is an integration test I think it would be ideal to keep our assertions to what the user sees on screen. i.e. rather than reloading the activity and making an assertion on value of the `#policy_market_gender` attribute, we could assert that we see the expected "Significant objective" label on the screen, e.g. with

```ruby
expect(page).to have_css(".policy_marker_gender dd", text: "Significant objective")
```

What do you think?

spec/features/staff/users_can_edit_an_activity_spec.rb
```diff
@@ -413,6 +413,42 @@
...
+        _report = create(:report, state: :active, organisation: user.organisation, fund: activity.associated_fund)
+
+        visit activity_step_path(activity, :policy_markers)
+
+        expect(page.find(:css, "#activity-policy-marker-desertification-significant-objective-field")).to be_checked
+
+        click_button t("form.button.activity.submit")
+        activity.reload
+
+        expect(activity.policy_marker_desertification).to eql("significant_objective")
```
edavey, line 450: as above, maybe:
```
        expect(page).to have_css(
          ".policy_marker_desertification dd", 
          text: "Significant objective"
        )
```
would be a better assertion here?

app/views/staff/shared/activities/_activity.html.haml
```diff
@@ -395,7 +395,7 @@
         = activity_presenter.policy_marker_nutrition
       %dd.govuk-summary-list__actions
         - if policy(activity_presenter).update? && step_is_complete_or_next?(activity: activity_presenter, step: :policy_markers)
-          = a11y_action_link(t("default.link.#{activity_presenter.call_to_action(:policy_marker_nutrition)}"), activity_step_path(activity_presenter, :policy_markers), t("summary.label.activity.policy_marker_nutrition"))
+          = a11y_action_link(t("default.link.#{activity_presenter.call_to_action(:policy_marker_nutrition)}"), activity_step_path(activity_presenter, :policy_markers, anchor: "nutrition"), t("summary.label.activity.policy_marker_nutrition"))
```
edavey, line 398: Do you think that the inclusion of these new `anchor` attributes in the links should be described in a spec to prevent regressions?

edavey (approved):
Nice work, very clear. (Even a newbie like me was able to follow along!)

I've made a couple of small suggestions regarding testing.

spec/features/staff/users_can_edit_an_activity_spec.rb
CristinaRO: I think it's great! I'm just surprised that it works, because the other week I lost half a day trying to debug a `StaleElementReferenceError` when I tried to test both scenarios in one.

Capybara didn't seem to like `page` being re-used after going into editing mode again - that's why I split them into two scenarios in the end. 🤯 

I thought the reference went stale because of executing another request, but it turns out I don't really understand what makes it stale and what doesn't.

spec/features/staff/users_can_edit_an_activity_spec.rb
CristinaRO: I'm not sure about the implementation detail of it being a `dd` - this would make the spec brittle to us changing the DOM  without really changing the meaning, whereas the class can be applied to any element so is a more durable (and semantic) assertion.

app/views/staff/shared/activities/_activity.html.haml
CristinaRO: Possibly - it's not a "core" feature IMO, but if accepted (I went a bit over the scope of the card because it was a quality-of-life improvement for me, and I projected its need onto our users) then yes, we should also try to maintain it.

PR #1196, opened 2021-06-24
Feature (1927) Record changes made to Activities via Wizard

edavey:
## Trello

https://trello.com/c/tt3mt1wQ/1927-record-activity-changes-made-via-wizard

### 1) `HistoryRecorder` decomposes 'changes' into historical events

`Staff::ActivityFormsController` uses new `HistoryRecorder` service to record changes made via wizard:

e.g.

```ruby
HistoryRecorder
  .new(user: current_user)
  .call(
    changes: @activity.changes,
    reference: "Update to Activity purpose",
    activity: @activity
  )
```

### 2) `HistoryRecord` represents 'before' and 'after' state
For each change in the list of `changes` given to `HistoryRecorder`, the recorder creates a `HistoricalEvent` record.

**HistoricalEvent**:

- `activity_id`: uuid
- `value_changed`: the attribute, e.g. "title",
- `new_value`: stored as YAML to handle various data types 
- `previous_value`: stored as YAML to handle various data types 
- `user_id`: UUID (but maybe should be email address or name?),
- `reference`: e.g 'Update to Activity purpose' or 'Upload of file123.csv',
- `created_at`: datetime timestamp

## Screenshots of UI changes

No UI yet

## Next steps

- [x] Migration needs to run as part of deployment to create new `historical_events` table

### DIFF
--- a/app/services/history_recorder.rb
+++ b/app/services/history_recorder.rb
@@ -4,6 +4,19 @@ class HistoryRecorder
   end
 
   def call(changes:, reference:, activity:)
-    # coming soon
+    changes.each do |value_changed, (previous_value, new_value)|
+      HistoricalEvent.create(
+        user: user,
+        activity: activity,
+        reference: reference,
+        value_changed: value_changed,
+        previous_value: previous_value,
+        new_value: new_value,
+      )
+    end
   end
+
+  private
+
+  attr_reader :user
 end
--- a/app/services/history_recorder.rb
+++ b/app/services/history_recorder.rb
@@ -1,0 +1,9 @@
+class HistoryRecorder
+  def initialize(user:)
+    @user = user
+  end
+
+  def call(changes:, reference:, activity:)
+    # coming soon
+  end
+end
--- a/spec/models/historical_event_spec.rb
+++ b/spec/models/historical_event_spec.rb
@@ -5,4 +5,74 @@ RSpec.describe HistoricalEvent, type: :model do
     it { should belong_to(:activity) }
     it { should belong_to(:user) }
   end
+
+  describe "flexible 'value' fields which handle a range of data types" do
+    let(:event) { HistoricalEvent.new }
+
+    context "when value is a String" do
+      let(:persisted_event) do
+        set_value_fields_to("String")
+      end
+
+      it "should read back a String" do
+        expect(persisted_event.new_value).to eq("String")
+        expect(persisted_event.previous_value).to eq("String")
+      end
+    end
+
+    context "when value is an Integer" do
+      let(:persisted_event) do
+        set_value_fields_to(101)
+      end
+
+      it "should read back an Integer" do
+        expect(persisted_event.new_value).to eq(101)
+        expect(persisted_event.previous_value).to eq(101)
+      end
+    end
+
+    context "when value is a boolean" do
+      let(:persisted_event) do
+        set_value_fields_to(true)
+      end
+
+      it "should read back an boolean" do
+        expect(persisted_event.new_value).to be true
+        expect(persisted_event.previous_value).to be true
+      end
+    end
+
+    context "when value is a datetime" do
+      let(:datetime) { Time.current }
+
+      let(:persisted_event) do
+        set_value_fields_to(datetime)
+      end
+
+      it "should read back a datetime" do
+        expect(persisted_event.new_value).to eq(datetime)
+        expect(persisted_event.previous_value).to eq(datetime)
+      end
+    end
+
+    context "when value is a date" do
+      let(:date) { Date.parse("01-Jan-2020") }
+
+      let(:persisted_event) do
+        set_value_fields_to(date)
+      end
+
+      it "should read back a date" do
+        expect(persisted_event.new_value).to eq(date)
+        expect(persisted_event.previous_value).to eq(date)
+      end
+    end
+
+    def set_value_fields_to(value)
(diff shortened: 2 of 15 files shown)


app/services/history_recorder.rb
```diff
@@ -1,9 +1,23 @@
 class HistoryRecorder
-  def initialize(user:)
+  def initialize(user:, history_model: HistoricalEvent)
```
CristinaRO, line 2: Are we ever going to record anything other than a `HistoricalEvent`? IMO *this* is a premature abstraction that adds a layer of indirection for no benefit.

edavey: Good question: I've added `HistoricalEvent` as a default parameter in the initialiser because it's a *dependency* (`HistoryRecorder` depends on `HistoricalEvent`) and I believe it makes sense to inject dependencies where possible. Two advantages:

1. testing is easier and faster. We can assert that a mock of `HistoricalEvent` receives the expected `::create` messages -- rather than by using the ActiveRecord stack and even using the database. I find this approach (mocking everything except the class under test) helps focus on the messages passed between objects. Rather than looking at the effect the class under test has had on a collaborating object.
2. If we do need to change or elaborate the dependency (unlikely in this case I think) then we know where to find it (at the top in the constructor) -- and importantly the tests for `HistoryRecorder` should be unaffected by the change

Does this sell it to you?

edavey:
### A note on the language used. 

I've departed from the suggestions in the Trello ticket (`ChangeLogEntry` and `ChangeLogEntryCreator`). 

The main reason is that we already use the term 'change log' to refer to the ChangeLog™️ system which we use to summarise changes to the application between releases (https://keepachangelog.com/en/1.0.0/). And I note that a lurker in our project Slack channel has already been confused in this way.

I'm convinced of the benefit of agreeing clear and unambiguous language for our code and our discussions of the "problem domain".  (See https://martinfowler.com/bliki/UbiquitousLanguage.html) So in this PR I'm suggesting:

- `HistoricalEvent` - a persisted record of a single change ('before' and 'after' state)
- `HistoryRecorder` - the service responsible to breaking down a group of changes and recording the constituent changes

app/services/history_recorder.rb
CristinaRO: Not really? Testing might be easier, but in my experience it also correlates with making it easier to write tests that still pass even when the functionality got broken. 🤷🏻‍♀️ 

For some reason it also seems to make debugging of live production bugs much harder.

pezholio: Yeah, I'm inclined to agree that making API changes to facilitate easier testing makes me a tad nervous. Could we not do something like:

```
expect(HistoricalEvent).to recieve(:create).with ...
```

spec/models/historical_event_spec.rb
```diff
@@ -5,4 +5,75 @@
...
+        expect(persisted_event.new_value).to eq(date)
+        expect(persisted_event.previous_value).to eq(date)
+      end
+    end
+
+    def set_value_fields_to(value)
+      event.tap do |e|
+        e.new_value = value
+        e.previous_value = value
+        e.save
```
pezholio, line 88: Should we call `reload` after save here, to make sure we've fetched the object from the database again?

pezholio (commented):
Couple of tiny things - looks like a great start overall though, ta! 👍

spec/models/historical_event_spec.rb
edavey: hmm, interesting question! 

On revisiting I see that actually we don't need `e.save` at all. The spec passes without that persistence method call.

The serialisation behaviour we're describing is achieved through assigning and reading back (not sure exactly where!) And if we remove:

```
  serialize :new_value
  serialize :previous_value
```

from `HistoricalActivity`, that's what causes the specs to fail. Persisting or not persisting isn't relevant. 

So the example is clearer without `e.save` (or any persistence) I think?

app/services/history_recorder.rb
edavey: OK, I've moved this dependency down into the class, hopefully that makes this class easier to read.

spec/models/historical_event_spec.rb
pezholio: Out of interest, do the tests pass if we do `e.save` and `e.reload`? If so, I'm happy to do it without `e.save`

app/services/history_recorder.rb
pezholio: Yeah, this is much better, and doesn't sacrifice performance too 👍

spec/models/historical_event_spec.rb
edavey: Yes, if we save the record and reload the serialisation is still in place.

But in order to do `save` and `reload` we'd need to make the subject valid:

```
    let(:event) do
      HistoricalEvent.new(
        user: create(:delivery_partner_user), 
        activity: create(:third_party_project_activity)
      ) 
    end
```

which is more noise in addition to the unnecessary db connections.

We do have an integration test `spec/features/staff/users_can_edit_an_activity_spec.rb` which demonstrates that the units all join together as expected, i.e. we get the expected `HistoryEvent` after editing an Activity using the wizard...

pezholio: Ah, cool, course. That's fine in that case 👍

pezholio (approved):
LGTM 👍

PR #1203, opened 2021-06-29
Feature (1928) Record activity changes made in bulk via CSV upload

edavey:
## Trello card
https://trello.com/c/BDdnYxcb/1928-record-activity-changes-made-in-bulk

## Changes in this PR
When processing bulk imports of activities in CSV form we now record any updates to existing activities using the new `HistoryRecorder` service, which creates a `HistoricalEntry` record for each attribute being updated. 

Those updated attributes are obtained using `Activity#changes` before the activity in question is saved.

## Next steps

### DIFF
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -122,10 +122,10 @@ module Activities
           end
         end
 
-        return set_errors unless @activity.valid?
+        changes = @activity.changes
+        return set_errors unless @activity.save
 
-        record_history
-        @activity.save
+        record_history(changes)
       end
 
       def find_activity_by_roda_id(roda_id)
@@ -137,11 +137,11 @@ module Activities
 
       private
 
-      def record_history
+      def record_history(changes)
         HistoryRecorder
           .new(user: @uploader)
           .call(
-            changes: @activity.changes,
+            changes: changes,
             reference: "Import from CSV",
             activity: @activity
           )
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -122,11 +122,10 @@ module Activities
           end
         end
 
-        return if @activity.save
+        return set_errors unless @activity.valid?
 
-        @activity.errors.each do |error|
-          @errors[error.attribute] ||= [@converter.raw(error.attribute), error.message]
-        end
+        record_history
+        @activity.save
       end
 
       def find_activity_by_roda_id(roda_id)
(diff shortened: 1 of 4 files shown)


edavey:
[Record Activity changes made in *bulk*](https://trello.com/c/BDdnYxcb/1928-record-activity-changes-made-in-bulk)

app/services/activities/import_from_csv.rb
```diff
@@ -122,11 +122,10 @@ def update
           end
         end
 
-        return if @activity.save
+        return set_errors unless @activity.valid?
 
-        @activity.errors.each do |error|
-          @errors[error.attribute] ||= [@converter.raw(error.attribute), error.message]
-        end
+        record_history
```
mec, line 127: Do you think there is any value in recording the history after the save, incase anyting goes wrong during the save and we write a untrue history?

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -924,9 +924,11 @@
 
   context "when updating and importing" do
     let(:activity_policy_double) { instance_double("ActivityPolicy", create_child?: true, update?: true) }
+    let(:history_recorder) { instance_double(HistoryRecorder, call: true) }
```
mec, line 927: I like this, I am always surprised we don't have more doubles.

spec/features/staff/users_can_upload_activities_spec.rb
```diff
@@ -207,6 +226,19 @@
     end
   end
 
+  def expect_change_to_be_recorded_as_historical_event(field:, previous_value:, new_value:, activity:)
+    historical_event = HistoricalEvent.find_by!(value_changed: field)
+
+    aggregate_failures do
```
mec, line 232: First time I have seen this as well! Very nice, I had to look it up! 

This makes rspec run all the expectations in the block even if one fails, is that right?

mec (changes requested):
Lovely work! Thank you! I learnt a bunch!

I have once question about when we recotd the changes, otherwise this is great!

app/services/activities/import_from_csv.rb
edavey: yes, good point!

I think this behaviour is rooted back in the earlier Wizard implementation where there's no clear way in `Staff::ActivityFormsController` to create the History after the Activity has been updated:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/master/app/controllers/staff/activity_forms_controller.rb#L74-L79

as the persistence is performed magically by the Wizard gem.

But in the case of the bulk importer (`Activities::ImportFromCSV::ActivityUpdater`) we **can** handle the creation of history after the activity is updated.

It's worth noting that the 'main'  `Activities::ImportFromCSV#import` method wraps the row-at-a-time import processes in an `ActiveRecord::Base.transaction`, meaning that if there's any exception, then all changes (including the creation of `HistoricalEvent`s) is rolled back.

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/master/app/services/activities/import_from_csv.rb#L32-L42

But yes what you suggest is definitely better -- clearer and safer. I've added a commit to add this behaviour.

spec/features/staff/users_can_upload_activities_spec.rb
edavey: yes this is handy for those specs where it makes sense to break the 'one assertion per test' convention...

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -966,6 +966,20 @@
           activity: existing_activity
         )
       end
+
+      context "when the activity fails to update" do
```
mec, line 970: 👍

mec (approved):
🚢

PR #1205, opened 2021-06-29
Document deleting the entire forecast history

mec:
## Changes in this PR

When an activity has been added which is not required, usually through
error, we need to delete it and all its associated entities.

This is the ONE time it is safe to delete `Forecast` rows in the database.

### DIFF
--- a/doc/forecasts-and-versioning.md
+++ b/doc/forecasts-and-versioning.md
@@ -358,3 +358,33 @@ The `ForecastOverview` class deals with the different versioning
 schemes at different levels, including if you pass a set of IDs for activities
 at various levels. Its `latest_values` method will always return the current
 versions for each activity.
+
+## Deleting the forecast history
+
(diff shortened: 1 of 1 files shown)


doc/forecasts-and-versioning.md
```diff
@@ -358,3 +358,33 @@ The `ForecastOverview` class deals with the different versioning
...
+so we need to delete the entire history of the forecasts__
+
+Use `unscoped` to bypass our warning on Forecasts.
+
+The steps are:
+
+Get your activity:
+
+```ruby
+activity = Activity.find([activity id])
```
pezholio, line 374: What's the rationale behind putting `[activity id]` in an array like that?

mec: Typo!

pezholio (approved):
LGTM  👍

PR #1209, opened 2021-06-30
(1930) Feature: associate Report with Wizard-orchestrated change to activity

edavey:
## Changes in this PR

When editing an `Activity` via the `ActivityFormController` we now:

- look up the appropriate `Report` 
- pass that report to the `HistoryRecorder`
- see the `HistoryRecorder` associate the given `Report` with each `HistoricalEvent` created

## Next steps

### DIFF
--- a/app/models/historical_event.rb
+++ b/app/models/historical_event.rb
@@ -1,6 +1,7 @@
 class HistoricalEvent < ApplicationRecord
   belongs_to :user
   belongs_to :activity
+  belongs_to :report, optional: true
 
   serialize :new_value
   serialize :previous_value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -719,6 +719,7 @@
 - Show actuals grouped by activity against a specific report
 - Show uploaded transactions after an upload
 - Record edits to Activities made in bulk via CSV imports, using the new Historical Event entity
+- Associate Report with HistoricalEvent when editing Activity via Wizard
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-59...HEAD
 [release-59]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-58...release-59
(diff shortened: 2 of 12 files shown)


edavey:
[Associate *Report* with wizard changes to Activity](https://trello.com/c/yZO3chyY/1930-associate-report-with-wizard-changes-to-activity)

app/models/historical_event.rb
```diff
@@ -1,6 +1,7 @@
 class HistoricalEvent < ApplicationRecord
   belongs_to :user
   belongs_to :activity
+  belongs_to :report, optional: true
```
pezholio, line 4: I think making this optional for now is fine, especially given the fact that it'll make tests fail on the imports. We can revisit this later if needs be

pezholio (approved):
Looks good to me, it'd be nice to confirm the correct report gets associated with the `HistoricalEvent` in the integration tests - just to give us a bit more confidence.

app/models/historical_event.rb
edavey: I've tweaked the integration test to make the assertion about the identity of the associated report more precise. In the case of the 'BEIS user' scenario, there is no report. I've added a note in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1209/commits/69b861a8d62998ffc30eef26429067eb3d27414d as I'm not sure what that signifies... Maybe @mec will have a view?

pezholio: Ah yeah, good call - BEIS users don't need a report to exist to make any changes, so it probably makes sense for the Report to be optional in this case

mec: Hey! Yep, totally agree, the report association has to be optional as BEIS users will make changes to their activites without a report.

edavey: thanks both, we can figure out how to make it a bit more explicit in the code/specs. Can we say:

- when using wizard: report is mandatory for DP and optional for BEIS
- when using upload CSV: report is always mandatory (it's a URL param after all!)
- when using 'rake' import facility: report is always optional

mec: - when using wizard: report is mandatory for DP and optional for BEIS: yep
- when using upload CSV: report is always mandatory (it's a URL param after all!): we will probably need it to be optional as BEIS will need to upload data as well, but we are not there yet.
- when using 'rake' import facility: report is always optional: as it's optional dependant on the user I think it would be the same here.

PR #1210, opened 2021-06-30
(1894) Feature: new BEIS user home page

mec:
## Changes in this PR

We have been using the organisation show page as the user 'home' page for some time, which never felt right.

This work introduces a distinct home page for BEIS users that lets them focus on common tasks and as a 'launching off point' for the applicaiton, this is the design in our last round of user testing.

Future work will do something similar for delivery partner users, but for now we redirect and keep the experience the same fo them.

### Before

Screenshot 2021-06-30 at 09-23-48 Organisation Department for Business, Energy and Industrial Strategy - Report your Offici

### After

Screenshot 2021-06-30 at 09-23-42 - Report your Official Development Assistance

### DIFF
--- a/app/views/staff/home/_delivery_partner_organisations_table.html.haml
+++ b/app/views/staff/home/_delivery_partner_organisations_table.html.haml
@@ -1,0 +1,16 @@
+%table.govuk-table#delivery_partner_organisations
+  %caption.govuk-table__caption.govuk-table__caption--m
+    Delivery partner organisations
+  %thead.govuk-table__head
+    %tr.govuk-table__row
+      %th.govuk-table__header{scope: "col"}
+        = t("table.header.organisation.name")
+      %th.govuk-table__header{scope: "col"}
+        = t("table.header.default.action")
+  %tbody.govuk-table__body
+    - @delivery_partner_organisations.each do |delivery_partner_organisation|
+      %tr.govuk-table__row
+        %td.govuk-table__cell
+          = delivery_partner_organisation.name
+        %td.govuk-table__cell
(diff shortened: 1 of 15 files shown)


app/views/staff/home/_delivery_partner_organisations_table.html.haml
```diff
@@ -0,0 +1,14 @@
+%table.govuk-table#delivery_partner_organisations
+  %caption.govuk-table__caption.govuk-table__caption--m
+    Delivery parnter organisations
```
pezholio, line 3:
```suggestion
    Delivery partner organisations
```

app/views/staff/home/_delivery_partner_organisations_table.html.haml
```diff
@@ -0,0 +1,14 @@
...
+  %thead.govuk-table__head
+    %tr.govuk-table__row
+      %th.govuk-table__header{scope: "col"} Name
+      %th.govuk-table__header{scope: "col"} Action
```
pezholio, line 7: Should we use translations for these? (or do we already have some we can reuse?)

pezholio (approved):
LGTM - just a couple of tiny bits

app/views/staff/home/_delivery_partner_organisations_table.html.haml
mec: Oh no! One of my pet peeves! 😅 

Yes, your right, for consistency with our code we should (I am against translations for uses like this one generally!) - I was looking at the export tables that have been introduce lately and they didn't use the locales so I think I wanted to see what would happen if I did the same!

I'll go and see what I can do. 👍

mec:
@pezholio I've used the existing `table.header.organisation.name` for one string and a new `table.header.default.action` for the other as i feel like 'action' has no specifity to anything so is a more generic term like 'continue'.

pezholio:
Cool cool - you may fire when ready! 👍

PR #1211, opened 2021-06-30
(1937) Budget type should be required and default to the direct type of the current activity fund

pezholio:
When tweaking the allowed budget types, we stopped defaulting to a particular type, and this highlighted the fact that we didn't actually validate the `budget_type` was present. This PR adds a validation to make sure a Budget Type is selected, as well as setting the default `budget_type` to `direct`.

### Before

image

### After

image

image

### DIFF
--- a/spec/presenters/budget_presenter_spec.rb
+++ b/spec/presenters/budget_presenter_spec.rb
@@ -7,6 +7,11 @@ RSpec.describe BudgetPresenter do
     it "returns the name of the budget type" do
       expect(described_class.new(budget).budget_type).to eq("Direct")
     end
+
+    it "returns nothing when the budget type is not present" do
+      budget.budget_type = nil
+      expect(described_class.new(budget).budget_type).to eq(nil)
+    end
   end
 
   describe "#iati_type" do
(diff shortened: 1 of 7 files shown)


pezholio:
[Budget type should be required and default to the direct type of the current activity fund](https://trello.com/c/MDn4MiA5/1937-budget-type-should-be-required-and-default-to-the-direct-type-of-the-current-activity-fund)

spec/presenters/budget_presenter_spec.rb
```diff
@@ -7,6 +7,11 @@
     it "returns the name of the budget type" do
       expect(described_class.new(budget).budget_type).to eq("Direct")
     end
+
+    it "returns nothing the the budget type is not present" do
```
edavey, line 11: tiny typo here I think. 

Probably you wanted "returns nothing *when* the budget type is not present"?

edavey (approved):
Nice!

One tiny typo I picked up...

spec/presenters/budget_presenter_spec.rb
pezholio: Cool - good catch 👍

PR #1214, opened 2021-07-01
(1934) Feature: create new delivery partner home page

mec:
## Changes in this PR

Following #1210 this work intoduces the delivery partner home page.

Delivery partners see the search widget and the hierarchical activity view on their home page, allowing them to locate their most common activities right from the home page.

The activity views are split by fund and progressively enchanced into tabs, this allows the users to view both funds activities without using a great deal of vertical space, for users with large numbers of activities this stops them potentially having to scroll a long way to locate activities.

The last commit 346384d contains a change to the JS pack as it seems the GOVUK JS was never being initialised correctly that makes the GOVUK JS tabs work.

### Before
Screenshot 2021-07-01 at 16-22-23 Organisation Academy of Medical Sciences - Report your Official Development Assistance

### After
Screenshot 2021-07-01 at 16-52-27 - Report your Official Development Assistance

### DIFF
--- a/app/controllers/staff/home_controller.rb
+++ b/app/controllers/staff/home_controller.rb
@@ -7,7 +7,22 @@ class Staff::HomeController < Staff::BaseController
       authorize @delivery_partner_organisations
       render :service_owner
     else
+      @grouped_programmes = fetch_grouped_programmes_for(current_user.organisation, :current)
       render :delivery_partner
     end
+  end
+
+  private def fetch_grouped_programmes_for(organisation, scope)
+    activities = policy_scope(
+      Activity.includes(
+        :organisation,
+        parent: [:parent, :organisation],
+        child_activities: [:child_activities, :organisation, :parent]
+      ).programme
+       .send(scope)
+    )
+    activities = activities.where(extending_organisation: organisation)
+    activities.order(:roda_identifier_compound)
+      .group_by(&:parent)
   end
 end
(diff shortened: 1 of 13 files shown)


app/controllers/staff/home_controller.rb
```diff
@@ -5,8 +5,24 @@ def show
...
+        :organisation,
+        parent: [:parent, :organisation],
+        child_activities: [:child_activities, :organisation, :parent]
+      ).programme
+       .send(scope)
+    )
+    activities = activities.where(extending_organisation: organisation)
+    activities.order(:roda_identifier_compound)
+      .group_by(&:parent)
+  end
```
pezholio, line 27: It looks like there's almost identical code being used in the ActivitiesController. Can we break this out into a service or similar?

pezholio (approved):
Only one teeny thing - feel free to ignore, just made my DRY senses tingle, especially as it's reasonably complex code

app/controllers/staff/home_controller.rb
mec: Yeah, I plan to, just as other work. I thought I covered this in a comment? Anyway - it is also not quite working how I think it probably should, but I want to keep to the acceptance on this card and make this better and probably add it to Activity.

Are you okay to let this in if I make sure the card is written up?

mec: Oop! I have already: https://trello.com/c/II0qSP5w

pezholio: Ah, so you did - sorry. Missed that! Yep - go for it!

PR #1215, opened 2021-07-01
Feature/1950 view change history via a tab on activity page

edavey:
## Changes in this PR

- Add a "Change history" tab to the Activity view
- Display a table of historical events

## Screenshots of UI changes

change_history

### Group by reference. 

This needs a tweak to how `HistoricalEvent#reference` is set (see last commit message). My idea is to present each group as a list item containing:

- a definition list giving the properties common to the *group*:
  + reference
  + user email
  + link to report
  + time of change

- a table showing for each historical element in the group:
  + property 
  + previous value
  + new value
 

## Next steps

## Trello

https://trello.com/c/8AxbdARO/1950-view-change-history-via-a-tab-on-activity-page

### DIFF
--- a/app/controllers/staff/activity_historical_events_controller.rb
+++ b/app/controllers/staff/activity_historical_events_controller.rb
@@ -7,5 +7,6 @@ class Staff::ActivityHistoricalEventsController < Staff::BaseController
     activity = Activity.find(params[:activity_id])
     authorize activity
     @activity = ActivityPresenter.new(activity)
+    @historical_events = @activity.historical_events.includes([:user, :report])
   end
 end
--- a/app/controllers/staff/activity_historical_events_controller.rb
+++ b/app/controllers/staff/activity_historical_events_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class Staff::ActivityHistoricalEventsController < Staff::BaseController
+  include Secured
+
+  def show
+    activity = Activity.find(params[:activity_id])
+    authorize activity
+    @activity = ActivityPresenter.new(activity)
+  end
+end
--- a/app/views/staff/activity_historical_events/show.html.haml
+++ b/app/views/staff/activity_historical_events/show.html.haml
@@ -1,0 +1,19 @@
+= content_for :page_title_prefix, t("document_title.activity.historical_events", name: @activity.title)
+
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-full
+      %h1.govuk-heading-xl
+        = @activity.title
+
+  .govuk-grid-row
+    .govuk-grid-column-full
+      .govuk-tabs
+        = render "staff/shared/activities/tab_nav"
+
+        .govuk-tabs__panel
+          %h2.govuk-heading-l
+            = t("page_title.activity.change_history")
+
+          %p.govuk-body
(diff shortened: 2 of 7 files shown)


edavey:
[View change history via a tab on "Activity" page](https://trello.com/c/8AxbdARO/1950-view-change-history-via-a-tab-on-activity-page)

app/controllers/staff/activity_historical_events_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Staff::ActivityHistoricalEventsController < Staff::BaseController
+  include Secured
```
mec, line 4: I think this is included already in `BaseController` - no biggie, just checking my understanding!

config/locales/models/activity.en.yml
```diff
@@ -22,6 +22,7 @@ en:
       children: Activity %{name} - Child activities
       comments: Activity %{name} - Comments
       transfers: Activity %{name} - Transfers
+      historical_events: Activity %{name} - Change history
```
mec, line 25: 👍

app/views/staff/activity_historical_events/show.html.haml
```diff
@@ -17,3 +17,26 @@
...
+              - @historical_events.each do |event|
+
+                %tr.govuk-body-s.govuk-table__row[event]
+                  %td.govuk-table__cell.reference= event.reference
+                  %td.govuk-table__cell.property= event.value_changed
+                  %td.govuk-table__cell.previous-value= event.previous_value
+                  %td.govuk-table__cell.new-value= event.new_value
+                  %td.govuk-table__cell.user= event.user.email
+                  %td.govuk-table__cell.timestamp= event.created_at.strftime("%d %b %Y at %R")
+                  %td.govuk-table__cell.report= link_to(event.report.financial_quarter_and_year, report_path(event.report))
```
mec, line 42: We might want to test for nil `event.report` here as I think there are cases when it is optional? I certainly have `HistoricalEvent` with report set to nil which causes this to fail.

mec (changes requested):
Seeing this working makes me so happy! I have been worried about this for so long, it is great to see!

Only one thing I think you should have a look at and a couple of comments.

app/views/staff/activity_historical_events/show.html.haml
edavey: thanks, yes just came across this myself as I was playing with editing 'geography'! 

I'll get a fix pushed this afternoon.

app/controllers/staff/activity_historical_events_controller.rb
edavey: yes you are right. I was just copy/pasting from other similar controllers like `Staff::ActivityTransfersController`, `Staff::ActivityChildrenController` etc which also inherit from `Staff::BaseController`...

I see that the inclusion of the `Secured` module is an original feature of the `Staff::BaseController` dating  back to Nov 2019 (https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/679316fe11526f129fd2ca99501b8196efe3dcbd).

It looks like more than half the controllers inheriting from `Staff::BaseController` are performing the same extra include.

So I think a general refactoring on a different PR would be the neatest way to tidy this up generally?

mec (approved):
🚢

PR #1216, opened 2021-07-02
(1899) Report summary includes financial totals

pezholio:
This PR adds some methods to gather up some basoc information about a report, as well as adding a new tab to display this as Summary data. Rather than redirecting to the variance, this tab is shown by default.

image


edavey:
Just a tiny formatting suggestion: should these 'currency' elements be right-aligned?

right_align_currency_values

spec/models/report_spec.rb
```diff
@@ -182,6 +182,27 @@
...
+      _project_outside_reporting_period = create(:project_activity, parent: programme, organisation: @report.organisation)
+      _third_party_project_outside_reporting_period = create(:third_party_project_activity, parent: @project_in_reporting_period, organisation: @report.organisation)
+
+      expect(@report.activities_created).to match_array([
+        @project_in_reporting_period,
+        @third_party_project_in_reporting_period,
+      ])
+    end
+  end
+
```
edavey, line 205: This is v clear, but I do think it ought to be unnecessary to be create all these objects and make actually database queries. i.e. for this to be more of a unit test and less of an integration test. 

I say "ought" because a bit of untangling would be needed for us to be able describe the way that `Report` communicates with other classes without actually exercising those classes or doing a load of unreadable mocking.

For example, if the `Report#reportable_activities` used a 'finder' service to perform its lookup on `Activity` rather than calling `Activity.reportable.projects_and_third_party_projects_for_report(self).with_roda_identifier` directly then our description of `Report#activities_created` could be limited to saying that the AR object returned by our putative new `ReportableActivityFinder` is sent the `.where(created_at: financial_period)` message.

e.g.
```
  describe "#activities_created" do
    it "filters the #reportable_activities using the financial period" do
      report = Report.new
      finder = instance_method(ReportableActivityFinder, where: double('AR scope'))
      allow(ReportableActivityFinder).to receive(:new).and_return(finder)

      report.activities_created

      expect(finder).to have_received(:where).with(created_at: financial_period)
    end
```

edavey (approved):
Really nice work. Very easy to follow. Loads of improvements on the way.

I have made a suggestion about how the unit tests on `Report` might be made simpler by decoupling some of the complex dependencies, e.g. using a `ReportableActivityFinder` -- which would then allow quite straightforward mocking to be done.

Just something for us to think about and discuss maybe?

spec/models/report_spec.rb
pezholio: I definitely think this merits further discussion, I think there's a bunch of places where we could use a similar pattern

PR #1218, opened 2021-07-05
Optimise projects_and_third_party_projects_for_report scope

pezholio:
The `projects_and_third_party_projects_for_report` scope is currently quite slow because we walk the entire tree to
get the activities. Now we have the `source_fund_code` attribute, we can get all these in one query.

I also had to make some tweaks to the Activity factory to get the parent fund by fetching or creating a parent fund.

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -211,13 +211,11 @@ class Activity < ApplicationRecord
   scope :with_roda_identifier, -> { where.not(roda_identifier_compound: nil) }
 
   scope :projects_and_third_party_projects_for_report, ->(report) {
-    programmes = where(level: :programme, parent_id: report.fund_id)
-    projects = where(level: :project, parent_id: programmes.pluck(:id))
-    third_party_projects = where(level: :third_party_project, parent_id: projects.pluck(:id))
-
-    for_organisation = where(organisation_id: report.organisation_id)
-
-    for_organisation.merge(projects.or(third_party_projects))
+    where(
+      level: [:project, :third_party_project],
+      organisation_id: report.organisation_id,
+      source_fund_code: report.fund.source_fund_code
+    )
   }
 
   scope :current, -> {
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -210,14 +210,6 @@ class Activity < ApplicationRecord
   scope :publishable_to_iati, -> { where(form_state: :complete, publish_to_iati: true) }
   scope :with_roda_identifier, -> { where.not(roda_identifier_compound: nil) }
 
-  scope :projects_and_third_party_projects_for_report, ->(report) {
-    where(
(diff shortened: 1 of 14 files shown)


edavey:
[Optimise projects_and_third_party_projects_for_report scope](https://trello.com/c/Qs1vrqPF/1988-optimise-projectsandthirdpartyprojectsforreport-scope)

app/models/activity.rb
```diff
@@ -211,13 +211,11 @@ class Activity < ApplicationRecord
...
-
-    for_organisation = where(organisation_id: report.organisation_id)
-
-    for_organisation.merge(projects.or(third_party_projects))
+    where(
+      level: [:project, :third_party_project],
+      organisation_id: report.organisation_id,
+      source_fund_code: report.fund.source_fund_code
+    )
   }
```
edavey: This is so much easer to understand (and no doubt faster too).

I feel like the ideal presentation of the changes would be to have a single commit where you perform this refactoring of `Activity:projects_and_third_party_projects_for_report` and the matching spec **remains unchanged**. That would make it super clear that there is no change in behaviour, just in implementation. I'm guessing that you might need an earlier commit in which you refactor the spec so it supports the new implementation, preemptively?

app/services/activity/projects_for_report_finder.rb
```diff
@@ -0,0 +1,18 @@
+class Activity
+  class ProjectsForReportFinder
+    attr_accessor :report, :scope
```
edavey, line 3: Is there are reason for these to be i) writers and ii) public? 

Keeping them private and immutable would I think be best unless there's a good reason not to?

edavey (approved):
👍 This is really good, brings lots more clarity to these complex 'finder' methods in `Activity`.

I've made a couple of minor comments/suggestions.

spec/models/report_spec.rb
```diff
@@ -157,86 +157,92 @@
...
-            report: report,
-            financial_quarter: 2,
-            financial_year: 2021,
-            user: create(:delivery_partner_user)
-          ).set_value(50_000)
-        end
-      }
+  describe "#forecasts_for_reportable_activities" do
+    let(:report) { build(:report) }
+    let(:reportable_activities) { build_stubbed_list(:project_activity, 5) }
```
edavey, line 204: I've not familiar with these `build_stubbed*` methods in FactoryBot. I need to have a proper look into their use...

thks!

app/services/activity/projects_for_report_finder.rb
pezholio: Yep, fixed that 👍

app/models/activity.rb
pezholio: Good call - I've done that now. Makes it much easier to follow I think

PR #1222, opened 2021-07-08
(1956) Improve the query that populates the activity tree view

pezholio:
This PR changes the way we fetch grouped activities by adding a new `Activity::GroupedActivitiesFetcher`, which we can
use in the homepage and activities page. It also improves on what we have by doing all the fetching of child activities in the class itself, rather than while rendering the page. It also fixes a bug where child activities didn't have the same scope as the programmes, so historic projects were showing up under currently running programmes.

I've also taken the opportunity to tweak some of the feature tests to use stubs and mocks, so we're not having to create so many objects in the database, hopefully speeding up the tests a little bit.

### DIFF
--- a/app/controllers/staff/activities_controller.rb
+++ b/app/controllers/staff/activities_controller.rb
@@ -3,11 +3,15 @@
 class Staff::ActivitiesController < Staff::BaseController
   include Secured
   after_action :verify_authorized, except: [:index, :historic]
-  after_action :verify_policy_scoped, only: [:index, :historic]
+  after_action :skip_policy_scope, only: [:index, :historic]
 
   def index
     @organisation = Organisation.find(organisation_id)
-    @grouped_programmes = fetch_grouped_programmes_for(@organisation, :current)
+    @grouped_activities = Activity::GroupedActivitiesFetcher.new(
+      user: current_user,
+      organisation: @organisation,
+      scope: :current
+    ).call
   end
 
   def show
@@ -34,7 +38,11 @@ class Staff::ActivitiesController < Staff::BaseController
 
   def historic
     @organisation = Organisation.find(organisation_id)
-    @grouped_programmes = fetch_grouped_programmes_for(@organisation, :historic)
+    @grouped_activities = Activity::GroupedActivitiesFetcher.new(
+      user: current_user,
+      organisation: @organisation,
+      scope: :historic
+    ).call
   end
 
   private
@@ -58,21 +66,5 @@ class Staff::ActivitiesController < Staff::BaseController
 
   def fund_id
     params[:fund_id]
-  end
-
-  def fetch_grouped_programmes_for(organisation, scope)
-    activities = policy_scope(
-      Activity.includes(
-        :organisation,
-        parent: [:parent, :organisation],
-        child_activities: [:child_activities, :organisation, :parent]
-      ).programme
-       .send(scope)
-    )
-    unless organisation.service_owner?
-      activities = activities.where(extending_organisation: organisation)
-    end
-    activities.order(:roda_identifier_compound)
-      .group_by(&:parent)
   end
 end
--- a/app/services/activity/grouped_activities_fetcher.rb
+++ b/app/services/activity/grouped_activities_fetcher.rb
@@ -1,0 +1,41 @@
+class Activity
+  class GroupedActivitiesFetcher
+    def initialize(user:, organisation:, scope: :all)
+      @organisation = organisation
+      @scope = scope
+      @activities = ActivityPolicy::Scope.new(user, Activity).resolve
+    end
+
+    def call
+      grouped_programmes.each_with_object({}) do |(fund, programmes), funds|
+        funds[fund] = programmes.each_with_object({}) { |programme, programmes|
+          programmes[programme] = scoped_child_activities_for(programme).each_with_object({}) { |project, projects|
+            projects[project] = scoped_child_activities_for(project)
+          }
+        }
+      end
+    end
+
+    private
+
+    def grouped_programmes
+      programmes = activities.includes(
+        :organisation,
+        parent: [:parent, :organisation],
+        child_activities: [:child_activities, :organisation, :parent]
+      ).programme.send(scope)
+
+      unless organisation.service_owner?
+        programmes = programmes.where(extending_organisation: organisation)
+      end
+
+      programmes.order(:roda_identifier_compound).group_by(&:parent)
+    end
+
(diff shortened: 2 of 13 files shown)


pezholio:
[Improve the query that populates the activity tree view](https://trello.com/c/II0qSP5w/1956-improve-the-query-that-populates-the-activity-tree-view)

app/services/activity/grouped_activities_fetcher.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      programmes.order(:roda_identifier_compound).group_by(&:parent)
+    end
+
+    def scoped_child_activities_for(activity)
+      activity.child_activities.send(scope).order(:created_at)
+    end
+
+    attr_reader :organisation, :activities, :scope
```
mec, line 39: Is there significance to this being here? I usually see this at the top of a file?

app/services/activity/grouped_activities_fetcher.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      unless organisation.service_owner?
+        programmes = programmes.where(extending_organisation: organisation)
+      end
+
+      programmes.order(:roda_identifier_compound).group_by(&:parent)
+    end
+
+    def scoped_child_activities_for(activity)
+      activity.child_activities.send(scope).order(:created_at)
```
mec, line 36: Confirming my understanding:

If `scope` here is `:all` we call:

`activity.child_activities.all.order(:created_at)`

the `send()` calls the method name passed in, is that right?

app/services/activity/grouped_activities_fetcher.rb
```diff
@@ -0,0 +1,41 @@
+class Activity
+  class GroupedActivitiesFetcher
+    def initialize(user:, organisation:, scope: :all)
+      @organisation = organisation
+      @scope = scope
+      @activities = ActivityPolicy::Scope.new(user, Activity).resolve
+    end
+
+    def call
+      grouped_programmes.each_with_object({}) do |(fund, programmes), funds|
```
mec, line 10: This is super clever, it took a while for me to unpick what was going on, but that is my ignorance! 👍 

Question: this kind of nested looping over the hierarchy is probably common, is there a way to split the nesting in to method calls to make it approachable?

app/controllers/staff/activities_controller.rb
```diff
@@ -3,11 +3,15 @@
 class Staff::ActivitiesController < Staff::BaseController
   include Secured
   after_action :verify_authorized, except: [:index, :historic]
-  after_action :verify_policy_scoped, only: [:index, :historic]
+  after_action :skip_policy_scope, only: [:index, :historic]
 
   def index
     @organisation = Organisation.find(organisation_id)
-    @grouped_programmes = fetch_grouped_programmes_for(@organisation, :current)
+    @grouped_activities = Activity::GroupedActivitiesFetcher.new(
```
mec, line 10: :heart:

app/views/staff/activities/historic.html.haml
```diff
@@ -30,4 +30,4 @@
           %h2.govuk-heading-l
             = t("page_title.activities.historic")
 
-          = render partial: "staff/shared/activities/tree_view/table", locals: { grouped_programmes: @grouped_programmes }
+          = render partial: "staff/shared/activities/tree_view/table", locals: { activities: @grouped_activities }
```
mec, line 33: :heart:

app/controllers/staff/activities_controller.rb
```diff
@@ -3,11 +3,15 @@
 class Staff::ActivitiesController < Staff::BaseController
   include Secured
   after_action :verify_authorized, except: [:index, :historic]
-  after_action :verify_policy_scoped, only: [:index, :historic]
+  after_action :skip_policy_scope, only: [:index, :historic]
```
mec, line 6: What do you think about adding one sentance to the commit message about the fact you have to skip the policy scope because we have `verify_policy_scoped` in the base controller so all our controllers have it switched on, I think we forget about this kind of inheritance on controllers (e.g. `include Secured`) sometimes?

mec (changes requested):
Great piece of work! Thank you for taking the time to show how we can write faster tests by following this 'fetcher' pattern! 👍 

I struggled a bit to follow, but that is my ignorance - I have left some questions for my own learnings so I am selfishly not approving until I have learned from you! 😄 

Happy to chat through my questions if it's easier than GH comments?

app/services/activity/grouped_activities_fetcher.rb
pezholio: This is a habit that @edavey has got me in to - if we don't need the `attr_reader`s to be publicly accessible (that is, outside the class itself), then it make sense for them to be private. Sadly `private attr_reader` (which IMO is more readable) is not valid Ruby 😞

app/services/activity/grouped_activities_fetcher.rb
pezholio: That's exactly right 👍

app/controllers/staff/activities_controller.rb
pezholio: Cool, done! 👍

app/services/activity/grouped_activities_fetcher.rb
pezholio: It's a bit tricky I think - we're acting on an ActiveRecord relation, and we're having to do something with each thing when we loop through the relation. It might be something we can improve on later, but it's making my head hurt! OK to let this go in and improve on it later?

pezholio:
Actually, can we hold off on this? I'm noticing quite a not insignificant performance hit, and I'm going to try and get to the bottom of it

mec:
> Actually, can we hold off on this? I'm noticing quite a not insignificant performance hit, and I'm going to try and get to the bottom of it

@pezholio I am leaving this as is until Monday 12th, we can catch up then as needed.

pezholio:
After agonising over this, I think this is fine because:

a) The performance hit is only small
b) It's only noticable when viewing all acitivities as a BEIS user; and
c) The number of users who meet this profile are small

mec:
@pezholio if it helps, I think your concerns are about the perf of the view when BEIS users view 'all' activities i.e `/activities` with the filter set to BEIS, this view will be going aways soon to be replaced by indivual 'delivery partner activity views'

pezholio:
@mec Sweet - that makes me feel much better 😃

mec (approved):
🚢

PR #1226, opened 2021-07-12
(2004) fix: orphanned historicevents cause activitiesupdated to fail

mec:
## Changes in this PR

We heard from a BEIS user that they could not view a report. The issue was that we had deleted an `Activity` from the report that had already created some `HistroricalEvents`. This meant the event stored the ID of an activity that no longer exists.

The `Report` method `activities_updated` uses the Activity IDs in the HistoricalEvents to get a list of 'activities which have been modified in this report`. Under the hood, this method called `find` passing in an array of activity IDs from the historical events.

In cases where the activity has been deleted, this causes `find` to raise an error.

To fix this we swap to a `where` as this will always return a enumerable, even if it is empty and so the view can be rendered in an empty state.

This work also adds `dependent: destroy` on `Activity` so that HIstoricalEvents are removed along with their associated activities - see the commit long for further thoughts on this, on balance, I think this is the best approach for now.

Finally there is a data migration to remove any orphened HIstoricalEvents, we have 10 in production at this time.

## Deployment warning!
This PR includes a manual data migration that has to be run in production once the code is deployed.

```
bin/rails runner db/data/20210712154115_remove_orphaned_historical_events.rb
```

### DIFF
--- a/db/data/20210712154115_remove_orphaned_historical_events.rb
+++ b/db/data/20210712154115_remove_orphaned_historical_events.rb
@@ -1,0 +1,4 @@
+# Run me with `rails runner db/data/20210712154115_remove_orphaned_historical_events.rb`
+#
+orphaned_events = HistoricalEvent.all.select { |event| Activity.find_by(id: event.activity_id).nil? }
+orphaned_events.destroy_all
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -735,6 +735,7 @@
 - Group Activity's Change history by "reference" and show newest first
 
 ## [unreleased]
+- fix guard against orphan HistoricalEvents
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-61...HEAD
 [release-61]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-60...release-61
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -736,6 +736,7 @@
 
 ## [unreleased]
 - fix guard against orphan HistoricalEvents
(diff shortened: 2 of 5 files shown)


db/data/20210712154115_remove_orphaned_historical_events.rb
```diff
@@ -0,0 +1,4 @@
+# Run me with `rails runner db/data/20210712154115_remove_orphaned_historical_events.rb`
+#
+orphaned_events = HistoricalEvent.all.select { |event| Activity.find_by(id: event.activity_id).nil? }
```
edavey, line 3: It's not important here, in a one-off script, but I think you can do what you want much more efficient in SQL with an "inner join":

```
HistoricalEvent.joins(:activity).count
   (29.9ms)  SELECT COUNT(*) FROM "historical_events" INNER JOIN "activities" ON "activities"."id" = "historical_events"."activity_id"
=> 2143
```
as opposed to a left outer joins which would return all historical events including those whose activity no longer exists?

```
 HistoricalEvent.left_joins(:activity).count
   (1.4ms)  SELECT COUNT(*) FROM "historical_events" LEFT OUTER JOIN "activities" ON "activities"."id" = "historical_events"."activity_id"
=> 2153
```

i.e. there are 10 historical records orphaned?

See https://blog.codinghorror.com/a-visual-explanation-of-sql-joins/ for a great visualisation of these joins....

edavey (approved):
Good solution! I've made a tiny comment on the SQL vs Ruby in the script.

db/data/20210712154115_remove_orphaned_historical_events.rb
mec: 🙏 Thanks @edavey I really appreciate deeper learning like this!

PR #1228, opened 2021-07-12
Fix (1995) Don't record changes in the Wizard's "Form state"

edavey:
## Trello
https://trello.com/c/Yr4Kxj3X/1995-change-history-includes-the-formstate-changes

## Changes in this PR

We no longer record HistoricalEvents for changes which are to `Activity#form_state`. This is an internal property of the Wizard system we are using for multi-step forms.

We also include a simple data migration to remove existing events of this type. It should be run after deployment with:

```
rails runner db/data/20210712154227_remove_historical_events_for_wizard_form_state_changes.rb
```

### Before

wizard_form_state

### After

no_more_Internal_form_state

## Next steps

- [ x] Run 'data migration' after deployment

### DIFF
--- a/app/services/history_recorder.rb
+++ b/app/services/history_recorder.rb
@@ -1,10 +1,15 @@
 class HistoryRecorder
+  INTERNAL_VALUES = ["form_state"].freeze
+
   def initialize(user:)
     @user = user
   end
 
   def call(changes:, reference:, activity:, report: nil)
-    changes.each do |value_changed, (previous_value, new_value)|
+    changes
+      .each do |value_changed, (previous_value, new_value)|
+      next if INTERNAL_VALUES.include?(value_changed)
+
       HistoricalEvent.create(
         user: user,
         activity: activity,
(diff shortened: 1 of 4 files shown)


edavey:
[Change history includes the form_state changes](https://trello.com/c/Yr4Kxj3X/1995-change-history-includes-the-formstate-changes)

app/services/history_recorder.rb
```diff
@@ -4,7 +4,9 @@ def initialize(user:)
   end
 
   def call(changes:, reference:, activity:, report: nil)
-    changes.each do |value_changed, (previous_value, new_value)|
+    changes
+      .delete_if { |value_changed, _| value_changed == "form_state" }
+      .each do |value_changed, (previous_value, new_value)|
```
pezholio, line 9: Would it not be clearer if we called `next` if the `value_changed` was `form_state`. Might be good if we had `"form_state"` in an array constant called something like `INTERNAL_VALUES` to make it clear why we're skipping this particular variable (we may also want to add others)

edavey: yes, both excellent suggestions, thank you!

mec (approved):
🚢

PR #1230, opened 2021-07-13
Refactor activity specs

mec:
## Changes in this PR

Ahead of some work on navigating acitivites I wanted to split the big `users_can_view_activities` into individual spec files.

No actual specs changed here other than moving specs that simply test that a signed out user is redirected to the sign in page to (faster) controller specs.


pezholio (approved):
LGTM - some really useful tweaks that should start improve things 👍

PR #1232, opened 2021-07-16
Send a notification on successful deploy

pezholio:
This adds a new job in the deploy Github Action to poll the Health Check URL and check the SHA returned matches the current SHA of the code we’ve deployed.

Once it does it sends a message to our Slack channel to inform the team that the code has been deployed.

This only runs on production deploys and times out after 30 minutes (in case something has gone wrong and we end up waiting forever for something that is never going to happen)

Here's a sample of how the Slack message will look

image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -77,3 +77,14 @@ jobs:
         run: |
           script/deploy-terraform
         if: github.ref == 'refs/heads/master'
+  check_and_notify:
+    needs: deploy
+    if: github.ref == 'refs/heads/master'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v2
+      - name: Notifying the team of a successful deploy
+        env:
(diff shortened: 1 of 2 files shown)


pezholio:
[Send a notification of a successful production deploy](https://trello.com/c/FMaEcp9K/2013-send-a-notification-of-a-successful-production-deploy)

CristinaRO (approved):
I think it looks great.

Wanted to check a couple of things:
- The script itself has the potential to be stuck in an infinite loop, and it's only the GH action that has a timeout; does this risk leaving a process running somewhere in the cloud? Or does the script only live for as long as the action tries to run it?
- Do we want to be notified when the timeout has been reached, so we remember to debug what went wrong?

The branch needs updating.

pezholio:
@CristinaRO Cool - ta! In answer to your question - as far as I understand it, the timeout on the job will kill the script and set the build to a failing state. This will then send an email to the person who's push kicked off the build (as it does with the CI script that runs the tests)

PR #1233, opened 2021-07-16
Create and populate an activity's RODA identifier automagically

pezholio:
This is a biggie. Once this is in play, we will no longer have to ask delivery partners for a RODA idenfitier. This now happens automatically at the point an activity is created.

We still have the `-` seperator for each level, but each activity has its own fragment generated using the [Nanoid gem](https://github.com/radeno/nanoid.rb).

This is a pretty big PR, but a good chunk of the work is deleting/tweaking code and tests, so hopefully I've removed a good deal of complexity too - it's a win/win!


pezholio:
[Create and populate`roda_identifier` on Activity](https://trello.com/c/CJuTNbhD/1992-create-and-populaterodaidentifier-on-activity)

app/services/activity/updater.rb
```diff
@@ -36,7 +36,6 @@ def set_identifier
 
     def set_roda_identifier
       assign_attributes_for_step("roda_identifier_fragment")
```
CristinaRO: Does this step still make any sense?

spec/features/staff/users_can_search_for_activities_spec.rb
```diff
@@ -2,15 +2,13 @@
...
+  let!(:project) { create(:programme_activity, roda_identifier: "roda-id", title: "Project A") }
 
-  scenario "searching by RODA identifier fragment" do
+  scenario "searching by RODA identifier" do
     visit "/"
     fill_in :query, with: "roda-id"
     click_button t("form.activity_search.submit")
 
     expect(page).to have_link project.title, href: organisation_activity_path(project.organisation, project)
-    expect(page).to have_link third_party_project.title, href: organisation_activity_path(third_party_project.organisation, third_party_project)
```
CristinaRO, line 14: From the POV of users, this equates to removing functionality without a workaround, unless we add partial string matching. Are we OK with this?

CristinaRO (commented):
This is a high speed review, while I try to get back into things, so it's not as exhaustive as I'd like. Good stuff, I do have some questions and confusion as noted.

spec/services/activity/roda_identifier_generator_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+
+RSpec.describe Activity::RodaIdentifierGenerator do
+  let(:activity) { build(:project_activity) }
+
+  before do
+    expect(activity).to receive(:roda_identifier) { "NF-ABC1234" }
```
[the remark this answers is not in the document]
pezholio: That's a good call 👍

spec/services/activity/roda_identifier_generator_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  describe "#generate" do
+    subject { described_class.new(activity).generate }
+
+    it "generates a RODA identifier" do
+      expect(Nanoid).to receive(:generate).with(
```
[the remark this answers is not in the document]
pezholio: This is expectation, because I have the specific arguments in play

app/models/activity.rb
```diff
@@ -455,14 +451,6 @@ def cache_roda_identifier!
     end
   end
 
-  def roda_identifier_compound=(roda_identifier)
```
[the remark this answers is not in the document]
pezholio: Good call, I think it's worth keeping in in that case

app/services/activity/updater.rb
pezholio: At this point in the feature, yeah, but it does need removing in a lter commit

spec/features/staff/users_can_search_for_activities_spec.rb
pezholio: I have been thinking about this, and I don't know. Do we need to keep the fragments around in that case? Is it even useful? I'll raise it with the rest of the team - thanks for picking it up 👍

db/schema.rb
```diff
@@ -66,15 +64,16 @@
     t.integer "sdg_2"
     t.integer "sdg_3"
     t.boolean "sdgs_apply", default: false, null: false
+    t.integer "covid19_related", default: 0
```
CristinaRO, line 67: The unrelated schema changes make it trickier to distinguish what's actually changed. 😬

db/schema.rb
```diff
@@ -10,13 +10,13 @@
...
 
-ActiveRecord::Schema.define(version: 2021_07_08_112144) do
+ActiveRecord::Schema.define(version: 2021_07_14_132606) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
 
-  create_table "activities", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+  create_table "activities", id: :uuid, default: -> { "public.gen_random_uuid()" }, force: :cascade do |t|
```
CristinaRO, line 19: Is this addition of a `public` prefix? namespace? necessary, or unrelated artefact of local dev stuff?

spec/features/staff/users_can_search_for_activities_spec.rb
CristinaRO: This appears solved by the partial string search you've added, right? I assume it's intentional that you haven't re-added it to the feature spec, and covered it in the unit spec.

app/models/activity.rb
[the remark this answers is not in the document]
CristinaRO: How much do we want to enforce the read-only-ness of the automagically RODA identifier? If I read the tests right, users cannot edit it via the UI, is that right? Do we want to reserve the right to be able to edit it via console?

CristinaRO (commented):
Couple more observations / questions.

db/schema.rb
pezholio: Aaaaarrrgh. That was a total accident. I hate when this happens.

db/schema.rb
pezholio: Same. This should have been stripped out by me, but got caught up in rebase hell

spec/features/staff/users_can_search_for_activities_spec.rb
pezholio: Yeah, the feature spec just covers the most common use case now

app/models/activity.rb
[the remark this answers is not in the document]
pezholio: I did put this line back in, but it borked a lot of tests. As you say, there is no longer any way to edit the RODA Identifier via the UI now, so I think it makes sense to not worry about guarding against it being changed, otherwise we'd have to add guards for a whole heap of other things we consider immutable.

CristinaRO (approved):
Alas, there are conflicts now, but it should be ready for you to rebase when you come back.

spec/services/activity_defaults_spec.rb
```diff
@@ -20,13 +20,15 @@
...
-    subject do
+    let(:activity_defaults) do
       described_class.new(
         parent_activity: parent_activity,
         delivery_partner_organisation: delivery_partner_organisation
-      ).call
+      )
     end
 
+    subject { activity_defaults.call }
```
mec, line 30: 👍

mec (commented):
It was a long, long road to get us here, but we did it! Woop 🥳 

I am really happy how this turned out, thank you for putting the work in to get this done.

Can I ask you to do one thing, can we update the 'activity identifiers' doc: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/activity-identifiers.md

I think it would be good to rename the current section on RODA Identifier to 'Legacy RODA identifier' and note there that Q1-2021-2022 was the last reporting in which that was used, and add a short, new section about the new identifier.

db/schema.rb
```diff
@@ -10,8 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_07_19_082105) do
-
+ActiveRecord::Schema.define(version: 2021_07_14_132606) do
```
edavey, line 13: @meyric we need the later timestamp here

PR #1235, opened 2021-07-19
(1958) Feature: rework activites navigation

mec:
## Changes in this PR
There is a fair bit of change here, thank you in advance for reviewing!

The goal of this work is to change the way multiple acitvities are navigated.

Overall:

- use explicit urls instead of paramaters

For BEIS user:

- show a list of delivery partners
- each links to their own activities

For delivery partners:

- retain the existing current/historic split
- use the hierarchical view of activities

Most of the changes here are to the specs as a result of changes to the routing. We rely heavily on feature tests that simulate a user navigating the app and verifying behaviour.  I made the decision to skip parts of the navigation and focused on the behaviour part, see commit messages for more details, it boils down to: we have test for the navigating so there is little value in testing that part over and over.

The first commit here breaks the test suite, but I deicded to keep the commits seperate to aid reviewing as most of the changes are just boilerplate changes for the routing, see above.

WIth the new routes and views, we no longer need the activities to be shown on the organisation pages. We also have the ideal location for the BEIS user 'add new level B acitivities' UI - I've made a first pass at this, replicating the two buttons.

Having a view for a delivery partners acitivites also allows us to link from a BEIS user home page to the activities themselves, creating a clear and strong link right on the home page.

All that is left on the organisation show page is now the IATI XML UI and the organisation details themselves. The IATI XML UI is set to move in other work, but I took the time to signpost it more clearly and change the layout.

## Trello
https://trello.com/c/3N9cOe6D

## Screenshots of UI changes
The new BEIS user activities view:
Screenshot 2021-07-20 at 09-02-47 - Report your Official Development Assistance

A deivery partners activities view:
Screenshot 2021-07-20 at 09-02-57 Activities for Academy of Medical Sciences - Report your Official Development Assistance

The resulting, cleaned up organisation view:
Screenshot 2021-07-20 at 09-04-16 Organisation Academy of Medical Sciences - Report your Official Development Assistance


spec/controllers/staff/organistations_controller_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe Staff::OrganisationsController do
+  context "when the user is not logged in" do
+    it "redirects the user to the root path" do
+      organisation = create(:delivery_partner_organisation)
+      get :index
+      expect(response).to redirect_to(root_path)
+      get :show, params: {id: organisation.id}
```
edavey, line 9: I think you don't need to go to the "expense" of creating an organisation object and saving it to the db here and that something like

```
get :show, params: {id: `double`}
```

will do better as it:

- avoids the object creation/persistence
- makes it clear that the organisation ID is not relevant to the access control mechanism and redirection

spec/features/staff/users_can_create_a_budget_spec.rb
```diff
@@ -59,9 +56,7 @@
         fund_activity = create(:fund_activity, organisation: user.organisation)
         programme_activity = create(:programme_activity, parent: fund_activity, organisation: user.organisation)
 
-        visit activities_path
-
-        click_on(programme_activity.title)
+        visit organisation_activity_path(programme_activity.organisation, programme_activity)
 
         click_on(t("page_content.budgets.button.create"))
```
edavey, line 61: I totally agree with your simplification of the test setup here: going straight to the start of the journey which you wish to exercise. 

**Given**: I am on the activity page
**When**: I fail to supply the required budget detail
**Then**: I see error messages describing what's missing

Where it aids readability, reuse, maintainability I would advocate using helper methods for any of these parts of the scenario...

edavey (approved):
👍  wow, that was quite a lot to follow. I think I was able to play along pretty closely.

A couple of tiny typos.

I've made a few comments  -- not suggesting any particular changes now, just thoughts...

spec/controllers/staff/organistations_controller_spec.rb
mec: Nice! Thank you.

spec/features/staff/users_can_create_a_transaction_spec.rb
```diff
@@ -43,9 +31,7 @@
       activity = create(:programme_activity, :with_report, organisation: user.organisation)
 
       PublicActivity.with_tracking do
-        visit activities_path
-
-        click_on(activity.title)
+        visit organisation_activity_path(activity.organisation, activity)
```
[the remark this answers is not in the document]
mec: Yes, I imaged it as part of the work that added it's replacement - could you write up a card on the backlog to do this?

spec/features/staff/users_can_view_activities_spec.rb
```diff
@@ -57,16 +57,37 @@
...
+      gcrf = create(:fund_activity, :gcrf)
+      programme = create(:programme_activity, parent: gcrf, extending_organisation: delivery_partner_organisation)
+      _report = create(:report, :active, fund: gcrf, organisation: delivery_partner_organisation)
+
+      visit organisation_activity_path(programme.organisation, programme)
+      click_on t("tabs.activity.children")
+
+      expect(page).to_not have_link(t("action.activity.add_child"), exact: true)
+    end
   end
```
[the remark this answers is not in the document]
mec: Epic good or epic bad?

spec/features/staff/users_can_create_a_budget_spec.rb
mec: Are you proposing a helper method along the lines of `navigate_to_activity(activity)`?

edavey: in retrospect I might be possible to have:

1. an initial commit  extracting all the preamble to get to the activity page to a method like that used in multiple places. No change to behaviour, just a refactoring of the test implementation

2. a following commit which then implements your change in routing behaviour in that new helper, **in one place** making the change much clearer

mec: I think I see what you are striking for and I agree, it would be much nicer.  Given I have the approval, are you happy to go ahead without attempting that - any other time I would have a go, I am struggling to find the energy on the project today though?

app/controllers/staff/activities_controller.rb
```diff
@@ -7,11 +7,16 @@ class Staff::ActivitiesController < Staff::BaseController
...
+    if @organisation.service_owner?
+      @delivery_partner_organisations = Organisation.delivery_partners
+      render "staff/activities/index_beis"
+    else
+      @grouped_activities = Activity::GroupedActivitiesFetcher.new(
+        user: current_user,
+        organisation: @organisation,
+        scope: :current
+      ).call
+    end
```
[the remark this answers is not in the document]
mec: I agree with everything you propose. Some of the urls look a bit off to me, but I may be missing something. 

I am most intereseted in where and when you would aim to get a piece of work like this into delivery? How would you sell this to the client to prioritise over a feature they feel they want?

When, when and how would you capture this thought for the future, I have a thousand things like this that I believe could improve things, but generally just can them in a 'focus on the problem infront of you' kind of way.

mec:
@edavey I've merged this but I would still love to hear you thoughts on the points raised?

PR #1243, opened 2021-07-22
(2022) Strip out legacy PublicActivity auditing

edavey:
Trello: https://trello.com/c/hZzWu3h5/2022-strip-out-legacy-publicactivity-code

As per ADR 31 we are now using our own system for tracking
changes (see `HistoricalEvent` and `HistoryRecorder`).

Public Activity was introduced a long time ago, before it
was understood what Delivery Partners and BEIS users needed
from an audit trail.

None of the information collected from the numerous tracking
points has ever been surfaced to users and it can be
considered to have no current value.

I noticed in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1235 that work was done to maintain several of the tests describing `PublicActivity`'s behaviour -- so in order to remove this unnecessary maintenance burden I think we should remove this crufty code asap.


edavey:
[Strip out legacy PublicActivity code](https://trello.com/c/hZzWu3h5/2022-strip-out-legacy-publicactivity-code)

CristinaRO (approved):
Thank you for tackling the cruft!

PR #1246, opened 2021-07-22
Fix: Load report activities correctly for spending report

edavey:
Spending reports were throwing a:

NoMethodError (undefined method `each' for #<Activity::ProjectsForReportFinder>):
  /app/controllers/staff/spending_breakdowns_controller.rb:24

error which was causing an empty CSV file to be downloaded.

We now include an Actual "transaction" in our feature test
and verify that the downloaded file is not empty.

Trello: https://trello.com/c/ceicvcWC/2010

spending_report

### DIFF
--- a/spec/features/staff/users_can_view_reports_spec.rb
+++ b/spec/features/staff/users_can_view_reports_spec.rb
@@ -143,7 +143,12 @@ RSpec.feature "Users can view reports" do
     end
 
     scenario "can download a spending breakdown CSV of the report" do
-      report = create(:report, :active)
+      fund = create(:fund_activity)
+      programme = create(:programme_activity, parent: fund)
+      dp_org = create(:delivery_partner_organisation)
+      project = create(:project_activity, organisation: dp_org, parent: programme)
+      report = create(:report, :active, organisation: dp_org, fund: programme.parent)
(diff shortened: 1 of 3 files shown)


edavey:
[The spending reports appears to be broken (empty file)](https://trello.com/c/ceicvcWC/2010-the-spending-reports-appears-to-be-broken-empty-file)

spec/features/staff/users_can_view_reports_spec.rb
```diff
@@ -156,6 +161,9 @@
       expect(page.response_headers["Content-Type"]).to include("text/csv")
       header = page.response_headers["Content-Disposition"]
       expect(header).to match(/#{ERB::Util.url_encode("#{report.organisation.beis_organisation_reference}-report.csv")}\z/)
+
+      contents = CSV.parse(page.body, headers: true).map { |column| column["Title"] }
+      expect(contents).not_to be_empty
```
CristinaRO, line 166: If we're testing it, should we test a bit more than just "not empty"? How do we know it actually contains what it should? 😬

edavey: Fair point! I've used a more exacting assertion now...

CristinaRO (approved):

PR #1250, opened 2021-07-23
(2009) Auto deploy to pentest environment

pezholio:
This updates the deploy script to deploy to the pentest (training) environment at the same time as we deploy to production to ensure parity of feature set across the two environments.

~I haven't yet added the environment variables to Github, so this needs to be done first. Will update the PR when I'm done.~

I've tweaked the steps to run as seperate jobs for a couple of reasons:

- I (think?) it's more readable
- It allows us to run the prod and training deploys in parallel, so the deploy shouldn't be much slower than what we have already.

Totally happy to have a chat about how this all fits together - I am aware that I am the Github Actions "expert" here, so keen to pass on more knowlege if I can!

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -92,6 +92,38 @@ jobs:
       - uses: actions/checkout@v2
       - name: Deploy latest code to Staging
         run: script/deploy-terraform
+  deploy_to_training:
+    needs: build_and_push_to_dockerhub
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
+    env:
+      TF_VAR_docker_image: ${{ needs.build_and_push_to_dockerhub.outputs.docker_tag }}
+      TF_VAR_docker_username: ${{ secrets.DOCKER_USERNAME }}
+      TF_VAR_docker_password: ${{ secrets.DOCKER_PASSWORD }}
+      TF_VAR_environment: "pentest"
+      TF_VAR_secret_key_base: ${{ secrets.TRAINING_SECRET_KEY_BASE }}
+      TF_VAR_auth0_client_id: ${{ secrets.TRAINING_AUTH0_CLIENT_ID }}
+      TF_VAR_auth0_client_secret: ${{ secrets.TRAINING_AUTH0_CLIENT_SECRET }}
+      TF_VAR_auth0_domain: ${{ secrets.TRAINING_AUTH0_DOMAIN }}
+      TF_VAR_notify_key: ${{ secrets.TRAINING_NOTIFY_KEY }}
+      TF_VAR_notify_welcome_email_template: ${{ secrets.TRAINING_NOTIFY_WELCOME_EMAIL_TEMPLATE }}
+      TF_VAR_notify_view_template: "b541df04-add8-458e-a7d3-2e156386e150"
+      TF_VAR_rollbar_access_token: ${{ secrets.ROLLBAR_ACCESS_TOKEN }}
+      TF_VAR_skylight_access_token: ${{ secrets.SKYLIGHT_ACCESS_TOKEN }}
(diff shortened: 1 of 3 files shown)


pezholio:
[Auto deploy to pentest environment](https://trello.com/c/BL0klSEF/2009-auto-deploy-to-pentest-environment)

.github/workflows/deploy.yml
```diff
@@ -15,70 +15,85 @@ env:
...
+      TF_VAR_papertrail_destination: ${{ secrets.STAGING_PAPERTRAIL_DESTINATION }}
+      TF_VAR_google_tag_manager_container_id: ${{ secrets.STAGING_GOOGLE_TAG_MANAGER_CONTAINER_ID }}
+      TF_VAR_google_tag_manager_environment_auth: ${{ secrets.STAGING_GOOGLE_TAG_MANAGER_ENVIRONMENT_AUTH }}
+      TF_VAR_google_tag_manager_environment_preview: ${{ secrets.STAGING_GOOGLE_TAG_MANAGER_ENVIRONMENT_PREVIEW }}
+      TF_VAR_custom_domain: ${{ secrets.STAGING_CUSTOM_DOMAIN }}
+      TF_VAR_custom_hostname: ${{ secrets.STAGING_CUSTOM_HOSTNAME }}
+      TF_VAR_robot_noindex: "true"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Deploy latest code to Production
```
mec, line 93: Staging not Production?

script/deploy-terraform
```diff
@@ -3,8 +3,6 @@
 # exit on error or if a variable is unbound
 set -eu
 
-export TF_VAR_docker_image=$DOCKER_TAG
```
mec, line 6: Not needed becasue of line 36 and 68 right and the whole output from `build_and_push`?

.github/workflows/deploy.yml
```diff
@@ -15,70 +15,85 @@ env:
...
-          TF_VAR_custom_domain: ${{ secrets.PROD_CUSTOM_DOMAIN }}
-          TF_VAR_custom_hostname: ${{ secrets.PROD_CUSTOM_HOSTNAME }}
-          TF_VAR_robot_noindex: "false"
-        run: |
-          script/deploy-terraform
-        if: github.ref == 'refs/heads/master'
+      - name: Set Docker tag as an output
+        id: set_output
+        run: echo "::set-output name=docker_tag::${{ env.DOCKER_TAG }}"
+  deploy_to_production:
```
mec, line 31: I like the seperate jobs more! 👍

.github/workflows/deploy.yml
```diff
@@ -15,70 +15,85 @@ env:
   AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
 jobs:
-  deploy:
+  build_and_push:
```
mec, line 18: Is this 'build and push to docker hub'? Could it be called that to be clear where the push is to?

mec (approved):
This all makes sense to me as far as I can follow it 👍 

Did you get a sense for if/how we could change the actual environment on GPaas to 'training' or is that going to be an Ops/TF thing?

mec:
@pezholio just seen this https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance#training-and-penetration-testing can we update in this work to reflect the change?

script/deploy-terraform
pezholio: Yeah, exactly 👍

.github/workflows/deploy.yml
pezholio: Done! 👍

.github/workflows/deploy.yml
pezholio: Whoops, good catch!

PR #1251, opened 2021-07-23
(1997) Allow budgets to be destroyed

pezholio:
We want to make sure if there is an editable report, then budgets created erroneously can be deleted.

# Screenshots

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -762,8 +762,8 @@
 - Content and layout improvements to the actual spend tab on Report
 - Content and layout improvements to the actual spend upload page
 - Rework the way reports are shown to Delivery Partners and Service Owners
-
 - Fix: 'spending breakdown' download is no longer empty
+- Allow budgets to be deleted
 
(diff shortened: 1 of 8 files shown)


edavey (approved):
This all looks good, though I am a little confused by the language in the Trello card which says that a budget must be associated with a report which is either "active" or "awaiting changes".

Do those "active" and "awaiting changes" terms map on the logic which I see described in the `BudgetPolicySpec`:

- DP: where report is the one in which the budget was created
- BEIS: where activity is a fund or a programme

pezholio:
@edavey Yeah, I think so. A report is only considered "editable" by a DP if it is either "active" or "awaiting changes", otherwise it is locked for editing.

edavey:
> @edavey Yeah, I think so. A report is only considered "editable" by a DP if it is either "active" or "awaiting changes", otherwise it is locked for editing.

👍 the policy and its tests are really clear at describing what's permitted...

PR #1253, opened 2021-07-27
(1946) IATI export iteration

pezholio:
This PR moves the location of the XML exports from the Organisation view page to an Organisation's export page. It also moves the generation of the XML to the exports controller and cleans up/rationalises a bunch of things (including the XML tests, which were quite slow)

## Screenshots

image


pezholio:
[IATI export iteration](https://trello.com/c/yQjzugEk/1946-iati-export-iteration)

mec:
@pezholio i've given this a first pass, how do you feel about the following general points before I give it my all:

- align the export titles to IATI as that is the only thing they are used for. e.g. "Global Challenges Research Fund (GCRF) IATI export for level B (programme) activities" (changing the values as needed)
- is there a way to not show a export where a DP has no activities (e.g UKSA do not report on newton)?
- a changelog entry for when the UI moves

pezholio:
Cool - all good suggestions. Will get on it 👍

pezholio:
@mec This is ready for another looksee now

config/routes.rb
```diff
@@ -29,6 +29,9 @@
     namespace :exports do
       resources :organisations, only: [:show] do
         get "transactions", on: :member
+        get "programme_activities", on: :member
```
mec, line 32: Is it worth adding 'iati' to these, jsut to be super clear and keep them in a namespace of sorts?

app/views/staff/organisations/show.html.haml
```diff
@@ -42,36 +42,3 @@
 
       - if policy(:organisation).edit?
         = link_to t("page_content.organisation.button.edit_details"), edit_organisation_path(@organisation_presenter), class: "govuk-button"
-
```
mec, line 45: :heart:

.gitignore
```diff
@@ -135,3 +135,5 @@ terraform/.terraform/*
 /yarn-error.log
 yarn-debug.log*
 .yarn-integrity
+
+!app/views/**/*.xml.haml
```
mec, line 139: 👍

mec (approved):
I have one comment about the routes, but it shouldn't block this. Sorry for the unclear ealy feedback - thanks for taking it onboard!

config/routes.rb
pezholio: Good plan 👍 I've also namespaced the model, just to be consistent

PR #1256, opened 2021-07-27
(1964) improve the budget report tab

pezholio:
This is just some small text tweaks to the budget tab within reports

image

### DIFF
--- a/config/locales/models/budget.en.yml
+++ b/config/locales/models/budget.en.yml
@@ -47,6 +47,7 @@ en:
     budgets:
       button:
         create: Add budget
+      no_budgets: There are no budgets created during this reporting cycle.
   page_title:
     budget:
       edit: Edit budget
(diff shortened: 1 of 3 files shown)


pezholio:
[Improve the budget report tab](https://trello.com/c/Xm873qWj/1964-improve-the-budget-report-tab)

config/locales/models/budget.en.yml
```diff
@@ -47,6 +47,7 @@ en:
     budgets:
       button:
         create: Add budget
+      no_budgets: There are no budgets created during this reporting cycle.
```
CristinaRO, line 50: Should this say "created or updated", in line with what the "what it shows" copy says? (If a budget had been created during *another* reporting cycle, and updated during the current one, would it be listed?)

CristinaRO (approved):
A non-blocking question about the copy.

config/locales/models/budget.en.yml
pezholio: I've just had a look, and the budget is always tied to the report, so the introductory copy is wrong! Will fix 👍

PR #1260, opened 2021-07-29
(1977) Improve the report variance tab

pezholio:
This improves the report variance tab to only show activities that actually have variance, as well as showing the total. I've also made a bunch of changes to the view to (hopefully) make the tab a bit more approachable.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -770,6 +770,7 @@
 - Move IATI XML exports to the exports section
 - Collect all benefitting countries in one step
 - Benefitting countries can be imported via CSV upload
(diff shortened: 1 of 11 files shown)


pezholio:
[Improve the report variance tab](https://trello.com/c/BVdHcYRG/1977-improve-the-report-variance-tab)

mec (dismissed):
I've got one comment about a method name that I feel might improve the readability.

It would also be good to add a changelog to the last commit where we change what the view shows so that BEIS can understand what has changed.

Other than that, this is great stuff!

CristinaRO (approved):
Thank you! 👍🏻

PR #1262, opened 2021-07-29
(2033) Remove role=navigation from reports tabs

pezholio:
This has the effect of invalidating the HTML code. Screen readers notify users when they come to a list, and tell them how many items are in a list. If you don't mark up a list using proper semantic markup in a hierarchy, list items cannot inform the listener that they are listening to a list when no parent is indicating the presence of a list and the type of list.


CristinaRO (approved):
I don't fully understand when and how it *would* be useful to mark something *as* navigation, and whether using lists, which then have to be announced as list items, for sub-navigation is OK accessibility-wise in the first place, but given that this requirement has come from our a11y experts I'm good to go with it and research more later. :)

PR #1264, opened 2021-08-02
Update Organisation format error message

pezholio:
The existing error message was misleading


CristinaRO:
[Add new delivery partner short name error is incorrect](https://trello.com/c/EXg9ZkoJ/1960-add-new-delivery-partner-short-name-error-is-incorrect)

CristinaRO (commented):
Would it be worth having a unit spec for this specific kind of validation? The current spec only checks that we have presence validation.

pezholio:
Good plan! Done 👍

CristinaRO (approved):
Thank you! 👍🏻

PR #1268, opened 2021-08-03
Release 65

pezholio:
- Move IATI XML exports to the exports section
- Collect all benefitting countries in one step
- Benefitting countries can be imported via CSV upload
- Tidy up Report Variance tab to only show activities that have a varaince


CristinaRO (approved):
If you have the time and inclination, you could fix the "varaince" typo in the changelog, but it's not a blocker. (I should have spotted it earlier!)

PR #1269, opened 2021-08-03
Make sure users are active even after login

pezholio:
Currently, users that are made inactive, but still logged in can access parts of the site while their session remains valid (i.e. they don't logout, or their session doesn't time out). To get around this edge case, when fetching the `current_user`, we limit our query to only current users. If an active user is not found, we invalidate the session and return an error message

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -774,6 +774,8 @@
 
 ## [unreleased]
 
+- Make sure users are active even after login
+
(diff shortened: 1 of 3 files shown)


pezholio:
[Inactive users can still view parts of the site](https://trello.com/c/bBeyo17T/1942-inactive-users-can-still-view-parts-of-the-site)

CristinaRO (approved):

PR #1271, opened 2021-08-04
(1982) Adjust XML exports to use new benefitting countries field

pezholio:
This adjusts the XML exports to use the new `benefitting_countries` field, rather than the old, legacy `recipient_country` / `recipient_region` fields

### DIFF
--- a/app/helpers/activity_helper.rb
+++ b/app/helpers/activity_helper.rb
@@ -1,4 +1,6 @@
 module ActivityHelper
+  include CodelistHelper
+
   def step_is_complete_or_next?(activity:, step:)
     steps = Activity::FORM_STEPS
 
@@ -32,5 +34,23 @@ module ActivityHelper
 
   def policy_markers_desertification_iati_codes_to_enum(code)
     Activity::DESERTIFICATION_POLICY_MARKER_CODES.key(code.to_i)
+  end
+
+  def benefitting_countries_with_percentages(benefitting_countries)
+    return [] if benefitting_countries.blank?
+
+    # Get an equal percentage split between all the countries
+    # (together with the remainder if possible)
+    percentage, remainder = 100.divmod(benefitting_countries.count)
+
+    benefitting_countries.map do |country|
+      # If we're at the last item, add the remainder to the percentage
+      # split
+      percentage += remainder if country == benefitting_countries.last
+
+      OpenStruct.new(code: country,
+                     name: country_name_from_code(country),
+                     percentage: percentage.to_f)
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


pezholio:
[Adjust all export project(s) as XML journeys to use new 'beneficiary_countries' field](https://trello.com/c/sOKmNGbf/1982-adjust-all-export-projects-as-xml-journeys-to-use-new-beneficiarycountries-field)

app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -39,12 +39,10 @@
     %website= I18n.t("contact_info.website")
     %mailing-address
       %narrative= I18n.t("contact_info.mailing_address")
-  - if activity.recipient_country?
```
CristinaRO, line 42: I think my original interpretation of these "additive" steps to implementing `benefitting_countries` was to use the info *if* we have it, but if `benefitting_countries` aren't present, default to showing the legacy `recipient_*` fields until we officially remove them.

This is intended to cope with the intermediate phase when we haven't backfilled the values for the new field.

I'd have done this by using `benefitting_countries` if present, and `elsif`ing to use `recipient_country` or `recipient_region`, rather than deleting them.

app/helpers/activity_helper.rb
```diff
@@ -33,4 +35,20 @@ def policy_markers_iati_codes_to_enum(code)
   def policy_markers_desertification_iati_codes_to_enum(code)
     Activity::DESERTIFICATION_POLICY_MARKER_CODES.key(code.to_i)
   end
+
+  def benefitting_countries_with_percentages(benefitting_countries)
+    # Get an equal percentage split between all the countries
+    # (together with the remainder if possible)
+    percentage, remainder = 100.divmod(benefitting_countries.count)
```
CristinaRO, line 42: At the moment we haven't added presence validation for `benefitting_countries`, while we have both the new and old fields being collected.

Can this cope with the number being `0`?

CristinaRO (commented):
I have a few questions before deciding how to go ahead.

app/helpers/activity_helper.rb
pezholio: I've tweaked this to return an empty array if `benefitting_countries` is blank now 👍

app/views/staff/shared/xml/_activity.xml.haml
pezholio: Yeah, good call. Makes sense to do all this piecemeal. Sorted now!

CristinaRO (approved):
Thank you for making the changes!

I've rebased to fix the conflict in the changelog.

PR #1272, opened 2021-08-04
(1985) Backfill benefitting_countries where possible

CristinaRO:
## Changes in this PR

**ETA**: There are activities which have `geography=recipient_region` but also have `recipient_country` and/or `intended_beneficiaries`.

There are activities which have `geography=recipient_country` but also have `recipient_region`.

While the two geography options should have been mutually exclusive, the reality is that users have managed to fill in more data than that, and we don’t want to risk losing it.

Our choices were to only use the data corresponding to the geography - and risk discarding some data; use any and all data we can find in the records - and risk having the wrong data; run a report for any ambiguous cases - and risk delaying the process indefinitely.

Following a discussion with Meyric, we have chosen the second option.

The legacy fields will not be removed until BEIS are satisfied that the data is in the shape they want.

Regardless of the value of `geography`, as long as it is not `nil`:

- If there is a `recipient_country`, and optionally one or more `intended_beneficiaries`, we populate `benefitting_countries` by concatenating all of them into one field.

- If there is a `recipient_region`, we *also* add to `benefitting_countries` all the countries in that region or its subregions.

The only case still uncovered is
```
 {"code"=>"998", "name"=>"Developing countries, unspecified"}]
```

We output a list of the activities where we couldn't automatically populate the field.

## Next steps

### DIFF
--- a/db/data/20210804154421_backfill_benefitting_countries.rb
+++ b/db/data/20210804154421_backfill_benefitting_countries.rb
@@ -10,18 +10,20 @@ activities_to_update.each_with_index do |activity, counter|
     puts "#{counter + 1} of #{total_activities_to_update}"
   end
 
-  if activity.geography == "recipient_country"
-    recipient_country = activity.recipient_country
-    intended_beneficiaries = activity.intended_beneficiaries || []
+  # We want all the stored information:
+  # - recipient country
+  # - intended beneficiaries
+  # - countries belonging to the recipient region
+  # regardless of whether the geography is recipient_country or recipient_region
 
-    activity.benefitting_countries = ([recipient_country] + intended_beneficiaries).compact.uniq
-    activity.save(validate: false)
-  elsif activity.geography == "recipient_region"
-    recipient_region = activity.recipient_region
+  recipient_country = activity.recipient_country
+  intended_beneficiaries = activity.intended_beneficiaries || []
 
-    next if recipient_region.nil? || recipient_region == "998"
-
+  recipient_region = activity.recipient_region
+  region_countries = []
+  unless recipient_region.nil? || recipient_region == "998"
     recipient_regions = if !benefitting_countries_hash.key?(recipient_region)
+      # We want to collect all the countries in the supra-region, by adding up the subregions' countries
       regions = [recipient_region]
       loop do
         subregions = regions.map { |r| region_to_subregion_hash.fetch(r) }.flatten
@@ -34,13 +36,14 @@ activities_to_update.each_with_index do |activity, counter|
       [recipient_region]
     end
 
-    countries = recipient_regions.map { |region|
+    region_countries = recipient_regions.map { |region|
       benefitting_countries_hash.fetch(region, [])
-    }.flatten
-
-    activity.benefitting_countries = countries.map { |c| c["code"] }
-    activity.save(validate: false)
+    }.flatten.map { |c| c["code"] }
   end
+
+  benefitting_countries = ([recipient_country] + intended_beneficiaries + region_countries).compact.uniq
+  activity.benefitting_countries = benefitting_countries if benefitting_countries.present?
+  activity.save(validate: false)
 end
 
 if (remaining_activities = Activity.where.not(geography: nil).where(benefitting_countries: nil)).present?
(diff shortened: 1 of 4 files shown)


db/data/20210804154421_backfill_benefitting_countries.rb
```diff
@@ -0,0 +1,51 @@
+# Run me with `rails runner db/data/20210804154421_backfill_benefitting_countries.rb`
+
+benefitting_countries_hash = Codelist.new(type: "intended_beneficiaries").list
+region_to_subregion_hash = Codelist.new(type: "region_to_subregion_mapping", source: "beis").list
+activities_to_update = Activity.where.not(geography: nil).where(benefitting_countries: nil)
+total_activities_to_update = activities_to_update.count
+
+activities_to_update.each_with_index do |activity, counter|
```
pezholio, line 8: I wonder if this might be quicker if we updated all the recipient countries first, then looped through the regions in turn and did a bulk update on each set of activities with a given region?

CristinaRO:
[Backfill 'beneficiary_countries' for the 92 activities lacking any country](https://trello.com/c/J83mB4M0/1985-backfill-beneficiarycountries-for-the-92-activities-lacking-any-country)

db/data/20210804154421_backfill_benefitting_countries.rb
pezholio: Something like this:

```
recipient_country_activities = Activity.where(geography: "recipient_country")

recipient_country_activities.each do |activity|
    recipient_country = activity.recipient_country
    intended_beneficiaries = activity.intended_beneficiaries || []

    activity.benefitting_countries = ([recipient_country] + intended_beneficiaries).compact.uniq
    activity.save(validate: false)
end

low_level_region_activities = Activity.where(geography: "recipient_region", recipient_region: benefitting_countries_hash.keys).group_by(&:recipient_region)

benefitting_countries_hash.each do |region, countries|
   next if low_level_region_activities[region].blank?

   activities = Activity.where(id: low_level_region_activities[region].pluck(:id))
   activities.update_all(benefitting_countries: countries.map { |c| c["code"] })
end

other_region_activites =  Activity.where(geography: "recipient_region", recipient_region: region_to_subregion_hash.keys).group_by(&:recipient_region)

region_to_subregion_hash.each do |region, subregions|
   next if other_region_activites[region].blank?

  activities = Activity.where(id: other_region_activites[region].pluck(:id))
  countries = subregions.map { |region| benefitting_countries_hash.fetch(region, {}) }.flatten
   activities.update_all(benefitting_countries: countries.map { |c| c["code"] }.reject(&:blank?))
end

```

CristinaRO: Please review the update, [in light of this thread](https://dxw.slack.com/archives/CN0H2L066/p1628602900114600?thread_ts=1628600547.107200&cid=CN0H2L066) 😅

pezholio: Just made a tweak that should do it

CristinaRO: What's the tweak?

pezholio: Just matching against the presence (or otherwise) of `recipient_country` / `recipient_region` and removing the specific check for geography

pezholio: Comments should have git history applied to them 😆

CristinaRO: Oh I see. That doesn't solve the issue that my latest commit was trying to address, which is to add up *all* the data we find, including when an activity has *both* recipient country and region.

If we set `benefitting_countries` when using the region, we overwrite whatever came from the recipient country and intended beneficiaries, which might not be perfectly included in the region.

pezholio: Oh course - that's annoying. I'll give it a bit more thought after lunch

pezholio (approved):
LGTM - I think, as you say, my suggestion falls down if we want to keep all the data around, so let's stick with how you've solved it. I don't think it's worth optimising any more for performance given it's reasonably fast anyway, and it's a one shot script

PR #1274, opened 2021-08-05
(1986) Remove legacy geography steps

CristinaRO:
## Changes in this PR
- No longer collect geography, recipient country, recipient region, and intended beneficiaries (deprecated by collecting benefitting countries)

NB: We are deliberately not removing them from the views and from the exports, to allow BEIS and the DPs to see the old data until everyone is satisfied that the data migration has not lost any valuable information.

## Next steps


CristinaRO:
[Remove "legacy" geography steps](https://trello.com/c/t6rVxUGs/1986-remove-legacy-geography-steps)

pezholio (approved):
LGTM - I think this is a good way forward and will allow us to verify that data is sound before we delete the columns once and for all

PR #1275, opened 2021-08-09
Add horizontal scroll inside wrapper

isratc:
## Changes in this PR
Table overflows the normal device veiwport, causing the browser to add a horizontal scroll on the website, this is an accessibility issue. I have styled the data tables so that it doesn't break from its container and scrolls within its wrapper div

trello card: https://trello.com/c/fh8On5Hg/2032-moderate-page-reflow-issues-horizontal-scroll

## Screenshots of UI changes
Screenshot 2021-08-09 at 13 36 26

### Before
Screenshot 2021-08-09 at 14 45 18

### DIFF
--- a/app/assets/stylesheets/partials/_table.scss
+++ b/app/assets/stylesheets/partials/_table.scss
@@ -1,0 +1,10 @@
+table.govuk-table {
+  @include govuk-media-query($until: tablet) {
+    display: block;
+  }
+  overflow-x: scroll;
+}
+
+.govuk-template {
+  overflow-x:hidden;
+}


app/assets/stylesheets/partials/_table_tree_view.scss
```diff
@@ -63,3 +63,13 @@ $table-tree-view-gutter: 1.5em;
...
+
+
+table.govuk-table {
+  display: block;
+  overflow-x: scroll;
+}
+
+.govuk-template {
+    overflow-x:hidden;
+}
```
pezholio, line 75: Could we have this in a different partial? (called _table.scss or similar). This partial refers to the tree view of activities (with the drop down list of activites), so there's a degree of misdirection here

pezholio:
Thanks for doing this. The only worry I have is that setting the tables to `display: block` causes them to contract in width, so they don't always take up the full width of the container - e.g:

## Before:

image

## After:

image

isratc:
I have updated against the feedback please have a look

app/assets/stylesheets/partials/_table.scss
```diff
@@ -0,0 +1,10 @@
+table.govuk-table {
+  @include govuk-media-query($until: tablet) {
+  display: block;
+}
```
pezholio, line 4: Can we just sort the spacing out for readability please?

```suggestion
  @include govuk-media-query($until: tablet) {
    display: block;
  }
```

app/assets/stylesheets/partials/_table.scss
```diff
@@ -0,0 +1,10 @@
...
+  overflow-x: scroll;
+}
+
+.govuk-template {
+    overflow-x:hidden;
```
pezholio, line 9: Also, another tiny spacing thing:

```suggestion
  overflow-x:hidden;
```

pezholio:
@isratc One more thing (sorry!) Can you squash these commits together please?

pezholio (approved):
LGTM! 👍

PR #1277, opened 2021-08-10
Release 66

pezholio:
- Make sure users are active even after login
- Add breadcrumb trail for activities
- Adjust XML exports to use new benefitting countries field


mec (approved):
:ship:

PR #1285, opened 2021-08-13
(2089) Breadcrumbs should be truncated if over a certain length number of items

pezholio:
This truncates breadcrumbs with an elipsis (`...`) if they are over a certain width, and adds a title to all the links, so hovering over a link shows the full name of the item.

# Screenshots

## Before

image

## After

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -789,6 +789,7 @@
 
 - Allow a refund to be posted against an active report
 - Activity breadcrumb trail for BEIS users includes the delivery partner organisation
+- Truncate breadcrumbs that are over a certain length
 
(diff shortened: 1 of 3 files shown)


pezholio:
[Breadcrumbs should be truncated if over a certain length / number of items](https://trello.com/c/9Qjg7MvA/2089-breadcrumbs-should-be-truncated-if-over-a-certain-length-number-of-items)

CristinaRO (approved):
Clever solution with the CSS ellipses! (Let's hope it won't randomly break words at embarrassing points... ;) )

Are there any solutions equivalent to the title-on-hover for people who use keyboard only? Any accessibility impact for screen readers?

pezholio:
Nah, this should be fine because the truncation is only set via CSS. The full text is still visible in the HTML, so will be read out in full by screen readers.

PR #1286, opened 2021-08-13
(2038) *Critical* ARIA attributes must conform to valid values

pezholio:
Previously, the `aria-controls` attribute of the button was set to a value that didn't exist on the page. The purpose of `aria-controls` is to show assistive technology what element(s) are controlled by a given element (https://www.w3.org/TR/wai-aria-1.1/#aria-controls).

In our case, this is the child activities that are shown when clicking on the "expand" button. With this in mind, I'm grabbing all the child activity elements, and adding a space seperated list of the child IDs (`aria-controls` accepts a list of IDs, as well as a single ID). This tells assistive technology that clicking this button will show/hide those elements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -790,6 +790,7 @@
 - Allow a refund to be posted against an active report
 - Activity breadcrumb trail for BEIS users includes the delivery partner organisation
 - Truncate breadcrumbs that are over a certain length
+- Fix `aria-controls` element in the tree view
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-67...HEAD
 [release-67]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-66...release-67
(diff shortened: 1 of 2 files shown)


pezholio:
[*Critical* ARIA attributes must conform to valid values](https://trello.com/c/B5v1JWHf/2038-critical-aria-attributes-must-conform-to-valid-values)

edavey (approved):
this looks good to me. I see the `aria-controls` attribute set as expected, e.g.
```
<button 
  type="button" 
  id="activity-dfb01990-0ca3-4066-a8f0-7256a8d20fa4-heading-1" aria-controls="activity-5dfdef4f-01d0-4275-82c5-c02d06b94bc6 activity-7df2e24b-fd42-423d-b00f-7dcc5c17b366 activity-03bfae40-b308-4ab4-a183-f4b328177d61" 
  class="table-tree-view__parent-button" 
  aria-expanded="false">
AMS Coherence and Impact - Global Health Policy Workshops
  <span class="table-tree-view__icon" aria-hidden="true">
  </span>
</button>
```

I haven't tested in a screen reader, but would be interested to see how these options are represented....

PR #1293, opened 2021-08-17
(2061) Show report breadcrumb scheme when drilling down from a report

pezholio:
This PR adds a breadcrumb when drilling down to a report from the report listing page, and also adds a `BreadcrumbContext` class that allows users to drill down to an Activity from a report whilst still keeping the breadcrumbs within the context of the report.

I've also made a not insignificant refactor to gather all the activity-related controllers and put them into one controller, which makes some of this work easier, as well as tidying up the codebase and making the location of the controller code for activities much easier to find.

# Screenshots

image

image

image


pezholio:
[Show "Report" breadcrumb scheme when drilling down from a report](https://trello.com/c/v2XTTo0t/2061-show-report-breadcrumb-scheme-when-drilling-down-from-a-report)

app/services/activity/tab.rb
```diff
@@ -0,0 +1,80 @@
+class Activity
+  class Tab
+    include Pundit
```
edavey, line 3: I'm a bit uneasy with such a complex and clever abstraction of the activity tabs:

- use of `send`
- use of `instance_variable_get`
- dynamically generated routes

BUT: you explain that this is a reaction to real pain experienced maintaining these views, so yes, let's give it a go!

edavey (approved):
👍 really nice work! 

I can't find anything to tweak. 

I do have some reservations about the slightly magical `Activity::Tab`, but very happy to see how it goes with less controllers.

PR #1299, opened 2021-08-19
Remove Bullet deprecation warning

pezholio:
`add_whitelist` has been replaced with `add_safelist`, so this removes the deprecation warning when running the tests.


CristinaRO (approved):
👍🏻  (needs rebasing, will attempt the automated one but just in case it fails)

PR #1300, opened 2021-08-19
(2064) Ensure all other views have a sensible breadcrumb

pezholio:
This looks like a massive PR, but the commits are all pretty tiny, so should be OK to review. This PR makes sure we have breadcrumbs for the other views in the application. I'd add screenshots, but there's rather a lot to show - it might be better to pull this branch down and click through the various different parts of the application to check I haven't missed anything.


CristinaRO (approved):
I have a comment/question, but on the whole this is great work.

I'm inclined to approve it as it is and let users surface any unforeseen needs that might differ from what we assumed was a useful crumb trail.

pezholio:
Cool, that sounds like a good idea - There's a bunch of things I'm a bit iffy about, so it would be good to get what we have so far in front of users, so we can iterate based on feedback. Ta for the review! 👍

PR #1306, opened 2021-08-23
Fix automatic rebase

pezholio:
We have an action that _should_ automatically rebase any PRs that are set to automerge against the develop branch. However, this has not been working, because of a [bug in the rebase action](https://github.com/cirrus-actions/rebase/issues/58) we use (example run here https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/runs/3398039357?check_suite_focus=true)

I've also tweaked the action to only trigger against the oldest PR set to automerge rather than needlessly rebasing all the PRs set to automerge (which will then go out of date). Once that one is merged, it will trigger a rebase on the second oldest one and cascade down, saving us time and resources.

### DIFF
--- a/.github/workflows/trigger_rebase.yml
+++ b/.github/workflows/trigger_rebase.yml
@@ -36,7 +36,7 @@ jobs:
           fetch-depth: 0
           ref: ${{ env.ref }}
       - name: Automatic Rebase
-        uses: dxw/rebase@master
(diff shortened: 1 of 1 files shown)


pezholio:
[Fix the autorebase action](https://trello.com/c/aghvV8y8/2114-fix-the-autorebase-action)

CristinaRO (approved):

PR #1308, opened 2021-08-23
Data migration sets nil policy markers to default

pezholio:
This adds a data migration that sets any policy markers set to `nil` to the default state, which should be `not_assessed`. This will fix the issue where we see blank policy markers in the XML.


CristinaRO:
[Empty policy markers in IATI XML](https://trello.com/c/g0Ram7p6/2057-empty-policy-markers-in-iati-xml)

edavey (approved):
👍 this looks good.

PR #1311, opened 2021-08-23
(2100) Rename "historic" to "approved" in the context of reports

CristinaRO:
## Changes in this PR
- Replace the use of "historic" with "approved" in the context of reports

### After
Screenshot 2021-08-24 at 14 18 23
Screenshot 2021-08-24 at 14 18 12

## Next steps

### DIFF
--- a/config/locales/models/report.en.yml
+++ b/config/locales/models/report.en.yml
@@ -10,7 +10,6 @@ en:
         awaiting_changes: Reports awaiting changes
         approved: Approved reports
         current: Current reports
-        historic: Historic reports
     header:
       report:
         financial_quarter: Financial quarter
@@ -41,8 +40,7 @@ en:
       variance: Variance
       budgets: Budgets
       current: Current
-      historic: Historic
-
+      approved: Approved
   page_content:
     reports:
       title: Reports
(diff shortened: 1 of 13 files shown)


CristinaRO:
[DP sees status module on Reports | Historic tab](https://trello.com/c/Q9nNngkF/2100-dp-sees-status-module-on-reports-historic-tab)

CristinaRO:
I'll fix the failing test if the feature as a whole is indeed needed.

config/locales/models/report.en.yml
```diff
@@ -42,7 +42,7 @@ en:
       budgets: Budgets
       current: Current
       historic: Historic
```
pezholio, line 44: Do we need this translation now?

pezholio (approved):
One tiny thing, but I could be missing some context, so feel free to merge if I'm way off 😃

config/locales/models/report.en.yml
CristinaRO: I've done a few searches and it looks like it's not used anymore, although of course it's sometimes hard to detect if a string isn't being invoked via interpolation.

I've also renamed some backend code, for consistency and to avoid building up debt with one hand as we remove it with other, if you wouldn't mind having another look to make sure it looks ok?

PR #1316, opened 2021-08-24
Switch out the aid type features for requests

pezholio:
This switches out the feature specs that test specific behaviours when aid types are set for much faster request specs. We have coverage for much of this behaviour elsewhere in the test suite and the feature specs are extremely slow.

I've also added some faster controller tests to test when particular steps get skipped or shown, which was previously tested in a bunch of conditionals in the form helper `fill_in_activity_form` method.

## Before

`Finished in 4 minutes 49.4 seconds (files took 8.06 seconds to load)`

## After 

`Finished in 4 minutes 13.7 seconds (files took 8.03 seconds to load)`


pezholio:
[Switch out the aid type features for requests](https://trello.com/c/4E9H37d7/2128-switch-out-the-aid-type-features-for-requests)

edavey (approved):
This all looks good to me, clearly you've considered carefully where behaviour is already (or can be) tested more efficiently.

PR #1323, opened 2021-08-27
Email notification documentation

mec:
I wanted to see the emails the applicaiton sends so I looked for our
docs on the subject, as there weremn't any, I wrote this up as a
starter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,11 @@ The app is currently hosted on GPaaS: [https://www.report-official-development-a
 
 The `master` branch is deployed to production after a successful build via GitHub Actions.
 
+## Email notifications
+The service sends email notifications see the
+[documentation](/doc/email-notificaitons.md) for more
+details.
+
 ## DNS
 
 The DNS for the service is hosted and managed by [dxw](https://dxw.com) the
--- a/doc/email-notifications.md
+++ b/doc/email-notifications.md
@@ -1,0 +1,58 @@
+# Email notifications
+The application sends various notifications by email, this document outlines how
+this works from a technical standpoint.
+
+## Account emails
+Most emails that relate to user accounts are sent by Auth0.
+
+All others are sent by the applicaiton via GOV.UK Notify.
+
+## Mail Notify Gem
+We use this gem to interact with the GOV.UK Notify service.
+
+[https://github.com/dxw/mail-notify](https://github.com/dxw/mail-notify)
+
+## GOV.UK notify service
+[https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e](https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e)
+
+BEIS owned so access can be granted by them.
+
+You will need the API key in the NOTIFY_KEY environment variable. The key for
+each environment can be found here https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e/api/keys
+
+## Templates
+We use two templates to send email notifications:
+
+- Welcome email
+- View email
+
+See the templates at https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e/templates
+
+The templates ids are supplied as environment variables:
+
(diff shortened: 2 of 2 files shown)


doc/email-notificaitons.md
```diff
@@ -0,0 +1,45 @@
...
+You will need the API key in the NOTIFY_KEY environment variable. The key for
+each environment can be found here https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e/api/keys
+
+## Templates
+We use two templates to send email notifications:
+
+- Welcome email
+- Notifcation email
+
+See the tempaltes at https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e/templates
```
pezholio, line 29:
```suggestion
See the templates at https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e/templates
```

doc/email-notificaitons.md
```diff
@@ -0,0 +1,45 @@
...
+
+## GOV.UK notify service
+[https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e](https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e)
+
+BEIS owned so access can be granted by them.
+
+You will need the API key in the NOTIFY_KEY environment variable. The key for
+each environment can be found here https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e/api/keys
+
+## Templates
```
pezholio, line 23: It might be worth mentioning how we use templates here (i.e. we send the entire template to Notify, rather than having a template with variables stored in Notify)

mec: I didn't know this? I thought we sent the entire body but there is still a templat in Notify?

pezholio: Oh there is, but there's just one variable in it called `body` or something similar

pezholio: https://github.com/dxw/mail-notify#with-a-view

edavey:
@mec it looks like you've committed an empty file called `.quarterly_transaction_export_spec.rb@neomake_65300_1.rb` by mistake?

doc/email-notificaitons.md
```diff
@@ -0,0 +1,58 @@
+# Email notifications
```
edavey, line 1: tiny typo on the name of the markdown file itself...

doc/email-notificaitons.md
```diff
@@ -0,0 +1,58 @@
...
+each environment can be found here https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e/api/keys
+
+## Templates
+We use two templates to send email notifications:
+
+- Welcome email
+- View email
+
+See the templates at https://www.notifications.service.gov.uk/services/15322373-3ee0-4c9c-aef7-be3e967f2f3e/templates
+
```
edavey, line 30: is it ok to expose this ID in the repo?

doc/email-notificaitons.md
```diff
@@ -0,0 +1,58 @@
...
+variables set.
+
+### Welcome email
+The welcome email is sent to new users and takes a series of variables, changes
+to the content are made in the Notify service.
+
+### View email
+The Mail Notify gem allows us to assemble the entire body of the email in our
+code which gives us greater flexibility of the content, see
+[https://github.com/dxw/mail-notify#with-a-view](https://github.com/dxw/mail-notify#with-a-view)
```
pezholio, line 46: 👍

doc/email-notificaitons.md
mec: I think so, my rational is:

- it is behind a secure login
- the id is for the service account and could just as easily be 'beis-roda'
- linking to the exact templates benefits the user, they can be confident they have the right account and it would idicate if they do not have access to the right account

Happy to take it out if you are converned though?

pezholio: I think this is fine. We have the UUIDs exposed in the example `.env` files. There's nothing any potential attacker can do with these.

edavey (approved):

PR #1326, opened 2021-08-27
(2117) BEIS users can export external income

CristinaRO:
## Changes in this PR
- BEIS users can export external income per organisation and fund
 
## Next steps

### DIFF
--- a/app/controllers/staff/exports/organisations_controller.rb
+++ b/app/controllers/staff/exports/organisations_controller.rb
@@ -58,6 +58,20 @@ class Staff::Exports::OrganisationsController < Staff::BaseController
     render_xml
   end
 
+  def external_income
+    fund = Fund.new(params[:fund_id])
+
+    respond_to do |format|
+      format.csv do
+        export = QuarterlyExternalIncomeExport.new(@organisation, fund)
+
+        stream_csv_download(filename: export.filename, headers: export.headers) do |csv|
+          export.rows.each { |row| csv << row }
+        end
+      end
+    end
+  end
+
   private
 
   def fund_code
--- a/app/services/quarterly_external_income_export.rb
+++ b/app/services/quarterly_external_income_export.rb
@@ -1,0 +1,80 @@
+class QuarterlyExternalIncomeExport
+  HEADERS = [
+    "RODA identifier",
+    "Delivery partner identifier",
+    "Delivery partner organisation",
+    "Title",
+    "Level",
+    "Providing organisation",
+    "ODA",
+  ]
+
+  def initialize(organisation, source_fund)
+    @organisation = organisation
+    @source_fund = source_fund
+  end
+
+  def headers
+    return HEADERS if external_incomes.empty?
+
+    HEADERS + financial_quarter_range.map(&:to_s)
+  end
+
+  def rows
+    external_incomes.map do |record|
+      activity_data(record.activity) + external_income_data(record) + fq_data(record)
+    end
+  end
+
+  def filename
+    "#{@source_fund.short_name}_#{@organisation.beis_organisation_reference}_external_income.csv"
+  end
+
+  private
+
+  def activity_data(activity)
+    [
+      activity.roda_identifier,
+      activity.delivery_partner_identifier,
+      @organisation.name,
+      activity.title,
+      I18n.t("table.body.activity.level.#{activity.level}"),
+    ]
+  end
+
+  def external_income_data(external_income)
+    [
+      external_income.organisation.name,
+      I18n.t("table.body.external_income.oda_funding.#{external_income.oda_funding?}"),
+    ]
+  end
+
+  def fq_data(external_income)
+    return [] if external_incomes.empty?
+
+    financial_quarter_range.map do |quarter|
+      value = if external_income.own_financial_quarter == quarter
+        external_income.amount
+      else
+        0
+      end
+      "%.2f" % value
+    end
+  end
+
+  def external_incomes
+    @_external_incomes ||= ExternalIncome.includes(:activity, :organisation).where(activity_id: activity_ids).order(:activity_id, :financial_year, :financial_quarter)
+  end
+
+  def activity_ids
+    Activity.where(organisation_id: @organisation.id, source_fund_code: @source_fund.id).pluck(:id)
+  end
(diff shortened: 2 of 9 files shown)


app/services/quarterly_external_income_export.rb
```diff
@@ -0,0 +1,58 @@
...
+        record.organisation.name,
+        I18n.t("table.body.external_income.oda_funding.#{record.oda_funding?}"),
+      ] + fq_row(record)
+    end
+  end
+
+  private
+
+  def fq_row(activity_external_income)
+    # return [] if @activity_external_income.empty?
```
pezholio, line 43: Did you mean to leave this in?

app/services/quarterly_external_income_export.rb
```diff
@@ -0,0 +1,58 @@
...
+    "Providing organisation",
+    "ODA",
+  ]
+
+  def initialize(organisation, source_fund_code)
+    @organisation = organisation
+    @source_fund_code = source_fund_code
+    activity_ids = Activity.where(organisation_id: @organisation.id, source_fund_code: @source_fund_code).pluck(:id)
+    @activity_external_income = ExternalIncome.includes(:activity, :organisation).where(activity_id: activity_ids).order(:activity_id, :financial_year, :financial_quarter)
+    @financial_quarters = @activity_external_income.map(&:own_financial_quarter).uniq
```
pezholio, line 17: I prefer to not do anything that calls the database in an initializer, so here `activity_external_income` and `financial_quarters` would be private methods.

app/services/quarterly_external_income_export.rb
```diff
@@ -0,0 +1,58 @@
...
+    @activity_external_income.map do |record|
+      [
+        record.activity.roda_identifier,
+        record.activity.delivery_partner_identifier,
+        @organisation.name,
+        record.activity.title,
+        I18n.t("table.body.activity.level.#{record.activity.level}"),
+        record.organisation.name,
+        I18n.t("table.body.external_income.oda_funding.#{record.oda_funding?}"),
+      ] + fq_row(record)
```
pezholio, line 36: At first, I missed the call to `fq_data` here. It might be clearer to break this array out into a method of its own (called something like `activity_data`), so you have something like:

```
@activity_external_income.map do |record|
  activity_data(record) + fq_data(record)
end
```

app/controllers/staff/exports/organisations_controller.rb
```diff
@@ -58,6 +58,20 @@ def third_party_project_activities
...
   end
 
+  def external_income
+    fund = Fund.new(params[:fund_id])
+
+    respond_to do |format|
+      format.csv do
+        export = QuarterlyExternalIncomeExport.new(@organisation, fund.id)
+
+        stream_csv_download(filename: "#{fund.short_name}_#{@organisation.beis_organisation_reference}_external_income.csv", headers: export.headers) do |csv|
```
pezholio, line 68: Could we have the filename as a method in the `QuarterlyExternalIncomeExport` class?

config/locales/models/activity.en.yml
```diff
@@ -385,7 +385,7 @@ en:
     activity:
       index: Activities
       implementing_organisation:
-        edit: Edit implemening organisation
+        edit: Edit implementing organisation
```
pezholio, line 388: 👍

spec/services/quarterly_external_income_export_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  let :external_income_data do
+    export.rows.map { |row| row.take(1) + row.drop(7) }
+  end
+
+  it "exports one quarter of external income for a single project" do
+    project = create(:project_activity, :newton_funded, organisation: delivery_partner)
+
+    ei1 = create(:external_income, activity: project, financial_year: 2014, financial_quarter: 1, amount: 10)
+    create(:external_income, activity: project, financial_year: 2014, financial_quarter: 1, amount: 20)
```
pezholio, line 18: In the spirit of keeping our unit tests quick, can we use mocks and stubs rather than creating real objects in the database?

app/services/quarterly_external_income_export.rb
CristinaRO: No 😬 It was part of debugging, but it should be uncommented.

app/services/quarterly_external_income_export.rb
CristinaRO: I don't mind doing the changes, but I'd like to understand and learn why. ;)

pezholio: It's more of a feeling than a hard and fast rule TBH - the way I see it is that when calling a `call` method or similar, it should do everything at once. If you call an initializer and have it make a database query, you'll have to keep that data in memory for longer, whereas if you call it only when you need it, it can be thrown away quicker.

spec/services/quarterly_external_income_export_spec.rb
pezholio: Something like:

```
project = build(:project_activity, :newton_funded, organisation: delivery_partner)

relation = double("ActiveRecord::Relation")
allow(ExternalIncome).to receive(:activity, :organisation).and_return(relation)
allow(relation).to receive(:where).with(activity_id: [project.id]).and_return(relation)
allow(relation).to receive(:order).with(:activity_id, :financial_year, :financial_quarter).and_return([
  build(:external_income, activity: project, financial_year: 2014, financial_quarter: 1, amount: 10)
  build(:external_income, activity: project, financial_year: 2014, financial_quarter: 1, amount: 20)
])
```

app/controllers/staff/exports/organisations_controller.rb
```diff
@@ -58,6 +58,20 @@ def third_party_project_activities
...
   end
 
+  def external_income
+    fund = Fund.new(params[:fund_id])
+
+    respond_to do |format|
+      format.csv do
+        export = QuarterlyExternalIncomeExport.new(@organisation, fund)
+
+        stream_csv_download(filename: export.filename, headers: export.headers) do |csv|
```
pezholio, line 68: 👍

pezholio (approved):
LGTM :shipit:

PR #1325, opened 2021-08-27
(2144) Chore improve channel of delivery and collaboration type beis code lists

mec:
## Changes in this PR

This is a leave it better PR. Whilst looking at work to add a new channel of delivery code (https://trello.com/c/VKuRpXtM) I felt these two code lists could use improvements to the file names and some documentation on how they relate to make it more explicit what they are doing.

PR #1331, opened 2021-09-02
(1862) Infer region from new beneficiary countries information

pezholio:
This adds a new BEIS codelist that contains everything we know about benefitting countries, and uses this to infer a region when given an activity's benefitting countries. This is then presented to the user in the activity details page, as well as the report download and XML view.

### DIFF
--- a/app/models/benefitting_country.rb
+++ b/app/models/benefitting_country.rb
@@ -7,6 +7,16 @@ class BenefittingCountry
       @all ||= Codelist.new(type: "benefitting_countries", source: "beis").map { |country| new_from_hash(country) }
     end
 
+    def find_by_code(code)
+      all.find { |country| country.code == code }
+    end
+
+    def region_from_country_codes(codes)
+      codes.map { |c| find_by_code(c) }
+        .map { |c| c.regions }
+        .inject(:&).last || Region::DEVELOPING_COUNTRIES_REGION
+    end
+
     private
 
     def new_from_hash(country)
@@ -28,6 +38,12 @@ class BenefittingCountry
   class Region
     include ActiveModel::Model
     attr_accessor :name, :code, :level
+
+    DEVELOPING_COUNTRIES_REGION = new(
+      name: "Developing countries, unspecified",
+      code: "998",
+      level: 1,
+    )
 
     def hash
       code.hash
--- a/app/models/benefitting_country.rb
+++ b/app/models/benefitting_country.rb
@@ -14,7 +14,8 @@ class BenefittingCountry
     def region_from_country_codes(codes)
       codes.map { |c| find_by_code(c) }
         .map { |c| c.regions }
-        .inject(:&).last || Region::DEVELOPING_COUNTRIES_REGION
+        .inject(:&)
+        .max_by { |r| r.level.code } || Region.find_by_code(Region::DEVELOPING_COUNTRIES_CODE)
     end
 
     private
@@ -25,11 +26,7 @@ class BenefittingCountry
         code: country["code"],
         recipient_code: country["recipient_code"],
         regions: country["regions"].map { |region|
-          Region.new(
-            name: region["name"],
-            code: region["code"],
-            level: region["level"],
-          )
+          Region.find_by_code(region)
         }
       )
     end
@@ -39,11 +36,27 @@ class BenefittingCountry
     include ActiveModel::Model
     attr_accessor :name, :code, :level
 
-    DEVELOPING_COUNTRIES_REGION = new(
-      name: "Developing countries, unspecified",
-      code: "998",
-      level: 1,
-    )
+    DEVELOPING_COUNTRIES_CODE = "998"
+
+    class << self
+      def all
+        @all ||= Codelist.new(type: "benefitting_regions", source: "beis").map { |region| new_from_hash(region) }
+      end
+
+      def find_by_code(code)
+        all.find { |region| region.code == code }
+      end
+
+      private
+
+      def new_from_hash(region)
+        new(
+          name: region["name"],
+          code: region["code"],
+          level: Level.find_by_code(region["level"]),
+        )
+      end
+    end
 
     def hash
       code.hash
@@ -51,6 +64,27 @@ class BenefittingCountry
 
     def eql?(other)
       code == other.code
+    end
+
+    class Level
+      include ActiveModel::Model
+      attr_accessor :name, :code
+
+      class << self
+        def all
+          @all ||= Codelist.new(type: "benefitting_region_levels", source: "beis").map { |level| new_from_hash(level) }
+        end
+
+        def find_by_code(code)
+          all.find { |level| level.code == code }
+        end
+
+        private
+
+        def new_from_hash(level)
+          new(name: level["name"], code: level["code"])
+        end
+      end
     end
   end
 end
--- a/app/models/benefitting_country.rb
+++ b/app/models/benefitting_country.rb
@@ -1,0 +1,40 @@
+class BenefittingCountry
+  include ActiveModel::Model
+  attr_accessor :name, :recipient_code, :code, :regions
+
+  class << self
+    def all
+      @all ||= Codelist.new(type: "benefitting_countries", source: "beis").map { |country| new_from_hash(country) }
+    end
+
+    private
+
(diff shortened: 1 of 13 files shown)


mec:
This is really solid work, thank you! I have been waiting so long for this simplefied version of this to see the light of day, it makes me super happy!

As we have attempted this problem space so many times before I very much want us to get this as right as possible, so I have some comments/questions about our approach rather than your code specifically that I would like to talk about before commiting.

I don't want us to validate that the conversion from the spreadsheet to yaml is correct, BEIS shoud do that, but we should make sure it gets done, I think it can be as simple as providing the yaml and some guidance and letting Dan and Meena go over it - its also a great way to help them learn about this stuff.

Looking at the yaml file, I wonder if there is an approach that defines the regions on their own, It feels like this will be a trade off between duplication but readable vs less duplication with less readability, I am not sure which is right, but I would like to at least consider it, I think it could be something like:

```yaml
---
data:
  countries:
  - name: Algeria
    recipient_code: "130"
    iso_code: DZ
    status: active
    regions:
      - level: "region"
         code: "298"
      - level: "sub region 1"
         code: "189"
…
  regions:
    - name: Africa, regional
      code: "298"
    - name: North of Sahara, regional
      code: "189"
…
```
We could split the files if we wanted too, for me it is about exposing how the data fits together in an approachable way - the way IATI do, I don't think there is a right and wrong way, but I want us to have considered these things.

I am concerned about the user of numeric `level` 1, 2 and 3. We have learnt so often the issues that arsie when changing the lanaguage we use and I would like us to think about how we can retain BEIS `region`, `sub region 1` and `sub region 2` domain to keep it super tight - especially in this code list that we want to encourage BEIS user to at least grasp what this does.

In a similar vein, the country codes are ISO codes, so lets say so (total nitpick!)

I would like us to get into the habit of including meta data in the code list where we can, especially where there is documentation for the codes and their use, in this case there probably isn't any but we could do the minimum, there is an example here: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1325 I believe this to be so valuable that we should stub a page on zendesk and link to that and then let meena know that is where she needs to document the new data.

I have a (perhaps miss guided) need to include everything BEIS define in the code lists, by not including 'Developing countries, unspecified' as an entry in the yaml, are we hiding some logic in the code? As a BEIS users who comes to the code list later, would they expect to see it? Whilst BEIS logic does not always follow my own, I think we should again consider this from all angles.

I would like us to be sure of the expectation for the IATI export, I am not sure the BEIS and IATI region codes are the same so we should check if we haven't already (happy to split that out to a new card). My current understanding is we only use the region when there are no `benefitting countries` as the 'recipient region` will be from the IATI list, but our calculated region may not always be a valid IATI one.

Happy to talk about these here, in-Slack or in-virtual-person - just let me know what works.

app/models/benefitting_country.rb
```diff
@@ -0,0 +1,40 @@
+class BenefittingCountry
```
CristinaRO, line 1: I really like the approach of having a class for this entity!

app/models/benefitting_country.rb
```diff
@@ -0,0 +1,40 @@
...
+  include ActiveModel::Model
+  attr_accessor :name, :recipient_code, :code, :regions
+
+  class << self
+    def all
+      @all ||= Codelist.new(type: "benefitting_countries", source: "beis").map { |country| new_from_hash(country) }
```
CristinaRO, line 7: This is my confusion: would this `all` be calculated on every request, or would it get memoised on application load?

I remember something about codelists being loaded on application load (in production environments), for efficiency. Would the same happen to this codelist?

app/views/staff/shared/xml/_activity.xml.haml
```diff
@@ -43,6 +43,8 @@
     - benefitting_countries_with_percentages(activity.benefitting_countries).each do |country|
       %recipient-country{"code" => country.code, "percentage" => country.percentage}
         %narrative= country.name
+    %recipient-region{"code" => activity.benefitting_region.code, vocabulary: "1"}
```
CristinaRO, line 46: I think Meyric asked this too, but I didn't want this question to get lost in the massive list of other things he raised:

Are we sure we want to expose the computed region in the XML even where we have country-level precision? I thought the region would only be used for "legacy" activities where the users intentionally gave a `recipient_region`, **or** for activities where we allow users to not select any `benefitting_countries`.

CristinaRO (commented):
A few specific questions, not overriding Mec's review. ;)

app/models/benefitting_country.rb
pezholio: Yeah, it'll be saved in memory on the first usage

app/models/benefitting_country.rb
pezholio: Thanks! 👍

app/views/staff/shared/xml/_activity.xml.haml
pezholio: Ah, you're right:

> Recipient-region must not be used merely to describe the region of a country reported in recipient-country, but ONLY if the region is a recipient IN ADDITION to the country.

app/models/benefitting_country.rb
```diff
@@ -7,6 +7,16 @@ def all
...
     end
 
+    def find_by_code(code)
+      all.find { |country| country.code == code }
+    end
+
+    def region_from_country_codes(codes)
+      codes.map { |c| find_by_code(c) }
+        .map { |c| c.regions }
+        .inject(:&).last || Region::DEVELOPING_COUNTRIES_REGION
```
mec: This is where I feel we could have an actual issue. if the `regions` array here is not in the expected order the logic we base this on could be broken, here we look for the last region that comes back from all of them and say 'no regions are the same across countries so it must be 998' - but they may not match because of an ordering problem?

vendor/data/codelists/BEIS/benefitting_countries.yml
```diff
@@ -4,2009 +4,1141 @@ data:
     recipient_code: "130"
     code: DZ
     regions:
-      - name: Africa, regional
-        code: "298"
-        level: 1
-      - name: North of Sahara, regional
-        code: "189"
-        level: 2
+      - "298"
```
mec, line 14: And now the ordering of the regions in the array is less obvious, if I put the two codes in the wrong order here, things will not work as expected.

spec/models/benefitting_country_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+RSpec.describe BenefittingCountry do
+  describe ".all" do
+    subject { described_class.all }
+
+    it "returns all the countries" do
+      expect(subject.count).to eq(177)
+    end
+
+    it "allows grouping by most specific region" do
```
mec, line 11: I would like this spec to better document what is happening here if we can. I do wonder if forcing the order would do that?

mec (changes requested):
I might be missing something vital here, but I am concerned we are relying on some array ordering that is not very obvious.

I wonder if we always sort the regions of a BenefittingCountry by level this is both fixed and documented in the code?

It looks like the IATI export commit has come out, which is a good thing! Could we make sure we add a card to capture that need to the back log if not already please.

app/models/benefitting_country.rb
pezholio: Yep, so it's memoized twice, once on the load of all the codelists and again when the `Codelist` class instance is instaniated, which I think is fine

pezholio:
@mec I've made a tweak now so the order of the regions doesn't matter at all, and we `max_by` the region's `level` to find the commonalities. The tests are also a lot more prescriptive, so we set up the countries in the tests, rather than relying on the YAML, which should make the tests easier to follow.

mec (approved):
I think it would keep all the code lists together conceptually if we made the levels file `benefitting_region_levels` with that done, lets ship this 🚢

vendor/data/codelists/BEIS/region_levels.yml
```diff
@@ -1,4 +1,10 @@
 ---
+metadata:
+  title: Region Levels
```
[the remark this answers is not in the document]
pezholio: Good plan 👍

PR #1333, opened 2021-09-06
(2118) A delivery partner user can export their own external income data

pezholio:
This builds on the existing work to allow DP users to see the export page, as well as download their own external income data. I had to make a teeny tweak to the policy we use for exports, but hopefully this is more logical now. There's also a bunch of view specs added to test the template logic.

image

### DIFF
--- a/app/views/shared/_navigation.html.haml
+++ b/app/views/shared/_navigation.html.haml
@@ -15,6 +15,9 @@
       - if policy([:export, Organisation]).index?
         %li{ class: navigation_item_class(exports_path) }
           = link_to t("page_title.export.index"), exports_path, class: "govuk-header__link"
+      - elsif policy([:export, current_user.organisation]).show?
+        %li{ class: navigation_item_class(exports_organisation_path(id: current_user.organisation_id)) }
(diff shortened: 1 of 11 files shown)


pezholio:
[A delivery partner user can export their own external income data](https://trello.com/c/XptcmQPz/2118-a-delivery-partner-user-can-export-their-own-external-income-data)

app/views/shared/_navigation.html.haml
```diff
@@ -12,7 +12,7 @@
       %li{ class: navigation_item_class(organisation_activities_path(organisation_id: current_user.organisation_id)) }
         = link_to t("page_title.activity.index"), organisation_activities_path(organisation_id: current_user.organisation_id), class: "govuk-header__link"
 
-      - if policy(:export).index?
+      - if policy([:export, Organisation]).index?
```
CristinaRO, line 15: (Just making notes commit by commit, ignore if not relevant by the final commit!)

Should we take the opportunity to check the policy against an actual organisation rather than the generic permission?

CristinaRO (approved):

app/views/shared/_navigation.html.haml
pezholio: Nah, the index policy is only for listing all organisations, so doesn't have an organisation to check against.

PR #1334, opened 2021-09-06
Feature/1897 rename transaction to actual

edavey:
## Trello

https://trello.com/c/T88mkx9T/1897-stop-using-the-term-transactions-confirm-the-replacement-language-and-terminology-scope-the-resulting-work and

##  A sequence of changes prior to merging "Adjustments"

This PR represents steps 1 and 2 in this sequence:

1. the term "Transaction" becomes "Actual" (or "Actual spend") throughout codebase
2. `Actual` class becomes a sub-class of `Transaction` using 'Single Table Inheritance'
3. `Refund` is refactored to be a sub-class of `Transaction` (PR to follow)
4. existing `Adjustment` PR (https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1319) is refactored to also be a subclass of `Transaction`

This PR does [1] and [2]. Apologies in advance for the length and size of this, but the Transaction class turns out to be closely coupled to many collaborators. It is also 'magically' (class name reflection) linked with ActiveRecord error translations and with the GOV.UK Design System Form Builder system.

## Categories of renaming within

The changes made in this PR are of the following types, though regrettably not presented as clearly as I'd like:

- user-facing "copy" in locale translation files values edited e.g `'Edit transaction'` -> `'Edit actual spend'`
- locale keys renamed e.g. `transaction:` - > `actual:`
- locale files renamed e.g. `transaction.en.yml` - > `actual.en.yml`
- language in specs renamed
- internal code in specs renamed
- internal language in classes renamed
- internal code in service classes renamed e.g. private methods, instance variable within `CreateTransaction`
- service classes renamed
- view, controllers and routes renamed
- local variables, HTML class name and DOM ids renamed
- creation of a new Actual class (subclass of Transaction)
- data migration to give existing Transaction records the `type` of `'Actual'`

## Release note
We need to remember to run the data migration to convert existing transactions to new-style actuals:

```
rails runner db/data/20210901143708_set_transaction_type_to_actual_for_pre_sti_records.rb
```


app/services/create_transaction.rb
```diff
@@ -4,7 +4,7 @@ class CreateTransaction
   def initialize(activity:, report: nil)
     self.activity = activity
     self.report = report || Report.editable_for_activity(activity)
-    self.transaction = Transaction.new
+    self.transaction = Activity.new
```
pezholio: Should this be `Actual.new`?

app/controllers/staff/activities_controller.rb
```diff
@@ -50,7 +50,7 @@ def show
       format.xml do |_format|
```
pezholio, line 50: Not about this LOC, but there's a typo in the commit message " Convert translations and switch controlle to 'Actual' "

app/services/create_transaction.rb
edavey: well spotted! I caught myself making that typo  a couple of times

app/services/quarterly_actual_export.rb
```diff
@@ -32,22 +32,22 @@ def transaction_row(activity)
...
-    txn_relation = Transaction
+    actual_relation = Actual
       .where(parent_activity: @activities)
       .group(group_columns)
       .select("#{group_columns}, SUM(value) AS value")
 
     @actuals = {}
     @financial_quarters = Set.new
 
-    txn_relation.each do |txn|
```
pezholio, line 46: I totally would have missed this one!

pezholio (approved):
LGTM - just a couple of things about commit messages and a missing unit test. Thanks so much for structuring it this way - made it much easier to review! I usually have a hard and fast rule that each commit must have a greenm test suite, but that would have been way too hard to review!

app/controllers/staff/activities_controller.rb
edavey: fixed, thks

app/models/historical_event.rb
```diff
@@ -6,4 +6,10 @@ class HistoricalEvent < ApplicationRecord
 
   serialize :new_value
   serialize :previous_value
+
```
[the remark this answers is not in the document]
edavey: yes thks for spotting that. I've now added a unit test in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/2cc59c65d4ea1b4375ac1012d94bb547bd63794d to demonstrate that the `HistoricalEvent#trackable_type` is first **incorrectly** set to `'Transaction'` on initialisation, and then **correctly** set to `'Actual'` with a `before_validation` callback. 

That's the best I could do for now. 

Naturally I'd like it to be set correctly on initialisation! 

Feel free to have a play if you like!

edavey:
I'm merging this now, as I can't face more horrible rebases on this!

PR #1340, opened 2021-09-13
(2150) Describe the intended behaviour for recipient_region

CristinaRO:
## Changes in this PR
- Add a spec to describe the intended behaviour of the legacy field `recipient_region`

The intended behaviour is already in place, but it needs to be explicitly described to prevent regressions.

The `recipient_region` field acts as a legacy field, in that it is not editable anymore, but it has to be shown unless the user specifically chooses to provide benefitting_countries.

This feature will become more visible once we run the data migration that undoes the previous data migration, which had populated the inferred countries from the given region.

## Next steps

### DIFF
--- a/spec/views/staff/shared/activities/activity_spec.rb
+++ b/spec/views/staff/shared/activities/activity_spec.rb
@@ -189,6 +189,26 @@ RSpec.describe "staff/shared/activities/_activity" do
     end
   end
 
+  describe "legacy recipient_region" do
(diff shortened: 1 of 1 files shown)


CristinaRO:
[Respect legacy region data](https://trello.com/c/1vyunQyx/2150-respect-legacy-region-data)

pezholio (approved):
Nice work - always good to have behaviour documented in tests 😃

PR #1341, opened 2021-09-14
(2119) A BEIS user can export external income for all delivery partner organisations in a single file

pezholio:
This builds on the external income export work to allow BEIS users to export data about External Income for all delivery partners. This is now listed on the BEIS users' export page like so:

image

### DIFF
--- a/app/controllers/staff/exports_controller.rb
+++ b/app/controllers/staff/exports_controller.rb
@@ -2,7 +2,7 @@ class Staff::ExportsController < Staff::BaseController
   include Secured
 
   def index
-    authorize [:export, Organisation]
+    authorize :export, :index?
 
     add_breadcrumb t("breadcrumbs.export.index"), :exports_path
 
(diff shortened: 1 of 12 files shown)


CristinaRO:
[A BEIS user can export external income for all delivery partner organisations in a single file](https://trello.com/c/KmZIrMHu/2119-a-beis-user-can-export-external-income-for-all-delivery-partner-organisations-in-a-single-file)

app/controllers/staff/exports_controller.rb
```diff
@@ -1,11 +1,28 @@
...
-    authorize [:export, Organisation]
+    authorize :export, :index?
 
     add_breadcrumb t("breadcrumbs.export.index"), :exports_path
 
     @organisations = policy_scope(Organisation).delivery_partner
   end
+
+  def external_income
+    authorize :export, :show_external_income?
```
edavey, line 14: interesting to see all the ways in which policies can be applied!

edavey (approved):
Really clear work: this will be v useful in all the upcoming "export {foo}" stories...

PR #1343, opened 2021-09-14
Increase deploy timeout

pezholio:
We've been having some issues deploying, with the odd timeout causing our deploys to fail. After speaking to the GOV.UK PasS team, they've advised us that they've increased the maximum deploy timeout at their end, so have advised us to try increasing our maximum timeout to see if that helps.


CristinaRO (approved):
Definitely worth trying, thank you!

PR #1344, opened 2021-09-15
(2173) Extend Benefitting Countries and region calculation

mec:
## Changes in this PR

ODA eligible countries 'graduate' and are no longer eligible. We have a number of activities that have this countries in their data so the application must handle them.

We do not want new activities to be created with graduated benefitting countries though.

To handle this we extend `BenefittingCountry` to handle graduated countries including or exluding them as needed for different contexts.

We also found a bug in the region calculation when handling these graduated countires and the possibliliy that there is a completely unexpected country in an activitiy somewhere - this work allows us to handle these situation should they arise, see the commits for details on decisions made.

The main users facing change here is that the web form step for benefitting countries now uses the code list, we took the time to updated the view with:

- country and region codes
- additional help
- more prominent region breaks

Note: the linked guidance is currently empty until we merge this work and help BEIS to document it!

We still need to ensure the IATI XML handles countries and regions as expected, this is captured here https://trello.com/c/z5XUUW4f

### After
Screenshot 2021-09-15 at 10-41-21 What are the countries benefitting from this activity - Report your Official Development

## Next steps


vendor/data/codelists/BEIS/benefitting_countries.yml
```diff
@@ -10,1142 +10,1347 @@ data:
   - name: Algeria
     recipient_code: "130"
     code: DZ
+    graduated: no
```
pezholio, line 13: Would it be better to have `true` and `false` for these values?

app/services/activities/import_from_csv.rb
```diff
@@ -581,11 +581,7 @@ def country_to_region_mapping
       end
 
       def validate_country(country, error)
-        yaml = YAML.safe_load(File.read("#{Rails.root}/config/locales/codelists/#{IATI_VERSION}/iati.en.yml"))
-        countries = yaml["en"]["activity"]["recipient_country"]
-
-        raise error unless countries.key?(country)
-
+        raise error unless BenefittingCountry.find_non_graduated_country_by_code(country)
```
pezholio, line 584: 🎉

spec/models/benefitting_country_spec.rb
```diff
@@ -170,3 +186,36 @@
     end
   end
 end
+
+RSpec.describe BenefittingCountry::Region do
```
pezholio, line 190: I wonder now, if `BenefittingCountry::Region` is becoming useful on its own, if we should break it out into its own file (and, conversely, test suite)

app/helpers/codelist_helper.rb
```diff
@@ -12,8 +13,12 @@ def currency_select_options
...
   end
 
-  def benefitting_countries_hash
-    Codelist.new(type: "intended_beneficiaries").list
+  def benefitting_regions_for_form
+    BenefittingCountry::Region.all_for_level_code(BENEFITTING_SUB_REGION_2_CODE)
+  end
+
+  def benefitting_countries_in_region_for_form(region)
+    BenefittingCountry.non_graduated_for_region(region)
```
pezholio, line 21: Now this stuff has a degree of seperation from the codelist, whether it would be sensible to have a `BenefittingCountryHelper` or similarly named thing?

config/locales/models/activity.en.yml
```diff
@@ -337,6 +337,7 @@ en:
       organisation: Organisation
     activity:
       unknown_country: Unknown country code
+      benefitting_region_check_box: Select or deselect all countries in %{region}
```
pezholio, line 340: Me likey! 👍

app/models/benefitting_country.rb
```diff
@@ -24,7 +24,7 @@ def non_graduated_for_region(region)
     end
 
     def region_from_country_codes(codes)
-      codes.map { |c| find_by_code(c) }
+      codes.filter_map { |code| find_by_code(code) if find_by_code(code).present? }
```
pezholio, line 27: 😍

pezholio (approved):
LGTM - just a couple of questions/suggestions. Feel free to push back / discuss / ignore as appropriate 😃

vendor/data/codelists/BEIS/benefitting_countries.yml
pezholio: Oh wait, this is YAML being weird and YAML-y isn't it? 🙄

mec: Yeah, they translate to true and false automagically, I like the language of 'Graduated? No' for the code list.

app/models/benefitting_country.rb
mec: Pretty nice eh! 😄

app/helpers/codelist_helper.rb
mec: We have a form helper - I should've started there! How does that sound?

pezholio: Yeah, that makes sense 👍

spec/models/benefitting_country_spec.rb
pezholio: Do it 👍

app/models/benefitting_country.rb
```diff
@@ -1,16 +1,24 @@
...
-  attr_accessor :name, :recipient_code, :code, :regions
+  attr_accessor :name, :recipient_code, :code, :graduated, :regions
 
   class << self
     def all
       @all ||= Codelist.new(type: "benefitting_countries", source: "beis").map { |country| new_from_hash(country) }
     end
 
+    def non_graduated
+      @non_graduated ||= all.select { |country| !country.graduated }
```
CristinaRO, line 11: Could this be more snappily expressed as
```
@non_graduated ||= all.reject(&:graduated)
```
?

spec/models/benefitting_country_spec.rb
mec: https://trello.com/c/8vPgSZyH

CristinaRO (commented):
Well, I'll drop my in-progress review anyway. ;)

app/models/benefitting_country.rb
mec: Yes I think it could!!

mec:
@CristinaRO if you assign yourself or request yourself for review I won't merge until I get you approal! 👍

PR #1346, opened 2021-09-16
(2188) Chore: Split Region from BenefittingCountry

mec:
In #1344 It was noted in a review that BenefittingCountry::Region was becoming a
first class citizen and we felt it should be split out in to its own
file.

We took the opportunity to rename it BenefittingRegion to keep the link
with the concept it is modelling as clear as we can.

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -51,7 +51,7 @@ module FormHelper
   end
 
   def benefitting_regions_for_form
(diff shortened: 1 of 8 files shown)


pezholio:
I don't think you've committed the new `BenefittingRegion` files have you? (also, there's a StandardRB fail)

mec:
There was no ☕ this morning! Sorry!

pezholio:
Happens to the best of us!

pezholio (approved):
LGTM - I was thinking about keeping the namespacing, but I think calling it `BenefittingRegion` works better 👍

PR #1353, opened 2021-09-20
Refactor report export

pezholio:
This refactors the Report export to use the approach we're trying to move towards, with an `Export` subclass and public methods for the filename, headers, and rows.


spec/services/report/export_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+          [activities[4].title],
+          [activities[5].title],
+        ])
+      end
+    end
+  end
+
+  describe Report::Export::Row do
+    # The row generates URLs, so we need to make sure the objects have UUIDs
+    let(:organisation) { build(:delivery_partner_organisation, id: SecureRandom.uuid) }
```
edavey, line 169: I think this is a big improvement: adding unit tests for `Export::Row` and then stubbing its use in the test for `Report::Export`.

edavey (approved):
This is really nice work: you've made this easier to understand and much more clearly described in tests.

However (no criticism of this work which does make the code easier) I'm still finding this exporting code really hard to understand -- but at the moment I'm struggling to find suggestions for how to make it simpler!

I do wonder if we need to take a fundamentally simpler approach to reporting, so that we don't have the need to produce totals for multiple financial periods "on the fly"...

app/controllers/staff/reports_controller.rb
```diff
@@ -125,49 +117,37 @@ def approved_reports(including: [])
...
-    ]
-    report_sets.sum.sort_by { |report| report.organisation.name }
-  end
-
-  def report_activities_sorted_by_level(report)
-    Activity::ProjectsForReportFinder.new(
-      report: report,
-      scope: Activity.includes(:organisation, :extending_organisation, :implementing_organisations)
-    ).call.sort_by { |a| a.level }
+    @downloadable_reports_for_beis_users ||= begin
```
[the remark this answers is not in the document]
pezholio: There seems to be a lot of oddness going on with this export. I'm tempted to leave it, because even a fixed/cleaned up version is going to take forever to export, and @mec's comment in https://trello.com/c/GfKh3G8I/1939-the-download-all-reports-csv-feature-takes-too-long-and-dies suggests we don't even need it anyway

spec/services/report/export_spec.rb
pezholio: Ta! 👍

PR #1357, opened 2021-09-21
(2195) Remove the all reports download

pezholio:
This currently takes too long and the app times out. We're working on a replacement which covers all the needs for SID reporting, so this will soon not be needed anyway, so worth stripping out (especially as it adds a fair degree of complexity and misdirection)


pezholio:
[Remove the All Reports download](https://trello.com/c/6BfxFQ2L/2195-remove-the-all-reports-download)

edavey (approved):
👍 good to remove this problem for now. We'll need to figure out a workable way to generate all the reports for download -- if this is indeed an important need.

(BTW the Changelog needs fixing up in a rebase)

PR #1359, opened 2021-09-22
(2051) Include Adjustments in CSV exports

pezholio:
This adds an argument to the `ActualOverview` class to include any adjustments, which is then used in the `Report::Export` class to make sure the calculations of Actuals are returned with the adjustments (either negative or positive) included. There was a bit of added complexity when I realised Adjustments could apply to Refunds as well as Actuals - detail is in the commit message.

### DIFF
--- a/app/services/actual_overview.rb
+++ b/app/services/actual_overview.rb
@@ -37,7 +37,7 @@ class ActualOverview
   def scope
     if @include_adjustments
       Transaction
-        .joins("LEFT OUTER JOIN adjustment_details ON transactions.id = adjustment_details.adjustment_id")
+        .with_adjustment_details
         .where(type: "Actual")
         .or(
           Transaction
--- a/app/services/actual_overview.rb
+++ b/app/services/actual_overview.rb
@@ -35,7 +35,20 @@ class ActualOverview
   end
 
   def scope
-    @include_adjustments ? Transaction.where(type: ["Actual", "Adjustment"]) : Actual
+    if @include_adjustments
+      Transaction
+        .joins("LEFT OUTER JOIN adjustment_details ON transactions.id = adjustment_details.adjustment_id")
+        .where(type: "Actual")
+        .or(
+          Transaction
+            .where(
+              type: "Adjustment",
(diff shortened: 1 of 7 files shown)


pezholio:
[Include Adjustments in CSV exports](https://trello.com/c/37LM2Ps1/2051-include-adjustments-in-csv-exports)

app/services/actual_overview.rb
```diff
@@ -37,7 +37,7 @@ def actual_relation
   def scope
     if @include_adjustments
       Transaction
-        .joins("LEFT OUTER JOIN adjustment_details ON transactions.id = adjustment_details.adjustment_id")
+        .with_adjustment_details
```
edavey, line 40: nice refactoring!

app/services/actual_overview.rb
```diff
@@ -1,7 +1,8 @@
 class ActualOverview
-  def initialize(activity, report)
+  def initialize(activity:, report:, include_adjustments: false)
```
edavey, line 2: I wonder if we should maybe push ourselves to be clear on where it's appropriate for an `ActualOverview` to **not** include adjustments? 

I see that `Actual#actual_total_for_report_financial_quarter` is used in the "variances" tab, but is it really more correct for adjustments to be excluded from variances? i.e. I think need to see if we can simplify the business rules and the code by making the inclusion of adjustments a new element, rather than an _optional_ new element...

If what I suggest **does** make sense, probably that would be a clarification to make in a separate PR?

edavey (approved):
Nice work!

I've made one comment -- about whether the inclusion of adjustments in `ActualOverview` should be optional. i.e. could we, should we make things simpler?

app/services/actual_overview.rb
pezholio: I think the reason I did it this way is because this class gets used elsewhere, and I wanted to make sure we did the simplest thing possible in the first instance. We have a card elsewhere that covers including adjustments in calculations, and I wanted to make sure we didn't touch other parts of the codebase and complicating the feature. We can (and should) tidy this up later.

PR #1360, opened 2021-09-22
(2194) Chore: Upgrade Ruby and JS dependencies

mec:
We had 6 security alerts, 5 of which are high severity. All are caused by the dependencies in the JS part of Webpacker.

We update both the Webpacker Ruby gem and JS package to the latest version 5.4.3.

Not all dependencies can be updated to the recommend versions.

- ansi-html does not appear to have a patched version

These updates also caused the test environment to start outputting html comments on the partials and layout, something I can only assume was broken before, as this cauased a spec to fail. The spec and behaviour is questionable, but ultimately I decided that the test env should mirror prod and not include these comments regardless, so we turn them off.

I think next steps with this is to merge this and see how are alerts look once that is done, we can decide at that point how we plan to deal with any left or dismiss them.

You will need to run:

`bundle install`
`yarn install`

once this is merged or to try the branch to update all the dependencies.


mec:
@edavey yep `yarn.lock` is here - it's just so vast it is hidden in Github: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1360/commits/5cc8fb24a7692fe3c72cb8b921cd7092307397de#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de

edavey:
> @edavey yep `yarn.lock` is here - it's just so vast it is hidden in Github: [5cc8fb2#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/5cc8fb24a7692fe3c72cb8b921cd7092307397de#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de)

ah, yes, sorry. interesting that my lock file is different...

edavey (approved):
This looks good to me. It's slightly mysterious to me why my `yarn.lock` has this one difference, but it doesn't look significant.

mec:
@edavey can you confirm what version of yarn you are running?

edavey:
> @edavey can you confirm what version of yarn you are running? `yarn -v` mine is `1.22.11`

I've upgraded to `1.22.11` and run `yarn install --force` but get the same small diff in the lock file on the webpacker version...

mec:
@edavey I get it too if I run it today!?! I'll update this with it…

PR #1363, opened 2021-09-27
(2051) Cache all transactions in an ActualOverview

pezholio:
This updates the ActualOverview to removet he activity from the initializer. When returning `all_quarters`, this now groups each transaction by financial year, financial quarter AND activity ID. This means that rather than querying transactions on an activity by activity basis for the export, we can load all of a report's transactions into memory when we start building the report, and pluck them from the `AllQuarters` instance based on the financial period and the activity.

I've also had to update the `value_for_report_quarter` method to accept an activity as an argument, so we can reuse the relation in the class, but append the query to search for a specific activity.

### DIFF
--- a/spec/services/report/export_spec.rb
+++ b/spec/services/report/export_spec.rb
@@ -84,6 +84,7 @@ RSpec.describe Report::Export do
           report_presenter: report_presenter,
           previous_report_quarters: an_instance_of(Array),
           following_report_quarters: an_instance_of(Array),
+          actual_quarters: an_instance_of(ActualOverview::AllQuarters),
         ).and_return(stub)
         expect(stub).to receive(:call).and_return([activity.title])
       end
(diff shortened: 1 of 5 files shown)


spec/services/report/export_spec.rb
```diff
@@ -135,6 +136,15 @@
       ]
     end
 
+    let(:actuals) do
+      [
+        Transaction.new(financial_quarter: 1, financial_year: 2020, value: 20, parent_activity: activity),
+        Transaction.new(financial_quarter: 2, financial_year: 2020, value: 40, parent_activity: activity),
+        Transaction.new(financial_quarter: 3, financial_year: 2020, value: 80, parent_activity: activity),
```
edavey, line 143: Would it be clearer if you instantiated 3 `Actual`s objects here rather than using the `Transaction` class?

edavey (approved):
👍 

Just out of interest, did you see or measure a noticeable reduction in db activity / time?

pezholio:
There wasn't too much of an improvement strangely enough, though I think that might be more to do with the fact that the vast majority of the DB time is taken up with the forecasts.

PR #1366, opened 2021-09-28
(2158) Feature: spending breakdown includes forecasts

mec:
## Changes in this PR

Following on from #1361 this work introduces Forecasts to the Spending breakdown export.

As before this in not yet exposed to users, but you can interact with the class on the Rails console.

We want to display all Forecasts for the financial quarters not already covered by Actual spend and Refunds. We do this by taking the end of the range for Actual spend and Refunds and applying that as the start of the range of Forecasts.

Forecasts are already stored as a single total per financial quarter and activity so there is no need to sum them. We do have to utilise the `ForecastOverview` class which means we lose some control of the queries. Fortunately the `ForecastOverview` can be use with multiple activity ids allowing us to load all required latest values from the relevant forecasts - which keeps the class responsive loading the data.

To keep the code approachable we convert the Forecast data to the same hash structure as the ActiveRecord query on Transaction, which allows us to use `fetch` as our interface to the correct forecasts (also taking care of returning 0 when there is no forecast).

I thought a lot about IF the code should return 0 for a quarter with no forecast and decideded that this was the right thing to do. Whilst there is a difference between 'I have forecasted zero spend for that activity and quarter' and 'I have not forecastsed that activity and quarter' I felt from a finance persepective, who utiltise this report, the difference was not important.

Once the class is working we refactor to make improvements and then spend some time speeding up the spec.

### DIFF
--- a/spec/services/spending_breakdown/export_spec.rb
+++ b/spec/services/spending_breakdown/export_spec.rb
@@ -1,32 +1,26 @@
 RSpec.describe SpendingBreakdown::Export do
-  let!(:organisation) { create(:delivery_partner_organisation, beis_organisation_reference: "BC") }
-  let!(:activity) { create(:project_activity, organisation: organisation) }
-  let!(:source_fund) { Fund.new(activity.source_fund_code) }
-  let!(:actual) { create(:actual, parent_activity: activity, value: 100, financial_quarter: 1, financial_year: 2020) }
-  let!(:refund) { create(:refund, parent_activity: activity, value: -200, financial_quarter: 1, financial_year: 2020) }
+  before(:all) do
+    DatabaseCleaner.strategy = :transaction
+    DatabaseCleaner.start
 
-  let!(:positive_actual_adjustment) { create(:adjustment, :actual, parent_activity: activity, value: 200, financial_quarter: 1, financial_year: 2020) }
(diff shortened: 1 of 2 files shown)


spec/services/spending_breakdown/export_spec.rb
```diff
@@ -1,32 +1,26 @@
 RSpec.describe SpendingBreakdown::Export do
-  let!(:organisation) { create(:delivery_partner_organisation, beis_organisation_reference: "BC") }
-  let!(:activity) { create(:project_activity, organisation: organisation) }
-  let!(:source_fund) { Fund.new(activity.source_fund_code) }
-  let!(:actual) { create(:actual, parent_activity: activity, value: 100, financial_quarter: 1, financial_year: 2020) }
-  let!(:refund) { create(:refund, parent_activity: activity, value: -200, financial_quarter: 1, financial_year: 2020) }
+  before(:all) do
```
pezholio, line 2: This is a VERY cool approach!

pezholio (approved):
LGTM - particularly like the new approach with keeping all the DB objects around for the entirety of the test run. I reckon there's some other tests that could benefit from this.

spec/services/spending_breakdown/export_spec.rb
mec: It felt a bit old-skool, but then I realised that doesn't matter - it's just much, much better and faster in this use case! 🏎️

PR #1368, opened 2021-09-28
(2051) Include refunds in the report exports

pezholio:
This creates a Refund Overview class, which shares functionality with the Actual Overview to return the value of all of the refunds for an activity within the previous twelve quarters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -835,6 +835,8 @@
 
 ## [unreleased]
 
+- Include refunds in the report CSV exports
+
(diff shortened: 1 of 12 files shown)


pezholio:
[Include Adjustments in CSV exports](https://trello.com/c/37LM2Ps1/2051-include-adjustments-in-csv-exports)

pezholio:
Cool, ta! I have some plans in progress to make the export class simpler / faster. Working on it as we speak!

PR #1369, opened 2021-09-28
(2158) Feature: spending breakdown can be exported by users

mec:
## Changes in this PR
Following on from #1366 we now have a complete and working Spending breakdown export so we can offer it to our users.

This is mostly boiler plate to authorise the routes and wire everything up.

The final commit refactors the organisation export controller a little.

## Screenshots of UI changes
DP view:
Screenshot 2021-09-28 at 14-48-48 Royal Society - Report your Official Development Assistance

BEIS all organisations:
Screenshot 2021-09-28 at 14-49-43 Exports - Report your Official Development Assistance

BEIS specific organisation:
Screenshot 2021-09-28 at 14-49-52 Academy of Medical Sciences - Report your Official Development Assistance

PR #1370, opened 2021-09-28
(2158) Feature: remove old spending breakdown

mec:
## Changes in this PR

Following on from #1369 here we remove the old `ActivitySpendingBreakdown` and related code.

As a final 🎁 we finally improve the link to the Report CSV download to help users:

- find it
- understand what it is
- get additional help where required

This was discussed in Slack: https://dxw.slack.com/archives/CN0H2L066/p1632831443029400

Just thought this is probably a chore not a feature, sorry!

## Screenshots

Screenshot 2021-09-28 at 16-55-08 FQ1 2021-2022 Newton Fund report - Report your Official Development Assistance


app/views/staff/reports/_download.haml
```diff
@@ -1,2 +1,7 @@
-.govuk-grid-column-one-third.page-actions
-  = link_to t("action.report.download.button"), report_path(report_presenter, format: :csv), class: "govuk-link"
+.govuk-grid-column-two-thirds
+  %h3.govuk-heading-m
+    Download
+  %p.govuk-body
+    Download a CSV file to review offline. For guidance on reviewing and submitting your report, see the
+    = link_to_new_tab " guidance in the help centre", "https://beisodahelp.zendesk.com/hc/en-gb/articles/1500005604902-Review-and-Submit-your-Data"
+  = link_to t("action.report.download.button"), report_path(report_presenter, format: :csv), class: "govuk-button govuk-button--secondary"
```
edavey, line 7: do you think we might be missing an opportunity to add to `scenario "can download a CSV of the report` some detail on the additional value being added here, e.g.

- I see a link to guidance about the reviewing my submission
- I see that the CSV is offered for review purposes

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/14feb522f29189a1a59ac4144c3d2a394df95659/spec/features/staff/users_can_view_reports_spec.rb#L143-L157

edavey (approved):
nice work -- I make a small suggestion to add a little documentation of the new "download CSV" behaviour...

app/views/staff/reports/_download.haml
mec: Happy to. If I change the scenario to `they see helpful guidance about and can download a CSV of their own report` and check for the content and link, do you think that will do it?

edavey: that sounds good: I think there's generally a strong case for trying to ensure that our tests describe the improvements we make: i.e. how does the change in behaviour deliver value?

mec: I don't disagree, I do wonder if there is a balance to strike at some point, but that is definately a conversation for later and I am confident you will convince me!

PR #1371, opened 2021-09-29
Add data migration to fix up over-eager backfilling of benefitting_countries

edavey:
It seems we were over-eager in backfilling the new `Activity#benefitting_countries`
field, in that we added all of the countries in the legacy `recipient_region` even
when the legacy `recipient_country` field was present.

Example: Activity with RODA id `'GCRF-EP_C_MS-4_2018EP/T00343X/1'`:

```
   => {"recipient_region"=>"1027",
    "recipient_country"=>"UG",
    "intended_beneficiaries"=>nil,
    "requires_additional_benefitting_countries"=>false,
    "country_delivery_partners"=>[],
    "benefitting_countries"=>["UG", "BI", "KM", "DJ", "ER", "ET",
                              "KE", "MG", "MW", "MU", "MZ", "RW",
                              "SO", "SS", "SD", "TZ", "ZM", "ZW"]}
```

In this script, we find the approx 4,000 activities which are affected:

- legacy `recipient_region` is present, and
- legacy `recipient_country` is present, and
- there are no `intended_beneficiaries`, and
- more than 1 country is set in the new `benefitting_countries` field

and set the benefitting countries field to the legacy recipient_country value.
e.g. `["UG"]` in the example above.

It should produce output something like:

```
Fixing up 4130 activities...
-> there are now 0 activities still to fix
```

### DIFF
--- a/db/data/20210929_clear_additional_benefitting_countries_where_backfilling_not_appropriate.rb
+++ b/db/data/20210929_clear_additional_benefitting_countries_where_backfilling_not_appropriate.rb
@@ -1,0 +1,41 @@
+# Run me with `rails runner db/data/20210929_clear_additional_benefitting_countries_where_backfilling_not_appropriate.rb`
+
+# It seems we were over-eager in backfilling the new `Activity#benefitting_countries`
+# field, in that we added all of the countries in the legacy `recipient_region` even
+# when the legacy `recipient_country` field was present.
+#
+# Example: Activity with RODA id 'GCRF-EP_C_MS-4_2018EP/T00343X/1':
+#
+#     => {"recipient_region"=>"1027",
+#      "recipient_country"=>"UG",
+#      "intended_beneficiaries"=>nil,
+#      "requires_additional_benefitting_countries"=>false,
+#      "country_delivery_partners"=>[],
+#      "benefitting_countries"=>["UG", "BI", "KM", "DJ", "ER", "ET",
+#                                "KE", "MG", "MW", "MU", "MZ", "RW",
+#                                "SO", "SS", "SD", "TZ", "ZM", "ZW"]}
+#
+# In this script, we find the approx 4,000 activities which are affected:
+#
+# - legacy `recipient_region` is present, and
+# - legacy `recipient_country` is present, and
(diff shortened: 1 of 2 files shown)


edavey:
[Updates in Geography on RODA appear to have added additional benefitting countries on activities that have previously included just one benefitting country.](https://trello.com/c/VkkK6nSK/2217-updates-in-geography-on-roda-appear-to-have-added-additional-benefitting-countries-on-activities-that-have-previously-included-j)

pezholio:
Just looking at this now, but there's a StandardRB error that needs fixing 👍

db/data/20210929_clear_additional_benefitting_countries_where_backfilling_not_appropriate.rb
```diff
@@ -0,0 +1,41 @@
...
+
+finder = Activity
+  .where.not(recipient_country: nil)
+  .where.not(recipient_region: nil)
+  .where(intended_beneficiaries: nil)
+  .where("array_length(benefitting_countries, 1) > 1")
+
+puts "Fixing up #{finder.count} activities..."
+
+finder.load.each do |activity|
```
pezholio, line 36: I've never seen `.load` before, what's the benefit of using these, compared to just using the relation? Or is it just a semantic thing?

pezholio (approved):
LGTM 👍

mec:
@edavey looks great! I can see how we got here and how this fixes the issue. 🚢

db/data/20210929_clear_additional_benefitting_countries_where_backfilling_not_appropriate.rb
edavey: hum, in fact it's not needed here! `ActiveRecord::Relation#load` (https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-load) bypasses eager loading and loads up all the resulting records as an array. Sometimes I've found it's handy if you need to obtain all the records 'in' the relation right now...

But this isn't one of those cases! After we've done the remedial work we `reload` the relation and that's enough to verify that now there aren't any remain problem records.

Thanks for asking / pointing it out!

PR #1372, opened 2021-09-29
(2158) Chore: Spending breakdown export developer doc

mec:
## Changes in this PR
Here it is! The final PR for this work! 

To help developers and anybody else understand this feature, some brief guidance around it.

I was not 100% sure how to pitch this, but I am confident we we learn as we write more docs and can come back and improve.

### DIFF
--- a/doc/exports.md
+++ b/doc/exports.md
@@ -1,0 +1,45 @@
+# Exports
+
+## Spending Breakdown
+A report primarily aimed at BEIS finance needs.
+
+- Includes all __Actual spend__ at the time the export is created
+- Includes all __Refunds__ at the time the export is created
+- Actual spend includes adjustments for the same financial quarter and of `type`
+  `Actual`
+- Refunds includes adjustments for the same financial quarter and of `type`
+  `Refund`
+- Actual spend, Refunds and Adjustments share the same `Transaction` base class
+  and database table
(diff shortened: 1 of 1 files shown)


doc/exports.md
```diff
@@ -0,0 +1,44 @@
...
+### Example calculation
+
+```
+Q1 2021-2022 values
+---------------------------------
+Actual spend:              10,000
+Actual spend:               5,000
+Adjustment (type: Actual): -1,000
+Refund:                    -5,000
+Adjustment (type: Refund):  1,000
```
edavey, line 28: I suggest, for completeness: 

- another line for a negative Refund Adjustment, and
- another line for a positive Actual Adjustment

e.g.

```
Adjustment (type: Refund):  -2,000
Adjustment (type: Actual):   2,000
```

doc/exports.md
```diff
@@ -0,0 +1,44 @@
...
+Adjustment (type: Actual):   -500
```
edavey, line 29: and we could then lose this line as we already have a negative Actual Adjustment in the example

edavey (approved):
I think this is a really good level of detail into the report and the technical implementation. I've made a few small comments:

- nick-picking on consistency in pluralisation
- suggesting covering all the adjustment types in the example

PR #1373, opened 2021-09-29
(2194) Chore: Update Node JS and Webpacker JS

mec:
## Changes in this PR
This is a second run to update the JS dependancies. 

First we really need a newer version of Node, we are running 10 now and the latest long term support version is 14, this is not the end of the world as I believe Node is only used in Webpacker to compile JS and assets, but we should stay current regardless, note: we added a `.node-version` file check the commit message.

In #1360 we missed running the required commands:

```
bin/rails webpacker:install
yarn upgrade @rails/webpacker --latest
yarn upgrade webpack-dev-server --latest
```

which update various JS components, so we've done so here.

This should clear some of the Dependabot secuity warnings.

There is a large diff here on `yarn.lock` that will not be shown in the Github UI by default!


pezholio (approved):
LGTM - I'm not entirely convinced any of these security issues will cause a problem for us, given that most of this tooling is used at build time, and not in production per se, but if it works and makes the annoying message go away then I'm all for it!

mec:
@pezholio me either - but I think this is all worthwhile.

PR #1374, opened 2021-09-29
(2216) Chore: remove geography fields no longer required  from interface

mec:
## Changes in this PR
The `requires_additional_benefitting_countries` is no longer valuable since we replaced `recipient_region`, `recipient_country` and `intended_beneficiaries` with `benefitting_countries` and `benefitting_region`.

Here we remove the column and any associated items.

We also decided to always show ALL the related geography attributes regardless of their state to help users:

- understand what data an activity does or does not have, rather than hiding that
- understand how they could migrate the data to the new values

We also added a 'read only` flag to the label to help users understand why the values in the old fields can longer been changed.

## Screenshots of UI changes

Screenshot 2021-09-29 at 17-31-32 Activity AMS Coherence and Impact - Global Health Policy Workshops - Details - Report you

### DIFF
--- a/config/locales/models/activity.en.yml
+++ b/config/locales/models/activity.en.yml
@@ -256,7 +256,7 @@ en:
         geography: Benefitting recipient geography
         delivery_partner_identifier: Delivery partner identifier
         delivery_partner_organisation: Delivery partner organisation
-        intended_beneficiaries: Intended beneficiaries
+        intended_beneficiaries_html: 'Intended beneficiaries</br><span class="govuk-body-s">Legacy field: not editable</span>'
         objectives: Aims/Objectives
         oda_eligibility: ODA eligibility
         oda_eligibility_lead: ODA eligibility lead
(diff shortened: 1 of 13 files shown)


config/locales/models/activity.en.yml
```diff
@@ -277,8 +277,8 @@ en:
         policy_marker_nutrition: Nutrition Policy
         programme_status: Activity status
         projects: Projects (level C)
-        recipient_country: Recipient country
-        recipient_region: Recipient region
+        recipient_country: Recipient country (read only)
```
pezholio, line 276: Do we want to use `read only` here? It seems like an overly technical term. How about `Legacy field: not editable` or something? I'm also wondering if we can put it in smaller text - say wrapped in a span with a `govuk-body-s` class?

pezholio (approved):
LGTM - just a tiny suggestion about wording/typography

config/locales/models/activity.en.yml
mec: Can do.

PR #1378, opened 2021-09-30
(2221) Re-populate benefitting countries

mec:
## Changes in this PR
It seems we were over-eager in backfilling the new
`Activity#benefitting_countries` field, in that we added all of the
countries in the legacy `recipient_region` even when the legacy
`intended_beneficiaries` field was present and populated with a more
granular set of country data

Example: Activity with RODA id 'GCRF-UKR_ARE_C19-Ag2020EP/V044613/1':

     => {"recipient_region"=>"289",
      "recipient_country"=>"nil",
      "intended_beneficiaries"=>["NG", "ZA"],
      "requires_additional_benefitting_countries"=>false,
      "benefitting_countries"=>["PH", "IN", "KH", "CN", "KP", "ID",
      "LA", "MY", "MN", "TH", "TL", "VN", "IR", "IQ", "JO", "LB", "SY",
      "PS", "YE", "AF", "BD", "BT", "MV", "MM", "NP", "PK", "LK"]

In this script, we find the approx 400 activities which are affected:

 - legacy `recipient_region` is present, and
 - legacy `recipient_country` is empty, and
 - there are `intended_beneficiaries`, and
 - the number of countries in `benefitting_countries` is greater than
   those in `intended_beneficiaries`

and set the benefitting countries to that of intended_beneficiaries
e.g. `["NG", "ZA"]` in the example above.

Obviously credit to @edavey for the original work this is shamelessly taken from! 👍

### DIFF
--- a/db/data/20210930151236_intended_benefeciaries_to_benefitting_countries.rb
+++ b/db/data/20210930151236_intended_benefeciaries_to_benefitting_countries.rb
@@ -1,0 +1,39 @@
+# Run me with `rails runner db/data/20210930151236_intended_benefeciaries_to_benefitting_countries.rb`
+
+# It seems we were over-eager in backfilling the new `Activity#benefitting_countries`
+# field, in that we added all of the countries in the legacy `recipient_region` even
+# when the legacy `intended_beneficiaries` field was present and populated with a more
+# granular set of country data
+#
+# Example: Activity with RODA id 'GCRF-UKR_ARE_C19-Ag2020EP/V044613/1':
+#
+#     => {"recipient_region"=>"289",
+#      "recipient_country"=>"nil",
+#      "intended_beneficiaries"=>["NG", "ZA"],
+#      "requires_additional_benefitting_countries"=>false,
+#      "benefitting_countries"=>["PH", "IN", "KH", "CN", "KP", "ID", "LA", "MY", "MN", "TH", "TL",
+#      "VN", "IR", "IQ", "JO", "LB", "SY", "PS", "YE", "AF", "BD", "BT", "MV", "MM", "NP", "PK", "LK"]
+#
+# In this script, we find the approx 400 activities which are affected:
+#
+# - legacy `recipient_region` is present, and
+# - legacy `recipient_country` is empty, and
+# - there are `intended_beneficiaries`, and
+# - the number of countries in `benefitting_countries` is greater than those in `intended_beneficiaries`
+#
+# and set the benefitting countries to that of intended_beneficiaries
+# e.g. `["NG", "ZA"]` in the example above.
+
+finder = Activity
+  .where.not(recipient_region: nil)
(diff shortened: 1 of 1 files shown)


db/data/20210930151236_intended_benefeciaries_to_benefitting_countries.rb
```diff
@@ -0,0 +1,39 @@
...
+
+finder = Activity
+  .where.not(recipient_region: nil)
+  .where(recipient_country: nil)
+  .where("array_length(benefitting_countries, 1) >  array_length(intended_beneficiaries(, 1)")
+
+puts "Fixing up #{finder.count} activities..."
+
+finder.each do |activity|
+  activity.benefitting_countries = [activity.iintended_beneficiaries]
```
edavey, line 35: little typo here in `= [activity.iintended_beneficiaries]`

db/data/20210930151236_intended_benefeciaries_to_benefitting_countries.rb
```diff
@@ -0,0 +1,39 @@
...
+# - there are `intended_beneficiaries`, and
+# - the number of countries in `benefitting_countries` is greater than those in `intended_beneficiaries`
+#
+# and set the benefitting countries to that of intended_beneficiaries
+# e.g. `["NG", "ZA"]` in the example above.
+
+finder = Activity
+  .where.not(recipient_region: nil)
+  .where(recipient_country: nil)
+  .where("array_length(benefitting_countries, 1) >  array_length(intended_beneficiaries(, 1)")
```
edavey, line 30: I think an extra parens has slipped into the SQL fragment here and that we actually want:

`.where("array_length(benefitting_countries, 1) > array_length(intended_beneficiaries, 1)") `

otherwise I get:

```
ERROR:  syntax error at or near "," (PG::SyntaxError)
LINE 1: ...g_countries, 1) >  array_length(intended_beneficiaries(, 1))
```

db/data/20210930151236_intended_benefeciaries_to_benefitting_countries.rb
edavey: also `Activity#intended_beneficiaries` returns an array so I think we need to avoid wrapping it in another array here...

i.e. the assignment should be:

`activity.benefitting_countries = activity.intended_beneficiaries`

edavey (commented):
a few tweaks needed, I think, in order to see the glad tiding of:

```
Fixing up 439 activities...
-> there are now 0 activities still to fix
```

db/data/20210930151236_intended_benefeciaries_to_benefitting_countries.rb
mec: That's odd, I thought I ran it?! 👍

edavey (approved):
🚀

PR #1380, opened 2021-10-01
(2203) Add comments tab to report

pezholio:
This adds a comments tab to the Report page, showing a list of all comments made in a reporting period, grouped by their activity

image

### DIFF
--- a/app/services/report/grouped_comments_fetcher.rb
+++ b/app/services/report/grouped_comments_fetcher.rb
@@ -1,0 +1,31 @@
+class Report
+  class GroupedCommentsFetcher
+    def initialize(report:, user:)
+      @report = report
+      @user = user
+    end
+
+    def all
+      comments.group_by(&:activity)
+    end
+
+    private
+
+    attr_reader :report, :user
+
+    def comments
+      if user.delivery_partner?
+        report.comments.includes(
+          owner: [:organisation],
+          activity: [
+            parent: [
+              parent: [:parent],
+            ],
+          ]
+        )
+      else
+        report.comments.includes(:activity)
+      end
+    end
+  end
+end
--- a/spec/features/staff/users_can_view_reports_spec.rb
+++ b/spec/features/staff/users_can_view_reports_spec.rb
@@ -209,6 +209,26 @@ RSpec.feature "Users can view reports" do
         click_on "Download report as CSV file"
       end
     end
+
+    scenario "they can view all comments made during a reporting period" do
+      report = create(:report)
+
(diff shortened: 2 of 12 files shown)


pezholio:
[Add 'Comments' tab to Report](https://trello.com/c/cfZ7Gga3/2203-add-comments-tab-to-report)

spec/features/staff/users_can_view_reports_spec.rb
```diff
@@ -209,6 +209,26 @@ def expect_to_see_grouped_rows_of_reports_for_an_organisation(organisation:, exp
...
+      ].flatten
+
+      page = ReportPage.new(report)
+      page.visit_comments_page
+
+      expect(page).to have_comments_grouped_by_activity(
+        activities: activities,
+        comments: comments_for_report
+      )
+      expect(page).to have_no_edit_buttons_for_comments(comments_for_report)
```
edavey, line 230: I think we could avoid have a negative version of the `has_edit_buttons_for_comments?` matcher if we removed the expectation from the body and use `Enumerable#all?` e.g. 

```ruby
  def has_edit_buttons_for_comments?(comments)
    within ".govuk-table" do
      comments.all? do |comment|
        page.has_link?(href: edit_activity_comment_path(comment.activity, comment))
      end
    end
  end

```

In this way the negative assertion in the feature spec could be made with

`expect(page).not_to have_edit_buttons_for_comments(comments_for_report)`

But I'm not sure it's worth it? What do you think?

edavey (approved):
Really clear. I have:

- one question on whether we can use `Report#comments` as a simplification
- one v minor (weakly held) suggestion around a negative assertion

spec/features/staff/users_can_view_reports_spec.rb
pezholio: Oh definitely. I did want to do it that way, but had some issues (I forget what now! 😆). This approach is much nicer.

app/services/report/grouped_comments_fetcher.rb
```diff
@@ -0,0 +1,27 @@
...
+  class GroupedCommentsFetcher
+    def initialize(report:)
+      @report = report
+    end
+
+    def all
+      Comment
+        .includes(:activity)
+        .where(activity_id: activity_ids, report: report)
+        .group_by(&:activity)
```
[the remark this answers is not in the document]
pezholio: Yeah, much better approach. We didn't have a `has_many :comments` for the Report class, so I've added that too.

edavey (approved):
Nice work!

pezholio:
Ta! 👍

PR #1381, opened 2021-10-04
Add a support doc for resetting a user's MFA setting

edavey:
In the scenario where a user can't complete MFA because their phone number has
changed, support will need to manually reset the user's MFA settings via the
Auth0 control panel.


mec (approved):
🚢

PR #1382, opened 2021-10-04
(2196) Feature: include adjustments in calculation of variance

mec:
## Changes in this PR
Include adjustments of type `Actual` in the variance calclulation.

Also moves some content to the locales.

https://trello.com/c/FX0NPZux

### DIFF
--- a/config/locales/models/report.en.yml
+++ b/config/locales/models/report.en.yml
@@ -33,15 +33,24 @@ en:
         comment: Comment
   tabs:
     report:
-      summary: Summary
-      activities: Activities
-      forecasts: Forecasts
-      actuals: Actuals
-      variance: Variance
-      budgets: Budgets
-      current: Current
-      approved: Approved
-      comments: Comments
+      summary:
+        heading: Summary
+      activities:
+        heading: Activities
+      forecasts:
+        heading: Forecasts
+      actuals:
+        heading: Actuals
+      variance:
+        heading: Variance
+      budgets:
+        heading: Budgets
+      current:
+        heading: Current
+      approved:
+        heading: Approved
+      comments:
+        heading: Comments
   page_content:
     reports:
       title: Reports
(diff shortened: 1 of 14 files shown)


edavey:
[Include adjustments in calculation of variance (Activity#variance_for_report_financial_quarter)](https://trello.com/c/FX0NPZux/2196-include-adjustments-in-calculation-of-variance-activityvarianceforreportfinancialquarter)

spec/models/activity_spec.rb
```diff
@@ -981,6 +981,25 @@
...
+
+    it "does NOT include Adjustments of type 'Actual' for a different financial quarter in the calculation" do
+      project = create(:project_activity, :with_report)
+      report = Report.for_activity(project).first
+      create(:actual, parent_activity: project, value: 130.20, report: report, **current_quarter)
+      create(:adjustment, :actual, parent_activity: project, value: -30, report: report, **current_quarter.pred)
+
+      expect(project.actual_total_for_report_financial_quarter(report: report)).to eq(130.20)
+    end
+
```
edavey, line 1002: super-clear description of this behaviour, thks!

As an aside, I note that the somewhat magical way that a report can be initialised with an optional financial quarter is not really described in the unit tests. I think I will have a go at that as a "documentation" task!

config/locales/models/report.en.yml
```diff
@@ -33,15 +33,25 @@ en:
...
+        heading: Summary
+      activities:
+        heading: Activities
+      forecasts:
+        heading: Forecasts
+      actuals:
+        heading: Actuals
+      variance:
+        heading: Variance
+        body: This tab shows the variance between forecasted and actual spend, including any adjustments, in this reporting cycle.
```
edavey, line 46: I suspect that the change in behaviour contained here would be much clearer if you broke this down into 2 commits:

1. refactor with no changes to content
2. change in content to add the variance body "This tab shows the..."

edavey (approved):
Nice!

Just one suggestion to maybe make the refactor element separate from the content-change element....

spec/models/activity_spec.rb
mec: Is it not covered in the `HasFinancialQuarter` spec?

PR #1383, opened 2021-10-04
Fix (2222) Populate currency and transaction_type properties for Refund and Adjustment records

edavey:
Refunds and Adjustments were being created with blank attributes:

- `currency` should be "GBP"
- `transaction_type` should be "3" for "Disbursement"

This isn't causing an immediate problem, but both refunds and adjustments are "transactions" and so
ought to conform to the normal transaction behaviour.

In this PR we:

- add attribute initialisers to `Refund` and `Adjustment`
- add a data migration to fix up existing records 
  (to be run with `rails runner db/data/20211004_backfill_missing_attrs_on_refunds_and_adjustments.rb`) 

## Next steps

### DIFF
--- a/app/models/adjustment.rb
+++ b/app/models/adjustment.rb
@@ -16,6 +16,9 @@ class Adjustment < Transaction
   validates_associated :comment
   validates_associated :detail
 
+  attribute :currency, :string, default: "GBP"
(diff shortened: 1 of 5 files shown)


edavey:
[Fixup missing Refund & Adjustment properties](https://trello.com/c/LlSJiAIp/2222-fixup-missing-refund-adjustment-properties)

pezholio (approved):
LGTM 👍

PR #1385, opened 2021-10-05
Document 2 support tasks

edavey:
- Pull back a submitted report
  Allow a delivery partner to resume editing a submitted report.

- Change a report's financial period
  If a report is created with the wrong financial period it
  is a little fiddly to change.


doc/support_tasks/change_financial_period.md
```diff
@@ -0,0 +1,17 @@
+# Change a report's financial period
+
+New reports are created automatically when earlier reports are approved. For
+example if I have a GCRF Q1 report in a submitted state, once it is approved, a
+new GCRF report for Q2 will be automatically created. However, if my Q1 report
+is not approved before the start of Q2, the automatically generated report will
```
mec, line 6: I think it would actually be 'before the start of Q3' i.e the report has not been approved inside the financial quarter following the one for which the report should contain the data for.

mec (approved):
I left a comment as I think this description makes it sound worse that it actually is!

PR #1388, opened 2021-10-06
Fix: Comments for report

mec:
## Changes in this PR
This is a fix to a bug we have in the new comment work that we found whilst working on the report csv.

The `comments_for_report` method on actvity still assumes that there is only ever one comment per activity, per report and so uses `find_by` returning the first found comment or nil.

In the report csv we attempt to map over the result of this method and if there are no comments, blow up sliently and produce an empty csv file.

As we now allow multiple comments per report, per activity, we have also updated the links against variance in a report as we want to direct uses to the activity comments rather than a specific comment, where they can then add or view the comments.

We have also styled the add comment link as a button as it is the primary action in the scope of the comment tab.

### Before
Screenshot 2021-10-06 at 08-46-55 FQ1 2021-2022 Q1 Report - FY 21 22 variance - Report your Official Development Assistance

### After
Screenshot 2021-10-06 at 08-46-23 FQ1 2021-2022 Q1 Report - FY 21 22 variance - Report your Official Development Assistance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -860,6 +860,8 @@
 
 ## [unreleased]
 
+- Consolidate links to commnents on the report, variance tab
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-78...HEAD
 [release-78]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-77...release-78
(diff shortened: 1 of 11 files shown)


edavey (approved):
👍 bravo! this is super-clear and thorough.

edavey:
[Fix: Comments for report](https://trello.com/c/x6xLJErn/2227-fix-comments-for-report)

PR #1411, opened 2021-10-18
2254 (Fix) ensure an adjustment's financial period is in the past (relative to this report)

edavey:
## Fixes to Adjustments

Trello: https://trello.com/c/WLTtbks2/2254-bug-allow-adjustment-to-be-associated-with-any-editable-report-and-more

### Ensure adjustment relates to an earlier (than this report's) time period

Adjustments exist to make corrections to financial periods which already 
have approved reports. Therefore an adjustment's financial period should
always be earlier than that of the report which which it's associated. If the
actual or refund which needs to be fixed is in the current report, then the
correction should be made by editing the relevant actual or refund in the
current report. If the current report is in the "submitted" state, then BEIS 
will need to move it back to "awaiting changes" in order to make this possible.

## Screenshot of validation error

adjustment_error

### DIFF
--- a/spec/models/export/all_activity_totals_spec.rb
+++ b/spec/models/export/all_activity_totals_spec.rb
@@ -5,23 +5,26 @@ RSpec.describe Export::AllActivityTotals do
     @activity = create(:project_activity)
 
     @q1_report = create(:report, financial_quarter: 1, financial_year: 2020)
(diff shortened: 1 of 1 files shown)


edavey:
[Bug: allow Adjustment to be associated with any editable report (and more)](https://trello.com/c/WLTtbks2/2254-bug-allow-adjustment-to-be-associated-with-any-editable-report-and-more)

PR #1413, opened 2021-10-18
(2259) Fix: always display adjustments on activity financials tab

edavey:
The "partial view" displaying adjustments was only being
rendered when it was permissible to *create* an adjustment
(i.e. when there's an editable report).

We now show the list of adjustments always.

Trello: https://trello.com/c/kh8HrhE8/2259-bug


edavey:
[Bug: Adjustments function not working](https://trello.com/c/kh8HrhE8/2259-bug-adjustments-function-not-working)

PR #1420, opened 2021-10-20
(2246) Refactor: use new activity totals class

mec:
## Changes in this PR

Welcome future folk! 🤖 

The real complexity for producing the actual totals (for me) is in the query used to get them from the database in a performant way.

We settled on the query that this work wraps in a class. I wanted to do this for a few reasons:

- to reuse the query in any export
- to document it
- to test it closer

The whole idea for how we can build exports from building blocks came from seeing how this query returns the data.

This class takes an activity and optional report.

### With no report 
It returns all acutals (actual spend, refunds and adjustments) for that activity summed by financial quarter.

### With a report
It returns all the actuals up to and including the report, anything added after the supplied report is not included and thus gives us a 'snapshot' in time of the finances as reported. I leverage the `Report.historically_up_to` class method to do this.

I spent a lot of time trying to make this elegant and in the end settled on this. I am not super in :heart: with the call method here, but it at least starts to breakup the giant query and document it.

I do like the slightly odd looking spec, for me it shows how the data is retruned, giving you the total for an activity, financial quarter and type.

By getting confidence here that this class is returning the correct values, we can ease the testing on the export itself, which I also like.

Here we use the class in the Spending breakdown export, with more to follow.

### DIFF
--- a/app/services/export/spending_breakdown.rb
+++ b/app/services/export/spending_breakdown.rb
@@ -50,12 +50,7 @@ class Export::SpendingBreakdown
   end
 
   def all_totals_for_activity(activity)
-    Transaction.joins("LEFT OUTER JOIN adjustment_details ON adjustment_details.adjustment_id = transactions.id")
-      .where(parent_activity_id: activity.id)
-      .select(:financial_quarter, :financial_year, :parent_activity_id, :value, :type, "adjustment_details.adjustment_type")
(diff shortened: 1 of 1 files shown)


mec:
Fully agree with you - once we get to the report csv and the actual motivation for the changes I think we'll be in a good place for even more refinements! 👍

PR #1436, opened 2021-10-27
(2246) Refactor: spending breakdown uses activity actuals columns

mec:
## Changes in this PR
Here we are at step 6️⃣ ! Now we have the `ActivityActualsColumns` class, we can use it, here in the spending breakdown report.

To complete the expected behaviour, we need to know the last finacial quarter we have actual data for and then use that to start showing forecast data. Here I extend 'ActivityActualsColumns` to return this.

In the last commit I take some steps to clean up the `spending_breakdown_spec` although the fixtures for forecasts are still a bit rough, they will be removed in coming work to split out a forecast columns class.

### DIFF
--- a/spec/services/export/spending_breakdown_spec.rb
+++ b/spec/services/export/spending_breakdown_spec.rb
@@ -7,13 +7,62 @@ RSpec.describe Export::SpendingBreakdown do
     @activity = create(:project_activity, organisation: @organisation)
     @source_fund = Fund.new(1)
 
-    create_q1_2020_actual_and_adjustments
-    create_q4_2020_actual_and_adjustments
+    q1_2019_report = create(
+      :report,
+      :approved,
+      organisation: @organisation,
+      fund: @activity.associated_fund,
+      financial_quarter: 1,
+      financial_year: 2019
+    )
 
-    create_q1_2020_refund_and_adjustments
-    create_q4_2020_refund_and_adjustments
+    ForecastHistory
+      .new(@activity, report: q1_2019_report, financial_quarter: 1, financial_year: 2020)
+      .set_value(10_000)
+    ForecastHistory
+      .new(@activity, report: q1_2019_report, financial_quarter: 4, financial_year: 2020)
+      .set_value(5_00)
+    ForecastHistory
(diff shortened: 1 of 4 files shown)


spec/services/export/spending_breakdown_spec.rb
```diff
@@ -7,13 +7,62 @@
...
+      .set_value(2_500)
+    ForecastHistory
+      .new(@activity, report: q4_2019_report, financial_quarter: 1, financial_year: 2021)
+      .set_value(20_000)
+    ForecastHistory
+      .new(@activity, report: q4_2019_report, financial_quarter: 4, financial_year: 2021)
+      .set_value(10_000)
+
+    @q1_report = create(:report, financial_quarter: 1, financial_year: 2020)
+    @q2_report = create(:report, financial_quarter: 2, financial_year: 2020)
```
edavey, line 55: do you think it would be clearer to present these 8 forecast fixtures as a table also? e.g.

```
    create_forecast_fixtures(
      <<~TABLE
        |report|financial_quarter|financial_year|value|
        |q1_19 |   1             |   2020       |10000|
        |q1_19 |   4             |   2020       | 5000|
        |q1_19 |   1             |   2021       |10000|
        |q1_19 |   4             |   2021       |20000|
        |q4_19 |   1             |   2020       | 5000|
        |q4_19 |   4             |   2020       | 2500|
        |q4_19 |   1             |   2021       |20000|
        |q4_19 |   4             |   2021       |10000|
      TABLE
    )
```

edavey (approved):
This is very clear. 

I'm just wondering if you think the forecast fixtures would be clearer in tabular form also?

I've pushed some example code to https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/tree/refactor/2246_experimental if you want to consider...

spec/services/export/spending_breakdown_spec.rb
mec: Can you wait? It is coming after the forecast work in #1437 is used in the spending breakdown?

edavey: yes, of course! I saw that you are using tables of forecast fixtures in the next PR!

PR #1437, opened 2021-10-27
(2246) Feature new forecast columns export

mec:
## Changes in this PR
Number 7️⃣ in the series!

We extract the forecast columns out of the spending breakdown class. 

We then extend it to allow it to produce a 'snapshot' of forecasts which shows the forecasts values as they were when reported.

We also allow it to take a 'starting financial quarter', this sets the first financial quarter in the series and can be used in exports that require the latest forecasted values but only for a specific set of financal quarters, for example in the spending breakdown export.

### DIFF
--- a/app/models/export/activity_forecast_columns.rb
+++ b/app/models/export/activity_forecast_columns.rb
@@ -1,7 +1,12 @@
 class Export::ActivityForecastColumns
-  def initialize(activities:, report: nil)
+  def initialize(activities:, report: nil, starting_financial_quarter: nil)
     @activities = activities
     @report = report
+    @starting_financial_quarter = starting_financial_quarter
+
+    if @report && @starting_financial_quarter
+      raise ArgumentError.new("Provide either a report for historical forecasts or a starting_financial_quarter for the latest forecast")
+    end
   end
 
   def headers
@@ -57,15 +62,25 @@ class Export::ActivityForecastColumns
     @_forecast_quarter_range ||= begin
       return [] if all_financial_quarters_with_forecasts.blank?
 
-      if @report.nil?
-        Range.new(all_financial_quarters_with_forecasts.min, all_financial_quarters_with_forecasts.max)
-      else
+      if exporting_most_recent_forecasts?
+        Range.new(@starting_financial_quarter, all_financial_quarters_with_forecasts.max)
+      elsif exporting_forecasts_for_given_report?
         Range.new(report_financial_quarter(@report), all_financial_quarters_with_forecasts.max)
+      else
+        Range.new(all_financial_quarters_with_forecasts.min, all_financial_quarters_with_forecasts.max)
       end
     end
   end
 
   def report_financial_quarter(report)
     FinancialQuarter.new(report.financial_year, report.financial_quarter)
+  end
+
+  def exporting_most_recent_forecasts?
+    @starting_financial_quarter.present?
+  end
+
+  def exporting_forecasts_for_given_report?
+    @report.present?
   end
 end
--- a/app/models/export/activity_forecast_columns.rb
+++ b/app/models/export/activity_forecast_columns.rb
@@ -1,6 +1,7 @@
 class Export::ActivityForecastColumns
-  def initialize(activities:)
+  def initialize(activities:, report: nil)
     @activities = activities
+    @report = report
   end
 
   def headers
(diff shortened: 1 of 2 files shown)


app/models/export/activity_forecast_columns.rb
```diff
@@ -0,0 +1,61 @@
...
+    @_forecast_quarter_range ||= begin
+      return [] if all_financial_quarters_with_forecasts.blank?
+
+      Range.new(all_financial_quarters_with_forecasts.min, all_financial_quarters_with_forecasts.max)
+    end
+  end
+
+  def financial_quarters
+    all_financial_quarters_with_forecasts
+  end
```
edavey, line 81: I don't think this method is being used?

Maybe you were planning to call this generally rather than `all_financial_quarters_with_forecasts`?

spec/models/export/activity_forecast_columns_spec.rb
```diff
@@ -3,60 +3,195 @@
...
+    )
+    forecasts_for_report_from_table(q1_2019_report,
+      <<~TABLE
+        |financial_quarter|financial_year|value|
+        |1                |2020          | 5000|
+        |4                |2020          | 2500|
+        |1                |2021          |20000|
+        |4                |2021          |10000|
+      TABLE
+    )
```
edavey, line 45: this is a nice way of presenting the forecast fixtures: one table per report.

spec/models/export/activity_forecast_columns_spec.rb
```diff
@@ -3,60 +3,195 @@
...
+    describe "#rows" do
+      it "contains the financial data for financial quarter of the report" do
+        expect(value_for_header("Forecast FQ1 2019-2020").to_s).to eql("0")
+      end
+
+      it "contains the financial data for financial quarter 1 2021-2022" do
+        expect(value_for_header("Forecast FQ1 2021-2022").to_s).to eql("20000.0")
+      end
+
+      it "contains the financial data for financial quarter 4 2021-2022" do
```
edavey, line 197: I wonder if it's worth being really explicit here that the forecast data returned is for the financial period, as reported in the given report as distinct from the forecast given for the same period in other reports?

e.g. something like:

```
"contains the the forecast for Q1 2021-2022 as provided in the given report"
```

app/models/export/activity_forecast_columns.rb
```diff
@@ -57,10 +62,12 @@ def forecast_financial_quarter_range
     @_forecast_quarter_range ||= begin
       return [] if all_financial_quarters_with_forecasts.blank?
 
-      if @report.nil?
-        Range.new(all_financial_quarters_with_forecasts.min, all_financial_quarters_with_forecasts.max)
-      else
+      if @starting_financial_quarter.present?
+        Range.new(@starting_financial_quarter, all_financial_quarters_with_forecasts.max)
+      elsif @report.present?
```
edavey, line 67: would it be helpful to wrap these in more descriptive method names -- so that it's clear why a caller would use one or other param? e.g.

```rb
def exporting_most_recent_forecasts?
  @starting_financial_quarter.present?
end

def exporting_forecasts_for_given_report?
  @report.present?
end
```

app/models/export/activity_forecast_columns.rb
```diff
@@ -1,7 +1,12 @@
 class Export::ActivityForecastColumns
-  def initialize(activities:, report: nil)
+  def initialize(activities:, report: nil, starting_financial_quarter: nil)
     @activities = activities
     @report = report
+    @starting_financial_quarter = starting_financial_quarter
+
+    if @report && @starting_financial_quarter
+      raise ArgumentError.new("Only use report OR starting_financial_quarter")
```
edavey, line 8: could we be a little more helpful with this error message, e.g. something like:

```
"Provide either a report for historical forecasts or a starting_financial_quarter for the latest forecasts"
```

?

edavey (approved):
Very easy to follow, thks.

I've made a couple of minor suggestions which might make the "past" versus "latest" options clearer...

spec/models/export/activity_forecast_columns_spec.rb
mec: I have you to thank for it! 👍

app/models/export/activity_forecast_columns.rb
mec: It is just a hangover from the spending breakdown report where it would have included 'financial quarters with acutals' in addition - it does nothing now, I'll remove it - thank you!

app/models/export/activity_forecast_columns.rb
mec: Yes! Thank you!

PR #1438, opened 2021-10-27
(2271) Data migration: fix implementing org on level B activities

mec:
## Changes in this PR
This is a short term fix for implementing organisation on all level B
activities that have an extending organisation (all of them).

We take the extending organisation and use it to create a new
implementing organisation on activities that do not have any.

This will mean IATI data at level B will include an implementing
organisation which is not currently the case.

There is a longer term issue to fix: When creating new level B
activities the implementing organisation should be set to the delivery
partner, but that work will take longer and so we are correcting the
data here to enable IATI publishing to happen in the short term.

https://trello.com/c/lirfmCWM


edavey (approved):
👍 this fixed up 774 programmes for me.

PR #1440, opened 2021-10-27
(2272) Fix: benefitting country percentage split calculation

mec:
## Changes in this PR
The old calculation using `divmod` fails for numbers > 100.

As we have ~180 countries and selecting all of them is something that
happens we need the percentage calculation to work.

What matters to us right now is that the IATI XML is valid and NOT that
the percentage is accurate (it is not and for BEIS is may be impossible
to make it accurate).

I tested what percentages can be valid in the IATI XML for both 3
countries and 180 and this simpler calculation satisfies the validator.

### DIFF
--- a/app/helpers/activity_helper.rb
+++ b/app/helpers/activity_helper.rb
@@ -39,18 +39,12 @@ module ActivityHelper
   def benefitting_countries_with_percentages(benefitting_countries)
     return [] if benefitting_countries.blank?
 
-    # Get an equal percentage split between all the countries
-    # (together with the remainder if possible)
-    percentage, remainder = 100.divmod(benefitting_countries.count)
-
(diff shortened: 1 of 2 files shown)


edavey (approved):
It seems wrong to be reporting with such an extreme level of accuracy when actually we ought to be as imprecise as possible to reflect the actual (lack of) information in RODA.

However, in the circumstances, this is a pragmatic solution.

mec:
> It seems wrong to be reporting with such an extreme level of accuracy when actually we ought to be as imprecise as possible to reflect the actual (lack of) information in RODA.
> 
> However, in the circumstances, this is a pragmatic solution.

I agree and did consider the alteratives for a long while - it is a really interesting problem!

PR #1443, opened 2021-11-01
(2276) Chore: delete activity rake task

mec:
## Changes in this PR

Every report results in a bunch of activities that a developer has to
delete as the application cannot do this yet.

This rake task attempts to aid developers in performing this support
task by running the steps for a given activity database ID.

Some associations are destroyed along with the `Activity` object, others
would be orphaned. I wanted to update the documentation we have on this
and through the process of finding out, I figured we may as well wrap
the knowledge in a rake task.

Developers are encouraged to `think` about what they are doing here as
ultimately this is destructive, but they save a great deal of work on
the console by running the task and interpreting the result before performing 
the destruvtive action.

### DIFF
--- a/doc/deleting-things-from-production.md
+++ b/doc/deleting-things-from-production.md
@@ -1,38 +1,35 @@
+## Deleting activities
 Users cannot delete activities via the web app.
 
-Requests to delete activities usually come in via Zendesk support tickets.
+Requests to delete activities should come via Zendesk at the end of reporting
+and a complete, approved list of activities to be deleted must be provided.
 
-Sometimes we'll be given the database ID of the activity to delete:
+A rake task is available to help you delete activities and their associations.
 
-```
-activity = Activity.find(id)
-```
+Connect to production to run the task, see [console access](./console-access.md)
 
-Most often we'll be given its RODA identifier:
+You will need the database ID of each activity, often we'll be given the activity's 
+RODA identifier, so you first need to locate the database ID on the Rails console:
 
-```
-activity = Activity.by_roda_identifier("REPLACE-ME")
-```
-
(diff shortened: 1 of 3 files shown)


edavey (approved):
Should we have any hard control on activity deletion? e.g. if there are financial data (actual, refund, adjustment, transfer etc) associated with an approved report, should we just refuse to delete the activity? It seems to me that if a mistake has got so far that it's been assured, approved and passed downstream to OECD / IATI then it can't just be "disappeared" and that so other corrective mechanism is needed...

mec:
I agree with you, anything that has been assured and approved should never be deleted - but I am not convinced we are in that world just yet:

If we imagine that we cannot delete an activity that has approved data in the current world we would have to delete the errant activities before the assurance is complete i.e. the final step in assurance is to delete the errant acticities. I think this would place too great a burden on the support team becoming the gate keepers to approval, which could only happen once the deleting has taken place.

My view is that right now, we have to rely on the BEIS folks supplying the correct list of activities for deletion and an developer running the task only has to deal with making sure the ID used is the correct one and not that the activity itself is correct to be deleted, which should not be our problem.

My hope is that as the service matures we will see less activities to be deleted or even better, move deletion into the application where constraints like the one proposed could be enforced - I think there is more thought to be put into exactly what an acitivity that can be deleted would look like, there may be more than we have considered. In that instance, the assurance team could delete the errant activity before it ever got approve - though I suspect mistakes will still get through!

We have even previously considered that activites are NEVER deleted and merely marked as 'errors' and not shown to anyone unless explicitally requested - unfortunately that thinking and work never made it to the top of the priority order! :(

As you have given approval I'll assume this is a thinking outloud comment and go ahead with this PR - but let me know if you think we should add something, or at least flag something in this task?

PR #1444, opened 2021-11-02
(2246) refactor: spending breakdown export uses new forecast column

mec:
## Changes in this PR
Nmber 8️⃣ in the series!

Now that the forecast columns behaviour has it's own class
(Export::ActivityForecastColumns from #1437) the spending breakdown report can be
generated using it.

This frees the spending breakdown export itself to only deal with
collating the requirements with which to build the export, but not
actually do the work.

I've also used the in-line fixtures seen elsewhere in the `Export` classes
to make the spec more readable.

### DIFF
--- a/spec/services/export/spending_breakdown_spec.rb
+++ b/spec/services/export/spending_breakdown_spec.rb
@@ -10,51 +10,43 @@ RSpec.describe Export::SpendingBreakdown do
     q1_2019_report = create(
       :report,
       :approved,
-      organisation: @organisation,
+      organisation: @activity.organisation,
       fund: @activity.associated_fund,
       financial_quarter: 1,
       financial_year: 2019
     )
-
-    ForecastHistory
-      .new(@activity, report: q1_2019_report, financial_quarter: 1, financial_year: 2020)
-      .set_value(10_000)
-    ForecastHistory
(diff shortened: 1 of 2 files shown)


spec/models/export/activity_forecast_columns_spec.rb
```diff
@@ -243,6 +243,8 @@
     end
   end
 
+  private
+
```
edavey, line 247: I don't think you intended this to be part of the commit?

spec/services/export/spending_breakdown_spec.rb
```diff
@@ -213,7 +205,9 @@ def value_for_header(header_name)
     end
   end
 
-  def create_fixtures(table)
+  private
```
edavey, line 208: I'm not sure there's any need to use the `private` keyword in a spec?

edavey (approved):
Very nice, it must be so satisfying to see this plan coming together so neatly!

One thing, as this is a refactoring -- or rather two refactorings:

1. the spending breakdown, and
2. its spec

it would perhaps be clearer that no behaviour has changed if this PR were in two commits:

1. refactoring of the spending breakdown
2. refactoring of the spec

Both will be green I believe?

PR #1446, opened 2021-11-03
(2246) feature: Variance column

mec:
## Changes in this PR
Welcome to 9️⃣ !

This is the variance column. At this point I kept this one simple with some baked in assumptions. The main one being that this is only ever used to export a report and so has a  fixed financial quarter and so the variance will be the last financial quarter from the actuals and the first financial quarter from the forecasts.

I can see nice ways to improve this later with further refactoring on the underlying data structure, but I would like to complete this Export epic work so we can deliver on the user need and come back if there is the opportunity (and yes I know that is a classic thing to say! 😄 )

### DIFF
--- a/app/models/export/activity_actuals_columns.rb
+++ b/app/models/export/activity_actuals_columns.rb
@@ -23,8 +23,7 @@ class Export::ActivityActualsColumns
 
   def rows
     return [] if @activities.empty?
-
-    @activities.map { |activity|
+    @_rows ||= @activities.map { |activity|
       actual_and_refund_data(activity)
     }.to_h
   end
--- a/app/models/export/activity_actuals_columns.rb
+++ b/app/models/export/activity_actuals_columns.rb
@@ -32,6 +32,10 @@ class Export::ActivityActualsColumns
     financial_quarter_range.max
   end
 
+  def rows_for_last_financial_quarter
+    rows.each_with_object({}) { |(key, values), obj| obj[key] = values.last }
+  end
+
   private
 
   def actual_and_refund_data(activity)
--- a/app/models/export/activity_forecast_columns.rb
+++ b/app/models/export/activity_forecast_columns.rb
@@ -25,6 +25,10 @@ class Export::ActivityForecastColumns
     }.to_h
   end
 
+  def rows_for_first_financial_quarter
+    rows.each_with_object({}) { |(key, values), rows| rows[key] = values.first }
+  end
+
   private
 
   def forecast_data(activity)
--- a/app/models/export/activity_forecast_columns.rb
+++ b/app/models/export/activity_forecast_columns.rb
@@ -48,9 +48,16 @@ class Export::ActivityForecastColumns
   end
 
   def forecasts_to_hash
-    @_forecasts_to_hash ||= forecasts.each_with_object({}) { |forecast, hash|
-      hash[[forecast.parent_activity_id, forecast.financial_quarter, forecast.financial_year]] = forecast.value
-    }
+    @_forecasts_to_hash ||=
+      forecasts.each_with_object({}) { |forecast, hash|
+        hash[
+          [
+            forecast.parent_activity_id,
+            forecast.financial_quarter,
+            forecast.financial_year,
+          ]
+        ] = forecast.value
+      }
   end
 
   def all_financial_quarters_with_forecasts
--- a/spec/models/export/activity_actuals_columns_spec.rb
+++ b/spec/models/export/activity_actuals_columns_spec.rb
@@ -222,7 +222,23 @@ RSpec.describe Export::ActivityActualsColumns do
     end
   end
 
-  private
+  context "when #rows is called multiple times" do
+    let(:breakdown) { false }
+    let(:report) { nil }
+
+    before do
+      totals_double = double(call: {"fake_activity_id" => []})
+      allow(Export::AllActivityTotals).to receive(:new).and_return(totals_double)
+
+      3.times { subject.rows }
+    end
+
+    it "builds totals for the activities only once" do
+      expect(Export::AllActivityTotals)
+        .to have_received(:new)
+        .exactly(@activities.count).times
+    end
+  end
 
   def value_for_header(header_name)
     values = subject.rows.fetch(@activity.id)
--- a/spec/models/export/activity_actuals_columns_spec.rb
+++ b/spec/models/export/activity_actuals_columns_spec.rb
@@ -119,6 +119,16 @@ RSpec.describe Export::ActivityActualsColumns do
         end
       end
     end
+
+    describe "rows_for_last_financial_quarter" do
+      it "returns the rows for the last column in the set (Q4 2020)" do
+        last_column_data = subject.rows_for_last_financial_quarter
+        value_for_activity = last_column_data.fetch(@activity.id)
+
+        expect(value_for_activity).to eq BigDecimal(300)
+        expect(last_column_data.count).to eq 5
+      end
+    end
   end
 
   context "when a breakdown is requested" do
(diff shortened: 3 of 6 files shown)


app/models/export/activity_actuals_columns.rb
```diff
@@ -23,10 +23,10 @@ def headers
...
   def rows
     return [] if @activities.empty?
-
-    @activities.map { |activity|
-      actual_and_refund_data(activity)
-    }.to_h
+    @_rows ||=
+      @activities.map { |activity|
+        actual_and_refund_data(activity)
+      }.to_h
```
edavey, line 29: did you consider describing (and protecting) this optimisation in a spec?

e.g. something along the lines of:

```rb
context "when  #rows is called multiple times" do
  before do
    allow(Export::FinancialQuarterActivityTotals).to receive(:new).and_return(fake_totals)

    3.times { subject.rows }
  end
  
  it "builds totals for actuals once only" do
    expect(Export::FinancialQuarterActivityTotals)
      .to have_received(:new)
      .with(hash_including(type: :actual))
      .once
  end
  
  it "builds totals for refunds once only" do
    expect(Export::FinancialQuarterActivityTotals)
      .to have_received(:new)
      .with(hash_including(type: :refund))
      .once
  end
end
```

app/models/export/activity_actuals_columns.rb
```diff
@@ -33,6 +33,10 @@ def last_financial_quarter
     financial_quarter_range.max
   end
 
+  def rows_for_last_financial_quarter
+    rows.each_with_object({}) { |(key, values), rows| rows[key] = values.last }
```
edavey, line 37: just a little suggestion of the naming of the block variable:

```txt
rows.each_with_object({}) { |(key, values), rows| rows[key] = values.last }
 ^                     ^                      ^     ^
 |                     |                      |     |
rows (1)                ----------rows (2)-------------
is a method                       is a block variable
 
These two "rows" don't stomp on each other as they have a different scope but 
perhaps you can make the expression easier on the eye if you use a different block variable?

```
Something like:
```
rows.each_with_object({}) { |(key, values), obj| obj[key] = values.last }

```
perhaps?

edavey (approved):
Nice work, this series is gripping!

I've made a couple of tiny suggestions...

app/models/export/activity_actuals_columns.rb
mec: I did consider it, but decided it was testing ruby behaviour - happy to add it in though - especially as I have a way to do so now - thank you!

mec: @ed, what do you think about this:

```ruby
    context "when #rows is called multiple times" do
      let(:breakdown) { false }
      let(:report) { nil }

      before do
        totals_double = double(call: {"fake_activity_id" => []})
        allow(Export::AllActivityTotals).to receive(:new).and_return(totals_double)

        3.times { subject.rows }
      end

      it "builds totals for the activities only once" do
        expect(Export::AllActivityTotals)
          .to have_received(:new)
          .at_least(@activities.count).times
      end
    end
```

`Export::AllActivityTotals` is called on each activity, so in the spec 5 activites means 5 calls when cached?

edavey: 👍 makes sense, very clear but I think `at_least(n)` is too lax? That will be as happy with 15 times as with 5 times. Can we be precise, e.g. `.exactly(5).times` to demonstrate that the expensive operation are performed only once?

mec: 🤦 yep, I meant exactly and was looking for the syntax, I meant to change it! 👍

app/models/export/activity_forecast_columns.rb
```diff
@@ -19,8 +19,7 @@ def headers
 
   def rows
     return [] if @activities.empty?
-
-    @activities.map { |activity|
+    @_rows ||= @activities.map { |activity|
```
mec, line 22: @ed, the forecasts are already cached pretty heavily, but it is even harder to test this, do you feel the same way about the forecasts class spec?

edavey: I certainly don't think it's essential!

Looking at `ActivityForecastColumns`, would it be a case of asserting that `ForecastOverview` receives `:new` `1 x n` rather than `3 x n` times?

Is the difficulty in stubbing out the values returned by `ForecastOverview.new(activity_ids).latest_values`?

edavey: but if the `ForecastOverview` already caches the expensive work (i.e. is safe to call repeatedly) then we shouldn't worry about memoising with `@_rows ||=` -- nor demonstrating that we have done so in a test!

edavey: looking at 

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/master/app/models/export/activity_forecast_columns.rb#L40-L48

I see that the calls to ForecastOverview are themselves memoised, so yes, that looks well and truly covered already!

mec: Yep, I think it is stubbing `latest_values` was my issue last night - looking at the whole class in the cold light of day this morning, the caching on rows is not really needed.

I've taken the caching off the rows method in `ActivityForecastColumns` and added coverage around the main caching of `ForecastOverview` as you mentioned - this has been pulled out into a seperate commit.

Would you mind giving it another look for my peace of mind?

app/models/export/activity_forecast_columns.rb
```diff
@@ -48,9 +52,16 @@ def forecasts
...
+    @_forecasts_to_hash ||=
+      forecasts.each_with_object({}) { |forecast, hash|
+        hash[
+          [
+            forecast.parent_activity_id,
+            forecast.financial_quarter,
+            forecast.financial_year,
+          ]
+        ] = forecast.value
+      }
```
edavey, line 64: I think this is just a formatting change now?

spec/models/export/activity_actuals_columns_spec.rb
```diff
@@ -222,7 +232,23 @@
...
+      totals_double = double(call: {"fake_activity_id" => []})
+      allow(Export::AllActivityTotals).to receive(:new).and_return(totals_double)
+
+      3.times { subject.rows }
+    end
+
+    it "builds totals for the activities only once" do
+      expect(Export::AllActivityTotals)
+        .to have_received(:new)
+        .exactly(@activities.count).times
```
edavey, line 249: nice!

spec/models/export/activity_forecast_columns_spec.rb
```diff
@@ -241,6 +255,24 @@
...
+        forecast_overview_double = double(latest_values: [])
+        allow(ForecastOverview).to receive(:new).and_return(forecast_overview_double)
+
+        3.times { subject.rows }
+      end
+
+      it "builds totals for the activities only once" do
+        expect(ForecastOverview)
+          .to have_received(:new)
+          .exactly(1).times
```
edavey, line 273: or `.once`!

spec/models/export/activity_variance_column_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      when "Actual"
+        create(:actual, fixture_attrs(row))
+      when "Adj. Act."
+        create(:adjustment, :actual, fixture_attrs(row))
+      when "Adj. Ref."
+        create(:adjustment, :refund, fixture_attrs(row))
+      when "Refund"
+        create(:refund, fixture_attrs(row))
+      else
+        raise "don't know what to do"
```
edavey, line 86: Are you retaining all these redundant "factory"  case clauses for adjustments and refunds because you're planning to extract these "build transactions from table" methods to a single helper method?

If not, you might consider:

a. removing the "transaction" column from the table and 
b. losing all the unused case pathways here?

edavey (approved):
looking great. I've made a couple of tiny comments.

app/models/export/activity_forecast_columns.rb
mec: yeah, trying to make it more readable!

edavey: definitely an improvement in my book. 

The purist might say this formatting refactoring would be clearer in its own commit -- that way there's no effort expended in looking to understand what (if anything) has changed in this diff...

PR #1447, opened 2021-11-03
(2246) feature: activity link column

mec:
## Changes in this PR

Returns a column of links to activities in the application for use in
exports.

Not yet used in the applicaiton but it is coming…


edavey (approved):
🚀

PR #1448, opened 2021-11-04
(2246) feature: Introduce Activity Implementing Organisation Column

mec:
## Changes in this PR
Allows Implementing organisations to be exported.

Returns the name only as the only use case we have for this
right now is the report csv which only requires the name.

Multiple implementing organisations are delimited with a pipe (|).

When the activity is at Level B (a programme) there is no guarantee
there will be any implementing organisations, but the
extending organisation will have the correct organisation details, so we
use that.

We pass in an ActiveRecord::Relation of activities so we can leverage
Rails lazy loading of the organisations, failing to pass in the correct
type would cause an error, failing to pass in the correct type would
cause an error, so we check this.

### DIFF
--- a/spec/models/export/activity_implementing_organisation_column_spec.rb
+++ b/spec/models/export/activity_implementing_organisation_column_spec.rb
@@ -1,0 +1,105 @@
+RSpec.describe Export::ActivityImplementingOrganisationColumn do
+  before(:all) do
+    DatabaseCleaner.strategy = :transaction
+    DatabaseCleaner.start
+
+    @activity_with_single_organisation = activity_with_single_organisation
+    @activity_with_multiple_organisations = activity_with_multiple_organisations
+    @level_b_activity_with_extending_organisation = level_b_activity_with_extending_organisation
+  end
+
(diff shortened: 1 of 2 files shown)


spec/models/export/activity_implementing_organisation_column_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+    subject { described_class.new(activities_relation: Activity.all) }
+
+    describe "#headers" do
+      it "returns the correct header" do
+        expect(subject.headers).to match_array(["Implementing organisations"])
+      end
+    end
+
+    describe "#rows" do
+      context "when the acitivity is level C (project)" do
```
edavey, line 25: tiny typo in "acitivity"

edavey (approved):
👍 v easy to follow!

spec/models/export/activity_implementing_organisation_column_spec.rb
mec: You would think I could type 'activity' by now! 🤦

PR #1450, opened 2021-11-04
(2246) feature: Introduce ActivityCommentsColumn

mec:
## Changes in this PR
This class is concerned with exporting a column of activity comments.

You can pass in a `Report` and only the comments added in that report
will be returned or you can not, in which case you will get all comments
for each activity.

In either case there are an unknown number of comments that must be added
into a single table cell, so we output the comments separated with the
following:

`----`

which is surrounded by new lines.

We would usually delimit with a pipe but that did not feel like enough
(currently comments are separated by a new line).

### DIFF
--- a/app/models/export/activity_comments_column.rb
+++ b/app/models/export/activity_comments_column.rb
@@ -1,0 +1,30 @@
+class Export::ActivityCommentsColumn
+  def initialize(activities:, report:)
+    @activities = activities
+    @report = report
+  end
+
+  def headers
+    ["Comments in report"]
+  end
+
+  def rows
+    return [] if @activities.empty?
+
+    @activities.map { |activity|
+      [activity.id, comments_for_activity(activity).join("\n----\n")]
+    }.to_h
+  end
+
+  private
+
+  def all_comments_for_report
+    @_comments ||= @report.comments.includes(:commentable)
+  end
+
+  def comments_for_activity(activity)
+    all_comments_for_report.select { |comment|
+      comment.associated_activity.id == activity.id
+    }.pluck :body
+  end
+end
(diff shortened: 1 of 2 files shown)


app/models/export/activity_comments_column.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  private
+
+  def all_comments_for_report
+    @_comments ||= @report.comments.includes(:commentable)
+  end
+
+  def all_comments
+    @_comments ||= Comment.all.includes(:commentable)
```
edavey, line 26: Are we able to constrain this `all_comments` branch in some way? If the exporter is BEIS, then is the export for one organisation at a time? Or is there the ability to download every single comment in one giant export? 

Is it maybe worth profiling memory usage based on some much larger quantity of comments (an estimate of what might be in RODA in a couple of years)?

Maybe there's no problem and this is all using lazy loading cleverness? I haven't had a chance to play with it...

edavey (approved):
Looks good -- I just had a reservation about the `Comment.all` part, but didn't have time to look at it properly...

app/models/export/activity_comments_column.rb
mec: To be honest, we don't have a use for the all right now - I can take it out and still deliver the report csv - I'll do that for now.

mec: All of the export parts are very expensive - I would love to have some time to look at how they are performing in terms of memory (and other metrics) - not sure we will get the opportunity to do so though.

I like the idea that this class is not concerned with constraints - perhaps that is naive of me, but I like the idea that the responsibility of this is simply to take the activities and give you back the comments for export as tabular data, as a developer it is our responsibility to understand the impact of doing so for all activities and the cost of doing so.

Another one for a good discussion over coffee? :)

mec:
I've taken the `all_comments` out of here as it of no value right now, I will merge this as there is approval and we need to move onto using this in the report csv, I will be aiming to look at the overall performance of all the `Export` classes when used together once we have them in the report csv.

PR #1451, opened 2021-11-04
Fix/2097 total of forecasts should not include past values

edavey:
Trello: https://trello.com/c/iIgqfYUE/2097-total-of-forecasts-should-not-include-past-values

## Changes in this PR

Previously an activity's "financials" summary showed a total of all forecasts, irrespective of whether ~~actual spend had been reported~~ a report has been made for any of those forecasted periods.

We now total only the forecasts for periods which have not ~~had spend reported~~ been reported on.

## Screenshots of UI changes

total_forecasted_spend

A good example activity is shown above (RODA ID NF-MOBRCSSP-512) 

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -262,10 +262,15 @@ class Activity < ApplicationRecord
     Budget.direct.where(parent_activity_id: id).sum(:value)
   end
 
-  def total_forecasted
+  def future_forecasts
     activity_ids = descendants.pluck(:id).append(id)
-    overview = ForecastOverview.new(activity_ids)
-    overview.latest_values.sum(:value)
+    ForecastOverview.new(activity_ids)
+      .latest_values
+      .select { |forecast| forecast.later_period_than?(latest_report) }
+  end
+
+  def total_forecasted
+    future_forecasts.sum(&:value)
   end
 
   def own_and_descendants_actuals
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -142,6 +142,9 @@ class Activity < ApplicationRecord
 
   has_one :commitment, dependent: :destroy
 
+  has_many :reports,
+    ->(activity) { unscope(:where).for_activity(activity).in_historical_order }
+
   enum level: {
     fund: "fund",
     programme: "programme",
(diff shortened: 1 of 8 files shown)


edavey:
[Total of forecasts should not include past values](https://trello.com/c/iIgqfYUE/2097-total-of-forecasts-should-not-include-past-values)

app/models/activity.rb
```diff
@@ -142,6 +142,13 @@ class Activity < ApplicationRecord
 
   has_one :commitment, dependent: :destroy
 
+  has_one :latest_actual,
```
mec, line 145: The only thing I am not clear on is the reason we go through actual to get the latest report? Given reports are sequential and this is an activity, which gives us the delivery partner and fund, could we not just find the lastest report from `Report`?

I could be missing something as it is Monday and early! :)

mec (commented):
This is all good, I just would like to understand the through actual part.

app/models/activity.rb
edavey: No, I think you are right, that it's enough to consider the reports associated with an activity and not get into the detail of the "Actual spend" recorded.

For 3 example activities, over 2 reports, lets say they have either reported actual spend or not reported spend as follows:

```
| Activity   | Q2 report | Q3 report |
| ---------- |:---------:|:---------:|
| Activity A |     Y     |     Y     |
| Activity B |   **n**   |     Y     |
| Activity C |     Y     |   **n**   |
```

i.e. that only Activity A reported spend in both reports. But: probably gaps in reporting don't matter. In the case of Activity C, in terms of "showing future forecases only", I think it's irrelevant that no spend was reported in Q3. It still makes sense to only sum up forecasts for later than **Q3** (the latest report in which actual spend **could have been** reported).

Is this correct and reasonable?

At any rate I've tweaked this PR to remove the journey through actuals, so that the `Activity#reports` associated now uses `Report#for_activity` -- i.e. find reports with matching fund and organisation.

mec: Yep, I think that is correct, I think there is a baked in assumption that 'I reported no spend' is the same as 'I reported £0 spend' (assuming that it is not a mistake!) - once framed like this I think it can feel more natural - it's not that a posistive actual spend is required to then not show that quarters forecasts for example I might forecast £10,000 spend and actually spend £0, in this case the £10,000 is no longer relevent as the £0 has replaced it.

mec (approved):
Super nice work - thank you!

PR #1452, opened 2021-11-08
Fix: Spending breakdown export not cached

mec:
## Changes in this PR
When we moved the spending breakdown report to the new `Export` classes
we missed that we only need to load the `row` data once and then loop
over the rows to collate the data together, instead we loaded the entire
data set for each wow which  makes the export take far too long as all
the expensive calls are happening for each row instead of once for all
rows!


edavey (approved):
nice work. The v clear spec should prevent regressions here.

PR #1455, opened 2021-11-09
Fix: Free standing technical cooperation in report

mec:
## Changes in this PR
The ActivityCsvPresenter was not rendering this in a useful way and the
locale was incorrect.


edavey (approved):
🚀

PR #1462, opened 2021-11-11
(2246) Allow users to download legacy report csv

mec:
## Changes in this PR

Whilst we are introducing the new report csv we want to offer the old
one to reassure users that any workflows will not be broken. Once they
have confidence in the new export, this work can be rollback.


edavey (approved):
🚀 nice, this will be easy to revert in a week or two once we get confirmation that the new report supersedes the old one in every way!

PR #1463, opened 2021-11-11
(2246) Refactor: clean up

mec:
## Changes in this PR
Here we are cleaning up the end of the big Export refactor. We do not remove the old `Report::Export` just yet as we are still offering it incase the new version is missing something that would effect BEIS workflows, see: #1462


mec:
I relealised we still need the filename in report presenter whilst we support the legacy report csv, I'll split this into two and close this one.

PR #1464, opened 2021-11-11
(2246) New export helper

mec:
## Changes in this PR

A bunch of the specs that exercise the `Export` classes use inline
fixtures. Here we consolidate the helper methods that dirve those into
one place.


edavey (approved):
👍

PR #1469, opened 2021-11-15
Stub out logging to stdout (via 'puts') in specs

edavey:
The rake tasks output lots of useful info to stdout via the
`puts` method.

But the info is not so useful to see in the middle of the
RSpec output.

So we now stub the calls to `Kernel.puts`.


mec (approved):
I had no idea this was a thing! I couldn't find a way to stop the output in the specs. 🚢

PR #1472, opened 2021-11-17
(2252) Feature: new actual history importer

mec:
## Changes in this PR
This is the new importer for historic actual spend that ony BEIS users can use, it is not currently available that will follow.

Historic actual spend is treated differently to regular actual spend in that we expect negative values and we expect financial quarters in the past, as already mentioned, we only expect BEIS users to perform this upload and we don't expect it to be something long term, it is only fir use during migration to the service.

The import follows the `Import::Commitments` in that it returns `true` or `false` and builds arrays of `errors` and `imported` actuals - I like this interface.

Here I start out by preventing negative values in 'regular' actuals and allowing them when the validation context is `:history`. The importer then creates it's actuals in this context and is therefore allowed to add negative values.

In the early commits I decided that I wouldn't worry about validating the report that is used to import against - I come to a better conculsion in the final commit.

Big thank you to @pezholio who spiked this work out and made it simpler for me to follow up.

### DIFF
--- a/spec/models/import/actual_history_spec.rb
+++ b/spec/models/import/actual_history_spec.rb
@@ -1,0 +1,256 @@
+RSpec.describe Import::ActualHistory do
+  before(:all) do
+    DatabaseCleaner.strategy = :transaction
+    DatabaseCleaner.start
+
+    @activity = create(:project_activity)
+    @report = create(
+      :report,
+      organisation: @activity.organisation,
+      fund: @activity.associated_fund
+    )
+    @user = create(:beis_user)
+
+    valid_data = <<~CSV
+      RODA identifier,Financial quarter,Financial year,Value
+      #{@activity.roda_identifier},1,2021,10000
+      #{@activity.roda_identifier},2,2021,20000
+      #{@activity.roda_identifier},3,2021,30000
+    CSV
+
+    invalid_data = <<~CSV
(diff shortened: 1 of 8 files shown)


spec/models/import/actual_history_spec.rb
```diff
@@ -235,5 +235,36 @@
...
+      it "does not create the actual" do
+        expect(subject.actual).to be_nil
+      end
+
+      it "has a helpful error message" do
+        subject.call
+        expect(subject.errors.first.message)
+          .to include("does not match the fund and organisation of this report")
+      end
+    end
```
edavey, line 268: did you consider merging this commit into the earlier "New Actual history importer"? Did you feel that it was important to show this change in direction? If this is strictly a "fix" to something which was wrong in the earlier commit, then one might take the view that it's clearer to only show a corrected version of history and do away with the mis-step in the earlier commit...

edavey (approved):
Really clear work, easy to follow.

I just have one question about whether or not in would be clearer if the history were re-written to cut out the "change" step...

spec/models/import/actual_history_spec.rb
mec: It's a fair comment. I think by the time I reached here and really thougt about what would constitue a 'valid' report, I couldn't face the fixup - pure laziness! Let me see what I can do before merging this…

PR #1474, opened 2021-11-18
(2286) Check delete activity rake task is recursive

mec:
## Changes in this PR
When I wrote the delete activity task, I thought afterwards that it
would not delete descendent activities. Looking at it here though, it is
recursive and deletes right down the hierarchy.

Here is a spec to confirm that behaviour.

https://trello.com/c/WJwzZHXX

### DIFF
--- a/spec/lib/tasks/delete_activity_spec.rb
+++ b/spec/lib/tasks/delete_activity_spec.rb
@@ -145,5 +145,13 @@ RSpec.describe "rake activities:delete", type: :task do
 
       expect { delete_activity(activity_id: activity.id) }.to change { Activity.count }.by(-2)
     end
+
+    it "deletes the descendant activities" do
+      programme = create(:programme_activity)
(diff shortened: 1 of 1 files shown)


spec/lib/tasks/delete_activity_spec.rb
```diff
@@ -145,5 +145,13 @@
 
       expect { delete_activity(activity_id: activity.id) }.to change { Activity.count }.by(-2)
     end
+
+    it "deletes the descendant acitivities" do
```
edavey, line 149: `<nit-pick alert>` typo in "acitivities"

edavey (approved):
👍 nice to have this clearly laid out

spec/lib/tasks/delete_activity_spec.rb
mec: Not a nit pick! I need the spelling help! 👍

PR #1475, opened 2021-11-18
(2324) Fix: show unknown country codes and data migration to remove those in production 

mec:
## Changes in this PR
We discovered we has some 'unknown country codes' - interestingly this is exactly why we put 'unknown country codes' in. to discover any in the data set!

I've done two things:

1. We now show the unknown code along with the text, this will help users (and us) 'debug' the data
2. A data migration to remove the codes from the activities that have them now.

###Data migrattion to run in this PR
`rails runner db/data/20211118113311_remove_incompatible_country_codes.r`

https://trello.com/c/tHbVco0U

### DIFF
--- a/app/presenters/activity_presenter.rb
+++ b/app/presenters/activity_presenter.rb
@@ -274,15 +274,15 @@ class ActivityPresenter < SimpleDelegator
 
   private
 
-  def translate(reference)
-    I18n.t(reference)
+  def translate(*args)
+    I18n.t(*args)
   end
 
   def sentence_of_benefitting_countries(country_code_list)
     return nil unless country_code_list.present?
     benefitting_country_names = country_code_list.map { |country_code|
       benefitting_country = BenefittingCountry.find_by_code(country_code)
-      benefitting_country.nil? ? translate("page_content.activity.unknown_country") : benefitting_country.name
+      benefitting_country.nil? ? translate("page_content.activity.unknown_country", code: country_code) : benefitting_country.name
     }
     benefitting_country_names.to_sentence
   end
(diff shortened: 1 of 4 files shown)


app/presenters/activity_presenter.rb
```diff
@@ -282,7 +282,7 @@ def sentence_of_benefitting_countries(country_code_list)
     return nil unless country_code_list.present?
     benefitting_country_names = country_code_list.map { |country_code|
       benefitting_country = BenefittingCountry.find_by_code(country_code)
-      benefitting_country.nil? ? translate("page_content.activity.unknown_country") : benefitting_country.name
+      benefitting_country.nil? ? I18n.t("page_content.activity.unknown_country", code: country_code) : benefitting_country.name
```
edavey, line 285: Nice, it makes so much sense to include full and helpful information in error messages!

edavey (approved):
Nice 🕵️ work and fix.

app/presenters/activity_presenter.rb
mec: I probably should have used that `translate` method even though I am not sure why it is there, I might just change that before I merge this!

edavey:
👍 if you are in polishing mode you could consider fixing up the tiny typo in `db/data/20211118113311_remove_incomnpatible_country_codes.rb`

mec:
@ed thanks for the help on the Postgres arrays! 👍

app/presenters/activity_presenter.rb
mec: I had to update the `translate` method as it wasn't handling passed in attribute options:

```ruby
def translate(*args)
  I18n.t(*args)
end
```

PR #1477, opened 2021-11-18
2098 Use unique list of implementing organisations

edavey:
### The problem

Organisations are not well modelled in RODA. 

Organisations can play multiple roles. Are per [IATI](https://iatistandard.org/en/iati-standard/203/codelists/organisationrole/) participation can be described as follows:

> 1. *Funding*: The government or organisation which  
   provides funds to the activity.  
>  
> 2. *Accountable*: An organisation responsible for  
    oversight of the activity and its outcomes  
>  
> 3. *Extending*: An organisation that manages the budget  
   and direction of an activity on behalf of the funding  
   organisation  
>  
> 4. *Implementing*: The organisation that physically carries out the activity or intervention.

At present RODA uses 2 db tables:

1. `organisations` for "extending organisations", aka "delivery partners"
2. `implementing_organisations` for "implementing organisations"

There are two duplication issues:

- i. Due to the uncontrolled way they are managed, there are currently an ever increasing number of duplicate implementing organisations, created with slightly different names.

- ii. Many of the extending organisations (delivery partners) also have entries in the implementing organisation table

### The solution

Introduce a new canonical table of unique organisations and a table of "organisations participations" to allow an organisation to play multiple roles in a particular activity:

#### 1. Activity's implementing organisations

Allow an `Activity` to have many implementing organisations through "organisation participations", e.g.

```rb
  has_many :implementing_org_participations,
    -> { where("org_participations.role = 'Implementing'") },
    class_name: "OrgParticipation"
```

#### 2. Activity's delivery partner
Move from `Activity#extending_organisation` to `Activity#delivery partner`.

As per Meyrics and Leeky's proposed rfactoring to [Replace 'organisation' and 'extending_organisation' with a 'delivery_partner' attribute](https://trello.com/c/kJG0qMT3/1647) to use an `Activity#delivery_partner_id` foreign key to link to the new `unique_organisations` table.

### Steps to get there

1. [x] introduce implementing organisations
2. [x] migrate existing implementing orgs to new membership system
3. [ ] allow DPs to set new implementing orgs
4. [ ] allow BEIS to manage unique organisations
5. [ ] refactor extending organisation to use unique organisations
6. [ ] rename `Activity#extending_organisation` to `Activity#delivery_partner`

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -126,6 +126,13 @@ class Activity < ApplicationRecord
   belongs_to :extending_organisation, foreign_key: "extending_organisation_id", class_name: "Organisation", optional: true
   has_many :implementing_organisations, dependent: :destroy
   validates_associated :implementing_organisations
+  has_many :implementing_org_participations,
+    -> { where("org_participations.role = 'Implementing'") },
+    class_name: "OrgParticipation"
+  has_many :unique_implementing_organisations,
+    -> { distinct },
+    through: :implementing_org_participations,
+    source: :organisation
 
   has_many :budgets, foreign_key: "parent_activity_id"
   has_many :actuals, foreign_key: "parent_activity_id"
--- a/app/models/org_participation.rb
+++ b/app/models/org_participation.rb
@@ -1,0 +1,35 @@
+class OrgParticipation < ApplicationRecord
+  belongs_to :activity
+  belongs_to :organisation, class_name: "UniqueImplementingOrganisation"
+
+  attribute :role, :string, default: "Implementing"
+
+  # At present this join model is only linking "Implementing org" and "Activity"
+  # but we plan to use this for all type of Organisation, e.g.
+  #
+  #  Activity#has_one
+  #    :delivery_partner
+  #    through: :extending_org_participation
+  #
+  # Activity#has_one
+  #   :extending_org_participations,
+  #   -> { where("org_participations.role = 'Extending'") },
+  #   class_name: "OrgParticipation"
+  #
+  # In this way there will be one table of Organisations and one table of
+  # OrgParticipations, with OrParticipation#role distinguishing the types
+  # of particpation defined by IATI:
+  #
+  # 1. *Funding*: The government or organisation which
+  #    provides funds to the activity.
+  #
+  # 2. *Accountable*: An organisation responsible for
+  #     oversight of the activity and its outcomes
+  #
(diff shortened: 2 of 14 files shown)


edavey:
[Change how we collect and store implementing organisation](https://trello.com/c/8teLzwsh/2098-change-how-we-collect-and-store-implementing-organisation)

mec:
I've had a first pass over this - it looks fantastic! Two things spring to mind as initial thoughts though before I really dig in:

- on the Trello card we say 'Activities in RODA will have only one implemention organisation' - should we set this up now with `has_one :implementing_org_participations` or do we need to support many for activities that already have multiple implementing organisations?

- should we consider using the existing `organisations` table instead of creating the `unique_implementing_organisations` table?

`organisations` already has a basic set of roles: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/e72bc8d15738a7153a27aa8b0111012ce8629569/app/models/organisation.rb#L8 (one of which is delivery partner) that it might be good to extend into something more elegant as proposed there?

It may be that you are trying to avoid the complexity at this early stage and this forms part of later work, I wanted to check.

edavey:
Thanks Meyric,

> * on the Trello card we say 'Activities in RODA will have only one implemention organisation' - should we set this up now with `has_one :implementing_org_participations` or do we need to support many for activities that already have multiple implementing organisations?

When I spoke with Dan, he was clear that its important that some activities can have multiple implementing orgs, so it seems that the earlier decision isn't fully accepted!

> * should we consider using the existing `organisations` table instead of creating the `unique_implementing_organisations` table?
> 
> `organisations` already has a basic set of roles:
> 
> https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/e72bc8d15738a7153a27aa8b0111012ce8629569/app/models/organisation.rb#L8
> 
> (one of which is delivery partner) that it might be good to extend into something more elegant as proposed there?
> It may be that you are trying to avoid the complexity at this early stage and this forms part of later work, I wanted to check.

good point, I think I will have a go at using the existing table. I was thinking that `UniqueImplementingOrganisation` would become `UniqueOrganisation` and then `Organisation` but now that I've got a clear plan it's actually not that complex and probably will be simpler and clearer to work directly on the existing `organisations` table...

edavey:
I've opened a new PR: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1496

PR #1479, opened 2021-11-19
Fix setup script

Stretch96:
## Changes in this PR

* Ensures Apple Command Line tools are installed
* Ensures ruby-build is up to date
* Launches postgres and redis containers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -933,12 +933,13 @@
 - Channel of delivery code 90000, 'Other' added to the accepted codes list
 - Fix deployment notification
 
-## [unreleased]
+## [unreleased]
 - Health check endpoint includes basic sidekiq stats
 - Show users a warning about appending data when uploading actuals data
 - Update Readme - Add process to get terraform AWS credentials
 - New reports have a stricter validation to financial period
 - Documentation about logging
+- Fix setup script
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-90...HEAD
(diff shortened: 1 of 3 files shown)


mec:
This all looks good, but it is not clear to me which 'approach' the scripts are trying to take and it is possibly my fault!

The `backing-services-docker-compose.yml` exists for my benefit, I don't use rvm so I don't run the `script/setup`, etc

Certainly, `bootstrap` seems to assume Redis is running locally (https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/e72bc8d15738a7153a27aa8b0111012ce8629569/script/bootstrap#L52) and this change seems to imply it is inside Docker.

I don't mind which route we take, but I do think the scripts should be consistent, so for example if we are running Postgres and Redis inside Docker, should be have Docker in the Brefile?

Stretch96:
The change we had in mind, was to enforce launching redis / postgres via docker-compose, rather than running the services unisolated locally

This PR removes the commands in the script that attempt to run `redis-server`

There is a check to ensure that docker-compose is installed - which halts the setup, and informs the user that docker-compose isn't installed. Putting it in the Brewfile may cause issues, as there are multiple ways that someone may choose to install it (and didn't want to enforce installing it via homebrew)

Stretch96:
That's been removed in this PR:
https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1479/files#diff-2ad9e188678c4f3bef503efff261066b6c9df3a948bcaabe8bc10bdd2f45404aL50-L54

mec (approved):
🚢

PR #1480, opened 2021-11-19
Fix deploy notification

olivia-work:
* When using the Slack API, `<!here>` is used in place of `@here`, otherwise the message just shows '@here' in plain text rather than notifying user.


mec (approved):
🚢

PR #1482, opened 2021-11-22
Fix rows_for_first_financial_quarter

mec:
## Changes in this PR
Whilst the last column of actuals will always have values, there is a
high chance that the first column of forecasts will have none, when this
happens the value pulled out when rendering data will be `nil` and when
the variance column tries to calculate, the nil causes an error.

As a general rule, where we have no data we have to assume zero, so here
we replace nil with zero if there is no value.


edavey (approved):
🚀

PR #1483, opened 2021-11-22
Activity summary view uses BEIS locale strings

mec:
## Changes in this PR
We are slowing refactor where and what we use to describe the various
attributes in use throughout the service.

With this change the Activity summary view and report csv file share the
same locales.

Here the Activity summary view is updated to use the
`activerecord.attributes.activity` namespace whenever possible. Not
everything is an activerecord attributes so I have had to use my best
judgement when to include a key and when to not.

### Before
Screenshot 2021-11-22 at 12-14-34 Activity Newton International Fellowships 2018 19 - India - Details - Report your Officia

### After
Screenshot 2021-11-22 at 12-14-18 Activity Newton International Fellowships 2018 19 - India - Details - Report your Officia

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -929,6 +929,7 @@
 
 ## [unreleased]
 - Uploaded actual history must be in past financial quarters
+- Activity summary view uses BEIS approved attribute names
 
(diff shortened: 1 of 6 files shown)


edavey (approved):
Fiddly!

mec:
> Fiddly!

It was and is horrible. :(

PR #1486, opened 2021-11-23
(2252) Feature: actual history must be in the past

mec:
## Changes in this PR
In #1478 we decided the actual history importer should onlt allow values from past financial quarters. Here we change the `Import::ActualHistory` to enforce this. See the commit message for details on what 'the past' means in this context.

I've had a go at making the help on the initial upload content a little clearer? 🧑‍⚖️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -928,6 +928,7 @@
 - Update postgres to version 13
 
 ## [unreleased]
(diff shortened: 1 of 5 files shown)


edavey (approved):
Nice, I think adding this constraint will make life simpler!

PR #1487, opened 2021-11-23
(support) Add channel of delivery code to accepted codes

mec:
## Changes in this PR
BEIS have request that 90000 'Other' be added to the codes that the
application accepts.

This was requested on Zendesk ticket #15439

https://dxw.zendesk.com/agent/tickets/15439


edavey (approved):
very neat 🚀

PR #1488, opened 2021-11-23
(chore) Update channel of delivery code play

mec:
Speculative - @edavey let me know what you think?


edavey:
hmm, I think that this kind of workflow guidance is quite generic and applies to much work which arises in support. Maybe instead of adding this detail to every run-book, we could include it in the "Ways of working" doc which is being drafted?

Also I think the step:

> once the PR is approved contact the requester for their approval to perform a release, if necessary agree a time for the release

is perhaps unnecessarily prescriptive and process-heavy -- and even contradicts https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/architecture/decisions/0033-move-towards-more-continuous-delivery.md where we resolve that there's no formal process around preparing a release:

> Deployments will now be made regularly at the discretion of the team.

Do you think putting something in the "Ways of working" would work?

mec:
Yep, suits me. 👍

PR #1491, opened 2021-11-23
(2282) Feature: Add basic Sidekiq stats to health check endpoint

mec:
We want to know if the Sidekiq worker has failed for any reason and raise
an alert.

We decided it was acceptable to add some very basic stats to the public
`health_check` endpoint for this purpose.

We leverage the Sidekiq API to return the two statistics we want:

- the size of `enqueued`
- the `retry_size`

See:

https://github.com/mperham/sidekiq/wiki/API

and

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/background-jobs.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -934,6 +934,7 @@
 - Fix deployment notification
 
 ## [unreleased]
+- Health check endpoint includes basic sidekiq stats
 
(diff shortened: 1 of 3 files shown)


Stretch96 (approved):
lgtm 👍

edavey (approved):
neat solution!

PR #1492, opened 2021-11-23
Release 90

mec:
- Uploaded actual history must be in past financial quarters
 - Activity summary view uses BEIS approved attribute names
 - Channel of delivery code 90000, 'Other' added to the accepted codes list
 - Fix deployment notification


edavey (approved):
🚀

PR #1493, opened 2021-11-23
Use postgres 13 in backing-services-docker-compose.yml

Stretch96:
## Changes in this PR

* The Postgres RDSs have been updated to version 13, so this should be
  the same to ensure the local development environment is the same


edavey (approved):
👍

PR #1500, opened 2021-12-02
(2098) Refactor: setting an Activity's implementing orgs now uses "org participations"

edavey:
## Background

To correct earlier proliferation of duplicate implementing organisations we are migrating existing data out of `implementing_organisations` table into a new "join" table `org_participations` which associates Activities with Organisations and defines the type of participation using the `OrgParticipation#role` attribute. The range of organisation participations is defined by IATI as:

- **Funding**: The government or organisation which
  provides funds to the activity.

- **Accountable**: An organisation responsible for
  oversight of the activity and its outcomes

- **Extending**: An organisation that manages the budget
  and direction of an activity on behalf of the funding
  organisation

- **Implementing**: The organisation that physically carries out
  the activity or intervention.

At present all OrgParticipations are "implementing". We plan to do away with the `Activity#extending_organisation` attr as per the suggested refactoring of https://trello.com/c/kJG0qMT3/1647 (summarised as "Replace 'organisation' and 'extending_organisation' with a 'delivery_partner' attribute")

```
Activity#has_one
   :delivery_partner
   through: :extending_org_participation
```

## Changes in this PR

- **via the UI**: a DP now can add an arbitrary number of implementing organisations to an activity, via a (rather long) select list. Implementing organisations can be removed one-by-one from the Activity | Details page. If the required implementing organisation isn't in the list, the DP will need to ask BEIS to add it. (The facility for BEIS to manage implementing organisation via the UI will follow in a separate PR)

- **via CSV upload**: a DP can now supply the "Implementing organisation name" in the activity upload template. If the given name isn't found, then a helpful error will be reported against the row in question. At present only one implementing organisation can be supplied per activity in the upload CSV. (Handling of multiple implementing organisation names will follow in a separate PR)

### Before 

before_new_implementing_org

### AFTER: Add an implementing org
add_implementing_organisation

### AFTER: View and remove implementing orgs
view_and_remove_implementing_orgs

## Notes

Apologies in advance for the size of this PR. The test suite is not green at each step, as I've opted to group the changes for ease of reviewing.

## Next steps

- [x] guidance needs to be updated to reflect the fact that implementing organisations are no longer created on demand 


edavey:
[Change how we collect and store implementing organisation](https://trello.com/c/8teLzwsh/2098-change-how-we-collect-and-store-implementing-organisation)

LornaHarwood (approved):
LGTM! :shipit:

PR #1501, opened 2021-12-03
Feature (2325) New reports have a valid financial period.

LornaHarwood:
## Changes in this PR

- Ensure new reports have a financial period no later than the period of their creation
-  Add has_many reports association to Organisation
- Ensure new report does not rewrite history. A new report will be invalid if there is a pre-existing report for a later period.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -893,7 +893,7 @@
 
 ## [release-85] 2021-11-09
 
-- Fix "Total forecasted" (in Activity financial summary) to exclude periods 
+- Fix "Total forecasted" (in Activity financial summary) to exclude periods
   already reported
 - Fix spending breakdown report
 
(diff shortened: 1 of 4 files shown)


edavey (commented):
It occurred to me that we didn't write any tests to describe how the new validation manifests itself thru the UI. 

We do already have an integration test which shows how a report is created (`spec/features/staff/beis_users_can_create_a_report_spec.rb`). 

Perhaps we could add a new "unhappy path" scenario to this feature test where we say something like:

```
When I try to create a report for period earlier than the latest report
Then I see an error message telling me that a report already exists for a later period
And the financial quarter and year fields are highlighted as errors
```

What do you think?

LornaHarwood:
> It occurred to me that we didn't write any tests to describe how the new validation manifests itself thru the UI.
> 
> We do already have an integration test which shows how a report is created (`spec/features/staff/beis_users_can_create_a_report_spec.rb`).
> 
> Perhaps we could add a new "unhappy path" scenario to this feature test where we say something like:
> 
> ```
> When I try to create a report for period earlier than the latest report
> Then I see an error message telling me that a report already exists for a later period
> And the financial quarter and year fields are highlighted as errors
> ```
> 
> What do you think?

I feel this may not be necessary as we have enough coverage for this from the UI and within our unit tests.

edavey (approved):
Nice work, and on reflection I agree that the validation messages shown in the UI are adequate.

👍

PR #1506, opened 2021-12-07
Release 92

edavey:
- Refactor setting of implementing organisations to use
  unique organisations through new join table


edavey:
This will be followed by a 3rd PR which will include:

- a new feature to allow BEIS to create and edit implementing orgs (as DPs can no longer make them willy-nilly)
- dropping the old `implementing_organisations` table. The data was migrated earlier today after deploying Release 91.

LornaHarwood (approved):
👍  
Yup, looks look to me :shipit:

PR #1518, opened 2021-12-22
Replace ImplementingOrganisation with OrgParticipation in delete activity task

lozette:
https://trello.com/c/ElWiw4K1/26-beis-roda-delete-activity-rake-task-is-broken

ImplementingOrganisation has been replaced with the OrgParticipation model. This
was updated everywhere except for the delete activity task, and as a result
the task was not functioning correctly.

## Next steps


edavey (approved):
👍

PR #1522, opened 2022-01-04
Fix 3 spec failures due to change in year

edavey:
These specs failed once the year changed from 2021 to 2022.

We now use the Rails time-travelling test helper
(`ActiveSupport::Testing::TimeHelpers#travel_to`) to lock
our examples to mid-Dec 2021.


shuldt (approved):
Took me a second to figure out why the timezone was being passed in hardcoded but yup, that's fine

PR #1533, opened 2022-01-19
Upgrade standard and parser to avoid warning

rgarner:
## Changes in this PR

Upgrade from very old `0.1.8` `standard` to `1.5.0`. We were manually pinning `parser` to `~> 2.6.3.0` and that was generating the parser warning on the beginning of every test suite.

Avoid bumping Ruby to 2.7.5 right this second; this will make the job a bit easier when we come to do it. Acknowledge that pinning `parser` has the effect of potentially generating the parser warning whenever we upgrade Ruby (although Rubocop has since been patched to silence the warning so will no longer be a source), and we'll need to manually move it along in the Gemfile when we do so.

See https://github.com/rubocop/rubocop/issues/9700#issuecomment-821301333 for details

## Next steps


edavey (approved):
thanks for fixing this super-annoying issue and moving us forward to the latest "standard" syntax! 

It seems like these style issues (e.g. is a comma after the last item in a list a good thing?) can never be settled definitively.

rgarner:
@edavey you're welcome. I'm not sure it can be merged though. It doesn't look like the build works when the PR comes from a fork. Might have to re-open local to this repo tomorrow now I have full push access.

rgarner:
[Ruby parser](https://trello.com/c/DvrE8UEi/2292-ruby-parser)

rgarner:
Closing due to fork build issue in favour of #1537

PR #1534, opened 2022-01-19
Add IF EXISTS to DROP commands for databases/extensions

dragon-dxw:
If a database or extension does not exist, attempting to DROP it
fails and halts the restore process. Specifically, `postgis` did
not exist as an extension when installing a local development
environment for me.

```
==> Removing extraneous Postgres extensions...
NOTICE:  extension "postgis" does not exist, skipping
DROP EXTENSION
```

## Next steps

None.


edavey (approved):
Thanks for fixing this -- one less gotcha to remember to play past...

PR #1535, opened 2022-01-19
Add additional local setup steps to README

dragon-dxw:
There's probably a lot of discussion that can be had about exactly what should be documented where, but here is a first attempt at getting the challenges we faced today in getting a local environment set up documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,15 @@ Once setup has been completed, you can start the server with
 script/server
 ```
 
-## Running the tests
+and the tests should pass
 
 ```bash
 script/test
 ```
+
+### Creating a development database
+
+See "PaaS to Local" at the end of the (Database Backup and Restore) guide.
 
 ## Running backing services with Docker compose
 
@@ -119,5 +123,4 @@ This repository was bootstrapped from [dxw's `rails-template`](https://github.co
 
 ## Documentation
 
-[Detailed documentation](./doc/0_index_of_contents.md) for developers and support staff is found in `doc/`. 
-See [`doc/0_index_of_contents.md`](./doc/0_index_of_contents.md)
+[Detailed documentation](./doc/0_index_of_contents.md) for developers and support staff is found in `doc/`.
--- a/README.md
+++ b/README.md
@@ -54,12 +54,12 @@ We use ADRs to document architectural decisions that we make. They can be found 
 
 We use [Dotenv](https://github.com/bkeepers/dotenv) to manage our environment variables locally.
 
-The repository will include safe defaults for development in `/.env.example` and for test in `/.env.test`. We use 'example' instead of 'development' (from the Dotenv docs) to be consistent with current dxw conventions and to make it more explicit that these values are not to be committed.
+The repository will include safe defaults for development in `/.env.example` and for test in `/.env.test`. We use 'example' instead of 'development' (from the Dotenv docs) to be consistent with current dxw conventions and to make it more explicit that the real values are not to be committed.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -19,6 +19,24 @@ Once setup has been completed, you can start the server with
...
+* Install Cloud Foundry from [GPaaS](https://www.cloud.service.gov.uk/) after logging in.
+We needed to open it in Finder to be able to install it (a second button appears in grey)
+* `cf install-plugin conduit`
+* `cf login -a api.london.cloud.service.gov.uk -u your@account.name`
+* `script/db-restore`
+
+#### Tests
+
+* `RAILS_ENV=test bundle exec rake db:create` to create a test database
+* `bundle exec rake assets:precompile` (some tests failed without this step)
```
rgarner, line 38: This isn't something we should have to do in Rails dev (normally it's just for production – in dev because you're changing CSS and JS regularly, precompile is generally only something you do to get to the bottom of a problem on production). Might see if I can sit in on James's setup and see if it happens again and if we can work out why certain things aren't being fallen back on; it might be a problem with `environments/test.rb`, but also I suspect that we did our  `db-restore` and had it fail before it set the db `environment`, which broke the ability to create the test db via `db:test:prepare` (it would normally be automatic).

README.md
```diff
@@ -19,6 +19,24 @@ Once setup has been completed, you can start the server with
...
+We additionally needed to take the following steps to get a reasonable local install outside of Docker including
+a working database for running the server locally:
+
+#### Database
+
+* Install Cloud Foundry from [GPaaS](https://www.cloud.service.gov.uk/) after logging in.
+We needed to open it in Finder to be able to install it (a second button appears in grey)
+* `cf install-plugin conduit`
+* `cf login -a api.london.cloud.service.gov.uk -u your@account.name`
+* `script/db-restore`
```
rgarner, line 33: I think this is the purpose of `doc/database-backup-and-restore.md` – which is missing the login step that you're describing here. Maybe we can move the useful detail from this section over there and link to it from here?

rgarner:
Looks good – maybe we could also change the reference to `.env.development.local` to just `.env`? `.env.development.local` is not in the `.gitignore` so creating it would commit by default.

rgarner:
(the build failure looks transient I think; temporary Debian package server unavailability?)

README.md
dragon-dxw: Deleted this.

README.md
dragon-dxw: Rewrote this extensively, probably worth a second glance.

erbridge:
> .env.development.local is not in the .gitignore so creating it would commit by default.

FYI: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/.gitignore#L26-L27

README.md
```diff
@@ -19,12 +19,16 @@ Once setup has been completed, you can start the server with
...
-## Running the tests
+and the tests should pass
 
 ```bash
 script/test
 ```
 
+### Creating a development database
+
+See "PaaS to Local" at the end of the (Database Backup and Restore) guide.
```
rgarner, line 30: I think we can link this as [PaaS to Local](./doc/database-backup-and-restore.md#paas-to-local) `[PaaS to Local](./doc/database-backup-and-restore.md#paas-to-local)`. These links won't work from this comment because they're relative to the comment but they should work on commit.

rgarner (approved):
:100: much appreciated, thanks @dragon-dxw

PR #1536, opened 2022-01-19
Feature: move spending breakdown into async job

edavey:
### Background

The "BEIS" (admin) version of the "spending breakdown" export is a comprehensive output of all spending for all time periods for all organisations -- for a given fund. In recent history this export was considered "broken" because it timed out due to the volume of queries. By contrast, the "Delivery partner" version of the export (which only outputs details for the delivery partner's organisation) has been functional.

From Oct to Nov 2021 we undertook a major refactoring of "exports" (See https://trello.com/c/hbeQLje4/2255-export-improvements). At the end of this process we noted that the spending breakdown export was approx 35% faster. 

However, when the BEIS assurers attempted to run the export in December they experienced 504 (timeout) errors on production. On my (admittedly puny) development machine the exports take more than 10 mins to run so this is not surprising. See:

Zendesk: https://dxw.zendesk.com/agent/tickets/15546

As a first step in improving this area we've decided to hand off the export creation to a background job which is charged with running the report, writing to a CSV, putting the CSV in an S3 bucket and emailing the requester with a download link.

### AWS S3

Some info on using S3 with GPaaS:
 
-  S3 with GPaaS: https://docs.cloud.service.gov.uk/deploying_services/s3/#amazon-s3

- Connecting to a bucket: https://docs.cloud.service.gov.uk/deploying_services/s3/#connect-to-an-s3-bucket-from-your-app

We've initially implemented an `S3UploaderConfig` class to extract the necessary AWS config from Cloudfoundry's rather complex JSON formatted `VCAP_SERVICES` env var. However, we've since become aware of a `VcapParser` class which aleady exists in BEIS, but without S3 config/creds. And of a [RMI version](https://github.com/Crown-Commercial-Service/DataSubmissionServiceAPI/blob/352b545144653311c9b6a697b985ecbfccc3aa75/lib/vcap_parser.rb) which does include S3. 

### To do

A few things still to do:

- [x] create Terraform for S3 config as at present we only have a manually created test bucket in our staging environment. See [RMI for S3 on GPASS](https://github.com/Crown-Commercial-Service/DataSubmissionServiceAPI/blob/352b545144653311c9b6a697b985ecbfccc3aa75/terraform/terraform.tf) example. Credentials for the temp experimental bucket are in the RODA vault in 1Password in the "S3 for uploading exports (staging experiments)" secret.

#### To do later
- [ ] reconcile our `S3UploaderConfig` with the `VcapParser` pattern already in play on RODA and RMI. Consider whether there's code which ought be extracted out to a `gem`. We're going to this later as a learning exercise with a more junior team member.
- [ ] make sure temporary CSV file is always removed
- [ ] delete download from S3 -- e.g. after 7 days
- [ ] extract some copy from views and controllers to translation files.

### Exports page now reflects change in action offered

request_export_cta

### Download email sent to requester on success

Success

### Failure message sent to requester if there's a problem

failure_email

## Next steps

- [ ] complete "To do" list above

### DIFF
--- a/terraform/s3-export-download-bucket.tf
+++ b/terraform/s3-export-download-bucket.tf
@@ -1,0 +1,13 @@
+data "cloudfoundry_service" "aws-s3-bucket" {
+  name = "aws-s3-bucket"
+}
+
+resource "cloudfoundry_service_instance" "beis-roda-s3-export-download-bucket" {
+  name         = "beis-roda-${var.environment}-s3-export-download-bucket"
+  space        = cloudfoundry_space.space.id
+  service_plan = data.cloudfoundry_service.aws-s3-bucket.service_plans["default"]
+  json_params  = <<EOF
+  {"public_bucket":true}
+EOF
+}
+
--- a/terraform/worker.tf
+++ b/terraform/worker.tf
@@ -15,6 +15,12 @@ resource "cloudfoundry_app" "beis-roda-worker" {
   }
   service_binding { service_instance = cloudfoundry_service_instance.beis-roda-redis.id }
   service_binding { service_instance = cloudfoundry_service_instance.beis-roda-postgres.id }
+  service_binding {
+    service_instance = cloudfoundry_service_instance.beis-roda-s3-export-download-bucket.id
+    params = {
+      "permissions" = "read-write"
+    }
+  }
   service_binding { service_instance = cloudfoundry_user_provided_service.papertrail.id }
   environment = {
     "RAILS_LOG_TO_STDOUT"              = "true"
--- a/.env.example
+++ b/.env.example
@@ -21,3 +21,18 @@ NOTIFY_KEY=
 NOTIFY_WELCOME_EMAIL_TEMPLATE=b1282182-887e-4c8a-8a39-649c9215cd41
 NOTIFY_VIEW_TEMPLATE=b541df04-add8-458e-a7d3-2e156386e150
 NOTIFY_OTP_VERIFICATION_TEMPLATE=1a832f2e-b13b-47f0-b32a-9cd6672364d2
+
+VCAP_SERVICES='{
+   "aws-s3-bucket":[
+      {
+         "name":"s3-export-download-bucket",
+         "credentials":{
+            "bucket_name":"paas-s3-broker-prod-lon-id",
+            "aws_access_key_id":"KEY_ID",
+            "aws_secret_access_key":"SECRET",
+            "aws_region":"eu-west-2",
+            "deploy_env":""
+         }
+      }
+   ]
+}'
--- a/.env.test
+++ b/.env.test
@@ -13,3 +13,18 @@ REDIS_URL=redis://localhost:6379
 SECRET_KEY_BASE=abcdefghijklmnopqrstuvwxyz12345678
 NOTIFY_VIEW_TEMPLATE=b541df04-add8-458e-a7d3-2e156386e150
 NOTIFY_OTP_VERIFICATION_TEMPLATE=1a832f2e-b13b-47f0-b32a-9cd6672364d2
+
+VCAP_SERVICES='{
+   "aws-s3-bucket":[
+      {
+         "name":"s3-export-download-bucket",
+         "credentials":{
+            "bucket_name":"paas-s3-broker-prod-lon-id",
+            "aws_access_key_id":"KEY_ID",
+            "aws_secret_access_key":"SECRET",
+            "aws_region":"eu-west-2",
+            "deploy_env":""
+         }
+      }
+   ]
+}'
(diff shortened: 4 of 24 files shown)


terraform/s3-export-download-bucket.tf
```diff
@@ -0,0 +1,13 @@
+data "cloudfoundry_service" "s3_export_download_bucket" {
+  name = "s3-export-download-bucket"
+}
+
+resource "cloudfoundry_service_instance" "s3_export_download_bucket" {
+  name         = "beis-roda-${var.environment}-s3-export-download-bucket"
+  space        = cloudfoundry_space.space.id
+  service_plan = data.cloudfoundry_service.aws_s3_bucket.service_plans["default"]
```
rjw1, line 8: this needs to match what you called the data above.

terraform/s3-export-download-bucket.tf
```diff
@@ -0,0 +1,13 @@
+data "cloudfoundry_service" "s3_export_download_bucket" {
+  name = "s3-export-download-bucket"
```
rjw1, line 2: this proably needs to be the name of the service in the market place.

terraform/worker.tf
```diff
@@ -15,6 +15,12 @@ resource "cloudfoundry_app" "beis-roda-worker" {
   }
   service_binding { service_instance = cloudfoundry_service_instance.beis-roda-redis.id }
   service_binding { service_instance = cloudfoundry_service_instance.beis-roda-postgres.id }
+  service_binding {
+    service_instance = cloudfoundry_service_instance.beis-roda-s3-export-download-bucket.id
```
rjw1, line 19: this needs to match what you called  the service instance.

rjw1: if anthign thing the service instance needs to change name to match this.

terraform/s3-export-download-bucket.tf
rjw1: this is `aws-s3-bucket` for probably both name and the name of the resource

rjw1, line 2:
```suggestion
data "cloudfoundry_service" "aws-s3-bucket" {
  name = "aws-s3-bucket"
```

terraform/s3-export-download-bucket.tf
rjw1: this matches with the suggestion above.

terraform/s3-export-download-bucket.tf
```diff
@@ -0,0 +1,13 @@
...
+}
+
+resource "cloudfoundry_service_instance" "s3_export_download_bucket" {
```
rjw1, line 5:
```suggestion
resource "cloudfoundry_service_instance" "beis-roda-s3_export_download_bucket" {
```

terraform/worker.tf
rjw1: see suggestion above

terraform/worker.tf
```diff
@@ -15,6 +15,12 @@ resource "cloudfoundry_app" "beis-roda-worker" {
...
+    json_params      = <<EOF
+      {"permissions":"read-write"}
+    EOF
+  }
```
edavey, line 23: @rjw1 Thanks bob, I'm now past those naming issues.

I'm getting an error now trying to set permissions on this `service_binding`:

```
Error: Unsupported argument

  on worker.tf line 20, in resource "cloudfoundry_app" "beis-roda-worker":
  20:     json_params      = <<EOF

An argument named "json_params" is not expected here.
```
Do you have any ideas how I can include an instruction here equivalent to the `-c '{"permissions":"read-write"}'` in:

```
cf bind-service beis-roda-staging-worker \
  s3-export-download-bucket-staging \
  -c '{"permissions":"read-write"}'
```

rjw1: https://registry.terraform.io/providers/cloudfoundry-community/cloudfoundry/latest/docs/resources/app#params

shuldt (approved):
I'm satisfied that the structure of this does what we discussed, but I might not be the best person to give a detailed ruby code review here. Do let me know if you want to talk about how best to capture/scope the follow-ups here also

edavey:
> I'm satisfied that the structure of this does what we discussed, but I might not be the best person to give a detailed ruby code review here. Do let me know if you want to talk about how best to capture/scope the follow-ups here also

Thanks @shuldt If you can direct me to your Trello board for support tasks I could add these in?

The 4 "to do later" tasks break down into 1 (large) and 2-4 (minor):

1. reconcile our `S3UploaderConfig` with the `VcapParser` pattern already in play on RODA and RMI. Consider whether there's code which ought be extracted out to a gem. ➡️  We're going to this later as a learning exercise with a more junior team member.
2.  make sure temporary CSV file is always removed
3. delete download from S3 -- e.g. after 7 days
4. extract some copy from views and controllers to translation files.

- (1) is a mini-project for me to do over a period of time as a coaching project.
- (2-4) there may well be time to do these this week whilst we test/tweak the S3 Terraform config, assuming it doesn't work first time, that is!

PR #1538, opened 2022-01-20
Lint YAML files

dragon-dxw:
A number of duplicated keys in YAML files were discovered from running the `check-yaml` pre-commit hook (see pre-commit.com)

(further information: https://yaml.readthedocs.io/en/latest/api.html#duplicate-keys )

The following decisions were made whilst correcting them:

* Almost all duplicated keys with string values were identical duplicates; one was removed.
* The only exception is `State`/`Status` for the `state` key; State (the first key) was removed as there's evidence of an attempt to change from State to Status.
* All duplicated keys with hash values had no overlap in the keys of the hashes; the 'failure' subkeys were merged with the other subkeys from the duplicated keys. This *does* change the interpretation of the YAML file (the second, more substantial key with non-`failure` subkeys would have been hidden)

The impact of these changes is unknown but it seemed wise to bring them up for discussion.

### DIFF
--- a/config/locales/models/activity.en.yml
+++ b/config/locales/models/activity.en.yml
@@ -434,7 +434,6 @@ en:
         finance: Finance type
         aid_type: Aid type
         collaboration_type: Collaboration type
-        channel_of_delivery_code: Channel of delivery code
         policy_marker_gender: Gender - policy marker
         policy_marker_climate_change_adaptation: Climate Change - Adaptation policy marker
         policy_marker_climate_change_mitigation: Climate Change - Mitigation policy marker
@@ -449,15 +448,12 @@ en:
         originating_report_id: Originating report identifier
         source_fund_code: Fund code
         call_present: Call present?
-        transparency_identifier: IATI identifier
         parent_id: Parent activity identifier
         parent: Parent activity
         publish_to_iati: Include in IATI XML export?
-        sector_category: Sector category
         previous_identifier: Previous IATI identifier
         geography: Recipient geography
         extending_organisation_id: Extending organisation identifier
-        sector: Sector
         organisation_id: Organisation identifier
         tied_status: Tied status
         organisation: Organisation
--- a/config/locales/models/report.en.yml
+++ b/config/locales/models/report.en.yml
@@ -13,7 +13,6 @@ en:
     header:
       report:
         financial_quarter: Financial quarter
-        state: State
         description: Description
         organisation: Organisation
         fund: Fund (level A)
--- a/config/locales/models/report.en.yml
+++ b/config/locales/models/report.en.yml
@@ -186,14 +186,8 @@ en:
           title: Report activation complete
           body: "%{report_financial_quarter} %{report_description}"
         failure: Report could not be activated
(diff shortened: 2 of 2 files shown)


erbridge (commented):
This lint looks like a lot of deletions. Would you be able to split it up into separate commits and explain in the commit message why the changes you've made don't impact the service, please?

config/locales/models/activity.en.yml
```diff
@@ -434,7 +434,6 @@ en:
         finance: Finance type
         aid_type: Aid type
         collaboration_type: Collaboration type
-        channel_of_delivery_code: Channel of delivery code
```
dragon-dxw, line 437: `channel_of_delivery_code` also appears on line 432

config/locales/models/activity.en.yml
```diff
@@ -449,15 +448,12 @@ en:
...
         call_present: Call present?
-        transparency_identifier: IATI identifier
         parent_id: Parent activity identifier
         parent: Parent activity
         publish_to_iati: Include in IATI XML export?
-        sector_category: Sector category
         previous_identifier: Previous IATI identifier
         geography: Recipient geography
         extending_organisation_id: Extending organisation identifier
-        sector: Sector
```
dragon-dxw, line 460: L430

config/locales/models/activity.en.yml
```diff
@@ -449,15 +448,12 @@ en:
         originating_report_id: Originating report identifier
         source_fund_code: Fund code
         call_present: Call present?
-        transparency_identifier: IATI identifier
         parent_id: Parent activity identifier
         parent: Parent activity
         publish_to_iati: Include in IATI XML export?
-        sector_category: Sector category
```
dragon-dxw, line 456: L431

config/locales/models/activity.en.yml
```diff
@@ -449,15 +448,12 @@ en:
         originating_report_id: Originating report identifier
         source_fund_code: Fund code
         call_present: Call present?
-        transparency_identifier: IATI identifier
```
dragon-dxw, line 452: L394

config/locales/models/report.en.yml
```diff
@@ -18,7 +18,6 @@ en:
         organisation: Organisation
         fund: Fund (level A)
         deadline: Deadline
-        state: Status
```
dragon-dxw, line 21: L16; note the textual change

dragon-dxw:
@erbridge Refactored that into three commits: straight duplication of text; duplication with a textual change (I've changed my mind on which one we preserve based on the git blame) and merging hashes with non-overlapping subkeys.

config/locales/models/report.en.yml
```diff
@@ -13,7 +13,6 @@ en:
     header:
       report:
         financial_quarter: Financial quarter
-        state: State
```
erbridge, line 16: I wouldn't necessarily infer the change was intentional, especially if it wouldn't have worked, btw.

CristinaRO (approved):

PR #1542, opened 2022-01-24
Mini Racer 0.6.2 and other dependencies.

dragon-dxw:
Needs tidying, but got mini_racer 0.6.2 to work.


dragon-dxw:
@rgarner Would you be happy if I just completely flatten these commits and documented it all in the git commit?

rgarner:
[Why do the AWS credentials in terraform fail randomly in CI?](https://trello.com/c/v94L9ZQZ/2146-why-do-the-aws-credentials-in-terraform-fail-randomly-in-ci)

rgarner:
@dragon-dxw definitely, this is easily small enough.

dragon-dxw:
"ruby" is probably a "just pure ruby" platform that can't compile C code; "darwin" is suitable for running on MacOS natively, and it runs on "linux" inside Docker.

CristinaRO (approved):
This looks good, and such excellent sleuthing work!

For my understanding, why did we need to add back `x86_64-linux` but not the `-darwin` variant?

rgarner:
@CristinaRO from what I can tell it's that the `PLATFORM` section in `Gemfile.lock` gives hints to `make`/build processes for gems using native (e.g. C/C++) code. `x86_64-linux` is needed as such a hint for `mini_racer` to compile/link its version of `musl` (a libc variant).  When running locally on Mac, XCode command line tools provides the hints for you. You *can* add `x86_64-darwin18` variants, but you may need a lot of them to cover every developer's situation.

PR #1543, opened 2022-01-24
Upgrade bundler & rubygems for terraform/dependabot issue

rgarner:
We have a dependabot/terraform PR problem: all the dependabot-triggered
builds fail. A propos of

rubyjs/mini_racer#220 (comment)

...update the rubygems and bundler that Docker (and hence CI) uses.

Try this on its own first due to rubygems [patch](https://github.com/rubygems/rubygems/pull/4082) (check build runs, merge, maybe see new dependabot PR builds work?), then try

`bundle lock --add-platform x86_64-linux` as per the comment in the link above if still no joy

## Next steps


rgarner:
Closing in favour of #1542

PR #1545, opened 2022-01-25
Add note about Implementing Organisation to fill a hole in the glossary

dragon-dxw:


edavey (approved):
Thanks for this!

CristinaRO (approved):
Thanks for tending to the glossary!

PR #1549, opened 2022-01-25
Strip control characters from some Activity fields

dragon-dxw:
For fields which are of type 'string' or 'text' (and not an array
of those), before we validate we delete almost all control characters to
avoid XML documents we generate being malformed due to containing
unescaped control characters.

(We don't remove tab, carriage return or line-feed as they're
commonly found within normal text, and are permitted as-is within XML
documents. https://www.w3.org/International/questions/qa-controls#support )

## Next steps

- ❌  Do any environment variables need amending or adding?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -960,6 +960,7 @@
 ## [unreleased]
 
 - Moved deactivated users to the bottom of the users list
+- Remove control characters from input before validation
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-94...HEAD
 [release-94]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-93...release-94
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -67,6 +67,8 @@ class Activity < ApplicationRecord
 
   FORM_STATE_VALIDATION_LIST = FORM_STEPS.map(&:to_s).push("complete")
 
+  before_validation :strip_control_characters_from_fields!
+
   strip_attributes only: [:delivery_partner_identifier]
 
   validates :level, presence: true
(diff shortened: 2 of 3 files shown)


CristinaRO:
[IATI XML export issues Met Office level D and Innovate GCRF level D](https://trello.com/c/8t9fTxiv/2326-iati-xml-export-issues-met-office-level-d-and-innovate-gcrf-level-d)

rgarner:
[IATI XML export issues Met Office level D and Innovate GCRF level D](https://trello.com/c/8t9fTxiv/2326-iati-xml-export-issues-met-office-level-d-and-innovate-gcrf-level-d)

rgarner:
Hi – not reviewing this as it's not marked ready for it but two related questions:

1. will we need a [data migration](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/tree/develop/db/data) to retrospectively clean up the control characters such that IATI XML can be exported?
2. will such a data migration leave some records un-cleaned because of unrelated validation errors (I recall that we had at least one older record failing to save because newer validations had made it invalid)?

dragon-dxw:
>     1. will we need a [data migration](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/tree/develop/db/data) to retrospectively clean up the control characters such that IATI XML can be exported?

Running `Activity.find_each {|a| a.touch; a.validate; a.save} ` should be sufficient; that's probably convieniently wrapped up in a migration.

>     2. will such a data migration leave some records un-cleaned because of unrelated validation errors (I recall that we had at least one older record failing to save because newer validations had made it invalid)?

That is a risk; the known affected record doesn't but it looked like this problem might have occured in a handful of other places judging from update messages scrolling by in the logs.

dragon-dxw:
Open questions:
* should we apply this more broadly throughout the code
* should we make a report of all the data that get changed by this?
* are there fields we shouldn't touch even if they do contain control codes (e.g. identifiers)
* should we save touched files even if they don't validate?
* should we warn in the UI if we do mutate an entry like this?

rgarner (approved):

dragon-dxw:
Closed in favour of #1556 (which includes this plus a database migration)

PR #1550, opened 2022-01-26
Move deactivated users to the bottom of the list of users

dragon-dxw:
`includes` doesn't like it if you try to sort on the users table first, so we `joins` as well.
`joins` on its own raises an error saying to add includes:
```
     Bullet::Notification::UnoptimizedQueryError:
       user: dragon
       GET /users
       USE eager loading detected
         User => [:organisation]
         Add to your query: .includes([:organisation])
```

### Before
Mikey Middleton would appear in the middle.

### After
image

## Next steps

- ❌ Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- ❌ Do any environment variables need amending or adding?
- ❌ Have any changes to the XML been checked with the IATI validator? See [XML Validation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/xml-validation.md)
- [ ] Update the "how to use this website" guidance for Sam

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -959,6 +959,8 @@
 
 ## [unreleased]
 
+- Moved deactivated users to the bottom of the users list
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-94...HEAD
 [release-94]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-93...release-94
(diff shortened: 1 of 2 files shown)


CristinaRO:
[Order users by active/not active [2]](https://trello.com/c/MmloZfCL/2345-order-users-by-active-not-active-2)

CristinaRO (approved):
Have you considered adding an automated test for this new functionality? (I'm not implying you should, I just wanted to know if that's been considered and decided against.)

PR #1553, opened 2022-01-26
(2348) Refactor policy specs to minimise setup costs

CristinaRO:
## Changes in this PR
- Refactor policy specs to decrease the setup costs of individual examples

On my machine, the comparative gains when running all policy specs are as such:
Initial:
```
Finished in 1 minute 48.85 seconds (files took 10.44 seconds to load)
446 examples, 0 failures
```

After refactoring:
```
Finished in 1 minute 6.61 seconds (files took 4.49 seconds to load)
263 examples, 0 failures
```

## Next steps

### DIFF
--- a/spec/policies/actual_policy_spec.rb
+++ b/spec/policies/actual_policy_spec.rb
@@ -11,44 +11,52 @@ RSpec.describe ActualPolicy do
     context "when the activity is a fund" do
       let(:activity) { create(:fund_activity, organisation: user.organisation) }
 
-      it { is_expected.to permit_action(:show) }
-      it { is_expected.to forbid_action(:create) }
-      it { is_expected.to forbid_action(:edit) }
-      it { is_expected.to forbid_action(:update) }
+      it "only permits show" do
+        is_expected.to permit_action(:show)
+        is_expected.to forbid_action(:create)
+        is_expected.to forbid_action(:edit)
+        is_expected.to forbid_action(:update)
 
-      it { is_expected.to forbid_action(:destroy) }
+        is_expected.to forbid_action(:destroy)
+      end
     end
 
     context "when the activity is a programme" do
       let(:activity) { create(:programme_activity, organisation: user.organisation) }
 
-      it { is_expected.to permit_action(:show) }
-      it { is_expected.to permit_action(:create) }
-      it { is_expected.to permit_action(:edit) }
-      it { is_expected.to permit_action(:update) }
-      it { is_expected.to permit_action(:destroy) }
+      it "permits all actions" do
+        is_expected.to permit_action(:show)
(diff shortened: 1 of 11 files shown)


CristinaRO:
[Refactor policy specs to avoid multiple setup costs](https://trello.com/c/e9sSeJYF/2348-refactor-policy-specs-to-avoid-multiple-setup-costs)

rgarner:
40-odd seconds is not to be sniffed at – thank you!

spec/policies/actual_policy_spec.rb
```diff
@@ -11,44 +11,52 @@
     context "when the activity is a fund" do
       let(:activity) { create(:fund_activity, organisation: user.organisation) }
 
-      it { is_expected.to permit_action(:show) }
-      it { is_expected.to forbid_action(:create) }
-      it { is_expected.to forbid_action(:edit) }
-      it { is_expected.to forbid_action(:update) }
+      it "controls actions as expected" do
```
rgarner, line 14: It occurs to me (much later, after writing the card – my apologies) that we can improve the readability of these specs (and consequently the `rspec --format=doc` output we see in CI) if we're able to be more specific about what happens. In this case it would be `it "allows show and forbids everything else"`. Elsewhere is could be `it "allows everything"`. I don't know if people feel it's worth the effort; it might not be, but the CI output reads:

## Before this PR
```
when the activity is a fund
  is expected to permit action 'show'
  is expected to forbid action 'create'
 ... etc
```
## After this PR
```
when the activity is a fund
  controls actions as expected
```

I'm on the fence as to whether it's worth more effort! If we feel it's useful and we read the output in CI, then by all means upgrade the strings. But it's probably the case that we've already had the benefit in performance and these will be read tremendously infrequently.

rgarner (approved):

spec/policies/actual_policy_spec.rb
CristinaRO: I did think about this from the first time this optimisation was introduced! It makes it harder to spot exactly what broke when a test starts failing, but writing the full details gets awkward in cases where it's more of a fifty-fifty split between "allow"s and "don't allow"s, such as https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1553/files#diff-349429e14f9f4750a25e95f05f1b286909c8d815ee6203637854ecc608acc0a7R25.

Is it worth doing this only for those clusters where it's easier to say "permits everything" or "forbids everything" (with maybe one, max two exceptions)?

rgarner: I leave it to your discretion. Personally I'm not getting too much value from that part of the CI output, and if it fails it will fail at the right line so will give us the right info.

PR #1556, opened 2022-01-27
Remove control characters from Activities

dragon-dxw:
For fields which are of type 'string' or 'text' (and not an array
of those), before we validate we delete almost all control characters to
avoid XML documents we generate being malformed due to containing
unescaped control characters.

(We don't remove tab, carriage return or line-feed as they're
commonly found within normal text, and are permitted as-is within XML
documents. https://www.w3.org/International/questions/qa-controls#support )

## Next steps

- ❌ Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- ❌  Do any environment variables need amending or adding?

(The XML doesn't pass, due to a schema validation error, which we'll look into next)

- [ ] Perform database migration on live by running `
❯ bundle exec rails runner db/data/20220127173100_correct_control_code_activity.rb`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -960,6 +960,7 @@
 ## [unreleased]
 
 - Moved deactivated users to the bottom of the users list
+- Remove control characters from input before validation
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-94...HEAD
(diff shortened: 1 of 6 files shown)


dragon-dxw:
Additional context from an older incarnation of this PR: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1549

rgarner (approved):

PR #1557, opened 2022-01-31
Release 95

dragon-dxw:
- Moved deactivated users to the bottom of the users list
- Added pre-validation hook for Activities, to prevent malformed XML being generated


rgarner (approved):
:100: :shipit: 🥳  🎈

PR #1559, opened 2022-01-31
(2360) Include date validation step when importing

CristinaRO:
## Changes in this PR
- Ensure that activities validate the presence of `planned_start_date` (or `actual_start_date`) and `planned_end_date` (or `actual_end_date`)

Previously, imported activities were validated in a context that was missing the `:dates_step` validation step.

This has allowed a few activities to be imported without any of `planned_start_date`, `actual_start_date`, `planned_end_date`, `actual_end_date`. These activities go on to generate an invalid IATI XML, which is missing the `iso-date` attribute from the `activity-date` field.

This fix ensures that both imported activities and activities created via the UI form will be validated against the `:dates_step`.

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -47,6 +47,7 @@ class Activity < ApplicationRecord
     :sector_step,
     :call_present_step,
     :call_dates_step,
+    :dates_step,
     :total_applications_and_awards_step,
     :programme_status_step,
     :country_delivery_partners_step,
--- a/spec/services/activities/import_from_csv_spec.rb
+++ b/spec/services/activities/import_from_csv_spec.rb
@@ -712,6 +712,23 @@ RSpec.describe Activities::ImportFromCsv do
       end
     end
 
+    it "has an error if it fails dates_step validation" do
+      new_activity_attributes["Planned start date"] = nil
+      new_activity_attributes["Actual start date"] = nil
+
+      expect { subject.import([new_activity_attributes]) }.to_not change { Activity.count }
+
+      expect(subject.created.count).to eq(0)
+      expect(subject.updated.count).to eq(0)
+
+      expect(subject.errors.count).to eq(2)
+      expect(subject.errors.first.csv_row).to eq(2)
+      expect(subject.errors.first.csv_column).to eq("Planned start date")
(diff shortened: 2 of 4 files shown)


CristinaRO:
[Failing IATI validation](https://trello.com/c/xkPMh7ko/2360-failing-iati-validation)

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -712,6 +712,23 @@
       end
     end
 
+    it "has an error if both planned_start_date and actual_start_date are missing" do
```
dragon-dxw, line 715: Do we want to test fror end_dates?

app/models/activity.rb
```diff
@@ -47,6 +47,7 @@ class Activity < ApplicationRecord
     :sector_step,
     :call_present_step,
     :call_dates_step,
+    :dates_step,
```
dragon-dxw, line 50: I can't find where this lives!

spec/services/activities/import_from_csv_spec.rb
CristinaRO: Hm, I wondered that, but decided against it because what I'd really want to test is that the validation includes the `dates_step`, the meaning of which is tested elsewhere (in `spec/models/activity_spec.rb`, contexts `"when planned start and actual start dates are blank"`, `"when planned_start_date is blank but actual_start_date is not nil"`, etc.)

app/models/activity.rb
CristinaRO: Not sure what you mean - if you mean looking for a definition like `def dates_step`, there isn't one. This is not a method, it's a symbol that gets its meaning from being used in the model validations:
```
validates :planned_start_date, presence: {message: I18n.t("activerecord.errors.models.activity.attributes.dates")}, on: :dates_step, unless: proc { |a| a.actual_start_date.present? }
```
and then being checked against in the validation context:
```
activity.valid?(:dates_step)
```

spec/services/activities/import_from_csv_spec.rb
dragon-dxw: Cool! I feel like the name reflecting that decision would make a lot of sense.

dragon-dxw (approved):

PR #1566, opened 2022-02-02
🔑 : 2361 : Login with local devise

dragon-dxw:
## Changes in this PR

* Remove Auth0 and Omniauth gems, and everywhere that explicitly requires them
* Install Devise and add to User model
* Be able to log in with a console-created Username/Password

https://trello.com/c/gLQcXB94/2361-login-with-local-devise

## What aren't we doing and what is intentionally broken?

* 2FA is out of scope for this branch
* We have broken inviting/adding/editing users in the respect of leaving some Auth0 calls in (albeit commented out for context). We made the specs involving these `pending` specs.

### Removals 
We aren't removing everything in relation to Auth0, because that would make this PR hard to review. These things include but are not limited to:

* env vars for dev
* GitHub CI env vars
* production env vars in Terraform
* The declarations that the Auth0 env vars are required for the app to boot
* the `User#identifier` column, which appears to be in use only to match `User` records to `auth0`. This is a critical element should we wish to roll back temporarily for any reason so we should do this removal much later.

## TODO

- [ ] ~~Add an ADR about how we're accepting red commits on this long-running integration branch.~~ << we probably won't do this. We're only accepting RED commits on *this* branch – by the time they get to `devise-integration` they'll be green.


dragon-dxw:
[Login with local devise](https://trello.com/c/gLQcXB94/2361-login-with-local-devise)

app/controllers/concerns/secured.rb
```diff
@@ -2,23 +2,11 @@ module Secured
   extend ActiveSupport::Concern
 
   included do
-    before_action :logged_in_using_omniauth?
+    before_action :redirect_unauthenticated
```
james, line 5: Is there a reason we don't just use [`authenticate_user!` from devise](https://github.com/heartcombo/devise/blob/8593801130f2df94a50863b5db535c272b00efe1/lib/devise/controllers/helpers.rb#L17) here?

app/models/user.rb
```diff
@@ -1,4 +1,8 @@
 class User < ApplicationRecord
+  # Include default devise modules. Others available are:
+  # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
+  devise :database_authenticatable, :registerable,
```
james, line 4: I don't think we want `registerable`, as only admins will be creating user accounts, no self service here.

app/models/user.rb
```diff
@@ -1,4 +1,8 @@
...
+    :recoverable, :rememberable, :validatable
```
james, line 5: I may be splitting hairs a bit here seeing as I know this will be the next ticket, but I wouldn't normally want to add the recoverable module in this commit/PR, I'd want it in the chunk of work we're doing that section of work

app/views/shared/_navigation.html.haml
```diff
@@ -28,6 +28,4 @@
           = link_to t("page_title.users.index"), users_path, class: "govuk-header__link"
 
       %li{ class: navigation_item_class(sign_out_path) }
-        %a{ href: sign_out_path, class: 'govuk-header__link' }
-          = t("header.link.sign_out")
-
+        = link_to t("header.link.sign_out"), destroy_user_session_path, method: :delete, class: "govuk-header__link"
```
james, line 31: I'm suspicious of `link_to ... method: :delete` - I may be out of date here, but an `a` tag should be an immutable get request, and I think it can cause unexpected issues if this assumption isn't kept.

config/routes.rb
```diff
@@ -120,8 +121,6 @@
   get "/pages/*id" => "pages#show", :as => :page, :format => false
 
   # Authentication
-  get "auth/oauth2/callback" => "auth0#callback"
-  get "auth/failure" => "auth0#failure"
   get "sign_out" => "application#sign_out"
```
james, line 124: I think we can delete this route too?

spec/features/staff/users_can_sign_out_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.feature "Users can sign out" do
+  scenario "success" do
+    ClimateControl.modify BULLET_DEBUG: "false" do
```
james, line 5: Out of interest, what's this for?

app/controllers/concerns/secured.rb
rgarner: At present our behaviour is "if you're not logged in (via any controller `include`ing the `Secured` module) we redirect you to `/` (effectively an interstitial with a "sign in" button) as opposed to `sessions#new` and squirrel away the place you were going to in `session[:redirect_path]` before redirecting you to that place via [`Auth#after_sign_in_path_for(_user)`](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/dde6e35cd4abdc365121120488faf7b094819b54/app/controllers/concerns/auth.rb#L29-L31)

By contrast `authenticate_user!` raises an uncaught Warden error and doesn't appear to have any redirect semantics. Is there a way of being more "idiomatic Devise" than the above while preserving the redirect behaviour?

app/models/user.rb
rgarner: Noted with thanks, will remove.

app/models/user.rb
rgarner: I don't think this is splitting hairs – it's entirely reasonable, as we want the meaning in the next piece of work's commits "we had to add the `:recoverable` module to achieve this" to be present there, not here.

app/views/shared/_navigation.html.haml
rgarner: I think from a screen reader semantics point of view, yes. Originally @CristinaRO and I had changed this to a button, which then left us with a styling problem!

config/routes.rb
rgarner: Will do, thanks

rgarner: @CristinaRO note: this is in use in [`_navigation.html.haml`](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/ebaea76e659dadcfe298f9f9df36ec9e194d9667/app/views/shared/_navigation.html.haml#L30) on the `li`; we'll need to change this to the `path` on the link below it.

app/controllers/concerns/secured.rb
rgarner: (I suspect understanding what the idiomatic version looks like is key to understanding how we make our locked users work as `active_for_authentication?` is used by the [`activatable` hook](https://github.com/heartcombo/devise/blob/8593801130f2df94a50863b5db535c272b00efe1/lib/devise/hooks/activatable.rb#L7-L11)

spec/features/staff/users_can_sign_out_spec.rb
rgarner: All I can tell you is it turns off the Bullet gem's debug output, which as far as I can tell we aren't turning on explicitly. If I remove it, the spec still passes. @CristinaRO any clues?

rgarner: Ah ok. I think we can lose it then.

rgarner: [This commit](a5a016a2567a0ec280b5a2e48dd8a3798cd983df) has a bit more detail but I think it applies only to the `AUTH0` vars.

app/models/user.rb
rgarner: Removed

james (approved):
Thanks for addressing my comments. Looks great to me. Major chunk of work!

config/initializers/devise.rb
```diff
@@ -0,0 +1,311 @@
...
+  # If true, extends the user's remember period when remembered via cookie.
+  # config.extend_remember_period = false
+
+  # Options to be passed to the created cookie. For instance, you can set
+  # secure: true in order to force SSL only cookies.
+  # config.rememberable_options = {}
+
+  # ==> Configuration for :validatable
+  # Range for password length.
+  config.password_length = 6..128
```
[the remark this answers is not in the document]
rgarner: Great spot, thanks. Have opened https://trello.com/c/3ED5g72M/2386-update-password-policy-to-match-auth0-requirements

PR #1570, opened 2022-02-09
(2381) Style the login page

CristinaRO:
## Changes in this PR
- Style the login page to be consistent with the rest of the site

### After
Screenshot 2022-02-09 at 12 37 05

PR #1571, opened 2022-02-10
(2379) Resetting user password

CristinaRO:
## Changes in this PR
- Allow users to reset their password
- Use GOV.UK Notify mailer

### After
Screenshot 2022-02-10 at 15 31 19
Screenshot 2022-02-10 at 15 31 10

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -13,6 +13,14 @@ class UserMailer < ApplicationMailer
     )
   end
 
+  def reset_password_instructions(user, token, opts = {})
+    @token = token
+
+    view_mail(ENV["NOTIFY_VIEW_TEMPLATE"],
+      to: user.email,
+      subject: t("devise.mailer.reset_password_instructions.subject"))
+  end
+
   def password_change_link(user:)
     Auth0Api.new.client.post_password_change(
       user_id: user.identifier,
--- a/app/views/devise/passwords/new.html.haml
+++ b/app/views/devise/passwords/new.html.haml
@@ -1,0 +1,13 @@
+= content_for :page_title_prefix, "Reset password"
+
+%h1.govuk-heading-l Forgot your password?
+
+.govuk-grid-row
+  .govuk-grid-column-one-third
+    = form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :post }) do |f|
+      = render "devise/shared/error_messages", resource: resource
+
+      = f.govuk_email_field :email, autofocus: true, autocomplete: "email"
+
+      .actions
+        = f.govuk_submit "Submit"
--- a/app/views/devise/sessions/new.html.haml
+++ b/app/views/devise/sessions/new.html.haml
@@ -17,3 +17,5 @@
               = f.govuk_check_box :remember_me, true, multiple: false, label: { text: "Remember me" }
       .actions
         = f.govuk_submit "Log in"
+
+    = link_to "Forgot password?", new_user_password_path
--- a/app/views/user_mailer/reset_password_instructions.text.erb
+++ b/app/views/user_mailer/reset_password_instructions.text.erb
@@ -1,0 +1,8 @@
+Hello,
+
+Someone (hopefully you) has requested a link to change your password. You can do this through the link below.
+
+<%= edit_user_password_url(reset_password_token: @token) %>
+
+If you didn't request this, please ignore this email.
+Your password won't change until you access the link above and create a new one.
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -24,10 +24,10 @@ Devise.setup do |config|
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = "please-change-me-at-config-initializers-devise@example.com"
+  # config.mailer_sender = "please-change-me-at-config-initializers-devise@example.com"
 
   # Configure the class responsible to send e-mails.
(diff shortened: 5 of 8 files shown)


CristinaRO:
[Resetting user password [5]](https://trello.com/c/HOyeoeRY/2379-resetting-user-password-5)

config/initializers/devise.rb
```diff
@@ -27,7 +27,7 @@
   config.mailer_sender = "please-change-me-at-config-initializers-devise@example.com"
```
rgarner, line 27: Is the notify mailer sender address configured in our Notify tenant? Now we're using Notify, do we have to change this or is it effectively unused/removable/renamable as `"i-am-irrelevant-please-ignore-me@example.com"`?

app/views/user_mailer/reset_password_instructions.text.erb
```diff
@@ -0,0 +1,8 @@
+Hello,
+
+Someone (hopefully you) has requested a link to change your password. You can do this through the link below.
+
+<%= edit_user_password_url(reset_password_token: @token) %>
+
+If you didn't request this, please ignore this email.
+Your password won't change until you access the link above and create a new one.
```
rgarner, line 8: Is this the default content from Devise? And I can't find any HAML/ERB references in the ADRs – is it just an ambient property of this repo that we:

- Use HAML for HTML and XML
- Use ERB for mailers?

If so should we make that explicit in an ADR separate from this? I'm sure there's Reasoning for it.

app/views/devise/passwords/new.html.haml
```diff
@@ -0,0 +1,13 @@
+= content_for :page_title_prefix, "Reset password"
```
rgarner, line 1: I notice elsewhere we use i18n for page titles. Should we be using one here (and presumably retrospectively for Log In?) as per [ADR 0011](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/2f1b3e2da49768074411df063d9a698d40a588e9/doc/architecture/decisions/0011-use-govuk-design-system-form-builder.md#L34-L35)? (apologies, did not notice the i18n stuff until today, and now having poked about, am more confused not less!)

app/views/devise/sessions/new.html.haml
```diff
@@ -17,3 +17,5 @@
               = f.govuk_check_box :remember_me, true, multiple: false, label: { text: "Remember me" }
       .actions
         = f.govuk_submit "Log in"
+
+    = link_to "Forgot password?", new_user_password_path
```
rgarner, line 21: If I do `rails g devise:views`, the `sessions#new` ERB comes with the following line:

`<%= render "devise/shared/links" %>`

I guess we've decided to drop it because this is more flexible when we have `rememberable` [turned on](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1570/commits/63d5c9db8a83f1b2ba8e43b90f099def3b3ab305#diff-5fe3b11f41a498ce1b5888e455dce9420646cfe49a8e100a39d31d31eb70e041R15)?

config/initializers/devise.rb
rgarner: Ah yeah, found the reply-to in Notify's settings. Maybe we should remove this.

rgarner:
Ran this in development; it looked great

app/views/devise/passwords/new.html.haml
```diff
@@ -0,0 +1,13 @@
...
+
+.govuk-grid-row
+  .govuk-grid-column-one-third
+    = form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :post }) do |f|
+      = render "devise/shared/error_messages", resource: resource
+
+      = f.govuk_email_field :email, autofocus: true, autocomplete: "email"
+
+      .actions
+        = f.govuk_submit "Submit"
```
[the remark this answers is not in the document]
rgarner: (looking elsewhere, I think the answer might be "no")

app/mailers/user_mailer.rb
```diff
@@ -13,6 +13,14 @@ def welcome(user)
     )
   end
 
+  def reset_password_instructions(user, token, opts = {})
+    @token = token
+
+    view_mail(ENV["NOTIFY_VIEW_TEMPLATE"],
+      to: user.email,
+      subject: "Reset password instructions")
```
rgarner, line 21:
```suggestion
      subject: t("devise.mailer.reset_password_instructions.subject")
```

app/views/devise/sessions/new.html.haml
CristinaRO: We decided that the links weren't serving us for now, so the simpler the better.

app/views/user_mailer/reset_password_instructions.text.erb
CristinaRO: More or less - I wanted to add the "hopefully you" parenthetical, because I like it and it sounds marginally less alarming than the default "Someone".

If there is reasoning, it might be lost to time. I don't know if we want to add a retroactive ADR, but whether or not we do doesn't seem to be tied to this feature branch (since the mixture is already there).

rgarner: I *thought* the wording was better than the default! (I also like your add). Do we think it likely that they'll want to edit this content themselves and hence might want to provision a Notify template?

CristinaRO: Let's ask them!

rgarner: Ok. Let's not let it block a merge; if it's important, it'll come back.

PR #1578, opened 2022-02-14
(2377) Admin invites a user

rgarner:
## Changes in this PR
- Admins can invite users
- Users receive the welcome email with a link to reset their password

### DIFF
--- a/spec/services/create_user_spec.rb
+++ b/spec/services/create_user_spec.rb
@@ -1,80 +1,36 @@
 require "rails_helper"
 
-RSpec.describe CreateUser, pending: "To be finished as part of the user invite card" do
-  let(:user) { build(:administrator) }
-  before(:each) do
-    stub_auth0_token_request
-  end
+RSpec.describe CreateUser do
+  let(:user) { build(:administrator, :new_user) }
 
   describe "#call" do
     it "returns a successful result" do
-      stub_auth0_create_user_request(email: user.email)
-
-      result = described_class.new(user: user, organisation: build_stubbed(:delivery_partner_organisation)).call
+      result = CreateUser.new(user: user, organisation: build_stubbed(:delivery_partner_organisation)).call
 
       expect(result.success?).to eq(true)
-      expect(result.failure?).to eq(false)
     end
 
     it "sends a welcome email to the user" do
-      stub_auth0_create_user_request(email: user.email)
-
       expect {
-        described_class.new(user: user, organisation: build_stubbed(:delivery_partner_organisation)).call
+        CreateUser.new(user: user, organisation: build_stubbed(:delivery_partner_organisation)).call
       }.to have_enqueued_mail(UserMailer, :welcome).with(user)
+    end
+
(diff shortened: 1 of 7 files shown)


CristinaRO:
[Admin invites a user [5]](https://trello.com/c/JQXINuhU/2377-admin-invites-a-user-5)

spec/services/create_user_spec.rb
```diff
@@ -1,81 +1,37 @@
...
-
       expect {
-        described_class.new(user: user, organisation: build_stubbed(:delivery_partner_organisation)).call
+        CreateUser.new(user: user, organisation: build_stubbed(:delivery_partner_organisation)).call
       }.to have_enqueued_mail(UserMailer, :welcome).with(user)
     end
 
+    it "creates a default password, known to no-one" do
+      CreateUser.new(user: user, organisation: build_stubbed(:delivery_partner_organisation)).call
+      expect(User.last.encrypted_password).not_to be_blank
```
dragon-dxw, line 21: Feels mildly wierd to depend on the new user being the last one -- seems fine, though.

dragon-dxw (approved):
Might want to wait for someone more qualified to review, but lgtm.

spec/services/create_user_spec.rb
CristinaRO: It felt weird to me too! We can rely on the testing env being artificial conditions though, where we know that's the only email that's going to be sent.

Gemfile.lock
```diff
@@ -180,7 +180,7 @@ GEM
     jwt (2.3.0)
     launchy (2.5.0)
       addressable (~> 2.7)
-    libv8-node (16.10.0.0)
+    libv8-node (16.10.0.0-x86_64-darwin)
```
james, line 183: What's going on with this commit? Something feels like its gone wrong?

CristinaRO: Fixed it!

rgarner: @CristinaRO @james worth noting that this *line* itself is not an error; it's pretty helpful. If we lose the `-x76_64-darwin` line, `bundle install` has to invoke the compiler for `libv8-node` instead of picking a pre-compiled platform-specific version for development.

Consequently, my machine has been compiling it for the last half-hour after pulling `develop`. I can't tell if @james was referring to the `RED` commit that this was in somehow having found its way onto this branch after rebase (because Github has no good way of commenting on a commit without referring to some file content) and @CristinaRO misunderstood and fixed the line @james was forced to refer to!

rgarner: We fixed this by finding the commit where we lost this line 183 on `devise-authentication`, rebasing interactively and editing that commit and running `git checkout HEAD -- Gemfile.lock && bundle install`.

This was caused by `rm Gemfile.lock && bundle install` recreating the Gemfile.lock without platform information. Bundler itself suggests the solution if you try and `bundle install` while a `Gemfile.lock` contains conflicts – it indicates that you should check out the last commit's version of Gemfile.lock and `bundle install` to resolve the conflict.

PR #1580, opened 2022-02-15
Validate XML against IATI schema on export

dragon-dxw:
** NOT TO BE MERGED **

When you download an organisation XML export, we automatically validate the XML
against the IATI schema, and fail hard if we fail the validation.

The failing URL and the errors that we caught are available as an uncaught
exception in Rollbar etc.

You can get to the failing XML by appending &skip_validation=1 to the failing URL.
(This is also available in the error message.)

We tested our validator against all the IATI example data, but this takes a few
seconds so we're not running them by default, or including them as a part of the repo.
Running rspec with `--tag full_iati_test` will download the data from github and
run them.

We're concerned about the number of failures that occur: a brief audit of delivery partners
beginning with A and B raised 6 failing documents, with errors exclusively of the forms:

```
XML validation errors were found in http://localhost:3000/exports/organisations/b51fa616-de29-46db-8af0-6e5f38e50591/iati/programme_activities.xml?fund=NF&skip_validation=1 
767:0: ERROR: Element 'activity-status': The attribute 'code' is required but missing. 
768:0: ERROR: Element 'contact-info': This element is not expected. Expected is ( activity-date ).
```

As such we propose that we don't merge this until we've investigated how to fix those issues, since
just breaking the XML output for a substantial proportion of exports isn't friendly.

## Next steps


rgarner:
[We should validate all XML documents generated and warn at least the developers (if not the end user) if validation fails.](https://trello.com/c/FTRivhfc/2364-we-should-validate-all-xml-documents-generated-and-warn-at-least-the-developers-if-not-the-end-user-if-validation-fails)

dragon-dxw:
TODO: understand why I can't test if Rollbar is called.

app/controllers/staff/exports/organisations_controller.rb
```diff
@@ -103,9 +105,19 @@ def fund_code
...
-    response.headers["Content-Disposition"] = "attachment; filename=\"#{@organisation.iati_reference}.xml\""
-
-    render "staff/exports/organisations/show"
+    output = render_to_string "staff/exports/organisations/show"
+    validator = IATIValidator::XML.new(output)
+    if validator.valid? || params[:skip_validation]
+      response.headers["Content-Disposition"] = "attachment; filename=\"#{@organisation.iati_reference}.xml\""
+      render body: output
+    else
+      # prevent Missing template pages/errors/internal_server_error :formats=>[:xml]
```
dragon-dxw, line 114: This comment seems wrong now.

app/controllers/staff/exports/organisations_controller.rb
```diff
@@ -103,9 +105,19 @@ def fund_code
...
+      # (we wanted XML, but the error's HTML-only)
+      request.format = :html
+      exception = IATIValidator::XML::InvalidError.new(validator.errors, request.url)
+      Rollbar.error(exception)
```
dragon-dxw, line 118: This needs testing

app/views/pages/errors/invalid_xml.html.haml
```diff
@@ -0,0 +1,19 @@
+= content_for :page_title_prefix, t("page_title.errors.invalid_xml")
+
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-full
+      %h2.govuk-heading-xl
+        = t("page_title.errors.invalid_xml")
+      %p.govuk-body
+        = t("page_content.errors.invalid_xml.explanation")
+        You can <a href='#{exception.skip_validation_url}'>download the invalid XML</a> anyway.
```
dragon-dxw, line 10: Fold this into the document string if you can

spec/support/error_responses.rb
```diff
@@ -0,0 +1,24 @@
+module ErrorResponses
+  # See: https://github.com/rails/rails/pull/11289#issuecomment-118612393
+  # See: https://github.com/eliotsykes/rails-testing-toolbox/blob/master/error_responses.rb
+
+  def respond_without_detailed_exceptions
```
dragon-dxw, line 5: This might not be necessary now we have a custom 500 page.

lib/iati_validator/xml.rb
```diff
@@ -0,0 +1,62 @@
...
+        "XML validation errors were found in #{skip_validation_url}\n#{@errors.join("\n")}"
+      end
+    end
+
+    SCHEMAS = {"iati-activities" => "iati-activities",
+               "iati-organisations" => "iati-organisations",
+               "registry-record" => "iati-registry-record"}
+    PATH = "vendor/data/IATI-Schemas-2.03/%s-schema.xsd"
+
+    attr_reader(:errors)
```
rgarner, line 29: duplicate of L6?

lib/iati_validator/xml.rb
```diff
@@ -0,0 +1,62 @@
...
+      return error("Not XML. #{doc.errors}") unless doc.errors.empty?
+      return error("No root node.") unless doc.root
+
+      schema = SCHEMAS[doc.root.name]
+      return error("Did not recognise opening tag #{doc.root.name}") unless schema
+
+      xsd = Nokogiri::XML::Schema(File.open(PATH % schema))
+
+      validation_errors = xsd.validate(doc)
+      unless validation_errors.empty?
```
rgarner, line 53: Prefer the positive case with the antonym of `#empty?`: `if validation_errors.any?`

rgarner: (or could go with `return true if validation_errors.empty?`)

rgarner: image
^^ this indentation, for the avoidance of doubt :wink:

dragon-dxw: :+1:

PR #1581, opened 2022-02-16
Create CSV report of missing data in activities via validation errors

dragon-dxw:
This adds a 'data_holes' rake task for generating CSV reports to hi. We play about with the errors a bit to get them all their own columns to ensure the data is sortable.

### DIFF
--- a/doc/utilities.md
+++ b/doc/utilities.md
@@ -9,12 +9,16 @@ each CSV row is:
   - Activity title
   - Activity level
   - Activity URL
+  - Validation Errors, one per column
+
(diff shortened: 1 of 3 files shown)


rgarner (approved):
:100: report for infrequent brushing up; doesn't need anything else if we're going to continue running it infrequently, and shouldn't see any more effort from us unless we're getting actionable fixes as a result of the report.

rgarner:
[Resolve XML issues identified during validation](https://trello.com/c/oZdZ2ChA/2397-resolve-xml-issues-identified-during-validation)

PR #1583, opened 2022-02-17
Fix/2393 missing orgs on user create

dragon-dxw:
## Changes in this PR

We ensure User#create has access to the BEIS/DP organisations to display them after an error.
We ensure the error message for those organisations renders correctly, highlighting the erroneous field and allowing the user to click the link to travel there.

### Before

image

### After first commit
image

### After
image

## Next steps

### DIFF
--- a/app/controllers/staff/users_controller.rb
+++ b/app/controllers/staff/users_controller.rb
@@ -26,7 +26,8 @@ class Staff::UsersController < Staff::BaseController
     @user = User.new(user_params)
     @user.active = params[:user][:active]
     authorize @user
-    @organisations = policy_scope(Organisation)
+    @service_owner = service_owner
+    @delivery_partners = delivery_partners
 
     if @user.valid?
       result = CreateUser.new(user: @user, organisation: organisation).call
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -974,6 +974,9 @@
 
 ## [unreleased]
 
+- No longer hide all the organisations when creating a user if one wasn't picked to start with
+- Correctly highlight the absence of an organisation when creating a user if one wasn't picked to start with
+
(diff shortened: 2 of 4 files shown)


rgarner:
[BUG: Organisations list missing when create user results in an error](https://trello.com/c/wssyxtSa/2393-bug-organisations-list-missing-when-create-user-results-in-an-error)

app/controllers/staff/users_controller.rb
```diff
@@ -27,6 +27,8 @@ def create
     @user.active = params[:user][:active]
     authorize @user
     @organisations = policy_scope(Organisation)
+    @service_owner = service_owner
+    @delivery_partners = delivery_partners
```
rgarner, line 31: How does this work to fix the error (I can't tell just by looking 😄)? Would saying how in the commit message help future us understand why this made things work? Or might it make sense to cover the case with a spec?

dragon-dxw: I've had a stab at this but I'm not happy with it. Would appreciate taking a look at it with someone.

rgarner: Fundamentally a `#create` in a controller won't ever have a template of its own; it's more usual to `redirect_to` a place where success is represented or `render` a different template for failure (as is the case here), so setting `@things = stuff` is odd in a `#create` (including that `@organisations = policy_scope(Organisation)` on line 29!)

rgarner: @dragon-dxw and I discovered that we need to leave these where they are for now as there are two code paths with a `render :new`. It will be possible to move them in `devise-authentication` where `users#create` is somewhat simplified.

rgarner (approved):
:100: thanks @dragon-dxw – needs a quick rebase

PR #1587, opened 2022-02-18
Fix seeds for devise users

james:
Since we switched to using local passwords for users, the seeds were broken.

This adds passwords to the seed users


rgarner (approved):
We're acknowledging a couple of things as we pass through here:

1. We're leaving auth0 parts. That's ok – we have a rollback route until devise is live
2. The passwords are temporary confections that will need updating when we address the [password policy](https://trello.com/c/3ED5g72M/2386-update-password-policy-to-match-auth0)
3. We'll be told about this in future due to the upgrade to `save!`

PR #1588, opened 2022-02-18
Admin can deactivate a user

rgarner:
A deactivated user does not get signed in, so we have removed the old
text and only slightly modified the default devise message.

We are not deleting unused translations at this stage. That has been
planned as a larger “mop-up” operation for later.

We also handle the user editing specs, because this seemed like the best
place to reinstate them.

Co-authored-by: rgarner <rgarner@zephyros-systems.co.uk>

### After

image

## Next steps


dragon-dxw (approved):

james (approved):
🎉

PR #1589, opened 2022-02-21
(2389) Devise two-factor login

rgarner:
## Changes in this PR

Two-phase Devise two-factor login. A user logs in with user name and password, and if they are set up for 2FA, receive a text message via Notify.

## Still to do:

- [x] Provision new env vars in Terraform
- [x] Adjust interval and drift for SMS ROTP
- [x] [Filter](https://github.com/tinfoil/devise-two-factor#filtering-sensitive-parameters-from-the-logs) OTP attempts from the logs

### After
Screenshot 2022-02-21 at 18 43 37

## Next steps


CristinaRO:
[Verify login with MFA](https://trello.com/c/4hHxXuul/2389-verify-login-with-mfa)

spec/lib/notify/otp_message_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "notify/otp_message"
+
+RSpec.describe Notify::OTPMessage do
+  subject(:message) { Notify::OTPMessage.new("+4477881234567", "123456") }
```
dragon-dxw, line 4: If you're worried about accidently spamming someone with text messages, 07700 900000 might be a plan.

https://www.numbersupermarket.co.uk/fake-number-ranges-here-are-the-numbers-reserved-exclusively-for-tv-radio-shows/

rgarner: Heh! Yes, I believe the GOV.UK docs mention this explicitly!

In this case though, everything is going to the [`fake_client`](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1589/commits/af6ca35e7c5adb30d85054367c75c296a6b8616b#diff-f39dd1bca5a438cf92e99327023c748d23a600c0545c96b26107e96f5e00a6dcR8)

rgarner: Yep, here they are in a page on [smoke testing](https://docs.notifications.service.gov.uk/ruby.html#smoke-testing). It's entertaining to note the provenance, mind 😆

dragon-dxw: Oh, didn't realise they'd got their own hooks on some of those numbers! Nice.

app/models/user.rb
```diff
@@ -22,6 +24,10 @@ def active_for_authentication?
 
   private
 
+  def ensure_otp_secret!
+    update!(otp_secret: User.generate_otp_secret) if otp_secret.nil?
```
rgarner, line 28: This is happening `before_save` so we don't need to `update!` as we'll make a second DB call; just `self.otp_secret=`

CristinaRO (approved):
Given that we have paired on this story, and that the whole feature will be reviewed before merging, we are confident to merge it ourselves without a third party review.

PR #1593, opened 2022-02-22
Delete activity:import rake task

dragon-dxw:
We used this Activity importer to import level B activities, which aren't importable via the web interface
because developing the web interface to do so would be a lot of effort for a one-time task.
This was for onboarding the delivery partner's historical data, which dxw ran manually.
Future onboarding should hopefully not need to run this task.

This code doesn't reference reports (useful for Level B, which doesn't use the concept).
This will break if an entry in the CSV has a comment, since comments require a report.

We could probably delete this.
-----
Also, it's a singleton which doesn't have corresponding other importers.


rgarner (approved):
'sgood. We might get away with just the single delete commit, since we have all our reasoning in the commit log?

PR #1597, opened 2022-02-28
2388 registering mfa number for the first time

CristinaRO:
## Changes in this PR

Collect mobile numbers for accounts that require OTP for login. Note that we aren't mandating that yet and it's `users.otp_required_for_login`.

### After

The link to check shows only until the mobile number is confirmed for the first time:

image

image

## Next steps

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -44,7 +44,7 @@ class Users::SessionsController < Devise::SessionsController
   end
 
   def user_params
-    params.require(:user).permit(:email, :password, :remember_me, :otp_attempt)
+    params.require(:user).permit(:email, :password, :remember_me, :otp_attempt, :mobile_number)
   end
 
   def valid_otp_attempt?(user)
@@ -56,10 +56,17 @@ class Users::SessionsController < Devise::SessionsController
 
     if user_params[:otp_attempt].present? && session[:otp_user_id]
       authenticate_user_with_otp_two_factor(user)
-    elsif user&.valid_password?(user_params[:password])
-      send_otp(user) if user.mobile_number
-
+    elsif user_params[:mobile_number].present? && session[:otp_user_id]
+      user.update!(mobile_number: user_params[:mobile_number])
+      send_otp(user)
       prompt_for_otp_two_factor(user)
+    elsif user&.valid_password?(user_params[:password])
+      if user.mobile_number.nil?
+        prompt_for_mobile_number(user)
+      else
+        send_otp(user)
+        prompt_for_otp_two_factor(user)
+      end
     end
   end
 
@@ -79,7 +86,7 @@ class Users::SessionsController < Devise::SessionsController
 
   # If you have extra params to permit, append them to the sanitizer.
   def configure_sign_in_params
-    devise_parameter_sanitizer.permit(:sign_in, keys: [:otp_attempt])
+    devise_parameter_sanitizer.permit(:sign_in, keys: [:otp_attempt, :mobile_number])
   end
 
   def find_user
@@ -99,6 +106,13 @@ class Users::SessionsController < Devise::SessionsController
 
     session[:otp_user_id] = user.id
     render "devise/sessions/two_factor"
+  end
+
+  def prompt_for_mobile_number(user)
+    @user = user
+
+    session[:otp_user_id] = user.id
+    render "devise/sessions/mobile_number"
   end
 
   def send_otp(user)
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -82,8 +82,9 @@ class Users::SessionsController < Devise::SessionsController
       session.delete(:otp_user_id)
 
       remember_me(user) if user_params[:remember_me] == "1"
-      user.mobile_number_confirmed_at = Time.zone.now
+      user.mobile_number_confirmed_at ||= Time.zone.now
       user.save!
+
       sign_in(user, event: :authentication)
     else
       flash.now[:alert] = t("devise.failure.invalid_two_factor")
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -18,6 +18,12 @@ class Users::SessionsController < Devise::SessionsController
 
   protect_from_forgery with: :exception, prepend: true, except: :destroy
 
+  def edit_mobile_number
+    @user = self.resource = find_user
+
+    render "devise/sessions/mobile_number"
+  end
+
   protected
 
   # we also need to completely override Devise's require_no_authentication such that
@@ -76,6 +82,7 @@ class Users::SessionsController < Devise::SessionsController
       session.delete(:otp_user_id)
 
       remember_me(user) if user_params[:remember_me] == "1"
+      user.mobile_number_confirmed_at = Time.zone.now
       user.save!
       sign_in(user, event: :authentication)
     else
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -13,7 +13,7 @@ require "notify/otp_message"
 class Users::SessionsController < Devise::SessionsController
   before_action :configure_sign_in_params, only: :create
 
-  prepend_before_action :authenticate_with_otp_two_factor,
+  prepend_before_action :handle_otp_flow,
     if: -> { action_name == "create" && otp_two_factor_enabled? }
 
   protect_from_forgery with: :exception, prepend: true, except: :destroy
@@ -57,7 +57,7 @@ class Users::SessionsController < Devise::SessionsController
     user.validate_and_consume_otp!(user_params[:otp_attempt])
   end
 
-  def authenticate_with_otp_two_factor
+  def handle_otp_flow
     user = self.resource = find_user
 
     if user_params[:otp_attempt].present? && session[:otp_user_id]
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -65,13 +65,13 @@ class Users::SessionsController < Devise::SessionsController
     elsif user_params[:mobile_number].present? && session[:otp_user_id]
       user.update!(mobile_number: user_params[:mobile_number])
       send_otp(user)
-      prompt_for_otp_two_factor(user)
+      prompt_for(user, "otp_attempt")
     elsif user&.valid_password?(user_params[:password])
       if user.mobile_number.nil?
-        prompt_for_mobile_number(user)
+        prompt_for(user, "mobile_number")
       else
         send_otp(user)
-        prompt_for_otp_two_factor(user)
+        prompt_for(user, "otp_attempt")
       end
     end
   end
@@ -87,7 +87,7 @@ class Users::SessionsController < Devise::SessionsController
       sign_in(user, event: :authentication)
     else
       flash.now[:alert] = t("devise.failure.invalid_two_factor")
-      prompt_for_otp_two_factor(user)
+      prompt_for(user, "otp_attempt")
     end
   end
 
@@ -108,18 +108,11 @@ class Users::SessionsController < Devise::SessionsController
     find_user&.otp_required_for_login
   end
 
-  def prompt_for_otp_two_factor(user)
+  def prompt_for(user, template)
     @user = user
 
     session[:otp_user_id] = user.id
-    render "devise/sessions/two_factor"
-  end
-
-  def prompt_for_mobile_number(user)
-    @user = user
-
-    session[:otp_user_id] = user.id
-    render "devise/sessions/mobile_number"
+    render "devise/sessions/#{template}"
   end
 
   def send_otp(user)
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -11,20 +11,94 @@ require "notify/otp_message"
 # Much of this is taken from step 4 of
 # https://www.jamesridgway.co.uk/implementing-a-two-step-otp-u2f-login-workflow-with-rails-and-devise/
 class Users::SessionsController < Devise::SessionsController
-  before_action :configure_sign_in_params, only: :create
-
-  prepend_before_action :handle_otp_flow,
-    if: -> { action_name == "create" && otp_two_factor_enabled? }
+  prepend_before_action :handle_mfa_flow, if: -> { action_name == "create" && otp_two_factor_enabled? }
 
   protect_from_forgery with: :exception, prepend: true, except: :destroy
 
   def edit_mobile_number
-    @user = self.resource = find_user
-
+    self.resource = find_user
     render "devise/sessions/mobile_number"
   end
 
   protected
+
+  # Given a login +mfa_phase+, take the appropriate action.
+  def handle_mfa_flow
+    case mfa_phase
+    when :needs_mobile_number then prompt_for "mobile_number"
+    when :has_mobile_number then send_and_prompt_for_otp
+    when :updating_mobile_number
+      user.update!(mobile_number: user_params[:mobile_number])
+      send_and_prompt_for_otp
+    when :validating_otp then complete_sign_in_with_otp? || prompt_for("otp_attempt")
+    end
+  end
+
+  def mfa_phase
+    if user_params[:otp_attempt] && session[:otp_user_id]
+      :validating_otp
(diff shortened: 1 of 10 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -56,10 +62,17 @@ def authenticate_with_otp_two_factor
...
+      user.update!(mobile_number: user_params[:mobile_number], mobile_number_confirmed_at: DateTime.now)
+      send_otp(user)
       prompt_for_otp_two_factor(user)
+    elsif user&.valid_password?(user_params[:password])
+      if user.mobile_number.nil?
+        prompt_for_mobile_number(user)
+      else
+        send_otp(user)
+        prompt_for_otp_two_factor(user)
+      end
```
rgarner, line 75: This block is cyclomatic complexity hell. Can we make the conditionals read more naturally to indicate what we're doing?

app/views/devise/sessions/two_factor.html.haml
```diff
@@ -1,15 +1,28 @@
 = content_for :page_title_prefix, "Log in"
```
rgarner, line 1: We should probably rename this file `otp_attempt.html.haml`

app/controllers/users/sessions_controller.rb
rgarner: Suggestions are very welcome.

spec/features/staff/users_can_sign_in_spec.rb
```diff
@@ -42,34 +42,140 @@ def log_in_via_form(user, remember_me: false)
...
+        # Then I should not be logged in
+        expect(page).to have_content("Invalid two-factor verification code")
       end
+    end
 
-      # Then I should be logged in.
-      expect(page).to have_link(t("header.link.sign_out"))
-      expect(page).to have_content("Signed in successfully.")
+    context "user fails to confirm their mobile number" do
+      scenario "a successful login happens but the mobile number has not been confirmed" do
```
dragon-dxw, line 97: Clarifying what we mean by "successful login" here would be useful: I assume we don't mean "can use RODA", but "we have a valid username/password"

dragon-dxw (approved):

app/controllers/users/sessions_controller.rb
```diff
@@ -44,22 +50,29 @@ def require_no_authentication
...
-    params.require(:user).permit(:email, :password, :remember_me, :otp_attempt)
+    params.require(:user).permit(:email, :password, :remember_me, :otp_attempt, :mobile_number)
   end
 
   def valid_otp_attempt?(user)
     user.validate_and_consume_otp!(user_params[:otp_attempt])
   end
 
-  def authenticate_with_otp_two_factor
+  def handle_otp_flow
```
dragon-dxw, line 60: I think this method would benefit from a description (or link to the description) of the intended handling.

spec/features/staff/users_can_sign_in_spec.rb
rgarner: You're right, will fix

app/controllers/users/sessions_controller.rb
rgarner: I agree, but am wondering if we can factor it differently such that it explains itself a lot better. We're pushing a lot through a single POST to a `create` method and then `if`ing it up...

app/views/devise/sessions/mobile_number.html.haml
```diff
@@ -7,9 +7,9 @@
         %fieldset.govuk-fieldset
           %legend.govuk-fieldset__legend.govuk-fieldset__legend--l
             %h1.govuk-fieldset__heading
-              Verify your login
+              Enter your mobile number
 
-          = f.govuk_text_field :otp_attempt, label: { text: "Please enter your six-digit verification code" }
+          = f.govuk_text_field :mobile_number, label: { text: "Enter your mobile number" }, type: "tel", autocomplete: "tel"
```
CristinaRO, line 12: We need `autocomplete` there to comply with [WCAG 2.1 Level AA](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html), according to https://design-system.service.gov.uk/patterns/telephone-numbers/.

app/controllers/users/sessions_controller.rb
```diff
@@ -44,22 +50,29 @@ def require_no_authentication
   end
 
   def user_params
-    params.require(:user).permit(:email, :password, :remember_me, :otp_attempt)
+    params.require(:user).permit(:email, :password, :remember_me, :otp_attempt, :mobile_number)
```
james, line 53: Is `user_params` or `configure_sign_in_params` what we're using to add mobile_number to permitted params, or both?

My guess is we can delete line 14 and the configure_sign_in_params method entirely, seeing as if I comment it out, then `spec/features/staff/users_can_sign_in_spec` still passes 🤔

app/controllers/users/sessions_controller.rb
```diff
@@ -13,11 +13,17 @@
...
   before_action :configure_sign_in_params, only: :create
 
-  prepend_before_action :authenticate_with_otp_two_factor,
+  prepend_before_action :handle_otp_flow,
     if: -> { action_name == "create" && otp_two_factor_enabled? }
 
   protect_from_forgery with: :exception, prepend: true, except: :destroy
 
+  def edit_mobile_number
+    @user = self.resource = find_user
```
james, line 22: why do we need to set `self.resource` here?

app/controllers/users/sessions_controller.rb
```diff
@@ -13,11 +13,17 @@
 class Users::SessionsController < Devise::SessionsController
   before_action :configure_sign_in_params, only: :create
 
-  prepend_before_action :authenticate_with_otp_two_factor,
+  prepend_before_action :handle_otp_flow,
     if: -> { action_name == "create" && otp_two_factor_enabled? }
```
james, line 17: If we're looking to simplify things, can we just remove `otp_required_for_login` method and the `otp_required_for_login` field?

app/controllers/users/sessions_controller.rb
james: Here's an idea:

Get rid of `handle_otp_flow` altogether, and stick to (and admittedly push to the limits of) the approach that devise is expecting, which is to prepend an action to create for each thing we want to check before authenticating:

```ruby
  prepend_before_action :ask_for_mobile_number, only: :create, if: -> { user_missing_mobile_number? }
  prepend_before_action :save_new_mobile_number, only: :create, if: -> { user_provided_new_mobile_number? }
  prepend_before_action :prompt_for_otp, only: :create, if: -> { user_has_mobile_number? }
  prepend_before_action :validate_otp, only: :create
```

rgarner: Oh, that's nice, I see what you're driving at. I agree with you that this is perhaps pushing `prepend_before_action` a little far, though! How about something that takes symbols like that and makes our little state machine clear (and hopefully also debuggable!) in one method?

```ruby
  def handle_otp_flow
    case phase
    when :valid_password_needs_mobile_number then prompt_for "mobile_number"
    when :valid_password_has_mobile_number then send_and_prompt_for_otp
    when :updating_mobile_number
      user.update!(mobile_number: user_params[:mobile_number])
      send_and_prompt_for_otp
    when :validating_otp then authenticate_user_with_otp_two_factor
    end
  end
```

app/controllers/users/sessions_controller.rb
rgarner: Mm. I think you're right. I think this is still there in a cargo-culty sense; since we're rendering every phase of the process now and Devise isn't, I think it's a hangover from a time when we weren't and Devise needed To Be Told. Let's remove.

app/controllers/users/sessions_controller.rb
rgarner: Because the views are still in `devise/controllers` and use the Devise `resource` form rather than an explicit `@user`. That said, there's nothing to say we can't just move those to `users/session` now and use `@user`.

app/controllers/users/sessions_controller.rb
rgarner: It's part of the `two-factor` gem. I'm not certain it's worth the effort against defaulting the column to `true` in a `before_initialize` or similar, which at least gives us the option of turning it off to speed development.

rgarner:
Thanks @james, really helpful comments

app/controllers/users/sessions_controller.rb
james: I think that's definitely an improvement! Better names needed now I think. Maybe `phase` can become something like `mfa_state` and `handle_otp_flow` can become `mfa_authenticate`.

As for the state names, aren't they always going to have valid_password? Do we need to state that twice?

app/controllers/users/sessions_controller.rb
james: I understand that it seems foolish to disable something we get for free, but I really don't see us needing this feature (MFA is required for all users of the system), and I'm keen to remove as much logic as we can from an already overly confused process.

app/controllers/users/sessions_controller.rb
james: Ah, interestingly, when I removed this and ran `rspec spec/features/staff/users_can_sign_in_spec.rb` then all the tests still passed. Is this because the tests don't cover something?

app/views/devise/sessions/otp_attempt.html.haml
```diff
@@ -0,0 +1,28 @@
...
+            %h1.govuk-fieldset__heading
+              Verify your login
+
+          = f.govuk_text_field :otp_attempt,
+                                label: { text: "Please enter your six-digit verification code" },
+                                inputmode: "numeric",
+                                pattern: "[0-9]*",
+                                autocomplete: "one-time-code"
+
+  - unless resource.mobile_number_confirmed_at
```
james, line 18: worlds smallest thing, but could this be `resource.mobile_number_confirmed_at?` or `resource.mobile_number_confirmed_at.present?`

spec/factories/user.rb
```diff
@@ -28,5 +28,10 @@
     trait :mfa_enabled do
       otp_required_for_login { true }
     end
+
+    trait :no_mobile_number do
+      mobile_number { nil }
+      mobile_number_confirmed_at { nil }
```
james, line 34: here we're specifying that `mobile_number_confirmed_at` is nil, but we're not setting it for all other users, so its not really overriding anything. I think our default user factory should have this set to a date in the past.

rgarner: That is absolutely true but we're describing a known state we *want* as a result of the trait and indicating that these fields are to be taken together; this is largely because people can add things they believe to be unrelated in other FactoryBot places and break tests. It's as much a cue for the reader as a programmatic "must set".

spec/features/staff/users_can_sign_in_spec.rb
```diff
@@ -42,34 +42,140 @@ def log_in_via_form(user, remember_me: false)
...
-
-      # When I enter the one-time password before it expires
-      travel 3.minutes do
-        fill_in "Please enter your six-digit verification code", with: otp_at_time_of_login
+        # And I enter an incorrect OTP
+        fill_in "Please enter your six-digit verification code", with: "000000"
         click_on "Continue"
+
+        # Then I should not be logged in
+        expect(page).to have_content("Invalid two-factor verification code")
```
james, line 92: I may be being paranoid, but just because I can imagine something going wrong here with security, I'd like to double check we're not logged in here, and not just showing this error message. Maybe try and visit a page and ensure we're redirected to login?

spec/features/staff/users_can_sign_in_spec.rb
```diff
@@ -42,34 +42,140 @@ def log_in_via_form(user, remember_me: false)
...
+        # When I log in with that user's email and password
+        visit root_path
+        log_in_via_form(user)
+
+        # But I do not receive a message despite it having been sent
+        expect(notify_client).to have_received(:send_sms).once.with({
+          phone_number: incorrect_number,
+          template_id: ENV["NOTIFY_OTP_VERIFICATION_TEMPLATE"],
+          personalisation: {otp: user.current_otp}
+        })
```
james, line 111: Would notify respond with an error here? Do we handle notify errors is they exist?

app/views/devise/sessions/otp_attempt.html.haml
rgarner: Can add `present?` if you like but this is a DateTime column and can only be `nil` or an `ActiveSupport::TimeWithZone`. `#present?` is really more of use for strings where the blank case is also covered, though I do get that people like it as a visual treat 😄

app/controllers/users/sessions_controller.rb
rgarner: It's in use [here](https://github.com/tinfoil/devise-two-factor/blob/main/lib/devise_two_factor/strategies/two_factor_authenticatable.rb#L23)

app/controllers/users/sessions_controller.rb
rgarner: It's because in Devise they're [synonymous](https://github.com/heartcombo/devise/blob/main/app/controllers/devise_controller.rb#L38-L40)

spec/features/staff/users_can_sign_in_spec.rb
```diff
@@ -42,34 +42,140 @@ def log_in_via_form(user, remember_me: false)
...
+
+        # Then I should not be logged in
+        expect(page).to have_content("Invalid two-factor verification code")
       end
+    end
 
-      # Then I should be logged in.
-      expect(page).to have_link(t("header.link.sign_out"))
-      expect(page).to have_content("Signed in successfully.")
+    context "user initially entered an incorrect mobile number" do
```
james, line 96: I feel this context should be nested with the context below. They're both about if a user doesn't have a mobile number, one is about if they do it successfully, one if they don't?

Which makes me think maybe there should be a test for ensuring you can't edit your mobile number in this flow if you have a confirmed mobile number.

app/controllers/users/sessions_controller.rb
rgarner: (I removed the `@user = ` part this morning)

app/views/devise/sessions/otp_attempt.html.haml
james: Yup, I understand that, but I like being really explicit where I can be

spec/factories/user.rb
james: I'm not suggesting removing this trait, I agree it should exist, and I like it. I'm suggesting that the default user factory should have a mobile number confirmed, as that's our common case.

rgarner: Oh I see. In that case we'll add.

spec/features/staff/users_can_sign_in_spec.rb
rgarner: Doing.

spec/features/staff/users_can_sign_in_spec.rb
rgarner: Would you object to us adding such a test as an acceptance criterion to https://trello.com/c/2k94LHZX/2391-resetting-a-users-phone-number-for-mfa ?

rgarner: We'll add a test at the model level for the inability to edit confirmed mobile numbers.

rgarner: Hm. It's only possible to edit this were we to craft a POST to `users/session#create` with both a `session[:user_otp_id]` and `user_params[:mobile_number]`. This is difficult to do unless you're actually that user (And CSRF will catch it from the form flow anyway)

rgarner:
[Remove the `users.otp_required_for_login` column](https://trello.com/c/1b4Z4IJF/2432-remove-the-usersotprequiredforlogin-column)

app/controllers/users/sessions_controller.rb
CristinaRO: We've agreed to put [this card](https://trello.com/c/1b4Z4IJF/2432-remove-the-usersotprequiredforlogin-column) in the Tech debt column.

spec/features/staff/users_can_sign_in_spec.rb
rgarner: We have a test that the link to edit doesn't show up when you're confirmed. In light of that we can lock this down at model level on the Resetting MFA card, as it's next to impossible to craft a POST to edit your mobile number through this flow and it's not yet on the admin page.

PR #1598, opened 2022-03-01
2411: Comments in CSV import

dragon-dxw:
https://trello.com/c/MVWK01CQ/2411-add-comments-into-bulk-uploads

### Discovering Polymorphism The Hard Way

When we came to do things like

`@activity.comments.build(body: "Kittens", report: Report.first)` on an `Activity`, we would get this validation error on the newly created `Comment`:

`ActiveRecord::RecordInvalid: Validation failed: Commentable Commentable must exist`

This is because the scope `comments` uses the `Comment#for_activity` scope, which explicitly `LEFT JOIN`s some transaction types (refund and adjustment). Rails cannot then set the new comment's `commentable` property automatically as it would usually do in `@activity.comments.build`. We need to use the somewhat circuitous `@activity.comments.build(body: "Kittens", report: Report.first, commentable: @activity)`.

## Changes in this PR

* Add Comments column to Activities import CSV and import its contents
* Add comments column to Actuals importer, create DB handling for Actuals comments, and import their comments
* Add comments to `report/.../actuals` table.

## Screenshots of UI changes

The comments are new
image

`/reports/:report_id/comments` didn't support Actual comments, but does now.
image

## Next steps

- [X] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team


app/services/activities/import_from_csv.rb
```diff
@@ -204,7 +208,12 @@ def create
           @activity.implementing_org_participations = participations
         end
 
-        return if @activity.save(context: Activity::VALIDATION_STEPS)
+        if @activity.save(context: Activity::VALIDATION_STEPS)
+          if row["Comments"].present?
+            @activity.comments.create!(body: row["Comments"], report: @report, owner: @uploader)
+          end
```
dragon-dxw, line 214: Consider rewriting this to use build and not be reliant on saving

dragon-dxw:
Do we need to ensure `app/model/comments.rb` has been updated to reflect the newly commentable Actual type?

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -809,6 +814,18 @@
...
+    it "imports comments" do
+      rows = [existing_activity_attributes, new_activity_attributes]
+      subject.import(rows)
+      existing_activity = Activity.where(transparency_identifier: existing_activity_attributes["Transparency identifier"]).first
+      new_activity = Activity.where(transparency_identifier: new_activity_attributes["Transparency identifier"]).first
+      expect(subject.errors.count).to eq(0)
+      # If there is a comment text, we add it to the activity
+      expect(existing_activity.comments.first.body).to eq("Cat")
+      # If the comment is empty, there is no comment added, and no error
+      expect(new_activity.comments).to be_empty
```
dragon-dxw, line 826: We should *actually* test both creation and updating separately.

dragon-dxw: Done. This revealed a bug.

app/services/activities/import_from_csv.rb
dragon-dxw: No. Having it post-save is incorrect, and the new test catches this.

app/services/import_actuals.rb
```diff
@@ -161,6 +162,10 @@ def convert_to_attribute(attr_name, value)
       nil
     end
 
+    def convert_comment(string)
```
rgarner, line 165: Dunno why I put `string` here. My bad. `body` is what it is!

app/services/activities/import_from_csv.rb
```diff
@@ -17,10 +17,10 @@ def column_to_csv_column_mapping
     attr_reader :errors, :created, :updated
 
     def self.column_headings
-      ["Parent RODA ID"] + Converter::FIELDS.values - ["BEIS ID"]
+      ["Parent RODA ID"] + Converter::FIELDS.values - ["BEIS ID"] + ["Comments"]
```
dragon-dxw, line 20: Consider adding Comments to Converter::FIELDS

app/services/activities/import_from_csv.rb
```diff
@@ -128,6 +128,10 @@ def update
         else
           @activity.implementing_organisations = []
         end
+        # How do we handle errors?
```
dragon-dxw, line 131: Delete this comment.

app/services/activities/import_from_csv.rb
```diff
@@ -203,7 +208,13 @@ def create
           @activity.implementing_org_participations = participations
         end
 
-        return if @activity.save(context: Activity::VALIDATION_STEPS)
+        if row["Comments"].present?
+          @activity.comments.build(body: row["Comments"], report: @report, owner: @uploader, commentable: @activity)
+        end
+
+        if @activity.save(context: Activity::VALIDATION_STEPS)
```
dragon-dxw, line 215: Make a oneliner : return true if...

app/services/create_actual.rb
```diff
@@ -23,6 +23,11 @@ def call(attributes: {})
       actual.report = report
     end
 
+    if actual.comment
+      actual.comment.report = @report
+      actual.comment.commentable = actual
```
dragon-dxw, line 28: Add comment or commit message explaining commentable faff.

spec/services/activities/import_from_csv_spec.rb
```diff
@@ -809,6 +814,29 @@
       expect(subject.errors.count).to eq(0)
     end
 
+    it "imports comments" do
+      rows = [existing_activity_attributes, new_activity_attributes]
+      subject.import(rows)
```
dragon-dxw, line 819: new line after subject; before assertions too

dragon-dxw:
Add feature specs for both Activty and Actual cases. (Maybe new and updated)

app/models/actual.rb
```diff
@@ -1,4 +1,12 @@
 class Actual < Transaction
+  has_one :comment,
+    -> { where(commentable_type: "Actual") },
+    foreign_key: :commentable_id,
+    dependent: :destroy,
+    autosave: true
+
+  # validates_associated :comment
```
CristinaRO, line 8: We should either remove this comment or un-comment it. Was this a temporary measure during development?

dragon-dxw: Removing comment, rerunning tests.

app/services/activities/import_from_csv.rb
dragon-dxw: We can't trivially mass-assign Comments, because it's structurally different to all the other fields.
So it probably doesn't belong in that big list?

rgarner:
[Add comments into bulk uploads](https://trello.com/c/MVWK01CQ/2411-add-comments-into-bulk-uploads)

rgarner (approved):
:shipit:

PR #1599, opened 2022-03-01
Authenticate with Devise; remove Auth0 functional code

rgarner:
## Changes in this PR

This PR is a long-lived integration branch that holds the changes required to move from Auth0 to a self-hosted authentication solution.

## Things this PR doesn't do

In order to keep the size of the PR down, we've deliberately left some Auth0 artefacts behind. Those include but are not limited to:

- Translations
- Terraform ENV vars

## Next steps


rgarner:
[Authentication](https://trello.com/c/fuP2sGlq/2340-authentication)

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+      sign_in(user, event: :authentication)
+      true
+    else
+      flash.now[:alert] = t("devise.failure.invalid_two_factor")
+      false
+    end
+  end
+
+  def user
+    @user ||= (self.resource = find_user)
```
rgarner, line 75: No need for the `@user` as Devise's `#resource=` calls `instance_variable_set` on `@user` anyway

app/mailers/user_mailer.rb
```diff
@@ -1,22 +1,25 @@
-require "./lib/auth0_api"
+# require "./lib/auth0_api"
```
rgarner, line 1: We can remove this.

app/controllers/concerns/auth.rb
```diff
@@ -18,18 +18,6 @@ module Auth
     end
   end
 
-  def current_user
```
rgarner, line 21: We can also remove the `class UserNotAuthorised` in this file (GitHub doesn't let me point at it directly) which is now no longer in use

app/services/create_user.rb
```diff
@@ -9,29 +9,11 @@ def initialize(user:, organisation:)
...
-        result.error_message = extract_auth0_error_message(e)
-        Rails.logger.error("Error adding user #{user.email} to Auth0 during CreateUser with: #{result.error_message}")
-        raise ActiveRecord::Rollback
-      end
-    end
+    user.organisation = organisation
+    user.password = SecureRandom.uuid
+    result.success = user.save
 
     SendWelcomeEmail.new(user: user).call if user.persisted?
```
rgarner, line 16: Perhaps should read `if result.success?`

config/initializers/devise.rb
```diff
@@ -0,0 +1,318 @@
+# frozen_string_literal: true
```
rgarner, line 1: Again, can be removed

db/migrate/20220301145502_default_user_otp_required_for_login_to_true.rb
```diff
@@ -0,0 +1,14 @@
...
+    change_column :users, :otp_required_for_login, :boolean, default: true
+    sql = <<~SQL
+      UPDATE users SET otp_required_for_login = 't'
+    SQL
+
+    ActiveRecord::Base.connection.execute sql
+  end
+
+  def down
+    change_column :users, :otp_required_for_login, :boolean
```
rgarner, line 12: Should this read `default: nil` or is that implied?

app/controllers/concerns/auth.rb
```diff
@@ -18,18 +18,6 @@ module Auth
...
-    @current_user ||= if session.dig(:userinfo)
-      User.active.includes(:organisation).find_by!(identifier: signed_in_user_identifier) do |user|
-        user.name = session.dig(:userinfo, "info", "name")
-        user.email = session.dig(:userinfo, "info", "email")
-      end
-    end
-  rescue ActiveRecord::RecordNotFound
-    repudiate!
-    raise UserNotAuthorised
```
rgarner, line 30: Because we've removed the only thing that could `raise UserNotAuthorised` and the only usage of it is on L10 of this file, we should remove the class and also the branch of the `if` on L11 that does not respond to `:policy`

app/services/create_user.rb
rgarner: Opened https://trello.com/c/fODkynU1/2452-remove-usernotauthorised-paraphernalia to deal with this, as it's not as simple as it first appears.

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,125 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
rgarner: Not hugely important

app/controllers/users/sessions_controller.rb
rgarner: Rebased

james:
🥳

jessicalondon:
The CHANGELOG looks good to me!

JessBuildsTech (approved):
The CHANGELOG looks good to me, great work!

james (approved):

PR #1605, opened 2022-03-07
(2386) Update password policy to align with BEIS policy

CristinaRO:
## Changes in this PR
- Enforce password requirements as close to BEIS requirements and GOV.UK guidelines as possible
- Turn devise's `paranoid` mode on, to avoid email enumeration attacks

### After

Screenshot 2022-03-08 at 14 18 29
Screenshot 2022-03-08 at 14 19 11
Screenshot 2022-03-08 at 14 19 25
Screenshot 2022-03-08 at 14 20 01

### DIFF
--- a/app/views/devise/passwords/new.html.haml
+++ b/app/views/devise/passwords/new.html.haml
@@ -5,7 +5,7 @@
 .govuk-grid-row
   .govuk-grid-column-one-third
     = form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :post }) do |f|
-      = render "devise/shared/error_messages", resource: resource
+      = f.govuk_error_summary
 
       = f.govuk_email_field :email, autofocus: true, autocomplete: "email"
 
--- a/db/seeds/development_users.rb
+++ b/db/seeds/development_users.rb
@@ -9,8 +9,8 @@ delivery_partner = User.find_or_initialize_by(
   identifier: "auth0|5e5e1ee731555a0cb0ab5a75"
 )
 
-administrator.password = "letmein"
-delivery_partner.password = "letmein"
+administrator.password = "LlEeTtMmEeIiNn!1"
+delivery_partner.password = "LlEeTtMmEeIiNn!1"
 
 beis = Organisation.service_owner
 administrator.organisation = beis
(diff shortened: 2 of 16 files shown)


CristinaRO:
[Update password policy to align with BEIS policy](https://trello.com/c/3ED5g72M/2386-update-password-policy-to-align-with-beis-policy)

james:
Looking good! We should update db/seeds/development_users.rb too so that reseeding the database continues to work

james:
One solution for the error message is to create a single one that covers all possibilities. "Password must be at least 9 characters and include at least 1 one digit, 1 lower case and 1 upper case character and 1 symbol". This should also be given as a hint on the form field as well I think.

CristinaRO:
> One solution for the error message is to create a single one that covers all possibilities. "Password must be at least 9 characters and include at least 1 one digit, 1 lower case and 1 upper case character and 1 symbol". This should also be given as a hint on the form field as well I think.

Thanks for the hint idea! I'm okay with the first error message being shown when we have the full requirements listed in the hint.

app/views/devise/passwords/new.html.haml
```diff
@@ -5,7 +5,7 @@
 .govuk-grid-row
   .govuk-grid-column-one-third
     = form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :post }) do |f|
-      = render "devise/shared/error_messages", resource: resource
+      = f.govuk_error_summary
```
james, line 8: I'm not entirely sure why this change was made. It may just need adding to the commit message.

CristinaRO: Rewrote two commit messages, if you'd like to re-re-review please?

james (approved):

spec/models/user_spec.rb
```diff
@@ -25,4 +25,33 @@
     it { is_expected.to delegate_method(:service_owner?).to(:organisation) }
     it { is_expected.to delegate_method(:delivery_partner?).to(:organisation) }
   end
+
+  describe "#email_validation" do
```
rgarner, line 29: Is this still a method name we're interested in now we're using the `EmailValidator`?

db/seeds/development_users.rb
```diff
@@ -9,8 +9,8 @@
   identifier: "auth0|5e5e1ee731555a0cb0ab5a75"
 )
 
-administrator.password = "letmein"
-delivery_partner.password = "letmein"
+administrator.password = "LlEeTtMmEeIiNn!1"
```
rgarner, line 12: I can *hear* this 🙀

rgarner:
😹

PR #1611, opened 2022-03-09
2431: Add Other ODA fund into RODA

dragon-dxw:
## Changes in this PR

We add the Other ODA (OODA) fund, both as a Fund in the codelist and as an Activity with level: 'fund'

For the Activity, we generated the values in the data migration by: 

Sector and Sector Category are identical to GCRF (after discussion from BEIS)
Description and start date are directly from BEIS
Other values generally there to pass validation 
We removed GDI validation from Activities at Fund Level since it's nil for both GCRF/NF

We've done some testing that if we swap the :newton_funded traits in the activities factories to reference OODA activities, any failing tests are understood not to be due to OODA.

- [ ] Data Migration has occurred after deployment

### DIFF
--- a/db/data/20220308163700_add_ooda_fund_activity.rb
+++ b/db/data/20220308163700_add_ooda_fund_activity.rb
@@ -1,0 +1,26 @@
+# Add the top level (A) Other ODA Activity (i.e. a fund-tier activity)
+
+beis = Organisation.service_owner
+
+description = <<~TXT.squish
+  Other ODA represents spending from ‘other non ODA budgets from BEIS’ which may be
+  classed as ‘ODA spend’ i.e. is ODA-eligible in the same way as spend from the
+  Newton Fund and GCRF.
+TXT
+
+Activity.create!(
+  source_fund_code: 3,
+  roda_identifier: "OODA",
+  title: "Other ODA",
+  organisation: beis,
+  level: "fund",
+  form_state: "complete", # wizard progress
+  fstc_applies: false, # free_standing_technical_cooperation: false,
+  delivery_partner_identifier: "OODA",
+  description: description,
+  sector_category: "998", # 510, 998 for NF/GCRF (based on GCRF)
+  sector: "99810", # focus area, # 51010, 99810 for NF/GCRF (based on GCRF)
+  programme_status: "spend_in_progress", # as NF/GCRF
+  actual_start_date: Date.new(2015, 4, 1),
+  aid_type: "C01" # same as NF/GCRF
+)
--- a/spec/models/iati/xml_download_spec.rb
+++ b/spec/models/iati/xml_download_spec.rb
@@ -43,14 +43,23 @@ RSpec.describe Iati::XmlDownload do
         allow(third_party_project_relation).to receive(:where).and_return(build_list(:project_activity, 5))
       end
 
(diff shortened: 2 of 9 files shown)


vendor/data/codelists/BEIS/fund_types.yml
```diff
@@ -5,3 +5,6 @@ data:
   - code: 2
     name: Global Challenges Research Fund
     short_name: GCRF
+  - code: 3
```
CristinaRO, line 8: Would the XML spec fail already as a result of this commit? (And if so, are you comfortable having a few "red" commits? ;) )

vendor/data/codelists/BEIS/fund_types.yml
```diff
@@ -5,3 +5,6 @@ data:
...
+    name: Other ODA
+    short_name: OODA
```
rgarner, line 10: Oops newline

vendor/data/codelists/BEIS/fund_types.yml
rgarner: Yes. 🙈

rgarner: (I mean yes it would fail; I still experience discomfort. But not much)

db/data/20220308163700_add_ooda_fund_activity.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_fund_code: 3,
+  roda_identifier: "OODA",
+  title: "Other ODA",
+  organisation: beis,
+  level: "fund",
+  form_state: "complete", # wizard progress
+  fstc_applies: false, # free_standing_technical_cooperation: false,
+  delivery_partner_identifier: "OODA",
+  description: description,
+  sector_category: "998", # 510, 998 for NF/GCRF (based on GCRF)
```
CristinaRO, line 21: Given that the sector and category have been confirmed with BEIS, I'm not sure the comments here add anything.

spec/models/iati/xml_download_spec.rb
```diff
@@ -43,14 +43,23 @@
...
+
+        expect(downloads.map { |p| p.fund.short_name }).to eq(%w[
+          NF GCRF OODA
+          NF GCRF OODA
+          NF GCRF OODA
+        ])
+
+        expect(downloads.map(&:level)).to eq(%w[
+          programme programme programme
+          project project project
```
CristinaRO, line 59: badger badger badger

vendor/data/codelists/BEIS/fund_types.yml
CristinaRO: Interesting! I should ask you to amend the commit title to say "RED:" :P (But I won't, I'll just let the idea float in the air ;) )

CristinaRO (approved):
It would make me a little more content if we had a project- or third party project-level spec for `:ooda_funded`. While I'm content with having run the switcheroo test, something more permanent that doesn't add minutes to the testing time would cement OODA as a proper fund in its own right.

spec/models/iati/xml_download_spec.rb
dragon-dxw: 🍄 🍄 🍄

PR #1614, opened 2022-03-11
2412: Bulk upload Refunds

dragon-dxw:
## Changes in this PR

    The chosen method is to bolt this functionality onto the existing Actuals
    upload, renaming the Value column to Actual Value and adding a new Refund Value
    column.
    
    Exactly one of these values must be filled in for an upload to occur.
    
    Create Actual now takes an optional parameter which allows it to generate a
    Refund instead, but is otherwise unchanged.
    
    Import Actual contains logic for determining whether a row is a Refund or
    Actual, based on the presence of the two types of value.
    
   
We also make comments visible on 'refunds by activity'
 
### Actuals copy: before
image

### Actuals copy: after
image

### Upload page: before
image

### Upload page: after
image

### After

CSV upload has Actual Value and Refund Value
image

Refunds have comments (still messy)
image

## Next steps

### DIFF
--- a/app/services/create_refund.rb
+++ b/app/services/create_refund.rb
@@ -13,11 +13,8 @@ class CreateRefund
     refund.parent_activity = activity
     refund.report = editable_report_for(activity)
     assign_refund_and_comment(attributes)
-    save_value = refund.save
-
-    raise Error, validation_error_message(refund) if refund.errors.any?
-
-    record_historical_event(refund)
+    save_successful = refund.save
+    record_historical_event(refund) if save_successful
 
     Result.new(save_successful, refund)
   end
--- a/app/services/create_refund.rb
+++ b/app/services/create_refund.rb
@@ -3,7 +3,7 @@ class CreateRefund
 
   attr_accessor :activity, :report, :refund, :user
 
-  def initialize(activity:, user:)
+  def initialize(activity:, user:, report: nil)
     self.activity = activity
     self.user = user
     self.refund = Refund.new
@@ -13,13 +13,13 @@ class CreateRefund
     refund.parent_activity = activity
     refund.report = editable_report_for(activity)
     assign_refund_and_comment(attributes)
-    refund.save
+    save_value = refund.save
 
     raise Error, validation_error_message(refund) if refund.errors.any?
 
     record_historical_event(refund)
 
-    refund
+    Result.new(save_successful, refund)
   end
 
   private
--- a/app/services/import_actuals.rb
+++ b/app/services/import_actuals.rb
@@ -88,8 +88,8 @@ class ImportActuals
       attrs = @converter.to_h
       assign_default_values(attrs)
 
-      creator = CreateActual.new(activity: @activity, report: @report)
-      result = creator.call(attributes: attrs)
+      creator = CreateActual.new(activity: @activity, report: @report, actual_class: @converter.actual_type)
+      result = creator.call(attributes: attrs.except(:refund_value))
       return unless result
 
       result.object.errors.each do |error|
@@ -113,12 +113,15 @@ class ImportActuals
       activity: "Activity RODA Identifier",
       financial_quarter: "Financial Quarter",
       financial_year: "Financial Year",
-      value: "Value",
+      value: "Actual Value",
+      refund_value: "Refund Value",
       receiving_organisation_name: "Receiving Organisation Name",
       receiving_organisation_type: "Receiving Organisation Type",
       receiving_organisation_reference: "Receiving Organisation IATI Reference",
       comment: "Comment"
     }
+
+    NON_VALUE_FIELDS = FIELDS.without :value, :refund_value
 
     attr_reader :activity, :errors
 
@@ -127,6 +130,16 @@ class ImportActuals
       @errors = {}
       @attributes = convert_to_attributes
       @activity = @attributes.delete(:activity)
+    end
+
+    # We must have precisely one value, and we will check if
+    # the other is empty when we validate the value it contains
+    def actual_type
+      return Actual if @row["Refund Value"].nil?
+      return Refund if @row["Actual Value"].nil?
+      # TODO: make this nicer, actually return value that's bad
+      @errors[:value] = [@row["Actual Value"], "One of Actual Value and Refund Value must be numeric and the other must be blank"]
+      @errors[:refund_value] = [@row["Refund Value"], "One of Actual Value and Refund Value must be numeric and the other must be blank"]
     end
 
     def raw(attr_name)
@@ -138,8 +151,19 @@ class ImportActuals
     end
 
     def convert_to_attributes
-      FIELDS.each_with_object({}) do |(attr_name, column_name), attrs|
+      attrs = {value: convert_values}
+      NON_VALUE_FIELDS.each_with_object(attrs) do |(attr_name, column_name), attrs|
         attrs[attr_name] = convert_to_attribute(attr_name, @row[column_name])
+      end
+    end
+
+    def convert_values
+      if actual_type == Actual
+        # convert_value(@row["Actual Value"])]
+        convert_to_attribute(:value, @row["Actual Value"])
+      else
+        convert_to_attribute(:refund_value, @row["Refund Value"])
+        # convert_value(@row["Refund Value"])
       end
     end
 
@@ -175,6 +199,8 @@ class ImportActuals
     rescue ConvertFinancialValue::Error
       raise I18n.t("importer.errors.actual.non_numeric_value")
     end
+
+    alias_method :convert_refund_value, :convert_value
 
     def convert_receiving_organisation_type(type)
       validate_from_codelist(
--- a/app/services/import_actuals.rb
+++ b/app/services/import_actuals.rb
@@ -1,3 +1,4 @@
+# This originally handled only Actuals, but now also handles Refunds as well.
 class ImportActuals
   Error = Struct.new(:row, :column, :value, :message) {
     def csv_row
@@ -137,7 +138,7 @@ class ImportActuals
     def actual_type
       return Actual if @row["Refund Value"].nil?
       return Refund if @row["Actual Value"].nil?
-      # TODO: make this nicer, actually return value that's bad
+      # TODO: for some reason I can't use t("importer.errors.actual.non_numeric_value") here.
       @errors[:value] = [@row["Actual Value"], "One of Actual Value and Refund Value must be numeric and the other must be blank"]
       @errors[:refund_value] = [@row["Refund Value"], "One of Actual Value and Refund Value must be numeric and the other must be blank"]
     end
--- a/app/services/import_actuals.rb
+++ b/app/services/import_actuals.rb
@@ -89,7 +89,11 @@ class ImportActuals
       attrs = @converter.to_h
       assign_default_values(attrs)
 
-      creator = CreateActual.new(activity: @activity, report: @report, actual_class: @converter.actual_type)
+      creator = if @converter.actual_class == Refund
+        CreateRefund.new(activity: @activity, report: @report, user: @uploader)
+      else
+        CreateActual.new(activity: @activity, report: @report, user: @uploader)
+      end
       result = creator.call(attributes: attrs.except(:refund_value))
       return unless result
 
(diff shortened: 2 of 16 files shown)


dragon-dxw:
Consider using app/services/create_refund.rb rather than mangling create_actual

app/services/import_actuals.rb
```diff
@@ -129,6 +133,16 @@ def initialize(row)
       @activity = @attributes.delete(:activity)
     end
 
+    # We must have precisely one value, and we will check if
+    # the other is empty when we validate the value it contains
+    def actual_type
+      return Actual if @row["Refund Value"].nil?
+      return Refund if @row["Actual Value"].nil?
+      # TODO: for some reason I can't use t("importer.errors.actual.non_numeric_value") here.
```
rgarner, line 141: `t` only gets mixed into view helpers. Try `I18n.translate` or `I18n.t`.

dragon-dxw:
[Bulk upload refunds](https://trello.com/c/FeCRz16U/2412-bulk-upload-refunds)

app/services/import_actuals.rb
```diff
@@ -176,6 +200,10 @@ def convert_value(value)
       raise I18n.t("importer.errors.actual.non_numeric_value")
     end
 
+    def convert_refund_value(value)
+      convert_value(value)
+    end
```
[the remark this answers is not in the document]
rgarner: (I assume it's a bit of metaprogramming calling it again)

dragon-dxw:
image
When a refund doesn't have a comment, the error bubbles all the way up.

app/services/import_actuals.rb
```diff
@@ -88,8 +89,12 @@ def create_actual
       attrs = @converter.to_h
       assign_default_values(attrs)
 
-      creator = CreateActual.new(activity: @activity, report: @report)
-      result = creator.call(attributes: attrs)
+      creator = if @converter.actual_class == Refund
+        CreateRefund.new(activity: @activity, report: @report, user: @uploader)
```
rgarner, line 93: Not directly related to this line, but now we're creating refunds, does the code/naming of the method on [L58](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/f914dc35b7dd7434ea6358f9a3c1de1b7d2a1761/app/services/import_actuals.rb#L58) still make sense?

app/services/import_actuals.rb
```diff
@@ -88,8 +89,12 @@ def create_actual
...
+      else
+        CreateActual.new(activity: @activity, report: @report)
```
rgarner, line 95: Just noticed that `CreateActual` can take a `user:` but is `nil` by default (hence the difference in params on L95 and L93). Is there a reason we've never set it on actuals before?

app/services/import_actuals.rb
```diff
@@ -129,6 +137,15 @@ def initialize(row)
       @activity = @attributes.delete(:activity)
     end
 
+    # We must have precisely one value, and we will check if
+    # the other is empty when we validate the value it contains
+    def actual_class
+      return Actual if @row["Refund Value"].nil?
+      return Refund if @row["Actual Value"].nil? || @row["Actual Value"] == "0.00"
+      @errors[:value] = [@row["Actual Value"], I18n.t("importer.errors.actual.non_numeric_value")]
+      @errors[:refund_value] = [@row["Refund Value"], I18n.t("importer.errors.actual.non_numeric_value")]
```
rgarner, line 146: Busy wondering if we'll get away with one of these, perhaps with a virtual attribute name like `@errors[:refund_or_actual_value]`?

app/services/import_actuals.rb
dragon-dxw: `zero_value?` does feel like a more reasonable name at this point

app/services/import_actuals.rb
dragon-dxw: No clue. I should probably set it?

app/services/import_actuals.rb
dragon-dxw: Might need a little help understanding the virtual attribute name, but sure!

rgarner: I mean: the attribute doesn't "exist" as such, it's just a placeholder meaning "these two fields". These aren't real ActiveModel errors, they're just our way of communicating a thing. At present we communicate "two of these things are wrong" for our upload errors UI (so two errors for the given row) where we might be better served by communicating "you only needed to supply one of these" as a single "error" applying to a "field" `:refund_or_actual_value` (which doesn't exist but would appear in the errors list as "Refund or actual value").

I don't honestly know if I've made it worse with this "explanation" 😆

spec/services/create_refund_spec.rb
```diff
@@ -68,6 +68,20 @@
...
+            financial_quarter: 1,
+            financial_year: 2020
+          }
+        end
+
+        it "has an error" do
+          expect(subject.object.errors.messages.values.to_s).to include("Enter a comment describing the need for the refund")
+        end
+      end
+
```
rgarner, line 84: I guess we want to squash this commit into the previous WIP?

app/services/import_actuals.rb
rgarner: Possibly yes. We should ask the question more widely I think.

app/services/import_actuals.rb
dragon-dxw: No, that's clear, and sounds good.

spec/features/staff/users_can_upload_refunds_spec.rb
```diff
@@ -66,6 +71,21 @@ def upload_csv(content)
...
   end
 
+  scenario "Upload a valid refund with a zero actual" do
+    click_link t("page_content.actuals.button.upload")
+    id = project.roda_identifier
+
+    # When I upload some refunds with comments
+    upload_csv <<~CSV
+      Activity RODA Identifier | Financial Quarter | Financial Year | Actual Value | Refund Value | Receiving Organisation Name | Receiving Organisation Type | Receiving Organisation IATI Reference | Comment
+      #{id}                    | 1                 | 2020           | 0.00         | 1234.56      | Example University          | 80                          |                                       | Ocelot
```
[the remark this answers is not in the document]
rgarner: Note to future readers: `CreateRefund` always `abs`'d them; a positive is equivalent to a negative in this column

spec/services/import_actuals_spec.rb
```diff
@@ -188,9 +213,12 @@
       end
     end
 
-    context "when the Value is blank" do
+    context "when the Actual Value and the Refund Value are blank" do
```
dragon-dxw, line 216: This test is misleading: the higher-level behaviour is that if these are blank, the row is skipped entirely.

spec/services/create_refund_spec.rb
dragon-dxw: Confused as to which commits you'd like squashing, but very happy to tidy up the commits once we're ready to go. (I've broadly left them as-is because losing the context for the changes is unhelpful right now, but yes, good to have tidy commits in future.)

rgarner: You've already squashed them! Resolving :)

app/services/import_actuals.rb
```diff
@@ -129,6 +137,15 @@ def initialize(row)
       @activity = @attributes.delete(:activity)
     end
 
+    # We must have precisely one value, and we will check if
+    # the other is empty when we validate the value it contains
+    def actual_class
+      return Actual if @row["Refund Value"].nil?
+      return Refund if @row["Actual Value"].nil? || @row["Actual Value"] == "0.00"
+      @errors[:value] = [@row["Actual Value"], I18n.t("importer.errors.actual.non_numeric_value")]
+      # @errors[:refund_value] = [@row["Refund Value"], I18n.t("importer.errors.actual.non_numeric_value")]
```
rgarner, line 146: Deletion?

app/services/import_actuals.rb
```diff
@@ -176,6 +198,10 @@ def convert_value(value)
       raise I18n.t("importer.errors.actual.non_numeric_value")
     end
 
+    def convert_refund_value(value)
```
CristinaRO, line 201: When we have converter methods that need to exist for the sake of metaprogramming, we tend to define just the one method and alias the other to it (example: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/app/services/activities/import_from_csv.rb#L392)

app/services/import_actuals.rb
```diff
@@ -88,8 +89,12 @@ def create_actual
...
       assign_default_values(attrs)
 
-      creator = CreateActual.new(activity: @activity, report: @report)
-      result = creator.call(attributes: attrs)
+      creator = if @converter.actual_class == Refund
+        CreateRefund.new(activity: @activity, report: @report, user: @uploader)
+      else
+        CreateActual.new(activity: @activity, report: @report, user: @uploader)
+      end
+      result = creator.call(attributes: attrs.except(:refund_value))
```
CristinaRO, line 97: I'm not following this very well: why are we discarding `:refund_value` at this point?

app/services/import_actuals.rb
CristinaRO, line 146: Does this need to be an error message in its own right, or can we delete the commented line?

CristinaRO (commented):
Partial review, I haven't reviewed the specs yet!

app/services/import_actuals.rb
dragon-dxw: :+1:

app/services/import_actuals.rb
dragon-dxw: :+1:

app/services/import_actuals.rb
dragon-dxw: Value is somewhat overloaded, as both the symbol of the Actual Value from the spreadsheet and the value of a field in eithe an Actual or Refund object. Giving these different names is a good idea.

app/services/create_refund.rb
```diff
@@ -13,13 +13,13 @@ def call(attributes:)
...
     assign_refund_and_comment(attributes)
-    refund.save
+    save_value = refund.save
 
     raise Error, validation_error_message(refund) if refund.errors.any?
 
     record_historical_event(refund)
 
-    refund
+    Result.new(save_value, refund)
```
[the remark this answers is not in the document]
dragon-dxw: Renamed and made record_historical_event conditional

dragon-dxw:
Because it might be hard to see the changes based on @CristinaRO 's comments once I've `git absorb`ed:

```
diff --git a/app/services/create_refund.rb b/app/services/create_refund.rb
index 4ff9671b..deca801e 100644
--- a/app/services/create_refund.rb
+++ b/app/services/create_refund.rb
@@ -13,10 +13,10 @@ class CreateRefund
     refund.parent_activity = activity
     refund.report = editable_report_for(activity)
     assign_refund_and_comment(attributes)
-    save_value = refund.save
-    record_historical_event(refund)
+    save_successful = refund.save
+    record_historical_event(refund) if save_successful
 
-    Result.new(save_value, refund)
+    Result.new(save_successful, refund)
   end
 
   private
diff --git a/app/services/import_actuals.rb b/app/services/import_actuals.rb
index 6428f75b..faf8ec92 100644
--- a/app/services/import_actuals.rb
+++ b/app/services/import_actuals.rb
@@ -143,7 +143,6 @@ class ImportActuals
       return Actual if @row["Refund Value"].nil?
       return Refund if @row["Actual Value"].nil? || @row["Actual Value"] == "0.00"
       @errors[:value] = [@row["Actual Value"], I18n.t("importer.errors.actual.non_numeric_value")]
-      # @errors[:refund_value] = [@row["Refund Value"], I18n.t("importer.errors.actual.non_numeric_value")]
     end
 
     def raw(attr_name)
@@ -198,9 +197,7 @@ class ImportActuals
       raise I18n.t("importer.errors.actual.non_numeric_value")
     end
 
-    def convert_refund_value(value)
-      convert_value(value)
-    end
+    alias_method :convert_refund_value, :convert_value
 
     def convert_receiving_organisation_type(type)
       validate_from_codelist(
```

app/services/import_actuals.rb
dragon-dxw: Changed a great many of the 'value' variable names to be clearer, explicitly making `:value` (the value of either the Refund or Actual) either `:actual_value` or `:refund_value` when referring to the initial spreadsheet, and only `:value` when it is both those concepts successfully combined (and the only difference is whether it's going into an Actual or a Refund). This is in a separate commit.

rgarner (approved):
Thank you for your patience getting this to a point, @dragon-dxw 🙏

PR #1616, opened 2022-03-11
(2438) Add and edit comments on actuals via UI

CristinaRO:
## Changes in this PR
- Users can add and edit comments on actuals through the form

### After
Screenshot 2022-03-11 at 17 24 38

## Next steps


CristinaRO:
[Edit actuals comments in website, not CSV](https://trello.com/c/PwtfPxf8/2438-edit-actuals-comments-in-website-not-csv)

dragon-dxw:
I'm currently doing some work on comments in create actuals, specifically on the interface of whether we expect Comments or Strings to be passed to it. I'm moving from Comments to Strings, this will impact this work (perhaps in a positive way).

rgarner:
@CristinaRO in answer to the question:

> Do we need more specs? For example create_actual_spec.rb and update_actual_spec.rb (We do have feature specs)

It looks like we've only changed the consistency/readability of these classes, so the existing specs appear to cover them

app/policies/actual_policy.rb
```diff
@@ -23,9 +20,15 @@ def destroy?
     can_update_or_delete?
   end
 
+  def create_comment?
+    return false if beis_user?
+
+    editable_report_for_organisation_and_fund.present?
+  end
```
rgarner, line 27: It doesn't look like we've specc'd this in `actual_policy_spec`?

app/controllers/staff/actuals_controller.rb
```diff
@@ -22,45 +24,45 @@ def new
...
-    if result.success?
-      flash[:notice] = t("action.actual.create.success")
-      redirect_to organisation_activity_path(@activity.organisation, @activity)
-    else
-      render :new
-    end
+    return render :new unless actual_created
+
+    flash[:notice] = t("action.actual.create.success")
+    redirect_to organisation_activity_path(@activity.organisation, @activity)
```
james, line 38: Using a `return unless` here reads funny to me. Why not:

```ruby
if actual_created
  flash[:notice] = t("action.actual.create.success")
  redirect_to organisation_activity_path(@activity.organisation, @activity)
else
  render :new
end

File: app/controllers/staff/actuals_controller.rb [hunk 3]
```diff
@@ -22,45 +24,45 @@ def new
...
-    authorize @actual
-
     @activity = Activity.find(activity_id)
+    @actual = ActualForm.new(attributes_for_editing)
+
+    authorize(@actual, policy_class: ActualPolicy)
 
     prepare_default_activity_trail(@activity)
-    add_breadcrumb t("breadcrumb.actual.edit"), edit_activity_actual_path(@activity, @actual)
+    add_breadcrumb t("breadcrumb.actual.edit"), edit_activity_actual_path(@activity, @actual.id)
```
james, line 48: why do we need to add .id here?

app/services/create_actual.rb
```diff
@@ -11,8 +9,14 @@ def initialize(activity:, report: nil, user: nil)
   def call(attributes: {})
     actual.parent_activity = activity
 
-    body = attributes.delete(:comment)
-    actual.build_comment(body: body, commentable: actual, report: @report) if body.present?
+    comment_body = attributes.delete(:comment).to_s.strip
+    if comment_body.present?
+      actual.build_comment(
+        body: comment_body,
+        commentable: actual,
```
james, line 16: do we need to specify `commentable: actual` if we're also doing `actual.build_comment`

app/helpers/activity_helper.rb
```diff
@@ -47,4 +47,13 @@ def benefitting_countries_with_percentages(benefitting_countries)
       )
     end
   end
+
+  def edit_comment_path_for(commentable, comment)
+    case commentable.class.name
+      when "Activity"
+        edit_activity_comment_path(commentable, comment)
+      when "Actual"
```
dragon-dxw, line 55: This feels like it's missing other commentables that might be relevant (e.g. Refunds). Activity is special (as something that is it's own activity) so I think this should be Activity vs. Everything Else.

james (approved):
A few non-blocking small questions, but nothing blocking. Looks good to me!

app/controllers/staff/actuals_controller.rb
CristinaRO: Because `@actual` isn't an instance of `Actual`, it's an `ActualForm`.

app/controllers/staff/actuals_controller.rb
CristinaRO: The succinctness seemed to help the readability of the controller, if it doesn't, we can expand it like you suggested.

app/services/create_actual.rb
CristinaRO: It looks like we do, when the actual itself isn't persisted yet, so the `commentable_id` doesn't end up `nil`.

app/helpers/activity_helper.rb
CristinaRO: They can't edit refund and adjustment comments from the activity view right now, so we thought it would be premature featurisation.

app/controllers/staff/actuals_controller.rb
james: 🙄  ok!

app/controllers/staff/actuals_controller.rb
james: I personally find the "defending return" (there's probably a proper name for it) is better if you're going to be wrapping a lot of LOC or complicated logic inside the `else`, but I find `if... else` easier to read and follow if its just a couple of lines of code like here.

app/services/create_actual.rb
james: 😣  ok

app/helpers/activity_helper.rb
CristinaRO: For example, the decision to allow editing comments on actuals by editing the entire actual was just that, a decision, which I wouldn't want to impose on all other similarly shaped objects. Edit paths for comments on refunds might end up linking to the comment rather than the actual, and I don't want to make that decision now.

dragon-dxw: :+1:

app/controllers/staff/actuals_controller.rb
```diff
@@ -22,45 +24,45 @@ def new
   def create
     @activity = activity
     authorize @activity
+    @actual = ActualForm.new(actual_params.merge({parent_activity: @activity}))
 
-    result = CreateActual.new(activity: @activity)
-      .call(attributes: actual_params)
-    @actual = result.object
```
rgarner, line 28: It looks like we might have changed the semantics with this deletion. Prior to the deletion, we were sending an activerecord object (unpersisted) to `render :new`, but now we're returning an `ActualForm`. I don't _think_ this has a material effect but we can note it.

app/controllers/staff/actuals_controller.rb
```diff
@@ -22,45 +24,45 @@ def new
...
-      redirect_to organisation_activity_path(@activity.organisation, @activity)
-    else
-      render :edit
-    end
+    ).call(attributes: actual_params).success?
+
+    return render :edit unless actual_updated
+
+    flash[:notice] = t("action.actual.update.success")
+    redirect_to organisation_activity_path(@activity.organisation, @activity)
```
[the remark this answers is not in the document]
rgarner: Acknowledge the "nothing blocking", and also acknowledge that we've just merged #1646 with the `return` style for better or worse given our PR backlog and GitHub downtime. Will accept PR backing these out if it's an issue!

app/controllers/staff/actuals_controller.rb
rgarner: (i.e. not a blocker but not going to resolve until seen 😄 )

rgarner: Narrator: they could not

PR #1617, opened 2022-03-14
Replace i18n references with actual text

james:
We've been finding it quite hard to understand what a test is actually trying to do due to the indirection created by using the translation.

While this may make the tests a bit more brittle, the cost of having to fix a spec if some copy changes is considerably smaller than a developer struggling to understand what a spec is testing.

### DIFF
--- a/spec/features/public/visitors/home_page_spec.rb
+++ b/spec/features/public/visitors/home_page_spec.rb
@@ -1,7 +1,7 @@
 RSpec.feature "Home page" do
   scenario "visit the home page" do
     visit root_path
-    expect(page).to have_content(t("app.title"))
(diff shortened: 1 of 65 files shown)


dragon-dxw:
I've been using a Visual Studio plugin called "Rails i18n" by Shane Hofstetter (v 0.4.0) which is significantly reducing the cost for me; I appreciate that's not an answer for everyone.

dragon-dxw (approved):
A quick glance suggest this is all fine, but haven't looked at in depth.

CristinaRO:
There are now conflicts with some new specs from the devise branch.

james:
I'll redo this after the backlog of other PRs have been merged

PR #1618, opened 2022-03-14
(2447) On password change, redirect to login

CristinaRO:
## Changes in this PR
- After a successful password change, redirect the user to log in again


CristinaRO:
[BUG: After resetting password, user does not join the SMS/OTP flow for logging in](https://trello.com/c/mvqCCPcY/2447-bug-after-resetting-password-user-does-not-join-the-sms-otp-flow-for-logging-in)

james (approved):
Nice!

PR #1625, opened 2022-03-17
rake sessions:delete_all -- delete sessions

dragon-dxw:
We needed to delete all active sessions when migrating from Auth0 / Omniauth to
Devise to avoid "OmniAuth: name not found" error on unmarshalling Redis
sessions

Note: this gave the following deprecation warning:
Passing 'select' command to redis as is; administrative commands cannot be
effectively namespaced and should be called on the redis connection directly;
passthrough has been deprecated and will be removed in redis-namespace 2.0

### Before
image

### After
It works

### DIFF
--- a/lib/tasks/sessions.rake
+++ b/lib/tasks/sessions.rake
@@ -1,0 +1,25 @@
+namespace :sessions do
+  desc "delete all sessions from REDIS"
+  task delete_all: :environment do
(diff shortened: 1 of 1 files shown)


rgarner:
[Fix "poison session" problem](https://trello.com/c/c5BQjvqt/2459-fix-poison-session-problem)

PR #1629, opened 2022-03-21
(2460) Catch GOV.UK mobile phone validation errors

CristinaRO:
## Changes in this PR
- Prevent server errors when a mobile number is invalid

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1000,6 +1000,7 @@
 
 - Add 'Other ODA' fund
 - Change the HTML title tag of the Users page to be Users, not Home
+- Prevent server errors when a mobile number is invalid
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-100...HEAD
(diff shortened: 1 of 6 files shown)


james (approved):
🎉

CristinaRO:
> Took me a while to get the logic of where the error was coming in and going out, but looks good.

What kind of detail in the commit message would make the logic clearer @dragon-dxw ? We can still amend the commit message, and I'd rather leave good enough breadcrumbs for future developers!

dragon-dxw:
> > Took me a while to get the logic of where the error was coming in and going out, but looks good.
> 
> What kind of detail in the commit message would make the logic clearer @dragon-dxw ? We can still amend the commit message, and I'd rather leave good enough breadcrumbs for future developers!

[caveat: most of this was me not being too familiar]
I had mistaken the bit that was getting a nice clean message and the bit that was mangling the notify string, somehow. I don't think there's anything worth changing or even elaborating on.

PR #1630, opened 2022-03-21
Title on Users index page no longer Home

dragon-dxw:
A truly minor bug that I discovered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -999,6 +999,7 @@
 ## [unreleased] ...
 
 - Add 'Other ODA' fund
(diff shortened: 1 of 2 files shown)


rgarner:
Cripes.

rgarner (approved):

rgarner:
[BUG: Make title on Users page "Users". Instead of "Home"](https://trello.com/c/CYMKC4MI/2464-bug-make-title-on-users-page-users-instead-of-home)

rgarner:
Technically, requires a CHANGELOG entry :pedant: 🚷

PR #1632, opened 2022-03-22
Remove UserNotAuthorised paraphernalia

dragon-dxw:
Context for change, from
https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1599#discussion_r818889659

Because we've removed the only thing that could raise `UserNotAuthorised` and the only usage of it is on L10 of `app/controllers/concerns/auth.rb`, we should remove the class and also the branch of the if on L11 that does not respond to `:policy`

We also update the text one spec expected to see.

@rgarner: The card implies that there's more to do, but I couldn't find a loose thread to pull, so I expect this isn't fully done.

### DIFF
--- a/app/controllers/concerns/auth.rb
+++ b/app/controllers/concerns/auth.rb
@@ -1,5 +1,3 @@
-class UserNotAuthorised < StandardError; end
-
 module Auth
   extend ActiveSupport::Concern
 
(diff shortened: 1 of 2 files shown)


dragon-dxw:
[Remove UserNotAuthorised paraphernalia](https://trello.com/c/fODkynU1/2452-remove-usernotauthorised-paraphernalia)

rgarner:
Hm. I wonder if it's possible that my Many Failing Specs at the time were down to a branch/database mismatch rather than anything related to this. You've looked at it more than I have!

rgarner (approved):

PR #1646, opened 2022-03-28
Make refund form validate numeric values

rgarner:
## Changes in this PR

This validation wasn't being copied up from the object, and so `CreateRefund`
was silently failing. Copy the validation up to fix this error.

In addition, make sure further mismatches between RefundForm and
CreateRefund can't silently fail by checking the return value of
CreateRefund#call.

Co-authored-by: Cristina Rotaru <cristina@dxw.com>

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1014,6 +1014,7 @@
 activity and on its child transactions (which can be actuals, refunds, and adjustments)
 - Remove Auth0 user identifier from user management
 - Add sector and sector category codes to radio button text when adding a new activity
+- Fix an error where a new refund creation would silently fail when the value was non-numeric
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-102...HEAD
 [release-102]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-101...release-102
--- a/app/controllers/staff/refunds_controller.rb
+++ b/app/controllers/staff/refunds_controller.rb
@@ -20,12 +20,12 @@ class Staff::RefundsController < Staff::ActivitiesController
     authorize @activity
     @refund = RefundForm.new(refund_params)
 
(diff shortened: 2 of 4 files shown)


rgarner:
[RefundForm doesn't pick up non-numerical amount](https://trello.com/c/fq6agKjg/2446-refundform-doesnt-pick-up-non-numerical-amount)

app/controllers/staff/refunds_controller.rb
```diff
@@ -20,12 +20,12 @@ def create
     authorize @activity
     @refund = RefundForm.new(refund_params)
 
-    return render :new unless @refund.valid?
-
-    CreateRefund.new(
+    refund_created = @refund.valid? && CreateRefund.new(
```
CristinaRO, line 23: How do you feel about applying the same succinct pattern to the `update` method in this same controller?

rgarner: I was busy thinking that we should become un-succinct in the same manner as `CreateActual` for consistency...

CristinaRO: Too late, I already changed the actuals controller to be more succinct, except James suggested some changes to the succinctness that will be a third way to which we can all converge. ;)

rgarner: Oh, well, alright then.

CHANGELOG.md
```diff
@@ -1014,6 +1014,7 @@
 activity and on its child transactions (which can be actuals, refunds, and adjustments)
 - Remove Auth0 user identifier from user management
 - Add sector and sector category codes to radio button text when adding a new activity
+- Fix an error where a new refund creation would silently fail when the value was non-numeric
```
CristinaRO, line 1017: Updating would also silently fail if you tried to change the value to something non-numeric. (Also fixed by adding validation to the form object.)

CristinaRO (approved):
Minor nitpick, not blocking.

PR #1647, opened 2022-03-28
Remove users.identifier / mashed potato from our hair*

rgarner:
## Changes in this PR

The last legacy of Auth0. Also remove:

- a mock still allowing it
- factory still setting it
- user view still displaying it
- a translation serving the view
- a spec still using one

\* Thanks @CristinaRO 
 
### Before

Imagine the identifier field, on the users show page

### After

Now imagine it gone

## Next steps


rgarner:
[Remove Auth0 users.identifier DB column](https://trello.com/c/bCAfPEib/2471-remove-auth0-usersidentifier-db-column)

CristinaRO (approved):
👍🏻 💆

PR #1649, opened 2022-03-29
Fix attempt to use an unloaded gem

CristinaRO:
## Changes in this PR

- The gem `bullet` is only bundled for test and development environments,
so when skipping it we first need to make sure it’s defined.

## Next steps


rgarner:
Hotfix, no CHANGELOG

rgarner (approved):

PR #1651, opened 2022-03-29
Delete duplicate organisations

dragon-dxw:
We deleted a number of organisations at BEIS' request. They had no
matched_efforts, reports, activities, users, or budgets referring to them
but did have org_participations. We deleted those.

There is an unused SQL query in the migration which we used to verify the above
and the deletion afterwards, left in the hopes that it might be useful.

A number of other organisations were already deleted.

See Zendesk ticket 16066.

## Next steps


db/data/20220329174300_delete_duplicate_organisations_2.rb
```diff
@@ -0,0 +1,44 @@
+query = <<~SQL
```
CristinaRO, line 1:
```
db/data/20220329174300_delete_duplicate_organisations_2.rb:1:1: Lint/UselessAssignment: Useless assignment to variable - `query`
```

Might get away with it by naming it `_query`?

CristinaRO (approved):

PR #1657, opened 2022-03-30
(2475) Fix some accessibility issues with tables and tab lists

CristinaRO:
## Changes in this PR
- Ensure table cells are not empty for screenreaders
- Remove invalid “aria-controls” attributes
- Add missing tablist role to tab lists
- Add landmarks to page contents that wasn't part of any landmarks

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1024,6 +1024,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 ## [unreleased]
 
 - Validate XML against IATI schema on export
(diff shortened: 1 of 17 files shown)


CristinaRO:
[Fix newly introduced accessibility issues](https://trello.com/c/x3mFfT7w/2475-fix-newly-introduced-accessibility-issues)

dragon-dxw (approved):
This all feels like really good work. There's more to do (isn't there always) but this, here, is good.

PR #1659, opened 2022-03-31
Hotfix puma

CristinaRO:
## Changes in this PR

Update puma to address https://github.com/advisories/GHSA-h99w-9q5r-gjq9

While the vulnerability can only be exploited under specific conditions, I've opened this PR in case we think it's better safe than sorry, to be able to expedite releasing it.


dragon-dxw (approved):
:+1:

PR #1665, opened 2022-04-04
Fix: fragile tests breaking on change of financial year

edavey:
Fix two fragile tests which broke over the weekend, when the financial year changed.


dragon-dxw (approved):
:+1:

PR #1673, opened 2022-04-06
Fix: tweak S3UploaderConfig to match Terraform naming

edavey:
We originally planned to have the buckets named:

`s3-export-download-bucket-{environment}`

but the Terraform convention which we've followed (see
`terraform/s3-export-download-bucket.tf`) gives us:

`beis-roda-{environment}-s3-export-download-bucket`

In this commit we relax the "matcher" regex in the
`S3UploaderConfig` helper and update our sample env
dotfiles.


shuldt (approved):
yup looks good

PR #1674, opened 2022-04-07
Fix: Include previous year in drop-down when posting external income

edavey:
On April 1st 2022 (financial year 2022-2023) I may be
needing to post external income for the period Q4 2021-2022.

Previously we were offering only the ten years starting from
the "currrent" financial year, i.e. 2022-2023. We now
include the previous year also.

## Screenshots of UI changes

years

## Next steps

- [  ] Identify other places suffering from the same problem. Where the "previous" financial year also needs to be displayed!


edavey:
See https://dxw.zendesk.com/agent/tickets/16151 for info on other similar problems to be addressed. But I think we ought to get this fix deployed.

shuldt (approved):
lgtm

PR #1687, opened 2022-05-04
Replace DB in 'training' env with copy from 'prod'

edavey:
This script is designed to update the 'training' environment
with a copy of the production data ('prod').

User credentials are scrubbed, requiring users to reset their
passwords after each 'sync'.

### HOW TO USE IT
It can be run from a developer or ops person's laptop
provided the user has:

- a GPaaS (https://docs.cloud.service.gov.uk) account for BEIS RODA
- the Cloudfoundry tools installed
  (https://docs.cloud.service.gov.uk/get_started.html#set-up-the-cloud-foundry-command-line)
- installed the Conduit plugin for Cloudfoundry/GPaaS
  (https://docs.cloud.service.gov.uk/guidance.html#using-the-conduit-plugin)

Environment variables expected:

- `GPAAS_CF_USER`
- `GPAAS_CF_PASSWORD`

Set these in your `.env` file or export to your shell
manually.

Run the script using the `bin/rails runner` command so that
the required libraries such as `Date`, `Open3` and
`FileUtils` are loaded:

`bin/rails runner script/training_db_sync.rb`

### WHAT IT DOES
See the `TrainingDbSync#call` method for a description of the
script's steps:

- print_plan
- capture_data_from_source
- copy_data_to_destination
- load_source_data_to_destination
- force_password_reset_for_users
- remove_temp_files

Please note a couple of details of the data dump and
data load steps which were found to work with the GPaaS
setup:

- the data is dumped from `prod` using `pg-dump --clean`: this
includes instructions in the dump file for db objects to be
dropped before importing new items

- the data dump is applied to `training` using
`psql < dumpfile` rather than using `pg_restore`.

### FUTURE PLANS
It is our intention to develop this script in the future to
perform the data sync automatically on a scheduled basis.

### DIFF
--- a/script/training_db_sync.rb
+++ b/script/training_db_sync.rb
@@ -1,0 +1,142 @@
+# Script to update the 'training/pentest' environment's DB
+# from a copy taken from 'prod'.
+#
+# Run with: `bin/rails runner script/training_db_sync.rb`
+
+class TrainingDbSync
+  attr_reader :source, :destination, :data_filename, :pw_reset_script
+
+  ORGANISATION_NAME = "beis-report-official-development-assistance"
+
+  def initialize
+    @source = :prod
+    @destination = :pentest # this is the old name for 'training'
+    @data_filename = "/tmp/#{source}-dump-#{timestamp}.dump"
+    @pw_reset_script = "/tmp/pw_reset_script.rb"
+  end
+
+  def call
+    print_plan
+    capture_data_from_source
+    copy_data_to_destination
+    load_source_data_to_destination
+    force_password_reset_for_users
+    remove_temp_files
+  end
+
+  private
+
+  def authenticate_to(env)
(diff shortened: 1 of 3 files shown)


script/training_db_sync.rb
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  def destination_service
+    "beis-roda-#{destination}-postgres"
+  end
+
+  def print_plan
+    plan =
+      "The *#{destination}* db will be dropped and replaced by a copy from *#{source}*. " \
+      "Users will need to reset their passwords on their next login to *#{destination}*."
```
dragon-dxw, line 49: consider adding: "and they will need to re-register their mobile phone" or however the 2FA is impacted.

dragon-dxw (approved):

PR #1696, opened 2022-06-13
Fix: 16160 - Bulk update of activities including implementing organisations

edavey:
## Summary

This PR fixes two bugs in the handling of implementing organisations when updating activities using the bulk upload method:

1. the bulk upload CSV template for activities was missing a column for describing each activity's implementing organisations. This resulted in any pre-existing associations being removed when an activity was updated in this way. We now have a `Implementing organisation names` column which takes a pipe-delimited list of organisation names
2. the update code (in `Activity::ImportFromCsv`) did not handle an activity being associated with multiple implementing organisations. 

## Details

This work results from a [bug report 16160](https://dxw.zendesk.com/agent/tickets/16160) showing that whenever an
activity with a pre-existing implementing organisation
association is updated via the bulk upload CSV mechanism,
the activity was stripped of its implementing organisations.

However, it makes sense that any pre-existing implementing organisation associations
which are not present in an update should be removed when the update is
applied. e.g.:

**pre-existing state:**
implementing orgs = [Org A, Org B]

**desired post-update state:**
implementing orgs = [Org A, Org C]

In the bulk upload CSV for the update, in column "Implementing organisation names",
I should expect to find: "Org A | Org C".

i.e. any implementing orgs not present in the update will be removed and any orgs
which are not present on the activity will be added.

### The fix

Multiple implementing orgs are now supplied in pipe-delimited
form, e.g. under the column "Implementing organisation names"
I can list:

```
Existing Org | Additional Org 1 | Additional Org 2
```

(NB: whitespace around the pipe is not signficant)

If one or more of the given organisation names does not
match up with a pre-defined organisation (that organisation
has not been set up in RODA) then I will be shown an error
message identifying the problem organisation(s), e.g.

"Additional Org 1 | Additional Org 2 is/are not a known implementing organisation(s)"

Error handling when 2 unknown implementing organisations are supplied in the CSV:

error_handling

### We will need guidance for delivery partners

We need *guidance* to let DPs know that failure to provide this
implementing organisation information will be destructive,
i.e. if they don't re-supply all the implementing orgs which
should be associated with an activity, then any associations
will be removed as part of the update.

## Next steps

- [x ] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1052,6 +1052,9 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 ## [release-110] 2022-06-21
 - Do not parse arbitary text in oda_eligibilities or programme_status as being 0.
 
+## [unreleased]
+- Fix handling of implementing organisations when updating activities through CSV upload
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-110...HEAD
 [release-110]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-109...release-110
 [release-109]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-108...release-109
--- a/app/services/activities/import_from_csv.rb
+++ b/app/services/activities/import_from_csv.rb
@@ -21,7 +21,7 @@ module Activities
         Converter::FIELDS.values -
         ["BEIS ID"] +
         ["Comments"] +
-        ["Implementing organisation name"]
+        ["Implementing organisation names"]
     end
 
     def initialize(uploader:, delivery_partner_organisation:, report:)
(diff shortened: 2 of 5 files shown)


dragon-dxw:
Does the CSV template pre-populate the pipe-separated implementing organisations? Can't see any code that looks like that, which would go a long way towards being less destructive (unless deliberate)...

It's possible I've missed it.

edavey:
> Does the CSV template pre-populate the pipe-separated implementing organisations? Can't see any code that looks like that, which would go a long way towards being less destructive (unless deliberate)...
> 
> It's possible I've missed it.

thanks @dragon-dxw good point! 

I wasn't aware that the Activities bulk upload CSV "template" which is offered as a download could contain activity rows as well as an initial headers row. But looking at the code it's plain that "reportable activities" are included:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/1e382c87d9af0d05a81fc9920ce149869808619f/app/controllers/staff/actual_uploads_controller.rb#L19-L30

I'll take another run at this. Thanks!

dragon-dxw (changes requested):
As per the comments, the CSV template should be pre-populated with the pipe-separated implementing organisations. Otherwise very happy with everything else.

dragon-dxw (approved):
I see this commit as resolving half the problem -- this enables users to bulk update and not erase all implementing orgs for an activity, if they type in the data manually, which is a significant step up from always erasing all implementing orgs for an activity.

It's not complete -- a complete handling would provide the implementing orgs in the activities template -- but it is a marked improvement on the current situation and provides excellent groundwork for a later PR which just populates that field in the template.

PR #1712, opened 2022-08-08
Don't setup backend for `terraform_test` script

Gweaton:
## Changes in this PR

Passes `-backend=false` to `terraform_test`'s `terraform init` command, to ensure we don't try and configure a backend for it.

This is currently failing on CI due to it trying to configure S3. Passing this flag should allow Dependabot to be able to run CI without us needing to set its own secrets e.g. CF username and password


CristinaRO (approved):
🤞🏻 🎉

PR #1727, opened 2022-08-11
(2509) Add banner to non production envs

CristinaRO:
## Changes in this PR
- Display a banner on non-production environments to make it clearer to users which site they're using

### After
Screenshot 2022-08-16 at 15 38 08

## Next steps

### DIFF
--- a/app/assets/stylesheets/partials/_alerts.css.scss
+++ b/app/assets/stylesheets/partials/_alerts.css.scss
@@ -22,3 +22,26 @@ ul.govuk-list.govuk-error-summary__list li {
   @include govuk-font($size: 19, $weight: bold)
   color:  $govuk-error-colour;
 }
+
+.environment-info-wrapper {
+  @include govuk-font($size: 19, $weight: normal);
+  width: 2em;
+  height: 1000px;
+  position: fixed;
+  left: 200px;
+  top: -25px;
+  background-color: orange;
+  transform: rotate(50deg);
+  transform-origin: top;
+  z-index: 998;
+  border-color: orangered;
+  border-width: 1px;
+  border-style: dotted;
+}
+
+.environment-info-content {
+  transform: rotate(-91deg);
+  position: relative;
+  bottom: -8em;
+  z-index: 999;
+}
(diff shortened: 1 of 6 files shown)


CristinaRO:
[Add banner to make it obvious that users are using "training" environment](https://trello.com/c/xnRsOC5c/2509-add-banner-to-make-it-obvious-that-users-are-using-training-environment)

app/assets/stylesheets/partials/_alerts.css.scss
```diff
@@ -22,3 +22,25 @@ ul.govuk-list.govuk-error-summary__list li {
   @include govuk-font($size: 19, $weight: bold)
   color:  $govuk-error-colour;
 }
+
+.environment-info-wrapper {
```
Gweaton, line 26: Shall we make the banner's font match the rest of the site here? Can do that with:

`@include govuk-font($size: 19, $weight: normal);`

Gweaton (approved):
Nice one!

This is totally a matter of personal style, but I wonder if it might be less intrusive (but still obvious) on the right hand side of the screen rather than the left hand side. Would that be a real pain to mock up?

app/assets/stylesheets/partials/_alerts.css.scss
```diff
@@ -22,3 +22,25 @@ ul.govuk-list.govuk-error-summary__list li {
...
+  width: 2em;
+  height: 1000px;
+  position: fixed;
+  left: 200px;
```
Gweaton, line 30: Here's how I made it go to the right, in case it's helpful:
```
  right: -32px;
  top: 150px;
  transform: rotate(130deg);
```

Gweaton: They're happy with the left banner.

app/assets/stylesheets/partials/_alerts.css.scss
Gweaton: Are you happy to make this change?

Gweaton:
Is this something we can test with a Capybara feature spec? Would be nice to have a catch in there in case it disappears for any reason in future.

Gweaton (approved):
Thanks for adding tests 🚢

PR #1746, opened 2022-08-25
(2517) Fix misspelling of “FSTC” in codelist

CristinaRO:
## Changes in this PR
- Fix the misspelling of "FSTC" in a codelist and in the service using the codelist

FSTC stands for “Free-standing Technical Cooperation”.[1]

The misspelling in the codelist is consistent with the code that checks
for that key, and the activity attribute is correctly spelled. However,
the misspelling is a source of confusion, so we want to fix it.

[1] If we were to use the official acronym, via OECD:
https://stats.oecd.org/glossary/detail.asp?ID=6022
we would be switching all of them to “FTC”. However, I feel that would
be more disruptive than helpful at this point.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1072,7 +1072,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 ## [unreleased]
 
 - Display total of refunds in the report summary
-
+- Fix the misspelling of "FSTC" in a codelist and in the service using the codelist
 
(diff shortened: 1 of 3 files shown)


CristinaRO:
[FSTC versus FTSC spellings in the code](https://trello.com/c/aJG2tV6M/2517-fstc-versus-ftsc-spellings-in-the-code)

Gweaton (approved):

PR #1747, opened 2022-08-25
Don't create rows with no export data

dragon-dxw:
## Changes in this PR

The previous implementation was calling `forecast_rows.any?` on some
forecast rows with the following structure:

```
{ 'activity-id'=>[100] }
```

This meant that we were calling `any?` on a hash value which was always
returning `true`, in contrast to the `headers` which would return an
empty array.

This led to the number of headers being out of sync with the number of
rows, causing columns to be out of sync from the `variance` row.

Now we check all the arrays in that hash are empty - if they are, we
don't render the variance row.

For the tests, We've added an entire new spec because the existing setup sets up a forecast, and
there's no easy way we could see to remove it or set it up at a later stage. It's intended
to otherwise be a complete copy of the setup for the with-forecast case.

Resolves https://dxw.zendesk.com/agent/tickets/16785

## Before 
image

## After
image

## Next steps

### DIFF
--- a/app/services/export/report.rb
+++ b/app/services/export/report.rb
@@ -54,8 +54,8 @@ class Export::Report
       row << delivery_partner_organisation_rows.fetch(activity.id, nil)
       row << change_state_rows.fetch(activity.id, nil)
       row << actuals_rows.fetch(activity.id, nil) if actuals_rows.any?
-      row << variance_rows.fetch(activity.id, nil) if actuals_rows.any? && forecast_rows.any?
-      row << forecast_rows.fetch(activity.id, nil) if forecast_rows.any?
+      row << variance_rows.fetch(activity.id, nil) if actuals_rows.any? && has_forecast_rows?
+      row << forecast_rows.fetch(activity.id, nil) if has_forecast_rows?
       row << comment_rows.fetch(activity.id, nil)
       row << link_rows.fetch(activity.id, nil)
       row.flatten
@@ -95,6 +95,10 @@ class Export::Report
 
   def forecast_rows
     @_forecast_rows ||= @forecast_columns.rows
+  end
+
+  def has_forecast_rows?
+    forecast_rows.values.flatten.any?
   end
 
   def variance_rows
--- a/spec/services/export/report_spec.rb
+++ b/spec/services/export/report_spec.rb
@@ -67,6 +67,72 @@ RSpec.describe Export::Report do
     DatabaseCleaner.clean
   end
 
+  context "when there are activities but no forecasts" do
+    before do
+      financial_quarter = FinancialQuarter.for_date(Date.parse("31-Mar-2022"))
+      financial_year = FinancialYear.for_date(Date.parse("31-Mar-2022"))
+
+      @report_without_forecasts = create(
+        :report,
+        financial_quarter: financial_quarter.to_i,
+        financial_year: financial_year.to_i
+      )
+
+      @project_for_report_without_forecasts = create(
+        :project_activity_with_implementing_organisations,
+        implementing_organisations_count: 2
+      )
+
+      @actual_spend_for_report_without_forecasts =
+        create(
+          :actual,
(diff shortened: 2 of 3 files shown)


spec/services/export/report_spec.rb
```diff
@@ -286,3 +288,102 @@ def link_for_row(row)
...
+        financial_year: @report.financial_year
+      )
+
+    previous_report = create(
+      :report,
+      financial_quarter: financial_quarter.pred.to_i,
+      financial_year: financial_year.to_i
+    )
+
+    @forecast =
```
CristinaRO, line 344: This is just from a first pass over this PR and not an in-depth review: do we need to setup this `@forecast` at all?

app/services/export/report.rb
```diff
@@ -54,8 +54,9 @@ def rows
       row << delivery_partner_organisation_rows.fetch(activity.id, nil)
       row << change_state_rows.fetch(activity.id, nil)
       row << actuals_rows.fetch(activity.id, nil) if actuals_rows.any?
-      row << variance_rows.fetch(activity.id, nil) if actuals_rows.any? && forecast_rows.any?
-      row << forecast_rows.fetch(activity.id, nil) if forecast_rows.any?
+      # can't call forecast_rows.any? because something is a map -- even if row value in each is (), still creates it, gives it ID, attaches empty list.
```
Gweaton, line 57: We've sort of got the same comment here twice (again on line 102-3), and it's quite well captured in the commit message. I'm not sure we need either comment here as I think `has_forecast_rows` on line 101 is probably descriptive enough and hides implementation details, but do you think this is something visitors to the codebase particularly need to be aware of?

spec/services/export/report_spec.rb
```diff
@@ -286,3 +288,102 @@ def link_for_row(row)
...
+  context "when there are activities but no forecasts" do
+    subject { described_class.new(report: @report) }
+
+    before do
+      relation = Activity.where(level: ["project", "third_party_project"])
+      finder_double = double(Activity::ProjectsForReportFinder, call: relation)
+      allow(Activity::ProjectsForReportFinder).to receive(:new).and_return(finder_double)
+    end
+
+    describe "#headers" do
```
Gweaton, line 372: Do we want a test for `#rows` here too?

spec/services/export/report_spec.rb
```diff
@@ -286,3 +288,102 @@ def link_for_row(row)
     row[@headers_for_report.length + 7]
   end
 end
+
+# This entire spec is for when there aren't forecasts but there are activities -- it's essentially
+# a copy of the setup for the above but we never add a value to the forecast.
+RSpec.describe Export::Report do
```
Gweaton, line 294: I think this would be better served in a `context "there are activities with no forecasts"` block, rather than duplicating the setup for this pretty complex test suite (which I'd love us to cut down in future).

Inspired by Cristina's comment, I've checked this out and run it locally and it looks like everything is passing without us setting the following variables:
- `@implementing_organisation`
- `@third_party_project`
- `previous_report`
- `@forecast`
- `@comment`

Is it worth us adding a new top level `@report_with_activities_but_no_forecasts` (or something) for use in this new context so we don't need to create and persist a lot more entities in the database for this test? The test suite is painfully slow at the moment as-is, so I'm reluctant to introduce more persistence!

Gweaton (commented):
This looks good - I've got a few comments re: testing & comments that would be good to resolve before this gets merged. Happy to chat through if things don't make sense.

app/services/export/report.rb
shuldt: no i think you're right, looking at the file as a whole and examining the commit message i think we're safe to get rid of both

spec/services/export/report_spec.rb
shuldt: so dragon and i tried a number of ways to avoid the duplicated set-up when we wrote this, but had no luck. i'll have another poke around today and see if i can get it to work another way. agree about the vars that don't need setting though

PR #1749, opened 2022-08-25
Provide specific error message to spec

CristinaRO:
## Changes in this PR

The test was throwing a warning:
“Using the `raise_error` matcher without providing a specific error or
message risks false positives, since `raise_error` will match when Ruby
raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially
allowing the expectation to pass without even executing the method you
are intending to call.”

Fix by matching against a specific error message.


Gweaton (approved):
Thanks for sorting this, has been blighting my terminal for too long!

PR #1753, opened 2022-09-06
(2540) Replace Partner Org "view details" link with clickable Partner Org name link

Gweaton:
## Changes in this PR

Replace "View details" link with link on Org name to free up some room in future for a link to "Upload activities".

I've also added a visually hidden span for screen readers here, although I wonder if the text could be a bit more obvious?

This PR also refactors text on links on this page to use translation files to be more consistent.

### Before

image

### After

image

## Next steps

We'll be adding the "Upload data" link to this table in the next chunk of work.

### DIFF
--- a/app/views/staff/home/_delivery_partner_organisations_table.html.haml
+++ b/app/views/staff/home/_delivery_partner_organisations_table.html.haml
@@ -11,9 +11,8 @@
     - @delivery_partner_organisations.each do |delivery_partner_organisation|
       %tr.govuk-table__row
         %td.govuk-table__cell
-          = delivery_partner_organisation.name
+          = a11y_action_link delivery_partner_organisation.name, organisation_path(delivery_partner_organisation), t("table.cell.organisations.details"), ["govuk-link--no-visited-state"]
         %td.govuk-table__cell
-          = a11y_action_link "View details ", organisation_path(delivery_partner_organisation), "for #{delivery_partner_organisation.name}", ["govuk-link--no-visited-state"]
-          = a11y_action_link "View activities", organisation_activities_path(delivery_partner_organisation), "for #{delivery_partner_organisation.name}", ["govuk-link--no-visited-state", "govuk-!-margin-left-7"]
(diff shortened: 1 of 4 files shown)


app/views/staff/home/_delivery_partner_organisations_table.html.haml
```diff
@@ -11,9 +11,8 @@
     - @delivery_partner_organisations.each do |delivery_partner_organisation|
       %tr.govuk-table__row
         %td.govuk-table__cell
-          = delivery_partner_organisation.name
+          = a11y_action_link delivery_partner_organisation.name, organisation_path(delivery_partner_organisation), t("table.cell.organisations.details"), ["govuk-link--no-visited-state"]
```
yndajas, line 14: What about "View [organisation name] organisation details"? Would need to extend the `a11y_action_link` functionality though, so might not be worth it

spec/views/home/service_owner_spec.rb
```diff
@@ -1,7 +1,8 @@
 RSpec.describe "staff/home/service_owner" do
   before do
     assign(:current_user, build(:beis_user))
-    assign(:delivery_partner_organisations, build_list(:delivery_partner_organisation, 2))
+    organisation = build(:delivery_partner_organisation, name: "Partner Org 1")
+    assign(:delivery_partner_organisations, [organisation, build(:delivery_partner_organisation)])
```
yndajas, line 5: Is the change here just so you can specify a name for the organisation and test against that on L19?

yndajas (approved):
Code looks good and tested in browser. Just a couple of questions/comments, but I think this is good to go

CristinaRO (approved):

spec/views/home/service_owner_spec.rb
Gweaton: Yep, that's right

app/views/staff/home/_delivery_partner_organisations_table.html.haml
Gweaton: This is a good suggestion, but I'm hesitant to make a change to this method here as it's used in a lot of places. I'm hoping this is usable enough. We'll be having an accessibility audit later in the year so we can revisit it then.

PR #1754, opened 2022-09-06
(2537) Use "Partner organisation" instead of "Delivery partner" in user-facing text

CristinaRO:
## Changes in this PR
- Replace "Delivery partner" with "Partner organisation" and "DP" with "PO" in user-facing text, including CSV column headers

### After
Screenshot 2022-09-06 at 16 47 05

This is just a sample, not all the pages in the app where this change has happened!

## Next steps


CristinaRO:
[Update text referring to DP -> Partner Organisation](https://trello.com/c/c4RqyX80/2537-update-text-referring-to-dp-partner-organisation)

Gweaton (approved):
Thanks for doing all this, can't imagine it was the most gripping task.

The one thing I'm thinking about is we're kind of inconsistent with case here - some places we have "Partner Organisation" some we have "Partner organisation", but we also use the term "PO" which makes me think it should be Title Case. I don't think should block us, as we're doing the same as we did with "Delivery (P/p)artner" but might be something we want to have a review of at some point.

CristinaRO:
> Thanks for doing all this, can't imagine it was the most gripping task.
> 
> The one thing I'm thinking about is we're kind of inconsistent with case here - some places we have "Partner Organisation" some we have "Partner organisation", but we also use the term "PO" which makes me think it should be Title Case. I don't think should block us, as we're doing the same as we did with "Delivery (P/p)artner" but might be something we want to have a review of at some point.

Haha, yes, I was faithful to the inconsistency I found along the way!

PR #1755, opened 2022-09-06
Add arm64 Bundler compatibility

yndajas:
## Changes in this PR

This fixes `bundle install` issues on ARM-based Macs (e.g. Macs with M1/M2 chips) by adding `arm64-darwin-21` to the platforms in Gemfile.lock. It also adds arm64 versions of a few gems, which were added automatically by bundler after fixing `bundle install`

@see https://github.com/rubyjs/mini_racer/issues/227#issuecomment-1014725749

## Comments

Not sure if this requires a changelog update, given it "has no impact outside the development team"?

## Next steps


Gweaton (approved):
Have tested this out on my non M1 mac and it seems to be working as expected. Might be worth @CristinaRO testing this out on an M1 mac as it might be more relevant?

And yep, no need to update changelog for this.

CristinaRO:
Oh no, my `pg` issue is back:

`An error occurred while installing pg (1.4.3), and Bundler cannot continue.`

PR #1756, opened 2022-09-06
(2532) Add hub page for being able to download template upload CSV file

yndajas:
## Changes in this PR

Add Level B activity bulk upload form view and a link to it from the [partner organisations table](http://localhost:3000/home). The view for the form includes a link to download the template CSV files, but it currently points nowhere

I'm not sure why the organisation names are title case on my end - the before screenshot is taken from #1753 (i.e. @Gweaton's setup), so it hopefully isn't a change I've introduced, just a local environment quirk (can look into this tomorrow)

### Before

image

### After

image

---

image

## Next steps

### DIFF
--- a/app/controllers/staff/level_b_activity_uploads_controller.rb
+++ b/app/controllers/staff/level_b_activity_uploads_controller.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Staff::LevelBActivityUploadsController < Staff::BaseController
+  include Secured
+
+  def new
+    @organisation ||= Organisation.find(params[:organisation_id])
+    authorize @organisation, :bulk_upload?
+  end
+end
--- a/app/views/staff/level_b_activity_uploads/new.html.haml
+++ b/app/views/staff/level_b_activity_uploads/new.html.haml
@@ -1,0 +1,31 @@
+= content_for :page_title_prefix, t("page_title.activity.upload_level_b", organisation_name: @organisation.name)
+
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %h1.govuk-heading-xl
+        = t("page_title.activity.upload_level_b", organisation_name: @organisation.name)
+
+      %p.govuk-body
+        Use the template to ensure your data is in the correct format.
+
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %h2.govuk-heading-m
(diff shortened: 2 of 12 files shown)


yndajas:
Looks like there's an MFA/Notify issue and some JSON parsing error, neither of which should relate to changes in this branch - curious as to why this is failing @Gweaton 🤔

Gweaton:
> Looks like there's an MFA/Notify issue and some JSON parsing error, neither of which should relate to changes in this branch - curious as to why this is failing @Gweaton 🤔

I re-ran the build and it seemed to pass. I think there are a couple of flaky tests that need re-running. I'll make a note of this one so we can pin down in case it's one of the regularly recurring ones.

config/routes.rb
```diff
@@ -61,6 +61,7 @@
           get tab, to: "activities#show", defaults: {tab: tab}
         end
       end
+      resources :level_b_activity_uploads, only: [:new]
```
Gweaton, line 64: Shall we make this `resource` here rather than `resources` as (as you so eloquently put it in a recent commit message):
> these routes aren't here to perform actions on database records of uploads (with upload IDs), but to upload activity data relating to organisations

app/controllers/staff/level_b_activity_uploads_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Staff::LevelBActivityUploadsController < Staff::BaseController
+  include Secured
+
+  def new
+    @organisation ||= Organisation.find(params[:organisation_id])
+    authorize @organisation, :show_bulk_upload?
```
CristinaRO, line 8: Do we foresee having different levels of bulk upload authorisations? I.e. some users can only `show_bulk_upload`, whereas others can `create_bulk_upload`?

app/views/staff/level_b_activity_uploads/new.html.haml
```diff
@@ -0,0 +1,31 @@
+= content_for :page_title_prefix, t("page_title.activity.upload_level_b", organisation: @organisation.name)
```
CristinaRO, line 1: Very small preference for naming the string parameter passed to the locale `organisation_name` or `org_name`, but it's not a big deal.

app/controllers/staff/level_b_activity_uploads_controller.rb
Gweaton: We've now renamed this to just `bulk_upload?` as there isn't a difference, and "showing" isn't really the important verb here.

PR #1758, opened 2022-09-08
(2538) Refactor, rename, and migrate code references to "delivery partners"

CristinaRO:
## Changes in this PR
- Replace references to `delivery_partner` with `partner_organisation` (or `partner_organisation_user` depending on context) across the code, including a database migration to rename `delivery_partner_identifier` to `partner_organisation_identifier`
- Replace references to `country_delivery_partners` with `country_partner_organisations`, including database migration

## Screenshots of UI changes

N/A

## Next steps


CristinaRO:
[Refactor/rename code references to Delivery Partners](https://trello.com/c/JjApfrls/2538-refactor-rename-code-references-to-delivery-partners)

spec/mailers/previews/report_mailer_preview.rb
```diff
@@ -8,7 +8,7 @@ def activated
     ).activated
   end
 
-  def submitted_delivery_partner
+  def submitted_partner_organisation
```
Gweaton, line 11: For my own understanding, is this rename part of:
> renaming all the methods, locale strings, partials etc. that use the role names instead
of their numeric values.

or did it sneak in this commit? No worries if it did, just want to make sure I know what's happening!

Don't worry about it i

Gweaton (approved):
Thanks for all this - a mammoth renaming task, but I really liked how you split it up into more focused commits e.g. renaming `Delivery partner user` rather than a single "find and replace all".

There was a lot to read through, which made any errors like finding a needle in a haystack, so I hope I haven't missed anything. I'd like us to do a bit of manual testing after this change on staging just to make sure everything is working as expected, but we're also well-covered by our tests.

spec/mailers/previews/report_mailer_preview.rb
CristinaRO: Yes, as far as I could identify what calls upon what, the name of this method is correlated with the partial `_submitted_delivery_partner.text.erb` also renamed in this commit to `_submitted_partner_organisation.text.erb`, which in turn is generated based on the action (`submitted`) and the role, which changed in this commit from `delivery_partner` to `partner_organisation`. 😅

PR #1764, opened 2022-09-13
Remove “PO Definition” from “Aims/Objectives”

CristinaRO:
## Changes in this PR
- Remove "PO Definition" from "Aims/Objectives" CSV headers

## Next steps


CristinaRO:
[Remove (PO definition) from the headers in the CSV](https://trello.com/c/jrJhwi0o/2559-remove-po-definition-from-the-headers-in-the-csv)

Gweaton (approved):
Nice work. Again, we should let the BEIS team know this header has changed in case they or POs have any scripts that rely on these headers - probably something we can notify Dan of in standup.

PR #1765, opened 2022-09-15
(2561) Make partner org identifier not be required

yndajas:
## Changes in this PR

This makes the partner organisation identifier optional for Level B activities

## Screenshots of UI changes

The only UI change is in the form for adding an individual Level B activity

### Before

image

### After

image

## Next steps


app/controllers/staff/level_b_activity_uploads_controller.rb
```diff
@@ -30,7 +30,8 @@ def update
       importer = Activities::ImportFromCsv.new(
         uploader: current_user,
         partner_organisation: organisation,
-        report: nil
+        report: nil,
+        validation_steps: Activity::LEVEL_B_VALIDATION_STEPS
```
Gweaton, line 34: Can we check this doesn't override the parent id uniqueness constraint here? We're assuming this only overrides the "not nil" requirement when skipping the `:identifier` step.

yndajas: Looks fine. I changed `valid_level_b_activities_upload.csv` to have `myui` as the `partner_organisation_identifier` for both activities and it came up with this error:

image

app/services/activities/import_from_csv.rb
```diff
@@ -203,14 +211,15 @@ def set_errors
     class ActivityCreator
```
yndajas, line 211: Does this need applying to `ActivityUpdater` too?

yndajas: I think it doesn't, since where `ActivityCreator` calls `return true if @activity.save(context: @validation_steps)`, `ActivityUpdater` just calls `return set_errors unless @activity.save` - no context argument to pass in the validation steps

app/services/activities/import_from_csv.rb
```diff
@@ -113,6 +120,7 @@ def sub_level_b_column_headings
           "DFID policy marker - Gender",
           "DFID policy marker - Nutrition",
           "ODA Eligibility Lead",
+          "Partner organisation identifier",
```
Gweaton, line 123: I'd like to check with the BEIS team whether this should be included in the template CSV or not - I think it should be included but not required, otherwise people uploading the CSV won't know it's something they can do. I'd like to confirm this with BEIS though.

Gweaton (commented):
This looks good to me. As part of this work, I think we should make this optional in the UI journey for adding an activity too, so the behaviour is consistent for both journeys.

yndajas:
[Make partner org identifier not be required](https://trello.com/c/UCXHQeiF/2561-make-partner-org-identifier-not-be-required)

spec/views/staff/shared/activities/activity_spec.rb
```diff
@@ -245,7 +245,10 @@
 
   RSpec::Matchers.define :show_basic_details do
     match do |actual|
-      expect(rendered).to have_content activity_presenter.partner_organisation_identifier
+      if activity_presenter.is_project? || activity_presenter.partner_organisation_identifier.present?
+        expect(rendered).to have_content activity_presenter.partner_organisation_identifier
+      end
```
yndajas, line 250: Perhaps we should abstract `activity_presenter.is_project? || activity_presenter.partner_organisation_identifier.present?` into a method (e.g. `partner_organisation_identifier_visible?`) on the model or the presenter, since it's used three times now - here and in the two partials in `app/views/staff/shared/activities`

yndajas:
I think this is ready for review again - it's actually a very small PR now with the latest decision on how this should work

CHANGELOG.md
```diff
@@ -1086,7 +1086,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Replace references to `country_delivery_partners` with `country_partner_organisations`, including database migration
 - Replace code references to `uk_dp_named_contact` with `uk_po_named_contact`, including database migration
 - Remove "PO Definition" from "Aims/Objectives" CSV headers
-- Add ability for BEIS users to create and update Level B activities in bulk by uploading a CSV 
+- Add ability for BEIS users to create and update Level B activities in bulk by uploading a CSV
```
yndajas, line 1089: Might make more sense to include this trailing space removal in a commit that actually updates the changelog (if it all), but that said, should we include a line for this since it's now beyond the scope of just bulk uploads, e.g. "Make partner organisation identifier optional for Level B activities (both via the individual new activity form and bulk upload)"

Gweaton: Yeah, I think it'd be good to include this as a new changelog item as it's changing the UI journey too.

Gweaton (approved):
Thanks for this, looks good!

PR #1770, opened 2022-09-20
(2541) Update Node to v16

Gweaton:
## Changes in this PR

Updates Node to v16.

This is the "LTS" version. However, it's going into Maintentance Mode in October this year, so we'll want to update it to 18 as soon as we can.


Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN mkdir -p ${DEPS_HOME}
 WORKDIR ${DEPS_HOME}
 # End
 
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash - \
+RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - \
```
CristinaRO, line 31: Should it be `setup_18` or are they not tied together?

Gweaton: We tried 18 but had some failures in the build, so thought we'd update to 16 for now rather than spend lots of time on it.

.node-version
```diff
@@ -1 +1 @@
-14.8.0
+18.9.0
```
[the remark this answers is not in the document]
Gweaton: Nope, forgot to put this back after we downgraded to 16!

CristinaRO (approved):

PR #1774, opened 2022-09-21
Reset env var post-spec

CristinaRO:
## Changes in this PR

The spec `it "loads redis URL to the ENV”` has side effects that can cause other specs to fail, depending on the order in which they are run.

`ClimateControl` can only reset `VCAP_SERVICES` for us, not the other env vars that are set during the spec running.

The default value of `VCAP_SERVICES` (set in .env.test) doesn’t have a value for `REDIS_URL`, so the value being set during this spec sticks, and overwrites the desired value of `REDIS_URL`, which is set in .env.test.

## Next steps

### DIFF
--- a/spec/lib/vcap_parser_spec.rb
+++ b/spec/lib/vcap_parser_spec.rb
@@ -23,6 +23,8 @@ RSpec.describe VcapParser do
     end
 
     it "loads redis URL to the ENV" do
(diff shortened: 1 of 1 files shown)


CristinaRO:
[Fix flaky spec](https://trello.com/c/1zzsMKsU/2575-fix-flaky-spec)

Gweaton (approved):

PR #1778, opened 2022-09-23
(2585) Add placeholder Level B budget upload page

yndajas:
## Changes in this PR

This adds a new view for bulk uploading Level B budget data via CSV. Currently the CSV template download and upload form are non-functional - these will be added in later commits

We added a new budget upload model and policy as this upload is not attached to a specific activity (or any entity) - budgets can be uploaded for multiple funds and organisations all in one CSV file

**Note:** linked in the parent Trello card is an design screenshot (`uploads_form.png`), which slightly differs in structure to the Level B activities upload form. We decided to keep it similar to the latter - we weren't sure how recent the design was - but we could change it to match the design more closely if preferred

## Screenshots of UI changes

image

## Next steps

### DIFF
--- a/app/policies/level_b_policy.rb
+++ b/app/policies/level_b_policy.rb
@@ -1,0 +1,5 @@
+class LevelBPolicy < ApplicationPolicy
+  def budget_upload?
+    return true if beis_user?
+  end
+end
--- a/app/policies/level_b_policy.rb
+++ b/app/policies/level_b_policy.rb
@@ -1,4 +1,8 @@
 class LevelBPolicy < ApplicationPolicy
+  def activity_upload?
+    return true if beis_user?
+  end
+
   def budget_upload?
     return true if beis_user?
   end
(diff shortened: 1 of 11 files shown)


app/policies/level_b_policy.rb
```diff
@@ -0,0 +1,5 @@
+class LevelBPolicy < ApplicationPolicy
```
Gweaton, line 1: Do you think this would be a place for the level b activity upload policy to live? If so, shall we add a new commit to this PR for refactoring that bit of functionality? Or at least add a technical debt card for it (though it'd be nice to sort rather than add to the pile!)

Gweaton (approved):
Looks good, thanks for this. Just one thing re: refactoring the previous policy

app/policies/level_b_policy.rb
yndajas: Done - I started writing an `activity_upload` method in the policy, but then changed tack, per the commit message

Gweaton: As discussed on Slack, we'll do what you originally did (an `activity_upload` method) as it'll allow us to be a bit more granular, just in case we want to split permissions for the different entities being uploaded in the future e.g. a finance team may be responsible for uploading budgets.

Gweaton (approved):

PR #1779, opened 2022-09-23
(2588) Add Level B budget bulk upload template CSV download action

yndajas:
## Changes in this PR

This adds a controller action to enable users to download the template CSV for Level B budget data bulk uploads, linked from the view/form created in the previous commit

The headers are as per the design collaboration with the client, but may still be subject to change. The naming of some columns differs slightly to that used in the activity templates

## Next steps

### DIFF
--- a/app/controllers/staff/level_b/budgets/uploads_controller.rb
+++ b/app/controllers/staff/level_b/budgets/uploads_controller.rb
@@ -1,7 +1,26 @@
 class Staff::LevelB::Budgets::UploadsController < Staff::BaseController
   include Secured
+  include StreamCsvDownload
 
   def new
     authorize :level_b, :budget_upload?
+  end
+
+  def show
+    authorize :level_b, :budget_upload?
+
+    headers = [
+      "Type",
+      "Financial year",
+      "Budget amount",
+      "Providing organisation",
+      "Providing organisation type",
+      "IATI reference",
+      "Activity RODA ID",
+      "Fund RODA ID",
+      "Partner organisation name"
+    ]
+
+    stream_csv_download(filename: "Level_B_budgets_upload.csv", headers: headers)
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/staff/level_b/budgets/uploads_controller.rb
```diff
@@ -1,7 +1,26 @@
...
+  include StreamCsvDownload
 
   def new
     authorize :level_b, :budget_upload?
   end
+
+  def show
+    authorize :level_b, :budget_upload?
+
+    headers = [
```
Gweaton, line 12: Do you think it's too early to encapsulate this logic in a `Budgets::ImportFromCSV` class like we do for other imports? This would give us an object we could interact with like `ImportFromCSV.column_headings`. Were you planning on heading in this direction in the next bit of work?

Gweaton (approved):
This looks good, one comment on keeping "logic" (if it can be called that) out of the controller action

app/controllers/staff/level_b/budgets/uploads_controller.rb
yndajas: I can't remember the specific thinking for this, but @CristinaRO and I discussed it. I think it was either that we thought it would come if/when needed, or that the budgets headers wouldn't vary (at least for this piece of work/Level B) so it might not need interacting with/modifying elsewhere?

It looks like we need it quite a bit in the `ImportFromCsv` class though, so I guess the question is whether we do that now (if it's not getting ahead of ourselves for this commit) or in the next PR/commit that is doing the work of introducing `ImportFromCsv`?

Gweaton: Let's leave this until the next PR that adds more functionality, this shouldn't block things for now.

PR #1783, opened 2022-09-27
(2591) Refactor the importers to have a unified approach

CristinaRO:
## Changes in this PR
- Place each upload controller in a hierarchy based on [level]/entity (pluralised), e.g. `level_b/activities/uploads_controller`, `actuals/uploads_controller`
- Place each import service in a hierarchy based on entity, e.g. `actual/import`
- Refactor the actuals history importer from a model into a service

NB: We have deliberately left the `Import::RowError` model where it was, to avoid scope creep and making this PR even harder to review. We will decide on what approach to take for modelling import errors in the (near) future.

## Screenshots of UI changes

N/A

## Next steps


CristinaRO:
[Spike: how easy would it be to refactor importers into namespaced structures?](https://trello.com/c/hMonBD6u/2591-spike-how-easy-would-it-be-to-refactor-importers-into-namespaced-structures)

app/controllers/staff/activities/uploads_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 require "csv"
 
-class Staff::ActivityUploadsController < Staff::BaseController
+class Staff::Activities::UploadsController < Staff::BaseController
```
yndajas, line 5: Are we level namespacing these or are we only level namespacing level B, with levels C and D as defaults?

CHANGELOG.md
```diff
@@ -1094,6 +1094,8 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 
 ## [unreleased]
 
+- Refactor the importers to use a unified approach for all entities being imported
```
yndajas, line 1097: `importers (including uploads controllers)`?

config/routes.rb
```diff
@@ -61,22 +61,30 @@
...
-        resource :actual_history, only: [:new, :update]
+      namespace :activities do
+        resource :upload, only: [:new, :show, :update]
+      end
+      namespace :actuals do
+        resource :upload, only: [:new, :show, :update]
+        resource :history_upload, only: [:new, :update]
+      end
+      namespace :forecasts do
+        resource :upload, only: [:new, :show, :update]
```
yndajas, line 87: You could group the upload resources. Only one line less, so it's whether it's worth grouping similar activities or sticking to more explicit entity groupings

```
%i(activities actuals forecasts).each do |report_namespace|
  namespace report_namespace do
    resource upload, only: [:new, :show, :update]
  end
end

namespace :actuals do
  resource :history_upload, only: [:new, :update]
end
```

Not sure if it's possible to encapsulate it into a method for reuse with the budgets too, since the method might not respect the resources nesting?

yndajas (approved):
Nice - this is much easier on the brain! Just a few small comments/queries but looks good

Gweaton (approved):
Nice work - a couple of minor comments.

app/controllers/staff/activities/uploads_controller.rb
yndajas: Makes sense - this question applies to namespacing across the board, but I think we can probably treat most/all of it the same way

CHANGELOG.md
CristinaRO: I tend to err on the side of having a changelog entry rather than not, to make things easier to trace across the history of the project. The checklist suggests "unless it's a small tweak" with no impact outside the dev team, so I tend to think that big tweaks should be logged, even if they have no impact outside the dev team.

app/services/forecast/import.rb
```diff
@@ -0,0 +1,172 @@
+require "csv"
+require "set"
+
+class Forecast
```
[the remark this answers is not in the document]
CristinaRO: I *thought* I had removed all the old services, but some may have slipped through!

app/controllers/staff/activities/uploads_controller.rb
yndajas: Yeah `sub_level_b` is a bit... something

Gweaton (approved):

PR #1785, opened 2022-09-28
(2590) Add budget importer

yndajas:
## Changes in this PR

This adds a budget importer for processing bulk uploads in a similar way to the activities importer

## Next steps

### DIFF
--- a/app/controllers/staff/level_b/budgets/uploads_controller.rb
+++ b/app/controllers/staff/level_b/budgets/uploads_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Staff::LevelB::Budgets::UploadsController < Staff::BaseController
   include Secured
   include StreamCsvDownload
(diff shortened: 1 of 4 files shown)


yndajas:
[Add `services/budgets/import_from_csv`](https://trello.com/c/3b5O78fk/2590-add-services-budgets-importfromcsv)

yndajas:
[Add `services/budgets/import_from_csv`](https://trello.com/c/3b5O78fk/2590-add-services-budgets-importfromcsv)

app/services/create_budget.rb
```diff
@@ -7,8 +7,8 @@ def initialize(activity:)
 
   def call(attributes: {})
     budget = Budget.new
-    budget.parent_activity = activity
     budget.assign_attributes(attributes)
+    budget.parent_activity = activity
```
Gweaton, line 11: ICan this ever actually be a different activity passed in via the attributes if we're passing in the activity _and_ the id that was used to fetch it?

Gweaton (approved):
Thanks for your hard work on this, it looks good to me. Just have a few questions/comments.

app/services/create_budget.rb
yndajas: I don't think my brain is working well enough to comprehend that question!

PR #1790, opened 2022-10-03
(2565) Build level b bulk budget upload functionality

yndajas:
## Changes in this PR

This adds the Level B budgets bulk upload functionality for BEIS users

### Before

image

### After

(This will fail the first time because of using the 'invalid' CSV, then the second time because the 'valid' CSV has a parent activity that isn't in my local database)

https://user-images.githubusercontent.com/40244233/193571743-33533669-e42b-471e-b8f8-e52df62d659f.mov

## Next steps

### DIFF
--- a/config/locales/models/budget.en.yml
+++ b/config/locales/models/budget.en.yml
@@ -96,3 +96,14 @@ en:
               blank: Enter the name of the providing organisation
             providing_organisation_type:
               blank: Select the type of the providing organisation
+  importer:
+    errors:
+      budget:
+        cannot_create: There is no activity RODA ID present, so cannot create a budget
+        invalid_budget_type: The budget type code is not valid
+        invalid_financial_year: The financial year is not valid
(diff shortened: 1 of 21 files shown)


config/locales/models/budget.en.yml
```diff
@@ -33,7 +42,12 @@ en:
         budget_type:
           direct: Budget allocated directly from the parent activity
           other_official: Budget allocated from an external organisation that is still considered ODA funding
+        csv_file: Upload a spreadsheet containing budget data in CSV format.
+        csv_file_recover_from_error_html: Upload a spreadsheet containing activity data in CSV format. We recommend <a href="%{link}" class="govuk-link">downloading this CSV template</a>.
```
CristinaRO, line 46: Ah, is this a copy-paste from activity? Should probably say "budget data"

CristinaRO (approved):
One thing that should be fixed before merging, but approving to not hold up the merge.

PR #1794, opened 2022-10-04
(2513) Make inactive organisations unusable

yndajas:
## Changes in this PR

This prevents users from adding inactive organisations as implementing organisations for activities

## Screenshots of UI changes

If the database is in sync with production, NERC CENTRE FOR ECOLOGY & HYRDOLOGY will appear before, but not after

### Before

image

### After

image

## Next steps

### DIFF
--- a/spec/models/org_participation_spec.rb
+++ b/spec/models/org_participation_spec.rb
@@ -3,4 +3,46 @@ require "rails_helper"
 RSpec.describe OrgParticipation, type: :model do
   it { should belong_to(:activity) }
   it { should belong_to(:organisation) }
+
+  context "when the org participation is unpersisted" do
+    context "when the role is implementing and the organisation is inactive" do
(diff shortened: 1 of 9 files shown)


spec/models/org_participation_spec.rb
```diff
@@ -3,4 +3,32 @@
 RSpec.describe OrgParticipation, type: :model do
   it { should belong_to(:activity) }
   it { should belong_to(:organisation) }
+
+  context "when role is implementing and organisation is inactive" do
+    subject(:org_participation) { build(:org_participation, :inactive_organisation) }
+
+    it "should not be valid" do
+      expect(org_participation.valid?).to be false
```
CristinaRO, line 11: Is this going to invalidate any existing records when an organisation becomes inactive? We should prevent invalidating any existing records.

Gweaton (approved):
Sorry for the delay on this approval, was taking my time to figure out how these entities interacted with one another. This looks good to me though.

PR #1795, opened 2022-10-05
(2179) Remove unused activity show view

yndajas:
## Changes in this PR

The activities show action renders the template for the current tab, with "Financials" set as the default tab if one is not specified via a URL param

## Next steps


yndajas:
[Activity show view is never rendered - let's remove it](https://trello.com/c/KG0AFKYa/2179-activity-show-view-is-never-rendered-lets-remove-it)

CristinaRO (approved):

PR #1796, opened 2022-10-05
(2555) Separate comments in reports CSV with a pipe

CristinaRO:
## Changes in this PR
- Comments in reports CSV are delimited by a pipe, instead of a newline, to enable BEIS users' QA workflow

### Before
Screenshot 2022-10-05 at 14 49 41

### After
Screenshot 2022-10-05 at 14 50 29

## Next steps


yndajas (approved):
Probably fine, but as per my comment, the leading and trailing spaces in the delimiter might not allow for cleanly converting text to columns in Excel

PR #1803, opened 2022-10-10
Fix training db sync script password reset command

Gweaton:
## Changes in this PR

After running the training db sync, which fetches the current users on production and writes them to the training environment database, we encountered the following error on logging in:

```
OpenSSL::Cipher::CipherError
```

This was a difficult one to debug, but after some digging, we discovered it was caused by users' passwords not being reset as expected in the `pw_reset_script` portion of the training db sync script.

This was a difficult one to debug, but we discovered that the `bin/rails` command wasn't being run (albeit silently, without any errors being surfaced to us running the command) due to the `ssh -c` command not placing the user in the `/app` directory where `bin/rails` is available.

Then, we discovered `ruby` wasn't available in the `PATH` as using double quotes around a `-c` argument tells the script to use your user's local PATH, rather than that on the machine running the command. To fix this, we've used single quotes and explicitly specified the Ruby path when running the script remotely.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1100,6 +1100,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Include refunds and adjustments in the calculation of an activity's total spend (previously only actuals were included)
 - Point the "Back to home" link on the Level B activities bulk upload to the home page instead of the organisations page
 - Comments in report CSVs are delimited by a pipe, instead of a newline, to enable BEIS users' QA workflow
+- Fix the command invoking the password reset script when updating training environment with the latest data from production
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-117...HEAD
 [release-117]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-116...release-117
--- a/script/training_db_sync.rb
+++ b/script/training_db_sync.rb
@@ -106,7 +106,7 @@ class TrainingDbSync
     Kernel.puts "Copying pw reset script to #{destination} ==> #{copy_cmd}"
     CmdRunner.run(copy_cmd)
 
-    reset_cmd = %(cf ssh beis-roda-#{destination} -c "bin/rails runner #{pw_reset_script}")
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1100,6 +1100,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Include refunds and adjustments in the calculation of an activity's total spend (previously only actuals were included)
 - Point the "Back to home" link on the Level B activities bulk upload to the home page instead of the organisations page
 - Comments in report CSVs are delimited by a pipe, instead of a newline, to enable BEIS users' QA workflow
+- Fix password reset script when updating training environment with the latest data from production
```
yndajas, line 1103: `script invoked when`?

yndajas, line 1103: `script invoked when`?

script/training_db_sync.rb
```diff
@@ -106,7 +106,7 @@ def call
     Kernel.puts "Copying pw reset script to #{destination} ==> #{copy_cmd}"
     CmdRunner.run(copy_cmd)
 
-    reset_cmd = %(cf ssh beis-roda-#{destination} -c "bin/rails runner #{pw_reset_script}")
+    reset_cmd = %(cf ssh beis-roda-#{destination} -c 'cd /app; PATH=$PATH:/usr/local/bin bin/rails runner #{pw_reset_script}')
```
yndajas, line 109: From the commit message, it sounds like single quotes should make it load the correct path, so what's the (extra) benefit of the explicit path?

Also, are there any docs for the single versus double quote distinction that could be linked to in the commit? The commit is a good explainer of why, but it feels like a pretty unusual distinction so some docs might be nice!

yndajas (approved):
Looks good. Just a couple of comments on the details. Also, your commit repeats "This was a difficult one to debug" - is this poetically motivated?

script/training_db_sync.rb
Gweaton: Yeah, I (lazily) put both in, but we don't need the single quotes here so have put them back to double quotes - I think the discovery that they use your local PATH was a bit of a red herring that wasn't actually required. I'm still not 100% sure why the PATH isn't loaded properly when running `ssh -c` but will see what I can find here.

yndajas (approved):
Looks good - tested the step with `User.find_by(email: "ynda@dxw.com")` and everything was reset successfully

PR #1806, opened 2022-10-10
(2617) Remove `parser` warning

yndajas:
## Changes in this PR

A previous change introduced a warning that would litter the console when running the server, tests etc. 4f1ffed had fixed `parser` to 3.0 due to Ruby 2.7.3 compatibility issues. We've since upgraded to 2.7.6, so this is no longer needed
- I've therefore removed the version specification and updated the bundle/lockfile, which removes the warning

## Next steps


yndajas:
[Remove parser warning](https://trello.com/c/oLFGPlhN/2617-remove-parser-warning)

Gweaton (approved):
Well spotted - thanks for sorting this, it's been bugging me for a while now!

PR #1813, opened 2022-10-12
(2619) Improve error message when trying to create a Level C activity on an incomplete Level B activity

Gweaton:
## Changes in this PR

Makes incomplete parent activity error more explicit

Users have reported being unable to add Level C and D activities to some Level B activities, as they were met with the error message: "The parent activity is not valid". The error was caused by the Level B activity not being in the correct `form_state` (it must be `completed`).

This message didn't feel clear enough, as it didn't tell the user how to rectify the issue - stating that the parent is actually "incomplete".

Hopefully this should make it clearer to the user that they have to go in and finish creating the Level B activity.

This PR also includes a couple of smaller bits of renaming - while I'd like to refactor the importer class, this felt like a bigger job than was necessary for solving this issue.

### Before

image

### After

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1110,6 +1110,8 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 
 ## [unreleased]
 
+- Improve error message when trying to create child activities on incomplete parent activities
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-119...HEAD
 [release-119]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-118...release-119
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1106,6 +1106,8 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 
 ## [unreleased]
 
+- Improve error message when trying to create level C activities on incomplete level B activities
```
yndajas, line 1109: `and D`?

yndajas (approved):
Looks good - just one comment about the changelog

yndajas:
[Level C activities can't be added to level B activities?](https://trello.com/c/CftZgcl5/2619-level-c-activities-cant-be-added-to-level-b-activities)

PR #1814, opened 2022-10-12
(2636) Sync budgets fixture with template

yndajas:
## Changes in this PR

The columns in the budgets template have changed (two removed, one moved) since the test fixture (CSV) was created. This brings it in sync to avoid confusion about the columns for future devs

## Next steps


Gweaton (approved):
Well noticed, thanks for sorting.

PR #1815, opened 2022-10-13
(2181) Add a helper method for download as XML permission

CristinaRO:
## Changes in this PR
- Refactor a spec to reflect the existing behaviour for downloading an individual project as XML
- Add a helper for the logic determining if download as XML link should be shown

## Screenshots of UI changes
N/A

## Next steps


spec/features/users_can_view_an_activity_as_xml_spec.rb
```diff
@@ -1,363 +1,283 @@
...
+    end
 
-      context "when the activity is a project" do
-        let(:activity) { create(:project_activity) }
-        let(:fund) { create(:fund_activity) }
-        let(:programme) { create(:programme_activity) }
-        let(:xml) { Nokogiri::XML::Document.parse(page.body) }
+    it "includes all the policy markers with reportable values for IATI" do
+      visit organisation_activity_path(organisation, activity, format: :xml)
+      expect(xml).to have_selector("iati-activity/policy-marker/@code", count: 4) # We do not report 'not_assessed' values to IATI
```
Gweaton, line 141: I wonder if the comment in this line belongs in the `it` description rather than a comment? Do you think that could be covered by `"It includes all IATI-assessed policy markers"`?

app/views/activities/children.html.haml
```diff
@@ -13,9 +13,8 @@
 
   .govuk-grid-row
     .govuk-grid-column-full
-      - if @activity.project? || @activity.third_party_project?
-        - if policy(:project).download? || policy(:third_party_project).download?
-          = link_to t("default.button.download_as_xml"), organisation_activity_path(@activity.organisation, @activity, format: :xml), class: "govuk-button"
+      - if can_download_as_xml?(activity: @activity, user: current_user)
```
Gweaton, line 16: 👍

Gweaton (approved):
A nice refactor 🏆

spec/features/users_can_view_an_activity_as_xml_spec.rb
CristinaRO: Huh, and there's already a spec for `it "does not include the policy markers with a value of 'not assessed'"`.

Gweaton: > 



> Do you think that's what "with reportable values for IATI" was intended to mean? ETA: and that it's not clear enough in its phrasing?

Yeah, I think so - agree with your edit, which is probably why the comment was added.

Interesting that there's already another test for it, maybe we can remove one?

CristinaRO: Eh, one spec is checking that the expected number of markers is present, but not the value of each specific marker, whereas the `not_assessed` spec is specifically checking for exclusion. Because `it "includes all the policy markers with reportable values for IATI"` only checks the total number of nodes, it could pass by accident.

Should we rework the first spec, and check for specific markers rather than the number of nodes? Removing one spec doesn't speed up all that much the test suite, but it *is* good to leave every spec in a better place than we found it...

Gweaton: Yeah, checking the values for the markers feels safer to me, as you said, checking the number of nodes could pass by accident. Definitely a fan of tidying up whenever there's a smell!

PR #1816, opened 2022-10-13
(2621) Allow comments on Level B activities

yndajas:
## Changes in this PR

This extends the activity comments functionality to allow BEIS users to add and edit comments at Level B

Partner organisations will be able to see comments on Level B activities, per client advice

### Before

image

### After

image

## Next steps

### DIFF
--- a/spec/controllers/activity_comments_controller_spec.rb
+++ b/spec/controllers/activity_comments_controller_spec.rb
@@ -1,0 +1,279 @@
+require "rails_helper"
+
+RSpec.describe ActivityCommentsController do
+  let(:beis_user) { create(:beis_user) }
+  let(:partner_organisation_user) { create(:partner_organisation_user) }
+
+  let(:programme_activity) { create(:programme_activity) }
+
+  let(:project_activity) { create(:project_activity, organisation: partner_organisation_user.organisation) }
+  let!(:project_activity_report) { create(:report, :active, fund: project_activity.associated_fund, organisation: partner_organisation_user.organisation) }
+
+  let(:existing_programme_activity_comment) { create(:comment, commentable: programme_activity, owner: beis_user) }
+  let(:existing_project_activity_comment) { create(:comment, commentable: project_activity, owner: partner_organisation_user, report: project_activity_report) }
+
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+  end
+
+  after { logout }
+
+  describe "#new" do
+    render_views
+
+    context "when the activity is a programme" do
+      context "when signed in as a BEIS user" do
+        let(:user) { beis_user }
+
+        it "shows the submit button" do
+          get :new, params: {activity_id: programme_activity.id}
+
+          expect(response.body).to include(t("default.button.submit"))
+        end
+      end
+
+      context "when signed in as a partner organisation user" do
+        let(:user) { partner_organisation_user }
+
+        it "responds with a 401" do
+          get :new, params: {activity_id: programme_activity.id}
+
+          expect(response.status).to eq(401)
+        end
+      end
+    end
+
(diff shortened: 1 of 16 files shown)


yndajas:
[Add UI journey for being able to add and view a comment for a level B activity](https://trello.com/c/0QjrMIGR/2621-add-ui-journey-for-being-able-to-add-and-view-a-comment-for-a-level-b-activity)

yndajas:
TODO:
- sort out edit and update actions
- sort out association tests on comment spec
- write create and update action tests
- clean up commits

spec/helpers/activity_helper_spec.rb
```diff
@@ -62,6 +65,137 @@
     end
   end
 
+  describe "#link_to_add_comment" do
+    context "when activity is a programme" do
+      context "when user is authorised to add programme comments" do
+        it "returns button to add comment" do
+          authorise_creating_comments(commentable_type: :programme_activity, authorise: true)
```
Gweaton, line 72: can we move these `authorise`s into the relevant `context` block? e.g. this one can go in `when user is authorised to add programme comments`

spec/helpers/activity_helper_spec.rb
```diff
@@ -62,6 +65,137 @@
...
+          expect(helper.link_to_add_comment(activity: programme_activity, report: nil)).to be_nil
+        end
+      end
+    end
+
+    context "when activity is a project" do
+      context "when user is authorised to add project comments" do
+        context "when a report exists" do
+          it "returns button to add comment" do
+            authorise_creating_comments(commentable_type: :programme_activity, authorise: false)
```
Gweaton, line 93: again, if we move this up to the `when user is authorised to add project comments` context, we don't need to repeat it in the test below.

spec/controllers/activity_comments_controller_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+          expect(response.status).to eq(401)
+        end
+      end
+    end
+  end
+
+  def post_comment(activity_id:, report_id:)
+    post :create, params: {
+      activity_id: activity_id,
+      comment: {body: "Nihilism slips on a banana peel.", report_id: report_id}
```
Gweaton, line 264: nice comment

Gweaton (commented):
This is a lot of work, well done for sticking with it.

I really like how you've dried up tests by adding re-usable helper methods here, it's something I've wanted to try for a while, and I think it works well.

I found this a little bit tricky to review due to the size of the first commit - it's a tough one to split out into smaller chunks though, I wonder if this would have a been a good example for `[RED]`-style commits that Cristina's mentioned in the past, where it's ok to have failing tests to keep the commit small. Alternatively, we could add pending tests to be addressed in a future commit?

spec/controllers/activity_comments_controller_spec.rb
yndajas: I used these nonsense generators for the comments 😅
- https://nonsense.x2d.org
- https://generatorfun.com/nonsense-generator

spec/helpers/activity_helper_spec.rb
yndajas: I'm not sure that this is possible - I think it needs to be inside `it`. I get an error about it not being defined on the class if I move the method out of the `it` block and similar if I put the `allow` code directly there:

```ruby
undefined local variable or method `view' for #<Class:0x0000000114b0b7f0>
```

yndajas: Actually, I could probably just stick it in a `before` block where repeated

Gweaton (approved):
Thanks for persevering - looks good to me!

PR #1822, opened 2022-10-18
(2661) Add and configure Rollout and Rollout UI

CristinaRO:
## Changes in this PR
- Configure Rollout and Rollout UI gems to allow BEIS users to manage feature flags

### Before
N/A

#### Overview
Screenshot 2022-10-18 at 19 01 11

#### Individual feature management
Screenshot 2022-10-18 at 19 00 52

## Next steps


CristinaRO:
[Add Rollout and Rollout UI](https://trello.com/c/50d0A9uK/2661-add-rollout-and-rollout-ui)

spec/rollout_spec.rb
```diff
@@ -0,0 +1,6 @@
+RSpec.describe "Rollout" do
+  it "has two custom groups for BEIS users and PO users" do
+    expect(ROLLOUT.groups).to include(:beis_users)
```
Gweaton, line 3: Do you think it'd be worth it/possible to test that a specific user belongs to a group here? e.g. create a BEIS user and check that they're in the `:beis_users`?

Gweaton, line 3: Do you think it'd be worth it/possible to test that a specific user belongs to a group here? e.g. create a BEIS user and check that they're in the `:beis_users`?

app/services/service_owner_constraint.rb
```diff
@@ -0,0 +1,9 @@
+class ServiceOwnerConstraint
+  def self.matches?(request)
+    id = request.session["warden.user.user.key"].try(:[], 0).try(:[], 0)
```
Gweaton, line 3: Did you find out if you could do `try(:first).try(:first)` here by the way?

Gweaton: Also, do you think being more explicit here and calling this `user_id_from_session` would make this clearer? Just thinking new readers (/mostly me) seeing `"warden.user.user.key"` might be confused about what that is!

Gweaton (approved):
Looks good - a couple of minor comments

spec/requests/rollout_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
+
+RSpec.describe "Rollout UI", type: :request do
+  before do
+    stub_const("ROLLOUT", Rollout.new(Redis.new))
```
Gweaton, line 5: 🥇

Gweaton (approved):
Oof, well done for figuring out that Redis stuff - saved me a lot of pain this morning!

yndajas (approved):
Looks good! Just a couple of minor comments and a question

spec/requests/rollout_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    Rollout::UI.configure do
+      instance { ROLLOUT }
+    end
+  end
+
+  context "for a visitor" do
+    it "returns 404" do
+      expect { get "/rollout" }.to raise_error(ActionController::RoutingError)
```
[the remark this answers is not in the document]
CristinaRO: It's not strictly equivalent, in that we couldn't test the `response` object in this case, as there is no response. The constraint acts in such a way that there is effectively no route defined when not logged in as a service owner, and calling `get "/rollout"` results in a routing error.

spec/rollout_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    expect(ROLLOUT.groups).to include(:partner_organisation_users)
```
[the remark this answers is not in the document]
CristinaRO: There's something about knowing exactly which part fails (in case there's a failure), but it doesn't really apply here, so I've changed it.

spec/requests/rollout_spec.rb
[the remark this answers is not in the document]
yndajas: Ahh, makes sense - thanks!

PR #1828, opened 2022-10-20
(2662) Ensure that feature flagged reports are not visible

CristinaRO:
## Changes in this PR
- Hide ISPF reports when the ISPF feature flag is enabled

### After
TBA

## Next steps


CristinaRO:
[Ensure that reports are not visible and cannot be created when feature flag is off](https://trello.com/c/na5jmqkV/2662-ensure-that-reports-are-not-visible-and-cannot-be-created-when-feature-flag-is-off)

PR #1834, opened 2022-10-24
(2662) Use feature flag to hide ISPF reports

CristinaRO:
## Changes in this PR
- Hide ISPF reports when the ISPF feature flag is enabled

### Before
Screenshot 2022-10-24 at 14 48 53
Screenshot 2022-10-24 at 14 51 31

### After
Screenshot 2022-10-24 at 14 49 38
Screenshot 2022-10-24 at 14 51 43

## Next steps

### DIFF
--- a/config/initializers/rollout.rb
+++ b/config/initializers/rollout.rb
@@ -13,3 +13,11 @@ end
 Rollout::UI.configure do
   instance { ROLLOUT }
 end
+
+def hide_ispf_for_group?(user_group)
+  ROLLOUT.get(:ispf_fund_in_stealth_mode).groups.include?(user_group)
+end
+
+def hide_ispf_for_user?(user)
+  ROLLOUT.active?(:ispf_fund_in_stealth_mode, user)
+end
--- a/spec/rollout_spec.rb
+++ b/spec/rollout_spec.rb
@@ -1,12 +1,12 @@
 RSpec.describe "Rollout" do
-  it "has two custom groups for BEIS users and PO users" do
+  it "has two custom groups for BEIS users and PO users", :use_original_rollout do
     expect(ROLLOUT.groups).to include(:beis_users, :partner_organisation_users)
   end
 
   context "a BEIS user" do
     let(:user) { build(:beis_user) }
 
-    it "is part of the beis_users group and not the partner_organisation_users group" do
+    it "is part of the beis_users group and not the partner_organisation_users group", :use_original_rollout do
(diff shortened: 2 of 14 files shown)


CristinaRO:
[Ensure that reports are not visible and cannot be created when feature flag is off](https://trello.com/c/na5jmqkV/2662-ensure-that-reports-are-not-visible-and-cannot-be-created-when-feature-flag-is-off)

config/initializers/rollout.rb
```diff
@@ -13,3 +13,11 @@
 Rollout::UI.configure do
   instance { ROLLOUT }
 end
+
+def ispf_in_stealth_mode_for_group?(user_group)
```
Gweaton, line 17: Would it be a pain to call these `show(/hide)_ispf_for_group?` and `show(/hide)_ispf_for_user?` There's something about my brain that requires a few re-reads `in_stealth_mode` to determine which way the boolean is pointing, but I can get over it if it's a faff.

spec/rollout_spec.rb
```diff
@@ -15,9 +15,29 @@
...
+      expect(ispf_in_stealth_mode_for_group?(:real_group)).to be true
+      expect(ispf_in_stealth_mode_for_group?(:fake_group)).to be false
+    end
+  end
+
+  describe "#ispf_in_stealth_mode_for_user?" do
+    let(:user) { double(:user) }
+
+    it "provides a more readable interface to check if the feature is enabled for that user" do
+      expect(ROLLOUT).to receive(:active?).with(:ispf_fund_in_stealth_mode, user)
```
Gweaton, line 38: Any reason you went for the `expect(x).to receive` rather than calling the method followed by `expect(x).to have_received`? More for my own curiosity than this not being the correct approach.

Gweaton (approved):
This is great - big fan of the `not_ispf` scope, it's nicely readable and does what it says on the tin.

I have a couple of pernickety comments but I'm not too precious about them - do as you see fit.

config/initializers/rollout.rb
CristinaRO: Not a faff, and it's a fair comment!

spec/rollout_spec.rb
CristinaRO: No particular reason, probably contagion from the spec above, and lack of a strong preference for `have_received`.

I used to be a proponent of `have_received` for reasons of following the arrange-act-assert order, but I think `have_received` didn't work for message chains so I started using it less. Not the case here, but the style stuck.

PR #1838, opened 2022-10-25
(2666) Exclude ISPF activities from search results

CristinaRO:
## Changes in this PR
- ISPF activities are not shown in search results when the feature flag is enabled

### Before
Screenshot 2022-10-25 at 15 20 14

### After
Screenshot 2022-10-25 at 15 20 27

## Next steps

- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [x] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.
- [ ] Do any environment variables need amending or adding?
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/xml-validation.md)

PR #1840, opened 2022-10-25
(2663) Exclude ISPF-related entities from exports when the feature flag is enabled

CristinaRO:
**NB* Builds upon #1838 

## Changes in this PR
- Exclude ISPF-related entities from exports when the feature flag is enabled

## Next steps


CristinaRO:
[Ensure that exports don't contain any ISPF-related entities when feature flag is off](https://trello.com/c/qaHjJptV/2663-ensure-that-exports-dont-contain-any-ispf-related-entities-when-feature-flag-is-off)

spec/models/fund_spec.rb
```diff
@@ -89,6 +89,18 @@
...
 
+  describe ".not_ispf" do
+    it "returns a Fund for every entry in the 'fund_types' codelist except for ISPF" do
+      codelist = Codelist.new(type: "fund_types", source: "beis")
+
+      funds = described_class.not_ispf
+      expect(funds.size).to eq(codelist.list.size - 1)
+
+      ids = funds.map(&:id)
+      expect(ids).to eq(codelist.values_for("code") - [4])
```
Gweaton, line 100: I know we'll be removing this in future so it's not a huge deal, but if this were going to be sticking around for a while, I think I'd rather this were an explicit list of the codelist values we're expecting. The values `- [4]` is pretty much the implementation code itself - we could store them in a `non_ispf_codelists` variable to make it clear what they mean.

Gweaton (approved):

PR #1841, opened 2022-10-25
(2677) Add Level B comments to budget exports

yndajas:
## Changes in this PR

This adds Level B activity comments to the budget exports in a new column. There is no demand for comments for other levels in budget exports at present

I've extended one test to include a programme activity with two comments (since it already created a programme activity), but not the test above it (`downloading budgets for a partner organisation`) - is it worth doing this?

For the `export_spec`, I've updated the `#activity` tests. I notice that the private methods don't have tests - is this standard? I haven't added tests for `#activity_comments` for now

### Before

image

### After

image

## Next steps

### DIFF
--- a/app/services/budget/export.rb
+++ b/app/services/budget/export.rb
@@ -1,5 +1,7 @@
 class Budget
   class Export
+    include CommentsHelper
+
     HEADERS = [
       "RODA identifier",
       "Partner organisation identifier",
@@ -56,10 +58,6 @@ class Budget
         activity.title,
         activity.programme? ? comments_formatted_for_csv(activity.comments) : ""
       ]
-    end
-
-    def comments_formatted_for_csv(comments)
-      comments.pluck(:body).map(&:strip).join("|")
     end
 
     def budget_data(budgets)
--- a/app/services/budget/export.rb
+++ b/app/services/budget/export.rb
@@ -5,7 +5,8 @@ class Budget
       "Partner organisation identifier",
       "Partner organisation",
       "Level",
-      "Title"
+      "Title",
+      "Level B activity comments"
     ]
 
     def initialize(source_fund:, organisation: nil)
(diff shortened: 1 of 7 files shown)


yndajas:
[Include Level B comments in budget reports](https://trello.com/c/p22vtdg0/2677-include-level-b-comments-in-budget-reports)

app/services/budget/export.rb
```diff
@@ -52,10 +55,17 @@ def activity_data(activity)
...
         I18n.t("table.body.activity.level.#{activity.level}"),
-        activity.title
+        activity.title,
+        activity.programme? ? activity_comments(activity) : ""
       ]
     end
 
+    def activity_comments(activity)
+      skip_bullet do
+        activity.comments.map { |comment| comment.body.strip }.join("|")
```
Gweaton, line 65: If we handle this in a similar way to how we do in `app/models/export/activity_comments_column.rb`(using`:pluck`) we don't need to skip bullet here. 

So we could do `activity.comments.pluck(:body).map(&:strip).join("|")`

Although we're `:pluck`ing then `map`ping, it appears to be the tiniest bit faster on average after running the tests a few times, but that could just be luck.

spec/features/beis_users_can_download_exports_spec.rb
```diff
@@ -169,6 +171,7 @@
         "Partner organisation" => partner_organisation.name,
         "Level" => "Project (level C)",
         "Title" => project.title,
+        "Level B activity comments" => "",
```
Gweaton, line 174: Might be worth adding a single comment here, just so it's covered for both types of download.

app/services/budget/export.rb
```diff
@@ -52,10 +55,17 @@ def activity_data(activity)
         activity.partner_organisation_identifier,
         activity.extending_organisation&.name,
         I18n.t("table.body.activity.level.#{activity.level}"),
-        activity.title
+        activity.title,
+        activity.programme? ? activity_comments(activity) : ""
       ]
     end
 
+    def activity_comments(activity)
```
Gweaton, line 63: As you mentioned in your PR description, we don't test private methods explicitly. Instead we test them indirectly via the methods they're called e.g. we're covering this with tests for `rows`.

Gweaton (approved):
Looks good to me - a couple of minor comments.

spec/features/beis_users_can_download_exports_spec.rb
yndajas: Left this in as a fixup for now - I think we should test this, but it does mean adding a programme, budget and comment

PR #1843, opened 2022-10-26
2669 add start of new ispf level b activity journey

CristinaRO:
## Next steps


CristinaRO:
[Add start of new ISPF level B activity journey and "Is ODA or Non-ODA" form step](https://trello.com/c/viE0TCPT/2669-add-start-of-new-ispf-level-b-activity-journey-and-is-oda-or-non-oda-form-step)

app/models/activity.rb
```diff
@@ -10,6 +10,7 @@ class Activity < ApplicationRecord
   DESERTIFICATION_POLICY_MARKER_CODES = Codelist.new(type: "policy_significance_desertification", source: "beis").hash_of_integer_coded_names
 
   FORM_STEPS = [
+    :is_oda,
```
Gweaton, line 13: I think this might have slipped into this commit by mistake?

app/views/activity_forms/is_oda.html.haml
```diff
@@ -0,0 +1,6 @@
+= render layout: "wrapper" do |f|
+  = f.govuk_error_summary
+  = f.govuk_radio_buttons_fieldset(:is_oda, legend: { text: t("form.legend.activity.is_oda"), size: "xl", tag: "h1" }) do
+    = f.hidden_field :is_oda, value: nil
```
Gweaton, line 4: I'm not sure I understand - what does this hidden field do?

config/locales/models/activity.en.yml
```diff
@@ -110,6 +110,9 @@ en:
         fstc_applies:
           "true": "Yes (1)"
           "false": "No (0)"
+        is_oda:
+          "true": "ODA (1)"
```
Gweaton, line 114: Is there anywhere else we display the numeric values other than `fstc_applies`? I can see it might be useful to associate the numbers with the bulk upload, but without that context, I wonder if it might be confusing to users? Maybe this is something we can check with BEIS or test as a post-MVP iteration.

config/locales/models/activity.en.yml
```diff
@@ -385,6 +392,7 @@ en:
         geography: Will the benefitting recipient be a region or country?
         identifier: Enter your unique identifier
         intended_beneficiaries: What are the intended beneficiaries?
+        is_oda: Is it ODA or Non-ODA?
```
Gweaton, line 395: Can we make this `"Is the activity ODA or Non-ODA"`? I appreciate we're working from minimal content in the designs, but I think `it` is a bit less specific.

app/presenters/activity_presenter.rb
```diff
@@ -197,7 +197,7 @@ def oda_eligibility
   end
 
   def call_to_action(attribute)
-    if send(attribute).blank?
+    if send(attribute).nil?
```
Gweaton, line 200: Would you mind adding some detail to the commit message about why this has changed? For future readers reading this commit message, a bit more context would be useful.

Gweaton (approved):
Looks good - a few things that would be good to tweak before merging this in.

app/views/activity_forms/is_oda.html.haml
CristinaRO: It makes sure that the `params` are submitted even if the user presses Continue without selecting a radio button. Otherwise we get no payload, and then `param is missing or the value is empty: activity`.

config/locales/models/activity.en.yml
CristinaRO: Does this mean "ok, keep", or "remove from current PR"?

app/views/activity_forms/is_oda.html.haml
CristinaRO: I've added some detail about this in the commit message.

PR #1847, opened 2022-10-31
(2671) Add Level B ISPF theme form step

yndajas:
## Changes in this PR

Adds Level B ISPR theme to the Activity model and to the UI multi-step form

### After

image

## Next steps

### DIFF
--- a/app/views/activity_forms/ispf_theme.html.haml
+++ b/app/views/activity_forms/ispf_theme.html.haml
@@ -1,0 +1,6 @@
+= render layout: "wrapper" do |f|
+  = f.govuk_collection_radio_buttons :ispf_theme,
+    ispf_theme_options,
+    :code,
+    :description,
+    legend: { tag: 'h1', size: 'xl', text: t("form.legend.activity.ispf_theme") }
--- a/db/seeds/activity.rb
+++ b/db/seeds/activity.rb
@@ -1,27 +1,27 @@
 beis = Organisation.service_owner
 
-gcrf_fund_params = FactoryBot.build(:fund_activity,
+gcrf_fund_params = FactoryBot.build(:fund_activity, :gcrf,
   roda_identifier: "GCRF",
   title: "Global Challenges Research Fund (GCRF)",
   organisation: beis).attributes
 
 gcrf_fund = Activity.find_or_create_by(gcrf_fund_params)
 
-newton_fund_params = FactoryBot.build(:fund_activity,
+newton_fund_params = FactoryBot.build(:fund_activity, :newton,
   roda_identifier: "NF",
   title: "Newton Fund",
   organisation: beis).attributes
 
 _newton_fund = Activity.find_or_create_by(newton_fund_params)
 
-ooda_fund_params = FactoryBot.build(:fund_activity,
+ooda_fund_params = FactoryBot.build(:fund_activity, :ooda,
   roda_identifier: "OODA",
   title: "Other ODA",
   organisation: beis).attributes
 
 _ooda_fund = Activity.find_or_create_by(ooda_fund_params)
 
-ispf_fund_params = FactoryBot.build(:fund_activity,
+ispf_fund_params = FactoryBot.build(:fund_activity, :ispf,
   roda_identifier: "ISPF",
   title: "International Science Partnerships Fund",
   organisation: beis).attributes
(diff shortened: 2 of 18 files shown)


spec/support/activity_form.rb
```diff
@@ -9,8 +9,12 @@ def initialize(activity:, fund:, level:)
...
   end
 
-  def complete!
-    send("fill_in_#{fund}_#{level}_activity_form")
+  def complete!(is_oda: nil)
+    if is_oda.nil?
+      return send("fill_in_#{fund}_#{level}_activity_form")
+    end
+
+    send("fill_in_#{fund}_#{level}_activity_form", is_oda: is_oda)
```
yndajas, line 17: I wonder if this should be the other way around (`if is_oda ... end ... default`), given `nil` is the default value?

yndajas (commented):
Looks good. Just a few small details in the comments. Also, do we need some project-level feature tests?

spec/controllers/activity_forms_controller_spec.rb
```diff
@@ -39,6 +39,18 @@
         end
       end
 
+      context "ispf_theme step" do
+        subject { get_step :ispf_theme }
+
+        it { is_expected.to skip_to_next_step }
```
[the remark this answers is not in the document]
Gweaton: I think it would be, but let's do it as a separate piece of work to keep this PR small. Would you mind adding a trello card to the technical debt column for this?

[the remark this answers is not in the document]
Gweaton: Ah, thanks for clearing that up Cristina - you're right. Sorry @yndajas I should've looked a bit more closely at that implementation before asking you to write up that card

app/views/activity_forms/ispf_theme.html.haml
```diff
@@ -0,0 +1,6 @@
+= render layout: "wrapper" do |f|
+  = f.govuk_collection_radio_buttons :ispf_theme,
```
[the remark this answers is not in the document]
CristinaRO: @Gweaton have you tested what happens if you submit the form without selecting any of the options? (I am assuming that no option is pre-selected.)

spec/controllers/activity_forms_controller_spec.rb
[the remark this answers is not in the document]
yndajas: Oh, very interesting! I did also wonder how it got the actual/activity, but it makes more sense if it's whatever action is being tested. Thanks!

spec/support/activity_form.rb
Gweaton: Have left this as-is for now.

db/seeds/activity.rb
```diff
@@ -21,7 +21,7 @@
 
 _ooda_fund = Activity.find_or_create_by(ooda_fund_params)
 
-ispf_fund_params = FactoryBot.build(:fund_activity,
+ispf_fund_params = FactoryBot.build(:fund_activity, :ispf,
```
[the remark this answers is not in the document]
Gweaton: I've added this for all funds.

spec/controllers/activity_forms_controller_spec.rb
```diff
@@ -39,6 +39,18 @@
...
+
+        context "when it's an ISPF activity" do
+          let(:activity) { create(:project_activity, organisation: organisation, parent: fund, source_fund_code: Fund.by_short_name("ISPF").id) }
```
yndajas, line 48: Given that the project side of this is coming in a later ticket, should this be a programme activity here? It seems like project might be correct in this context though

spec/presenters/activity_presenter_spec.rb
```diff
@@ -514,6 +514,15 @@ def cast_code_to_field(code)
     end
   end
 
+  describe "#ispf_theme" do
+    it "returns the code list description value for the stored integer" do
+      activity = build(:project_activity, ispf_theme: 1)
```
yndajas, line 519: Likewise here

yndajas (commented):
Just a question about the project activities in the tests given this is Level B. Otherwise looks good

spec/controllers/activity_forms_controller_spec.rb
Gweaton: Good point, this should be a `programme` which means the previous tests were incorrect too. I've updated them in their own commit.

spec/presenters/activity_presenter_spec.rb
Gweaton: I've made this `programme`

CristinaRO (approved):

PR #1851, opened 2022-11-01
(2642) Add new steps to the new activity form for Level C/D ODA ISPF activities

yndajas:
## Changes in this PR

Activities now inherit the `is_oda` attribute from their parent, so Level C/D activities will have the same `is_oda` value as was set on their parent Level B activity, and therefore be shown the correct ODA countries on the ISPF partner countries step

Other than skipping the ODA/non-ODA step, everything else is as per the Level B ISPF user journey

Re: feature tests: `spec/features/users_can_create_a_project_level_activity_spec.rb` isn't as comprehensive as `spec/features/beis_users_can_create_a_programme_level_activity_spec.rb` in testing different funds and the form steps, but perhaps we should expand it?

## Next steps

### DIFF
--- a/spec/services/activity_defaults_spec.rb
+++ b/spec/services/activity_defaults_spec.rb
@@ -9,6 +9,11 @@ RSpec.describe ActivityDefaults do
   let(:project) { create(:project_activity, :gcrf_funded, parent: programme) }
   let(:third_party_project) { create(:third_party_project_activity, :gcrf_funded, parent: project) }
 
+  let(:ispf_oda_programme) { create(:programme_activity, :ispf_funded) }
+  let(:ispf_non_oda_programme) { create(:programme_activity, :ispf_funded, is_oda: false) }
+  let(:ispf_oda_project) { create(:project_activity, parent: ispf_oda_programme, is_oda: true) }
+  let(:ispf_non_oda_project) { create(:project_activity, parent: ispf_non_oda_programme, is_oda: false) }
+
   let!(:current_report) { create(:report, :active, organisation: partner_organisation, fund: fund) }
 
   before do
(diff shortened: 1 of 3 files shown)


Gweaton (commented):
The card for this was a duplicate of the Level B one but I forgot to update the miro link and the content was misleading, massive apologies - the `is_oda` step isn't required here. I've updated the card with more details - sorry again about this.

spec/services/activity_defaults_spec.rb
```diff
@@ -9,6 +9,11 @@
   let(:project) { create(:project_activity, :gcrf_funded, parent: programme) }
   let(:third_party_project) { create(:third_party_project_activity, :gcrf_funded, parent: project) }
 
+  let(:ispf_oda_programme) { create(:programme_activity, :ispf_funded) }
+  let(:ispf_non_oda_programme) { create(:programme_activity, :ispf_funded, is_oda: false) }
+  let(:ispf_oda_project) { create(:project_activity, parent: ispf_oda_programme, is_oda: true) }
+  let(:ispf_non_oda_project) { create(:project_activity, parent: ispf_non_oda_programme, is_oda: false) }
```
yndajas, line 15: Without manually setting `is_oda` here the third party projects end up with `nil` as their `is_oda` value. My intuition is that this is because `activity_defaults` is only invoked in the UI user journey flow, so when manually creating a project in RSpec the defaults aren't added. Does that sound correct? If so, we'll need to work out how to make this behaviour the default in bulk upload as well, so I wonder if there's somewhere we can set even more default (baked in) parent-based defaults

yndajas:
[Add new steps to the new activity form for Level C & D ODA ISPF activities](https://trello.com/c/Mxc2Hifa/2642-add-new-steps-to-the-new-activity-form-for-level-c-d-oda-ispf-activities)

spec/services/activity_defaults_spec.rb
Gweaton: As discussed on slack:
> `activity_defaults` is called in `new_child` which is currently called by the (current) bulk importer, as well as the `activity_children_controller` so it should set those values by default. When we create an activity in the specs we're using the factory, so that method isn't called.

This means we should be covered when we do get around to bulk importing these activities

Gweaton (approved):

PR #1852, opened 2022-11-01
(2643) Add Level C/D non-ODA ISPF form journey

yndajas:
## Changes in this PR

- Adds Level C/D non-ODA ISPF form journey
- Adds feature tests covering both ODA and non-ODA ISPF Level C/D UI user journeys
- Adds unit tests for the public methods added to `Activity` - coverage here is still not incomplete, though

## Next steps

### DIFF
--- a/app/controllers/activity_forms_controller.rb
+++ b/app/controllers/activity_forms_controller.rb
@@ -40,7 +40,7 @@ class ActivityFormsController < BaseController
     when :aid_type
       skip_step unless @activity.requires_aid_type?
     when :collaboration_type
-      skip_step if @activity.fund?
+      skip_step unless @activity.requires_collaboration_type?
       skip_step unless Activity::Inference.service.editable?(@activity, :collaboration_type)
       assign_default_collaboration_type_value_if_nil
     when :sustainable_development_goals
--- a/app/controllers/activity_forms_controller.rb
+++ b/app/controllers/activity_forms_controller.rb
@@ -58,6 +58,8 @@ class ActivityFormsController < BaseController
       skip_step unless @activity.is_ispf_funded?
     when :benefitting_countries
       skip_step unless @activity.requires_benefitting_countries?
+    when :gdi
+      skip_step unless @activity.requires_gdi?
     end
 
     render_wizard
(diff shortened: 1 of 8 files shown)


yndajas:
[Add new steps to the new activity form for Level C & D Non-ODA ISPF activities](https://trello.com/c/pqjEpgJv/2643-add-new-steps-to-the-new-activity-form-for-level-c-d-non-oda-ispf-activities)

app/controllers/activity_forms_controller.rb
```diff
@@ -16,30 +16,47 @@ def show
     case step
     when :is_oda
       skip_step unless @activity.requires_is_oda?
+    when :identifier
```
Gweaton, line 19: 👍

app/models/activity.rb
```diff
@@ -497,20 +497,56 @@ def latest_forecasts
     ForecastOverview.new(self).latest_values
   end
 
+  def requires_objectives?
```
Gweaton, line 500: Do you think we could have some unit tests for the public methods in this file please? I know test coverage is a bit lacking for this file, but it'd be good to make sure we're covering all new methods.

I know it's going to be a bit of a pain, but for any with the same implementation, you could do it with handy loop e.g.

```
["requires_objectives", "requires_gdi"].each do |method|
  describe(method) do
    context "[the contexts]" do
       it "returns true" do
         # call the method with send
       end
    end
  end
end

Comment (Gweaton) [reply app/models/activity.rb:500]: Don't worry about rebasing this stuff in, as it'll be a bit of a pain - I'm happy for it to go in its own commit at the end

**Review by Gweaton — COMMENTED** [review]
This looks good to me, thanks for picking it up - I think it'd be good to add tests for this, but I hope that's not too horrible an imposition as they're all pretty similar!

File: spec/models/activity_spec.rb [hunk 3]
```diff
@@ -1813,4 +1813,208 @@ def create_report(organisation, fund)
...
+      context "when activity is a programme" do
+        let(:activity) { build(:programme_activity) }
+
+        it "returns true" do
+          expect(activity.send(method.to_sym)).to eq(true)
+        end
+      end
+
+      ["project", "third-party project"].each do |level|
+        factory = (level.underscore.parameterize(separator: "_") + "_activity").to_sym
```
Gweaton, line 1836: Shall we extract this into a helper method in this test file as we're using it in lots of these tests? (could call it something like`def activity_level_name(level)`)?

Gweaton (approved):
Thanks for writing all those tests - hope it wasn't too painful!

Just one thing we can extract, but otherwise this is good to go 👍

PR #1854, opened 2022-11-02
(2456) Require implementing org for ISPF level D activities

CristinaRO:
## Changes in this PR
- Require ISPF third-party projects (level D) to have at least one implementing organisation set through the new activity journey

### After
Screenshot 2022-11-02 at 18 32 51

## Next steps

### DIFF
--- a/app/controllers/activity_forms_controller.rb
+++ b/app/controllers/activity_forms_controller.rb
@@ -67,6 +67,9 @@ class ActivityFormsController < BaseController
       skip_step unless @activity.requires_oda_eligibility_lead?
     when :uk_po_named_contact
       skip_step unless @activity.is_project?
+    when :implementing_organisation
+      skip_step unless @activity.requires_implementing_organisation?
+      @implementing_organisations = Organisation.active.sorted_by_name
     end
 
     render_wizard
@@ -80,8 +83,15 @@ class ActivityFormsController < BaseController
     updater = Activity::Updater.new(activity: @activity, params: params)
     updater.update(step)
 
-    if step == :dates && @activity.errors.present?
-      render_step :dates
+    if @activity.errors.present?
+      case step
+      when :dates
+        render_step :dates
+      when :implementing_organisation
+        @implementing_organisations = Organisation.active.sorted_by_name
+        render_step :implementing_organisation
+      end
+
       return
     end
 
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -38,7 +38,8 @@ class Activity < ApplicationRecord
     :channel_of_delivery_code,
     :oda_eligibility,
     :oda_eligibility_lead,
-    :uk_po_named_contact
+    :uk_po_named_contact,
+    :implementing_organisation
   ]
 
   VALIDATION_STEPS = [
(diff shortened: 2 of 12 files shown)


CristinaRO:
[Select an implementing organisation (mandatory step) when adding a new ISPF activity](https://trello.com/c/PQeXnb5e/2456-select-an-implementing-organisation-mandatory-step-when-adding-a-new-ispf-activity)

app/controllers/activity_forms_controller.rb
```diff
@@ -74,6 +77,12 @@ def update
       return
     end
 
+    if step == :implementing_organisation && @activity.errors.present?
```
Gweaton, line 80: I'm trying to decide whether we want to have this behaviour covered in the controller tests - I noticed there aren't any feature tests for this new step in the form, is that because this is covered enough in the new `show` controller tests?

I think at the very least it would be good to have this step covered in `spec/features/users_can_create_a_third_party_project_level_activity_spec.rb` - what do you think?

Gweaton (commented):
Thanks for sorting this - it looks good to me. I think it'd be good to add a bit more test coverage for the new step in the form, but let me know if you have strong feelings the other way

app/controllers/activity_forms_controller.rb
CristinaRO: I'm trying to find the best balance between adding slow expensive tests and making sure all the paths are covered! This unhappy path in particular could only be reached if an org is made inactive while a user is in the process of assigning it as implem org, so a minimal amount of testing seems ideal.

app/models/activity.rb
```diff
@@ -537,6 +541,10 @@ def requires_country_partner_organisations?
     is_newton_funded? && programme?
   end
 
+  def requires_implementing_organisation?
+    is_project? && is_ispf_funded? && !has_implementing_organisations?
```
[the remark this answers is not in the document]
CristinaRO: Good catch, it's a leftover from the initial assumptions that it would cover C and D!

app/controllers/activity_forms_controller.rb
Gweaton: I think adding to the existing `spec/features/users_can_create_a_third_party_project_level_activity_spec.rb` to visit this page and set an implementing organisation (the happy path) is probably enough for me.

app/controllers/activity_forms_controller.rb
```diff
@@ -74,6 +77,12 @@ def update
...
+      @implementing_organisations = Organisation.active.sorted_by_name
+      render_step :implementing_organisation
+      return
+    end
```
yndajas, line 84: Is this essentially "repeat the step if there are errors"? Although I think all the steps do that by default, so maybe not. Is there a way of clarifying what this is doing/why it doesn't live inside the `case` statement, perhaps by abstracting into a method?

yndajas (approved):
Looks good. Just one comment on clarifying what's going on in the controller

app/controllers/activity_forms_controller.rb
Gweaton: Have added a test for this.

yndajas (approved):
Looks good - just one thing about some inconsistent test methods

spec/features/users_can_create_a_third_party_project_level_activity_spec.rb
```diff
@@ -202,7 +208,7 @@
         expect(created_activity.ispf_partner_countries).to match_array(activity.ispf_partner_countries)
         expect(created_activity.ispf_theme).to eq(activity.ispf_theme)
         expect(created_activity.uk_po_named_contact).to eq(activity.uk_po_named_contact)
-        expect(created_activity.implementing_organisations).to be_none
+        expect(created_activity.implementing_organisations).to eq(activity.implementing_organisations)
```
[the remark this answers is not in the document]
Gweaton: Ah well spotted - have updated both to use this implementation.

PR #1855, opened 2022-11-02
(2694) Do not allow removing the last implementing org

CristinaRO:
## Changes in this PR
- Do not allow removing the implementing organisation from a level C or D ISPF activity if it is the only one

### Before
Screenshot 2022-11-03 at 17 30 53

### After
Screenshot 2022-11-03 at 17 29 45

## Next steps

### DIFF
--- a/spec/features/users_can_view_an_activitys_details_spec.rb
+++ b/spec/features/users_can_view_an_activitys_details_spec.rb
@@ -17,7 +17,7 @@ RSpec.feature "Users can view an activity's details" do
         expect(page).to have_content "Details"
       end
       expect(page).to have_content activity.title
-      expect(page).to have_link t("page_content.activity.implementing_organisation.button.new")
+      expect(page).to have_content t("page_content.activity.implementing_organisation.heading")
     end
 
     scenario "activities have human readable date format" do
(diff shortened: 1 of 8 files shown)


CristinaRO:
[Do not allow removing the last implementing org](https://trello.com/c/3vlY1zJr/2694-do-not-allow-removing-the-last-implementing-org)

spec/features/users_can_view_an_activitys_details_spec.rb
```diff
@@ -17,7 +17,7 @@
         expect(page).to have_content "Details"
       end
       expect(page).to have_content activity.title
-      expect(page).to have_link t("page_content.activity.implementing_organisation.button.new")
+      expect(page).to have_content t("page_content.activity.implementing_organisation.heading")
```
Gweaton, line 20: Has this slipped in from another commit?

Gweaton (approved):

spec/features/users_can_view_an_activitys_details_spec.rb
CristinaRO: Nope! This spec doesn't set up an editable report, and after fixing the policy checks, implem orgs cannot be added while there's no editable report.

PR #1856, opened 2022-11-03
(2697) Update Level B ISPF journey

yndajas:
## Changes in this PR

This removes collaboration type and FSTC applies from the Level B UI user journey for creating ISPF activities (ODA and non-ODA). We also identified that channel of delivery code should not appear for such activities, however `requires_channel_of_delivery_code?` already narrows this step down to project/third-party project activities

## Next steps


yndajas:
[Confirm FSTC, collab and channel are not in ISPF Level B activity journey](https://trello.com/c/7J10dhPm/2697-confirm-fstc-collab-and-channel-are-not-in-ispf-level-b-activity-journey)

app/models/activity.rb
```diff
@@ -534,15 +534,22 @@ def comments_for_report(report_id:)
   end
 
   def requires_collaboration_type?
```
Gweaton, line 536: Do you think we could have some tests for these? Particularly where we have multiple conditions, it always feels good to have this covered and documented in a unit test.

Gweaton (approved):
Looks good - just one comment on tests but feel free to merge after that

PR #1863, opened 2022-11-07
(2618) Non-deterministic actuals spec

yndajas:
## Changes in this PR

[hoping this is the intended outcome for this ticket!]

Previously this spec test checked an attribute of the first and second elements in a collection (Active Record Relation), however it was identified that the order of the collection cannot always be guaranteed, therefore the tests need to be adaptable enough to deal with the collection being randomly ordered

This moves away from testing specific expectations on the nth element of the collection, and instead tests that the collection as a whole contains the expected values for the given attribute/association (comment)

## Next steps


yndajas:
[Non-deterministic spec](https://trello.com/c/ZH0pXnIY/2618-non-deterministic-spec)

Gweaton (approved):

PR #1864, opened 2022-11-07
Fix db-restore and set OTP

yndajas:
## Changes in this PR

- Fixes `db-restore` error and typo
- Integrates setting `otp_required_for_login` into local development user seeds, removing the need for devs to do this manually every time they sync with production data

## Next steps

### DIFF
--- a/db/seeds/development_users.rb
+++ b/db/seeds/development_users.rb
@@ -3,7 +3,7 @@ administrator = User.find_or_initialize_by(
   email: "roda@dxw.com"
 )
 partner_org_user = User.find_or_initialize_by(
-  name: "artner Org User",
+  name: "Partner Org User",
   email: "roda+dp@dxw.com"
 )
 
--- a/db/seeds/development_users.rb
+++ b/db/seeds/development_users.rb
@@ -7,13 +7,13 @@ partner_org_user = User.find_or_initialize_by(
   email: "roda+dp@dxw.com"
 )
 
-administrator.password = "LlEeTtMmEeIiNn!1"
-partner_org_user.password = "LlEeTtMmEeIiNn!1"
-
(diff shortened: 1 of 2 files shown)


db/seeds/development_users.rb
```diff
@@ -3,12 +3,13 @@
...
-  name: "artner Org User",
+  name: "Partner Org User",
   email: "roda+dp@dxw.com"
 )
 
-administrator.password = "LlEeTtMmEeIiNn!1"
-partner_org_user.password = "LlEeTtMmEeIiNn!1"
+[administrator, partner_org_user].each do |user|
+  user.update(password: "LlEeTtMmEeIiNn!1", otp_required_for_login: false)
+end
```
yndajas, line 12: Tried using `update_all` here but it didn't work - I think because this is a regular array, rather than an Active Record Relation/Collection

db/seeds/development_users.rb
```diff
@@ -3,7 +3,7 @@
   email: "roda@dxw.com"
 )
 partner_org_user = User.find_or_initialize_by(
-  name: "artner Org User",
+  name: "Partner Org User",
```
Gweaton, line 6: Hah, this has been bugging me for way too long. Thanks for doing something about it

db/seeds/development_users.rb
Gweaton: We've got a bit of a mix of approaches going on here - previously we set the fields individually and saved at the end (lines 16 & 19) whereas here we're doing a couple of saves. Feel free to ignore me here, but I wonder if it's worth picking one approach here: save at the end _or_ update all in one go.

Gweaton (approved):
Thanks for sorting

PR #1865, opened 2022-11-07
(2693) Hide fields from activity details based on `requires_X?` methods

yndajas:
## Changes in this PR

This uses all the new and existing `requires_X?` instance methods in the `Activity` class to determine whether to show fields in the activity details tab/page. One or two of these were already being used, while on some other fields the old logic of the `requires_X?` method was simply being replicated in the view - this brings the logic up to date and keeps the code DRY

It also updates the logic for displaying SDGs, however this does not use a `requires_X?` method. Since the validation of `sdg_1` is based on `sdgs_apply?` rather than a `requires_X?` method, the logic as to whether to skip this form step and show it in the view are kept separate/plain to avoid confusion about the difference between `sdgs_apply?` and `requires_sdgs?`

## Screenshots of UI changes

Example is non-ODA, which has the most changes

### Before

https://user-images.githubusercontent.com/40244233/200399448-d8961d61-73c3-44d0-9888-c8bdb64ff397.mov

### After

https://user-images.githubusercontent.com/40244233/200397516-5144a2c1-5a36-40e6-a90a-a8911f3de2f5.mov

## Next steps

### DIFF
--- a/app/views/shared/activities/_activity.html.haml
+++ b/app/views/shared/activities/_activity.html.haml
@@ -78,7 +78,7 @@
       - if policy(activity_presenter).update? && step_is_complete_or_next?(activity: activity_presenter, step: :purpose)
         = a11y_action_link(I18n.t("default.link.#{activity_presenter.call_to_action(:description)}"), activity_step_path(activity_presenter, :purpose), I18n.t("activerecord.attributes.activity.description").downcase)
 
-  - unless activity_presenter.fund?
+  - if activity_presenter.requires_objectives?
     .govuk-summary-list__row.objectives
       %dt.govuk-summary-list__key
         = t("activerecord.attributes.activity.objectives")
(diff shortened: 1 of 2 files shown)


yndajas:
[Remove ISPF-unused fields from activity details tab](https://trello.com/c/Gn1rjYyf/2693-remove-ispf-unused-fields-from-activity-details-tab)

app/views/shared/activities/_activity.html.haml
```diff
@@ -166,19 +166,18 @@
         - if policy(activity_presenter).update? && step_is_complete_or_next?(activity: activity_presenter, step: :programme_status)
           = a11y_action_link(t("default.link.#{activity_presenter.call_to_action(:programme_status)}"), activity_step_path(activity_presenter, :programme_status), t("activerecord.attributes.activity.programme_status"))
 
-  - if activity_presenter.is_newton_funded?
```
Gweaton, line 169: This should go back in now that we've gone back on `requires_country_partner_organisations?`

Gweaton (commented):
Just the one thing to fix

PR #1867, opened 2022-11-08
[Support] Set default currency & transaction_type on Actuals

Gweaton:
## Changes in this PR

Sets default currency on all classes that inherit from `Transaction` (specifically to set this on `Actual`s).

Previously, this was only done for `Adjustment`s and `Refund`s, but `Actuals` are also a type of `Transaction` and are exported in IATI XML.

Because `Actual`s didn't have a default value for `currency`, this meant it was outputting XML tags with an empty string value e.g.

```
<transaction>
<transaction-type code="3"/>
<value currency="" value-date="2022-03-31">1000</value>
```

I'm not sure why the default value wasn't added for Actuals [here](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1383), it may have been an oversight, but as we export all `Transaction`-inheritors to the XML, I thought it would be safest to put the default values on the `Transaction` class itself so we don't forget this in future.

Although the `transaction_type` doesn't appear to be an issue for IATI XML (as far as I'm aware) it feels best to set this to the default value ("3") as this seems to be the only real value we set it to in the codebase.

We'll want to re-run the data migration to backfill these values: `db/data/20211004_backfill_missing_attrs_on_refunds_and_adjustments.rb` on production to fix this.

### Before

image

### After (after running backfilling script)

image

## Next steps

- [ ] Run data migration script

### DIFF
--- a/db/data/20221108103513_backfill_missing_currencies_and_transaction_types.rb
+++ b/db/data/20221108103513_backfill_missing_currencies_and_transaction_types.rb
@@ -1,0 +1,33 @@
+# Run me with `rails runner db/data/20221108103513_backfill_missing_currencies_and_transaction_types.rb`
+
+# Transactions (actuals, refunds and adjustments) were being created with blank attributes:
+# - currency
+# - transaction_type
+#
+# This isn't causing an immediate problem, but they are "transactions" and so
+# ought to conform to te normal transaction behaviour.
+
+# This is adapted from db/data/20211004_backfill_missing_attrs_on_refunds_and_adjustments.rb
+
+finder = Transaction.where(currency: nil, transaction_type: nil)
+
+puts "Setting currency and transaction_type on #{finder.count} transactions..."
+
+finder.each do |transaction|
+  transaction.update_columns(
+    transaction_type: Transaction::TRANSACTION_TYPE_DISBURSEMENT,
+    currency: "GBP"
+  )
+end
+
+puts "-> there are now #{finder.reload.count} transactions where currency and transaction_type are nil"
+
+finder = Transaction.where(currency: nil)
+
+puts "Setting currency on #{finder.count} transactions..."
+
+finder.each do |transaction|
+  transaction.update_columns(currency: "GBP")
+end
+
+puts "-> there are now #{finder.reload.count} transactions where currency is nil"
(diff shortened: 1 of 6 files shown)


db/data/20221108103513_backfill_missing_currencies_and_transaction_types.rb
```diff
@@ -0,0 +1,35 @@
...
+puts "Setting currency and transaction_type on #{finder.count} transactions..."
+
+finder.each do |transaction|
+  transaction.update_columns(
+    transaction_type: Transaction::TRANSACTION_TYPE_DISBURSEMENT,
+    currency: "GBP"
+  )
+end
+
+puts "-> there are now #{finder.reload.count} transactions still to fix"
```
yndajas, line 23: Would this not be zero if it's the same query (at this point)? It says `-> there are now 0 transactions still to fix` but there are still transactions to fix using the second query

db/data/20221108103513_backfill_missing_currencies_and_transaction_types.rb
```diff
@@ -0,0 +1,35 @@
...
+
+finder = Transaction.where(currency: nil)
+
+puts "Setting currency on #{finder.count} transactions..."
+
+finder.each do |transaction|
+  # use `update_columns`` here as this bypasses validation errors
+  # where some older entities have fields with negative values
```
yndajas, line 31: Double backtick in the first line of this comment. Is the comment intentionally provided in this `each` block but not the earlier one?

db/data/20221108103513_backfill_missing_currencies_and_transaction_types.rb
```diff
@@ -0,0 +1,35 @@
...
+  transaction.update_columns(currency: "GBP")
+end
+
+puts "-> there are now #{finder.reload.count} transactions still to fix"
```
yndajas, line 35: Likewise - not always zero?

yndajas (approved):
Seems to work well - just a couple of communications things

db/data/20221108103513_backfill_missing_currencies_and_transaction_types.rb
Gweaton: Have reworded this

db/data/20221108103513_backfill_missing_currencies_and_transaction_types.rb
Gweaton: Have removed this after my own confusion in understanding what `update_columns` does.

PR #1870, opened 2022-11-08
(2706) Allow redacting programmes from IATI

yndajas:
## Changes in this PR

Previously redacting activities from IATI was only allowed to be done by BEIS users at Levels C/D. The client has indicated that it would also be useful to be able to do this at Level B, so this makes that possible

### Before

image

### After

image

## Next steps


yndajas:
[Do Not Publish button for Level B IATI XMLs](https://trello.com/c/JLnB5naX/2706-do-not-publish-button-for-level-b-iati-xmls)

PR #1871, opened 2022-11-08
Remove unnecessary conditional in ActivityForm

Gweaton:
## Changes in this PR

Removes unnecessary conditional in ActivityForm which was originally needed as we were passing in an argument to `send` to read from an `Activity`. It looks like we forgot to update this after removing that parameter from the method signature.

## Next steps


yndajas (approved):
Oops!

PR #1874, opened 2022-11-09
Update changelog for release 122

Gweaton:
(Tidies up the changelog, as I felt it was a bit too granular.)

- Update seeds and create data migration for adding ISPF fund entity to the service
- Add ISPF Level B UI user journeys for adding activities
- Add ISPF Level C/D UI user journeys for adding activities
- Inherit `is_oda` from the parent activity - Level C/D ISPF activities will therefore be ODA or non-ODA based on the (grand)parent Level B activity
- Do not allow removing the implementing organisation from a level C or D ISPF activity if it is the only one
- Require ISPF third-party projects (level D) to have at least one implementing organisation set through the new activity journey
- Hide ISPF-facing functionality in the service and exports when the feature flag is enabled
- Allow programmes to be redacted from IATI
- Set default values for currency and transaction type on Actual Transactions, output in IATI XML
- Remove reference to "ODA activities" in activity status descriptions

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1120,28 +1120,23 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Add Level B activity comments to budget exports
 - Strip leading/trailing whitespace and line breaks from comments in reports and exports
 
-## [unreleased]
+## [release-122] 2022-11-09
 
 - Update seeds and create data migration for adding ISPF fund entity to the service
-- Hide ISPF reports when the ISPF feature flag is enabled
-- ISPF programmes cannot be created when the feature flag is enabled
-- ISPF activities are not shown to users when the feature flag is enabled
-- ISPF activities are not shown in search results when the feature flag is enabled
-- Exclude ISPF-related entities from exports when the feature flag is enabled
-- The first step for an ISPF programme is now choosing between ODA and Non-ODA
-- Add ISPF theme form step
-- Add ISPF partner countries form step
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1120,28 +1120,23 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Add Level B activity comments to budget exports
 - Strip leading/trailing whitespace and line breaks from comments in reports and exports
 
-## [unreleased]
+## [release-122]
```
yndajas, line 1123: Today's date?

CHANGELOG.md
```diff
@@ -1120,28 +1120,23 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
...
 
 - Update seeds and create data migration for adding ISPF fund entity to the service
-- Hide ISPF reports when the ISPF feature flag is enabled
-- ISPF programmes cannot be created when the feature flag is enabled
```
yndajas, line 1127: "ISPF programmes cannot be created when the feature flag is enabled" might not be covered explicitly now, but I think it's implied by "Hide ISPF-related entities in the service and exports when the feature flag is enabled"

yndajas (approved):
Looks good - just a couple of small things

CHANGELOG.md
Gweaton: Do you think "Hide ISPF-facing functionality in the service and exports when the feature flag is enabled" would be explicit enough?

PR #1875, opened 2022-11-10
(2628) Add CSV upload for Level B ISPF ODA activities

yndajas:
## Changes in this PR

This enables BEIS users to download a CSV template for uploading Level B ISPF ODA activities, and then allows them to upload it

Uploads now have a `type` (currently `non_ispf` or `ispf_oda` - `ispf_non_oda` to follow), which is used to determine the filename and headings of the template, and whether and what to set `is_oda` (to)

Converters have been added to `Activity::Import::Converter` for ISPF fields, following the pattern of existing converters for code-based integer and string array fields

The upload views have been restructured to work with the different types

Level C/D-related code has been updated to reflect the various refactors that support this extension of functionality

### Before

https://user-images.githubusercontent.com/40244233/202456231-d9d6a6b7-aefe-4790-9cd4-96823045778d.mov

### After

https://user-images.githubusercontent.com/40244233/202456046-6e18ba23-87ee-45b9-a5ec-f5bbe1301022.mov

## Next steps


app/presenters/organisation_presenter.rb
```diff
@@ -17,10 +17,15 @@ def organisation_type
     I18n.t("organisation.organisation_type.#{super}")
   end
 
-  def filename_for_activities_template
+  def filename_for_activities_template(type:)
+    human_readable_type = {
+      ispf_oda: "ISPF_ODA",
+      non_ispf: "GCRF_NF_OODA"
+    }.fetch(type, "").tr(" ", "_")
```
yndajas, line 24: @Gweaton this `human_readable_type` will be reused elsewhere for Level C/D, but we could refactor when it comes to that or just repeat it, given it's a small bit of code?

spec/features/beis_users_can_upload_level_b_activities_spec.rb
```diff
@@ -212,6 +118,158 @@
...
+      visit organisation_activity_comments_path(organisation, new_activities.first)
+
+      expect(page).to have_text("A comment")
+
+      visit organisation_activity_comments_path(organisation, new_activities.second)
+
+      expect(page).to have_text("Another comment")
+    end
+
+    scenario "updating an existing activity" do
```
yndajas, line 184: I think this could possibly move back out of the fund/type-specific context(s), since it's not really testing anything fund-specific (just that uploads work from the error/validation view)

app/presenters/organisation_presenter.rb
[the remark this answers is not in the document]
yndajas: I've simplified it so we aren't adding a fallback, since we don't have a clear use case for needing to fall back as of yet. Currently it won't error - I moved the method for this formatting back into the importer (in anticipation of reusing it for levels C and D_, and by default it returns `nil` if the type doesn't exist, so this method will just have no type interpolated. Perhaps we should raise an error though

spec/features/beis_users_can_upload_level_b_activities_spec.rb
```diff
@@ -207,11 +158,137 @@
...
+        "ISPF partner countries",
+        "Comments"
+      ])
+    end
+
+    scenario "uploading a valid set of activities" do
+      old_count = Activity.count
+
+      within ".upload-form--ispf-oda" do
+        attach_file_and_click_submit(filepath: "spec/fixtures/csv/valid_level_b_ispf_ooda_activities_upload.csv")
```
CristinaRO, line 273: I was confused by the filename: `OODA` is a specific, non-ISPF fund. Should this be `ispf_oda`?

CristinaRO (approved):
Excellent work all around, just a couple of comments / questions!

spec/features/beis_users_can_upload_level_b_activities_spec.rb
yndajas: OOps

app/views/shared/activities/uploads/_upload_form.html.haml
```diff
@@ -7,7 +7,7 @@
         hint: { text: t("form.hint.activity.csv_file_recover_from_error_html", link: send(path_helper, instance, { type: type, format: :csv })) }
     - else
       = f.govuk_file_field :activity_csv,
-        label: { text: t("form.label.activity.csv_file"), hidden: true },
+        label: { text: t("form.label.activity.csv_file"), hidden: local_assigns[:show_hint] ? false : true },
```
yndajas, line 10: This can probably just be `!local_assigns[:show_hint]`

PR #1890, opened 2022-11-21
WIP use accessible autocomplete for linking activities

yndajas:
**Spike**

This adds accessible autocomplete to the activity linking form step

Wasn't sure where to put the options, since they don't really belong in the CodelistHelper

Currently if you use the form to change the linked activity, it will instead remove the link and set `has_linked_activity` to no, but maybe that's a quirk of how things are set up?

Played around with having a blank option equalling `nil`, which kind of works - it does the same as above, removing the link and setting `has_linked_activity` to no - but I haven't included this for now because it's difficult to differentiate nothing being selected and the blank `nil` option being selected. If we do go down this line, we should give `.autocomplete__option` a `min-height` of `1.5625rem` so that the blank option doesn't collapse to almost 0

Also, if you select something, empty out the input field, then click "continue" it will be as though the thing you initially selected is still selected, I think

If you have an option selected, you also need to know to clear it with your keyboard before you can see the full list again (although we might not want to show the full list ultimately anyway). Clicking in the input doesn't select all the text it's populated it with, so you need to Ctrl/Cmd + A and delete, or delete each character individually. It's not the most intuitive thing ever

I wonder if all these quirks are why this didn't end up getting used beyond the defunct (or renamed?) `country` field in #463. I think it could be viable, but it seems a little far from ideal

## Next steps

### DIFF
--- a/app/assets/javascripts/accessibleAutocomplete.js
+++ b/app/assets/javascripts/accessibleAutocomplete.js
@@ -1,0 +1,36 @@
+// adapted from https://github.com/OfficeForProductSafetyAndStandards/product-safety-database/blob/d070006b2d68ba28e6b204d9efb9807c241c1eeb/app/assets/javascripts/autocomplete.js
+
+function rodaAccessibleAutocomplete () {
+  const selectElement = document.getElementById("activity-linked-activity-id-field")
+
+  if (selectElement) {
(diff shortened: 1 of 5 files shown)


yndajas:
Folded into #1893 as bec0c0f

yndajas:
Folded into #1893 as bec0c0f

PR #1898, opened 2022-11-28
(2686) Part 4: enable level D activity linking

CristinaRO:
## Changes in this PR
- Enable linking third-party projects (AKA level C activities)

## Next steps

### DIFF
--- a/spec/views/shared/activities/activity_spec.rb
+++ b/spec/views/shared/activities/activity_spec.rb
@@ -71,10 +71,10 @@ RSpec.describe "shared/activities/_activity" do
   end
 
   context "when the fund is ISPF" do
-    before { render }
-
     context "when the activity is a programme activity" do
       let(:activity) { build(:programme_activity, :ispf_funded, ispf_theme: 1, ispf_partner_countries: ["IN"]) }
+
+      before { render }
 
       it { is_expected.to show_basic_details }
       it { is_expected.to show_ispf_specific_details }
@@ -91,6 +91,10 @@ RSpec.describe "shared/activities/_activity" do
 
         it "shows the linked programme" do
           expect(rendered).to have_content(activity_presenter.linked_activity.title)
+        end
+
+        it "does not show a link to edit the linked programme" do
+          expect(body.find(".linked_activity .govuk-summary-list__actions")).to_not have_content(t("default.link.edit"))
         end
       end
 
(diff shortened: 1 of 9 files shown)


spec/views/shared/activities/activity_spec.rb
```diff
@@ -112,46 +116,89 @@
...
         end
       end
     end
+
+    context "when the activity is a project that has any child projects that are linked" do
+      let(:linked_proj) { create(:project_activity, :ispf_funded) }
+      let!(:third_party_project) { create(:third_party_project_activity, parent: activity, linked_activity: create(:third_party_project_activity, parent: linked_proj)) }
+      let(:activity) { create(:project_activity, :ispf_funded, linked_activity: linked_proj, organisation: user.organisation, extending_organisation: user.organisation) }
+
+      before { create(:report, :active, fund: activity.associated_fund, organisation: user.organisation) }
```
yndajas, line 194: Likewise is this preferable to putting it in a `let` or `let!`? If so, should it be combined with the next `before`?

spec/views/shared/activities/activity_spec.rb
```diff
@@ -92,6 +92,10 @@
         it "shows the linked programme" do
           expect(rendered).to have_content(activity_presenter.linked_activity.title)
         end
+
+        it "does not show a link to edit the linked programme" do
+          expect(body.find(".linked_activity .govuk-summary-list__actions")).to_not have_content(t("default.link.edit"))
+        end
```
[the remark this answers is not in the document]
CristinaRO: Yes, this is implicitly a partner organisation user as set at the top of the file.

spec/views/shared/activities/activity_spec.rb
```diff
@@ -112,46 +116,91 @@
...
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.policy_marker_climate_change_mitigation"))
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.policy_marker_biodiversity"))
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.policy_marker_desertification"))
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.policy_marker_disability"))
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.policy_marker_disaster_risk_reduction"))
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.covid19_related"))
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.channel_of_delivery_code"))
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.oda_eligibility"))
+              expect(rendered).not_to have_content(t("activerecord.attributes.activity.oda_eligibility_lead"))
+            end
```
yndajas, line 164: Shouldn't these lists be the same, with the only difference being `to` or `not_to`? I think this is outside the scope of this PR though - looks like this change is just moving things around

spec/views/shared/activities/activity_spec.rb
```diff
@@ -112,46 +116,91 @@
...
-            expect(rendered).not_to have_content(t("activerecord.attributes.activity.oda_eligibility"))
-            expect(rendered).not_to have_content(t("activerecord.attributes.activity.oda_eligibility_lead"))
+        context "and it has a linked activity" do
+          let(:activity) {
+            create(factory_name, :ispf_funded,
+              extending_organisation: user.organisation,
+              organisation: user.organisation,
+              ispf_theme: 1,
+              ispf_partner_countries: ["IN"],
+              linked_activity: create(:third_party_project_activity, :ispf_funded))
```
yndajas, line 175: Given we're using `factory_name` here to cover both projects and third-party projects, does the `linked_activity` also need to use the factory name - otherwise we might be linking activities at different levels?

Wonder if this is getting missed because we don't have model validations?

spec/views/shared/activities/activity_spec.rb
yndajas: I think we might have discussed why this wouldn't be in a `let`? But it could probably still be combined with the `before` that follows

yndajas (approved):
Looks good. Just a few small questions left

PR #1899, opened 2022-11-29
(2719) Hide ISPF level B templates with feature flag

CristinaRO:
## Changes in this PR
- Hide the ISPF templates for level B bulk uploads when the ISPF feature flag is enabled.

I considered renaming the non-ISPF templates when the feature flag is hiding ISPF, but decided against it. It's not technically wrong, even if it might seem odd to name the template when there is no other template to distinguish it from.

### Before
Screenshot 2022-11-29 at 16 58 32

### After
Screenshot 2022-11-29 at 16 58 50

## Next steps


CristinaRO:
[Feature flag ISPF bulk upload forms](https://trello.com/c/0opHp7oH/2719-feature-flag-ispf-bulk-upload-forms)

yndajas (approved):
Looks good!

PR #1903, opened 2022-12-01
(2707) Exclude non ODA from IATI exports

yndajas:
## Changes in this PR

This filters out non-ODA activities from IATI exports

## Next steps

### DIFF
--- a/app/services/find_programme_activities.rb
+++ b/app/services/find_programme_activities.rb
@@ -1,16 +1,20 @@
 class FindProgrammeActivities
   include Pundit::Authorization
 
-  attr_accessor :organisation, :user, :fund_code
+  attr_accessor :organisation, :user, :fund_code, :include_ispf_non_oda_activities
 
-  def initialize(organisation:, user:, fund_code: nil)
+  def initialize(organisation:, user:, fund_code: nil, include_ispf_non_oda_activities: false)
     @organisation = organisation
     @user = user
     @fund_code = fund_code
+    @include_ispf_non_oda_activities = include_ispf_non_oda_activities
   end
 
   def call
-    programmes = ProgrammePolicy::Scope.new(user, Activity.programme)
+    is_oda = [nil, true]
+    is_oda << false if include_ispf_non_oda_activities
+
+    programmes = ProgrammePolicy::Scope.new(user, Activity.programme.where(is_oda: is_oda))
       .resolve
       .includes(
         :organisation,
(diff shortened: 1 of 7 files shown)


yndajas:
[Exclude non ODA Activities from IATI exports](https://trello.com/c/zYId3ftX/2707-exclude-non-oda-activities-from-iati-exports)

app/services/find_programme_activities.rb
```diff
@@ -1,16 +1,17 @@
 class FindProgrammeActivities
   include Pundit::Authorization
 
-  attr_accessor :organisation, :user, :fund_code
+  attr_accessor :organisation, :user, :fund_code, :is_oda
 
-  def initialize(organisation:, user:, fund_code: nil)
+  def initialize(organisation:, user:, fund_code: nil, is_oda: [nil, true, false])
```
Gweaton, line 6: Could we make `is_oda` a single boolean here and do this inside `#call`? I found it a little difficult to parse when reading where this is initialized in `programme_activities`

app/services/find_project_activities.rb
```diff
@@ -1,16 +1,17 @@
 class FindProjectActivities
   include Pundit::Authorization
 
-  attr_accessor :organisation, :user, :fund_code
+  attr_accessor :organisation, :user, :fund_code, :is_oda
 
-  def initialize(organisation:, user:, fund_code: nil)
+  def initialize(organisation:, user:, fund_code: nil, is_oda: [nil, true, false])
```
Gweaton, line 6: again, see my previous comment about passing an array of values here

app/services/find_third_party_project_activities.rb
```diff
@@ -1,16 +1,17 @@
 class FindThirdPartyProjectActivities
   include Pundit::Authorization
 
-  attr_accessor :organisation, :user, :fund_code
+  attr_accessor :organisation, :user, :fund_code, :is_oda
 
-  def initialize(organisation:, user:, fund_code: nil)
+  def initialize(organisation:, user:, fund_code: nil, is_oda: [nil, true, false])
```
Gweaton, line 6: you know what I'm going to say here!

Gweaton (commented):
This looks good - just one thing re: argument type to each of these methods.

app/services/find_programme_activities.rb
Gweaton: Maybe a verbose `include_ispf_non_oda_activities` that defaults to `false` to keep it in sync with the rest of the way the class has been structured?

Gweaton (approved):
Thanks for the change!

PR #1910, opened 2022-12-02
Update @rails/webpacker dependencies

Gweaton:
## Changes in this PR

Update @rails/webpacker dependencies

The most important one here is `loader-utils`.

This was done by running `yarn upgrade @rails/webpacker`

I've briefly smoke tested the UI locally and it seems to all be working as expected.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,396 +2,415 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


Gweaton:
The build is failing here. I'll investigate this next week.

Gweaton:
I think we should migrate away from Webpacker here, as it's now deprecated. Closing this and adding a ticket for that.

PR #1915, opened 2022-12-06
(2723) ISPF themes support multiple selection

CristinaRO:
## Changes in this PR
- ISPF activities can have multiple ISPF themes

### Before
Screenshot 2022-12-06 at 21 59 19
Screenshot 2022-12-06 at 21 59 13

### After
Screenshot 2022-12-06 at 21 49 33
Screenshot 2022-12-06 at 21 49 24

## Next steps


CristinaRO:
[Make ISPF themes multiple selection (checkboxes rather than radio buttons)](https://trello.com/c/TOknUp5f/2723-make-ispf-themes-multiple-selection-checkboxes-rather-than-radio-buttons)

db/migrate/20221206171338_change_ispf_theme_to_ispf_themes_array.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeIspfThemeToIspfThemesArray < ActiveRecord::Migration[6.1]
+  def change
+    rename_column :activities, :ispf_theme, :ispf_themes
+    change_column :activities, :ispf_themes, :integer, array: true, using: "(array[ispf_themes]::integer[])"
```
Gweaton, line 4: I've never had to do this before in Rails but I'm struggling to find documentation for how to use `using` here - did you find any?

This is more for my own understanding than anything else!

Gweaton (approved):
This looks good to me, I've tested locally to make sure existing activities with a single ISPF theme don't break after this change (which I thought we might have to handle manually but it looks like ActiveRecord does a lot of this for us?)

db/migrate/20221206171338_change_ispf_theme_to_ispf_themes_array.rb
CristinaRO: This is where I found how to convert the data type while keeping the value: https://stackoverflow.com/questions/13879713/postgresql-alter-column-data-type-from-integer-to-integer-array

PR #1916, opened 2022-12-07
(2727) Reject ODA fields in non-ODA activity upload

yndajas:
If a user uploads a completed ODA bulk upload template using the non-ODA form, we don't want to create a non-ODA activity with the additional fields, but instead reject the upload since it doesn't match out expectations for ODA typing. This safeguards against uploading activities as the wrong type by mistake. This only seems to be possible in one direction, since the ODA templates have all the non-ODA fields and more, while the non-ODA templates don't include various required fields for ODA

This compares the ODA and non-ODA attributes and adds an error if any unexpected attributes appear in the non-ODA upload

### After

image

## Next steps


app/services/activity/import.rb
```diff
@@ -219,7 +256,17 @@ def create
...
+        attributes = @converter.to_h
+
+        if @is_oda == false
+          incorrect_oda_type_errors = Activity::Import.incorrect_oda_type_errors(@activity, attributes)
+          @errors.merge!(incorrect_oda_type_errors)
+        end
+
+        return unless @errors.blank?
+
+        @activity.assign_attributes(attributes)
```
yndajas, line 269: If comparing this with the changes in `ActivityUpdater` above, you can see this isn't very DRY (except the `Activity::Import.incorrect_oda_type_errors` abstraction), but neither is some of the other code in these methods, so I've added a tech debt card

yndajas:
[Ensure you can't upload an ODA activity via the non-ODA bulk upload](https://trello.com/c/E8jAGO7O/2727-ensure-you-cant-upload-an-oda-activity-via-the-non-oda-bulk-upload)

app/services/activity/import.rb
```diff
@@ -30,6 +30,34 @@ def is_oda_by_type(type:)
           non_ispf: nil
         }[type]
       end
+
+      def unexpected_non_oda_attribute_errors(activity, converted_attributes)
+        level = {
+          "programme" => :level_b,
+          "project" => :level_c_d,
+          "third_party_project" => :level_c_d
+        }[activity.level]
```
yndajas, line 39: Maybe we should just pass in the level like we do with `filtered_csv_column_headings`

spec/services/activity/import_spec.rb
```diff
@@ -333,6 +333,107 @@
...
+    }
+
+    context "Level B" do
+      let(:activity) { create(:programme_activity, :ispf_funded, is_oda: true) }
+      let(:error_translation) { t("importer.errors.activity.oda_attribute_in_non_oda_activity") }
+
+      it "returns a hash with errors where ODA fields have been included in a non-ODA upload" do
+        expect(Activity::Import.unexpected_non_oda_attribute_errors(activity, converted_level_b_oda_attributes)).to eq(
+          {
+            aid_type: ["D01", error_translation],
```
yndajas, line 372: `converted_level_b_oda_attributes.aid_type` be cleaner than `"D01"`?

CristinaRO (approved):
One non-blocking comment/suggestion.

Gweaton (approved):
Looks good - I think it'd be good to, as Cristina says, move that calculation of the unexpected fields somewhere else. I found it a little hard to parse at first read too, so it might improve readability and simplify the method too.

app/services/activity/import.rb
```diff
@@ -30,6 +30,34 @@ def is_oda_by_type(type:)
...
+
+        non_oda_attributes = []
+        oda_attributes = []
+
+        ACTIVITY_CSV_COLUMNS.each do |attribute, column|
+          non_oda_attributes << attribute if column.dig(:inclusion, level, :ispf_non_oda)
+          oda_attributes << attribute if column.dig(:inclusion, level, :ispf_oda)
+        end
+
+        unexpected_attributes = oda_attributes - non_oda_attributes
```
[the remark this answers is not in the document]
yndajas: Per Slack, have added a commit for review showing what this might look like

app/services/activity/import.rb
```diff
@@ -131,7 +146,21 @@ def initialize(row:, uploader:, partner_organisation:, report:, is_oda:)
       def update
         return unless @activity && @errors.empty?
 
-        @activity.assign_attributes(@converter.to_h)
+        attributes = @converter.to_h
+
+        if @is_oda == false
+          level = @activity.programme? ? :level_b : :level_c_d
```
Gweaton, line 152: Sorry - I didn't realise we'd need to convert the level into a symbol here to then be checked in the method - I think it might be cleaner to pass in the `activity` itself here and call `.programme?` inside `unexpected_non_oda_attribute_errors` here.

config/initializers/activity_csv_columns.rb
```diff
@@ -2,3 +2,18 @@
...
+level_c_d_oda_attributes = []
+
+ACTIVITY_CSV_COLUMNS.each do |attribute, column|
+  level_b_non_oda_attributes << attribute if column.dig(:inclusion, :level_b, :ispf_non_oda)
+  level_b_oda_attributes << attribute if column.dig(:inclusion, :level_b, :ispf_oda)
+  level_c_d_non_oda_attributes << attribute if column.dig(:inclusion, :level_c_d, :ispf_non_oda)
+  level_c_d_oda_attributes << attribute if column.dig(:inclusion, :level_c_d, :ispf_oda)
+end
+
+INVALID_LEVEL_B_ISPF_NON_ODA_ATTRIBUTES = level_b_oda_attributes - level_b_non_oda_attributes
```
Gweaton, line 18: Shall we `.freeze` these to be extra safe?

Gweaton (approved):
Thanks for updating this - a couple of small changes but good to merge afterwards (after squashing!)

app/services/activity/import.rb
yndajas: I was conflicted about whether to pass it in as the symbol or `@activity.level` - I went with the symbol to keep it consistent with `filtered_csv_column_headings`, but it might be cleaner to pass the `@activity.level` and do the conversion inside (like before, but without the whole activity), or yeah, pass the whole activity and use the `programme?` method. Any preference?

config/initializers/activity_csv_columns.rb
yndajas: Sure!

yndajas: Does this approach more generally seem sensible to you? If so, I'll squash the commits

Gweaton: I think so - I sort of wish there was a way that didn't involve looping all over and subtracting (like if we could just do something like `columns.where(level_b: true, ispf_oda: true, ispf_non_oda: false)`, but I can't think of a neat way of doing that with the way our hash is structured right now. Let's go with this for now.

app/services/activity/import.rb
Gweaton: I think passing in and use the `programme?` would be my preference, but up to you - apologies again for flip-flopping on this!

config/initializers/activity_csv_columns.rb
yndajas: Yeah, `filtered_csv_column_headings` is essentially that kind of interface but for headers, not attributes. I suppose we could make an almost identical `filtered_csv_column_attributes` (or something more accurate) but the only use case right now is working out the difference

PR #1933, opened 2022-12-20
Refactor `Activity::Import` and specs

yndajas:
This was drafted before #2086, which addressed most of the `Activity::Import` refactor, but the specs (`import`, `import_level_b` and relevant feature specs) still need work. The Trello card 2722 has more information

This remains unclosed as a reference for some of the spec refactoring (this PR didn't finish the job)

## Next steps


mec:
As this is almost a year old at this point and #2086 seems to superceed it, I think it best to close this now, it won't go away so if we do need it for some reason we should be able to locate it again in the closed PRs.

PR #1951, opened 2023-01-09
(2747) Fix level B IATI export figures and inclusion

yndajas:
## Changes in this PR

- 2747: fixes the transactions reported in level B IATI exports so that they reflect a quarterly sum of actuals minus adjustments and refunds (not just actuals)
- 2706: fixes redaction of programmes from IATI exports, adding a scope query that was missed from #1870

Debugged and checked using the example from the ticket: the activity with RODA ID `NF-BRNIF`. The figures are now in line with the report CSV, which is the same as the spending breakdown except for FQ3 2021-2022

## Next steps

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -334,9 +334,9 @@ class Activity < ApplicationRecord
     model.where(parent_activity_id: activity_ids)
   end
 
-  def reportable_actuals_for_level
+  def reportable_transactions_for_level
     if programme?
-      spend_by_financial_quarter(own_and_descendants_associates(Actual).order("date DESC"))
+      spend_by_financial_quarter(own_and_descendants_associates(Transaction).order("date DESC"))
     else
       actuals.order("date DESC")
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1159,6 +1159,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Show an error message if trying to upload a level D ISPF activity without an implementing organisation
 - Remove legacy report CSV generator
 - Switch to using JSbundling and CSSbundling instead of now-deprecated Webpacker for our assets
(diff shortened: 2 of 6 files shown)


yndajas:
[Include adjustments and refunds in IATI calculations](https://trello.com/c/B95XRH1v/2747-include-adjustments-and-refunds-in-iati-calculations)

Gweaton (approved):
Thanks for this. Looks good to me - a couple of small things:

1. In the commit message for 3f1f9cb, it says:

> Curiously, the method only returns actuals for other levels, but this
> hasn't been highlighted as an issue. At those levels, the actuals aren't
> summarised by quarter, so perhaps the purpose is different and only
> requires actuals

I'm a bit unclear on what you mean by "other levels" here - is it sub-level B levels? I wonder if you could clarify this a bit in the commit message.

2. In the PR description, it says:

> The figures are now in line with the report CSV, which is the same as the spending breakdown except for FQ3 2021-2022.

Is there a specific reason why that value is different, or is it just how the data is supposed to look?

app/models/activity.rb
```diff
@@ -334,9 +334,9 @@ def own_and_descendants_associates(model)
     model.where(parent_activity_id: activity_ids)
   end
 
-  def reportable_actuals_for_level
+  def reportable_transactions_for_level
     if programme?
-      spend_by_financial_quarter(own_and_descendants_associates(Actual).order("date DESC"))
+      spend_by_financial_quarter(own_and_descendants_associates(Transaction).order("date DESC"))
     else
       actuals.order("date DESC")
```
CristinaRO, line 341: Is this what you meant that at other levels, it's summing up the actuals but the lack of adjustments/refunds hasn't been flagged? If so, do you think it's worth asking Dan in Slack to double-check?

yndajas:
Re: question two, no idea, but I'll ask Dan about both of these things

yndajas:
@Gweaton re: 1, I've updated the commit message after clarifying with Dan. Re: 2, it sounds like it's an expected difference between actual net spend and actual spend 🤯

PR #1954, opened 2023-01-10
Fix logout issues

yndajas:
37ce062 and fb834b8 from #1937 removed `@rails/ujs` in the process of moving to js-bundling. The logout link is created using the `link_to` helper but with `method :delete`. `a` tags with a `DELETE` method are not natively supported by browsers - the way it worked before was that `@rails/ujs` was intercepting link click events and making `DELETE` requests. Removing `@rails/ujs` has therefore broken the logout link

This brings the package back and loads it in `application.js`. Previuosly, and conventionally, when including `Rails` from `@rails/ujs`, somewhere in the same file `Rails.start()` is called. However, this doesn't appear to be necessary to restore the logout functionality

Alternatively, we could change the Devise initializer/config to allow logout requests to be processed via `GET`, but this might be more unconventional

Evidently, the `users_can_sign_out` spec didn't catch this, but I'm not sure why

#### Local

image

#### Staging

image

#### Local

Screenshot 2023-01-10 at 19 40 29

## Next steps


yndajas:
[Fix logout issues](https://trello.com/c/Omqv0Irg/2768-fix-logout-issues)

Gweaton (commented):
Aha! Well done for fixing this. I had no idea that was what we were using `@rails/ujs` for, and had incorrectly assumed it was deprecated from looking at https://github.com/rails/rails-ujs (I should have checked the npm package instead!).

I'm a little concerned about the test not failing locally. It'd be great if you could spend some time finding out why that's the case, and ideally getting it to fail and be fixed by this change. Is that something you'd have time for? I've had a quick go with `save_and_open_page` to see what's happening and it does seem to be signing out properly on `develop`, so it'd be good to dig in a bit more here so we can be confident this won't happen again.

Oh, and no one's reported this yet as we haven't deployed staging to production yet!

yndajas:
> I'm a little concerned about the test not failing locally. It'd be great if you could spend some time finding out why that's the case, and ideally getting it to fail and be fixed by this change. Is that something you'd have time for? I've had a quick go with `save_and_open_page` to see what's happening and it does seem to be signing out properly on `develop`, so it'd be good to dig in a bit more here so we can be confident this won't happen again.

So I can get it to fail on other branches by adding `js: true` to the `scenario`: `scenario "success", js: true do`. However, this results in a[ browser being opened](https://stackoverflow.com/questions/36523193/stopping-capybara-selenium-from-opening-a-browser-window-during-js-tests) (and very quickly closed) when the test runs, which isn't ideal. Also feels flimsy that we have to know to do this to make it fail, and counter-intuitive that we need to enable JavaScript to make a feature fail that should only work when we import a particular JavaScript package. Any thoughts?

Haven't tested in the CI yet to see if the opening in a browser method breaks it

---

> Aha! Well done for fixing this. I had no idea that was what we were using `@rails/ujs` for, and had incorrectly assumed it was deprecated from looking at https://github.com/rails/rails-ujs (I should have checked the npm package instead!).

Yeah it looks kinda deprecated when you search for it. Sounds like Rails 7 pushes towards replacing it with Turbo as well
 
> Oh, and no one's reported this yet as we haven't deployed staging to production yet!

Of course!

yndajas:
Actually we have a few other tests using `js: true`, and they all seem to quickly open and close a window when run locally, so maybe it's not a problem

Gweaton (approved):
Nicely done - looks good to me

PR #1948, opened 2023-01-05
(2739) Remove the legacy report CSV generator

CristinaRO:
## Changes in this PR
- Remove legacy report CSV generator

### Before
Screenshot 2023-01-05 at 13 36 34

### After
Screenshot 2023-01-05 at 13 36 22

## Next steps

- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [x] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.
- [ ] Do any environment variables need amending or adding?
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/xml-validation.md)

PR #1959, opened 2023-01-12
(2764) Remove unused report filename generation method

Gweaton:
## Changes in this PR

Removes unused report filename generation method.  

This was used for generating the filename for the legacy CSV, which was removed in #5bf495 (#1948)

I originally thought it would make sense to use the `ReportPresenter`'s method here, rather than the having the `Export:Report` class  have its own report generation method, but the other Export classes `Budget` and `ExternalIncome` have their own filename generation classes so it made sense to keep consistency across the three Export classes here.

## Next steps


Gweaton:
https://trello.com/c/UgiLmMMn/2764-remove-reportpresenterfilenameforreportdownload

PR #1960, opened 2023-01-12
(2767) Fix title tags

yndajas:
## Changes in this PR

This adds titles to the home and activities pages for all users, and also makes the titles for activity form pages match the main heading of the page (usually the question, except on pages with multiple inputs)

In doing so, it refactors and cleans up a bunch of code:
- less redundancy both in terms of repetiton (DRY) and old, now unsed translations
- alphabetical ordering of remaining translations

### Before

Screenshot 2023-01-12 at 16 29 44

### After

Screenshot 2023-01-12 at 16 29 09

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1166,6 +1166,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 
 - Upload the report CSV to S3 when the report is approved
 - For approved reports, provide download of stored report CSV file instead of generating it from live data
+- Add page titles for home and activities pages (useful for Google Analytics reporting and accessibility); make page titles and headings consistent on activity form pages
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-126...HEAD
 [release-126]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-125...release-126


yndajas:
[Fix <title> tags in the service](https://trello.com/c/FvN3ANmz/2767-fix-title-tags-in-the-service)

app/views/activity_forms/objectives.html.haml
```diff
@@ -1,5 +1,4 @@
 = render layout: "wrapper" do |f|
-  = f.govuk_fieldset legend: { text: t("form.legend.activity.objectives", level: t("page_content.activity.level.#{f.object.level}")), tag: :h1, size: "xl" } do
```
yndajas, line 2: This is intentional. The fieldset shouldn't be necessary when there's only one input. The label just said "Aims/Objectives of the activity", which adds nothing to the question "What are the aims/objectives of this %{level}?", so the label of the remaining text area is now the full question and the original label is removed

config/locales/models/implementing_organisation.en.yml
```diff
@@ -8,11 +8,6 @@ en:
         button: Remove
         success: Implementing organisation successfully removed
 
-  form:
```
Gweaton, line 11: This is the only one that's making me think twice - I wonder if we should move all of the translations in this file to the `activities` translations file if we're classifying Implementing orgs as part of the activity journey?

app/controllers/activity_forms_controller.rb
```diff
@@ -151,12 +151,15 @@ def assign_default_collaboration_type_value_if_nil
...
   def page_title(step)
-    if step == :linked_activity
-      @activity.is_oda ? t("page_title.activity_form.show.linked_non_oda_activity") : t("page_title.activity_form.show.linked_oda_activity")
-    elsif step == :identifier && @activity.programme?
+    if step == :identifier && @activity.programme?
       t("page_title.activity_form.show.identifier_level_b")
     else
-      t("page_title.activity_form.show.#{step}", sector_category: t("activity.sector_category.#{@activity.sector_category}"), level: t("page_content.activity.level.#{@activity.level}"))
+      t(
+        "page_title.activity_form.show.#{step}",
```
Gweaton, line 158: Do you think we could add a bit more to the commit message here to say that these arguments will be ignored in most cases and only used where we interpolate the `linked_activity_type` (for example)? At a quick glance, it feels as though we actually care about their values, when we don't really.

Gweaton (approved):
Thanks for this, a good refactor as well, I like the move towards more consistency. Just a couple of questions/thoughts

PR #1962, opened 2023-01-12
(2766) Reflect adjustments/refunds in level C/D IATI export

yndajas:
## Changes in this PR

4f17df3 updated level B IATI exports to account for adjustments and refunds. This brings levels C and D in line. The only remaining difference between the level B and levels C and D is that at level B it included all descendents' transactions

## Next steps


yndajas:
[Amend IATI export at levels C/D to account for adjustments (and possibly refunds)](https://trello.com/c/VChEuGdA/2766-amend-iati-export-at-levels-c-d-to-account-for-adjustments-and-possibly-refunds)

PR #1965, opened 2023-01-16
Feature: reports uploaded to s3

CristinaRO:
### After
Screenshot 2023-01-16 at 18 15 32

## Next steps


Gweaton (approved):
Approving this so we can test the S3 bucket on staging

PR #1974, opened 2023-01-19
Add new S3 bucket for private use

olivia-work:
## Changes in this PR

- create a new private S3 bucket for someone to deploy to staging
- bind the bucket to the app


DrizzlyOwl (approved):
LGTM. Worth noting that (as is intended) this will create a Private S3 bucket for each environment

PR #1977, opened 2023-01-19
(2779) Update links to organisation activities path

yndajas:
## Changes in this PR

The paths `/organisations/:organisation_id/activities` and `/activities?organisation_id=:organisation_id` are both served by the `activities#index` controller action and pass in the same params. For Google Analytics, it's better for such shows to use the same path. This updates the single use of the latter path to the former

It also updates the small number of specs using its path helper, and moves the link's text into a translation

## Next steps


yndajas:
[Update "View all activities" PO link to point to organisations/:id/activities](https://trello.com/c/c3QczLzH/2779-update-view-all-activities-po-link-to-point-to-organisations-id-activities)

Gweaton (approved):

PR #1978, opened 2023-01-19
(2736) Prefix all non-ODA identifiers with “NODA”

CristinaRO:
## Changes in this PR
- Prefix the RODA identifiers of all non-ODA activities with "NODA-"

### After
Screenshot 2023-01-19 at 19 17 18

## Next steps

### DIFF
--- a/spec/controllers/activity_forms_controller_spec.rb
+++ b/spec/controllers/activity_forms_controller_spec.rb
@@ -358,6 +358,16 @@ RSpec.describe ActivityFormsController do
       end
     end
 
+    context "when setting non-ODA on a programme" do
+      let(:activity) { programme }
+
+      it "updates the RODA identifier to start with 'NODA'" do
+        put_step(:is_oda, {is_oda: false})
+
+        expect(programme.reload.roda_identifier).to start_with("NODA-")
+      end
+    end
+
     context "when the activity is invalid" do
       before do
         allow(Activity).to receive(:find).and_return(activity)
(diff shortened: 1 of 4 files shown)


spec/controllers/activity_forms_controller_spec.rb
```diff
@@ -346,6 +346,14 @@
       end
     end
 
+    context "when setting non-ODA on a programme" do
+      it "updates the RODA identifier to start with 'NODA'" do
+        put :update, params: {activity_id: programme.id, id: :is_oda, activity: {is_oda: false}}
```
yndajas, line 351: You could use the `put_step` method here

yndajas (commented):
Code looks good - just a suggestion to use a private method that's been defined in the spec

Will this work with projects/3PP though, since (I believe) they inherit `is_oda` from their parent via `ActivityDefaults`, rather than setting it via `Activity::Updater` (triggered by a form step)?

spec/controllers/activity_forms_controller_spec.rb
CristinaRO: I could, but then I'd have to recreate the already defined `activity` to be a programme instead of using the `programme`. (`put_step` is hardcoded to use `activity.id`).

CristinaRO:
> 

Projects get their identifier by combining the parent programme's identifier with a "component" identifier, so as long as the programme is set up correctly, the child activities will inherit the string.

spec/controllers/activity_forms_controller_spec.rb
yndajas: Would `let(:activity) { programme }` in the context do the job? Although maybe it would be unclear what the purpose of that is when it's abstracted away in `put_step`

yndajas:
> > 
> 
> Projects get their identifier by combining the parent programme's identifier with a "component" identifier, so as long as the programme is set up correctly, the child activities will inherit the string.

Ah, nice! I guess the tests for that just cover that it includes the parent's identifier, in which case this doesn't change anything?

Gweaton (approved):
Nice work 🏆

PR #1983, opened 2023-01-25
(2709) Upload reports to private S3 bucket on approval

Gweaton:
## Changes in this PR

Uploads reports to a new private bucket on S3 when they're approved by the BEIS team, storing an `approved_at` datetime on the Report to display the date at which this happened to users.

We're doing this as reports have always been created with the most up-to-date data in the service at the point of clicking the "download" button. This meant the data for a report that was previously thought to approved (and closed for further editing) could be different if any future transactions were added that referred to that quarter, essentially changing historical data.

For any reports that haven't been approved yet, we pull them together with live data at the point of clicking the button, like we used to do.

### After

image

## Next steps

### DIFF
--- a/app/helpers/report_helper.rb
+++ b/app/helpers/report_helper.rb
@@ -1,0 +1,7 @@
+module ReportHelper
+  def report_download_link(report)
+    return download_report_path(report) if report.export_filename
+
+    report_path(report, format: :csv)
+  end
+end
--- a/app/presenters/report_presenter.rb
+++ b/app/presenters/report_presenter.rb
@@ -16,6 +16,18 @@ class ReportPresenter < SimpleDelegator
     I18n.l(super)
   end
 
+  def approved_at
+    return if super.blank?
+    I18n.l(super, {format: :detailed})
+  end
+
+  def uploaded_at
+    return nil if export_filename.nil?
+    uploaded_time_from_filename = Time.parse(export_filename[-18..-5])
+
+    I18n.l(uploaded_time_from_filename, {format: :detailed})
+  end
+
   def email_title
     return nil if financial_quarter_and_year.nil? || fund.nil? || organisation.nil?
     "#{financial_quarter_and_year} #{fund.roda_identifier} #{organisation.beis_organisation_reference}"
--- a/app/services/export/s3_downloader.rb
+++ b/app/services/export/s3_downloader.rb
@@ -1,0 +1,35 @@
+module Export
+  class S3DownloaderError < StandardError; end
+
+  class S3Downloader
+    def initialize(filename:)
+      @config = S3UploaderConfig.new(use_public_bucket: false)
+      @client = Aws::S3::Client.new(
+        region: config.region,
+        credentials: Aws::Credentials.new(
+          config.key_id,
+          config.secret_key
+        )
+      )
+      @filename = filename
+    end
+
+    def download
+      client.get_object(
+        bucket: config.bucket,
+        key: filename,
+        response_content_type: "text/csv"
+      ).body.read
+    rescue => error
+      raise_error(error.message)
+    end
+
+    private
+
+    attr_reader :config, :client, :filename
+
+    def raise_error(original_message = nil)
+      raise S3DownloaderError, [original_message, I18n.t("download.failure", filename: filename)].join(" ")
+    end
+  end
+end
--- a/app/services/export/s3_uploader.rb
+++ b/app/services/export/s3_uploader.rb
@@ -3,7 +3,7 @@ module Export
 
   class S3Uploader
     def initialize(file:, filename:)
-      @config = S3UploaderConfig.new
+      @config = S3UploaderConfig.new(use_public_bucket: true)
       @client = Aws::S3::Client.new(
         region: config.region,
         credentials: Aws::Credentials.new(
--- a/app/services/export/s3_uploader.rb
+++ b/app/services/export/s3_uploader.rb
@@ -3,11 +3,12 @@ module Export
 
   class S3Uploader
     def initialize(file:, filename:)
+      @config = S3UploaderConfig.new
       @client = Aws::S3::Client.new(
-        region: S3UploaderConfig.region,
+        region: config.region,
         credentials: Aws::Credentials.new(
-          S3UploaderConfig.key_id,
-          S3UploaderConfig.secret_key
+          config.key_id,
+          config.secret_key
         )
       )
       @file = file
@@ -18,7 +19,7 @@ module Export
 
     def upload
       response = client.put_object(
-        bucket: S3UploaderConfig.bucket,
+        bucket: config.bucket,
         key: filename,
         body: file
       )
@@ -34,6 +35,8 @@ module Export
 
     private
 
+    attr_reader :config
+
     def timestamped_filename(name)
       pathname = Pathname.new(name)
       basename = pathname.basename(".*")
@@ -48,7 +51,7 @@ module Export
 
     def bucket
       resource = Aws::S3::Resource.new(client: client)
-      resource.bucket(S3UploaderConfig.bucket)
+      resource.bucket(config.bucket)
     end
   end
 end
--- a/app/services/export/s3_uploader.rb
+++ b/app/services/export/s3_uploader.rb
@@ -2,8 +2,9 @@ module Export
   class S3UploadError < StandardError; end
 
   class S3Uploader
-    def initialize(file:, filename:)
-      @config = S3UploaderConfig.new(use_public_bucket: true)
+    def initialize(file:, filename:, use_public_bucket:)
+      @use_public_bucket = use_public_bucket
+      @config = S3UploaderConfig.new(use_public_bucket: use_public_bucket)
       @client = Aws::S3::Client.new(
         region: config.region,
         credentials: Aws::Credentials.new(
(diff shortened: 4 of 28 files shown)


db/schema.rb
```diff
@@ -287,6 +287,8 @@
     t.integer "financial_quarter"
     t.integer "financial_year"
     t.string "export_filename"
+    t.datetime "approved_at"
+    t.datetime "uploaded_at"
```
Gweaton, line 291: This needs to be removed

app/services/export/s3_downloader.rb
```diff
@@ -0,0 +1,37 @@
...
+    rescue => error
+      raise_error(error.message)
+    end
+
+    private
+
+    attr_reader :config
+
+    def raise_error(original_message = nil)
+      raise S3DownloaderError, [original_message, "Error generating download for #{filename}"].join(" ")
```
yndajas, line 34: Guessing we care less about translations here since it's a bit of an admin thing?

Also, do we know that `original_message` has punctuation at the end? If not, `.join(" ")` might make for a messy message

spec/features/users_can_view_reports_spec.rb
```diff
@@ -429,6 +443,19 @@ def expect_to_see_a_table_of_reports(selector:, reports:)
       end
     end
 
+    context "when the report has an export_filename" do
+      scenario "the link to download the report is the download path instead of the show path" do
+        report = create(:report, :approved, export_filename: "exported_csv", organisation: partner_org_user.organisation)
+
+        visit reports_path
+        within "##{report.id}" do
```
yndajas, line 451: Spacing here is different to the similar example earlier in this commit (in this file)

spec/services/export/s3_downloader_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      key_id: "key id",
+      secret_key: "secret key",
+      region: "region",
+      bucket: s3_bucket
+    )
+  }
+
+  let(:filename) { "Q1_report0101202312345.csv" }
+
+  let(:public) { true }
```
yndajas, line 22: I don't think we're using this

spec/services/export/s3_downloader_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+        expect(Aws::Credentials).to have_received(:new).with(
+          s3_uploader_config.key_id,
+          s3_uploader_config.secret_key
+        )
+        expect(Aws::S3::Client).to have_received(:new).with(hash_including(
+          credentials: aws_credentials
+        ))
+      end
+
+      it "sets region: using the region from the S3UploaderConfig" do
```
yndajas, line 48: Is there a reason for the colons in the test descriptions?

app/presenters/report_presenter.rb
```diff
@@ -16,6 +16,18 @@ def deadline
...
   end
 
+  def approved_at
+    return if super.blank?
+    I18n.l(super, {format: :detailed})
+  end
+
+  def uploaded_at
+    return nil if export_filename.nil?
+    timestamp = Time.parse(export_filename[-18..-5])
```
yndajas, line 26: This is an instance of `Time` parsed from a timestamp, so the variable name could be clearer

spec/presenters/report_presenter_spec.rb
```diff
@@ -39,6 +39,33 @@
...
+      now = Time.now
+      report = build(:report, approved_at: now)
+      result = described_class.new(report).approved_at
+      expect(result).to eql I18n.l(now, {format: :detailed})
+    end
+  end
+
+  describe "#uploaded_at" do
+    context "when the report has an `export_filename`" do
+      it "parses the timestamp off the uploaded report's filename" do
```
yndajas, line 53: `off` -> `from`?

yndajas (commented):
A few comments on small details, but in general it makes sense to me!

app/services/export/s3_uploader_config.rb
```diff
@@ -19,10 +23,18 @@ def secret_key
...
     rescue KeyError, NoMethodError => _error
       raise "AWS S3 credentials not found"
     end
+
+    private
+
+    attr_reader :public
+
+    def bucket_name_regex
+      public ? /s3-export-download-bucket$/ : /s3-export-download-bucket-private/
```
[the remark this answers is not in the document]
Gweaton: I was wondering that too, it's not listed [here](https://docs.ruby-lang.org/en/2.2.0/keywords_rdoc.html) but neither is `private`. I think I'll change this as even though it's working as expected, it might be confusing, and the syntax highlighting here does make it look reserved...

app/services/export/s3_downloader.rb
```diff
@@ -0,0 +1,37 @@
+module Export
+  class S3DownloaderError < StandardError; end
+
+  class S3Downloader
```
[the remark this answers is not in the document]
Gweaton: Let's keep this simple as, yeah, we'll be removing any reference to public buckets soon - means we don't have to remember to change it later.

app/services/export/s3_downloader.rb
Gweaton: This may have slipped in from the previous spike - I think it probably should belong in a translation here.

spec/services/export/s3_downloader_spec.rb
Gweaton: I suppose  it's a sort of shorthand way of saying "sets the region key...". Do you think that would be a bit clearer?

spec/services/export/s3_downloader_spec.rb
yndajas: I think that's in a different spec?

Gweaton: hah, I've just come across this in my renaming. You're right! Removing it now.

app/services/export/s3_uploader_config.rb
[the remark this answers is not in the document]
Gweaton: Have changed this to `use_public_bucket` (it's not the best name, but we'll be getting rid of it soon, once we move spending breakdown over)

spec/services/export/s3_downloader_spec.rb
yndajas: Ah got you, that or backticks maybe?

app/services/export/s3_uploader.rb
```diff
@@ -25,17 +26,23 @@ def upload
...
+      if public
+        OpenStruct.new(
+          url: bucket.object(filename).public_url,
+          timestamped_filename: filename
+        )
+      else
+        OpenStruct.new(
+          timestamped_filename: filename
+        )
+      end
```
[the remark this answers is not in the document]
Gweaton: I think I prefer treating the two return OpenStructs as kind of separate/distinct objects here rather than considering them the same object with different fields, if that makes sense at all?

We'll be removing this conditional behaviour pretty soon anyway, so it shouldn't be a concern in the codebase for too much longer.

app/services/export/s3_downloader.rb
Gweaton: Have updated this to use a translation here (and the existing S3 upload file). We don't know about punctuation here (haven't been able to find the messages with a quick google), but we're following the same approach taken for the S3 Uploader class, so at least we're doing the same for both. As this is an error, I don't think we need to worry too much about it being a bit messy.

app/presenters/report_presenter.rb
Gweaton: Have tried to make this clearer

app/helpers/report_helper.rb
```diff
@@ -0,0 +1,7 @@
+module ReportHelper
+  def report_download_link(report)
+    return download_report_path(report) if report.export_filename
+
+    report_path(report, format: :csv)
```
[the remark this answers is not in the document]
Gweaton: I don't think we have a preferred approach, but as far as I'm aware, `if`s are slightly preferable to ternaries in Ruby (in terms of its emphasis on human readability), so let's leave this as-is.

spec/jobs/report_export_uploader_job_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+          "Error uploading filename-xyz for roger@example.com",
+          error
+        )
+      end
+
+      it "does not re-raise the error as we don't wish to retry the job" do
+        expect {
+          ReportExportUploaderJob.perform_now(requester_id: double, report_id: double)
+        }.not_to raise_error
+      end
```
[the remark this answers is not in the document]
CristinaRO: Yeah, we decided to defer messing with ActiveJob so it wouldn't turn into a giant rabbit hole.

app/services/export/s3_uploader.rb
```diff
@@ -3,7 +3,7 @@ class S3UploadError < StandardError; end
 
   class S3Uploader
     def initialize(file:, filename:)
-      @config = S3UploaderConfig.new
+      @config = S3UploaderConfig.new(public: true)
```
CristinaRO: Should the argument be `use_public_bucket`?

CristinaRO (approved):
Only noticed one thing, which was probably changed in a subsequent commit, but if we want all-green individual commits... ;)

yndajas (approved):

app/services/export/s3_uploader.rb
Gweaton: ah, I think `git absorb` didn't do as I expected here, I missed that, well spotted!

yndajas: Ooh, didn't know about `git absorb`. Does it usually work well? Finding the right commits to make changes to can be a pain in bigger PRs (would be nice if GitHub gave you a pointer to the commit a comment was left on)

Gweaton: It's usually pretty good, particularly if you're making lots of small changes, but it's usually worth double checking (do as I say, not as I do!)

PR #1985, opened 2023-01-25
(2743) Decorate activity importer implementing organisation error

yndajas:
## Screenshots of UI changes

When there was an error with `implementing_org_participations` when importing activities, the column would be represented in snake case. This decorates the error so that it says "Implementing organisation names" instead

We may want to improve the error message at a later time - it currently says "Implementing org participations is invalid" (a default attribute error), which may be confusing to users

### Before

image

### After

image

## Next steps

### DIFF
--- a/app/services/activity/import.rb
+++ b/app/services/activity/import.rb
@@ -6,7 +6,10 @@ class Activity
       end
 
       def csv_column
-        return "Implementing organisation names" if column == :implementing_organisation_id
+        if [:implementing_organisation_id, :implementing_org_participations].include?(column)
+          return "Implementing organisation names"
+        end
+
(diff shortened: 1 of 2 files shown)


yndajas:
[Bulk upload error page: undecorated column name when implementing organisation is invalid](https://trello.com/c/YfB4tlN3/2743-bulk-upload-error-page-undecorated-column-name-when-implementing-organisation-is-invalid)

Gweaton (approved):
I was going to say, should we make that error message a bit clearer? Something like "Invalid implementing organisation name"?

yndajas:
@Gweaton how does that look?

Gweaton (approved):
Looks good, thanks for the change

PR #1987, opened 2023-01-25
(2782) Fix invalid comment update and other edge cases

yndajas:
## Changes in this PR

Previously:
- If when editing a comment, you submit the form with no body, an error would show up due to a missing instance variable
- Tests were missing for this in both the `update` and `create` controller actions (but the latter didn't cause an error)
- When an activity had no title, the heading when adding/editing comments was confusing: "Add a comment about"/"Edit comment about"
- Some comment tests were checking for the absence of the wrong translation
- If a create/update failed, the breadcrumbs would be lost when re-rendering the form

This fixes all of the above

### Before

image

### After

image

## Next steps

### DIFF
--- a/app/controllers/activity_comments_controller.rb
+++ b/app/controllers/activity_comments_controller.rb
@@ -71,6 +71,7 @@ class ActivityCommentsController < BaseController
       flash[:notice] = t("action.comment.update.success")
       redirect_to organisation_activity_comments_path(@comment.commentable.organisation, @comment.commentable)
     else
+      @activity = Activity.find(@comment.commentable_id)
       render :edit
     end
   end
--- a/app/controllers/activity_comments_controller.rb
+++ b/app/controllers/activity_comments_controller.rb
@@ -9,14 +9,12 @@ class ActivityCommentsController < BaseController
     if @activity.programme?
       @comment = @activity.comments.new
       authorize :level_b, :create_activity_comment?
-      prepare_default_activity_trail(@activity, tab: "comments")
     else
       @comment = @activity.comments.new(report_id: report_id)
       authorize @comment, policy_class: Activity::CommentPolicy
-      prepare_default_report_variance_trail(@comment.report)
     end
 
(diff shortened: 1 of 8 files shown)


app/controllers/activity_comments_controller.rb
```diff
@@ -93,4 +92,18 @@ def report_id
   def comment_params
     params.require(:comment).permit(:body, :report_id)
   end
+
+  def prepare_trail_and_breadcrumb(is_new:)
```
Gweaton, line 96: I'm finding this a little hard to parse - I wonder if it might be a bit clearer to opt for duplication here, and have a `prepare_new_trail_and_breadcrumb` and a `prepare_edit_trail_and_breadcrumb` method?

Also, are there already tests for this behaviour?

Gweaton (commented):
This looks good - just one thing re: that long method.

Also, let's add this to the CHANGELOG.

Gweaton (commented):
This looks good - just one thing re: that long method.

Also, let's add this to the CHANGELOG.

app/controllers/activity_comments_controller.rb
yndajas: I've split that method in two - looks better that way

I think the controller specs don't usually go into enough detail to cover trails/breadcrumbs. Can't see any existing tests in the specs I've touched referencing trails or breadcrumbs. Do you think we should add them, or leave the controller specs light?

Gweaton: It looks like we've tested breadcrumbs in feature specs in the past e.g. [here](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/a74265fed7afd7ae1341aa05019d3e0c89d279dd/spec/features/users_can_search_for_activities_spec.rb#L31). It might be worth adding specs like that in the new/edit comments feature spec, if that's not too much of a pain, so at least we can catch things if they go missing.

Gweaton (approved):
Sorry, meant to approve this but looks like I commented...

Gweaton (approved):
Nice one, thanks for adding those tests.

PR #1995, opened 2023-02-01
(2786) Fix `rows_for_last_financial_quarter`

Gweaton:
## Changes in this PR

A while back, we added protection to `rows_for_first_financial_quarter` to prevent it trying to convert `nil` values to BigDecimals when there weren't any found.

Although [the PR](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1482) in which that was added suggests that there should always be a value for the last row of actuals, on Staging, at least, this doesn't appear to be the case.

Although the above may be due to slightly wonky, possibly invalid data on Staging, this error caused a report (with ID
b45c5b36-e502-4406-ac66-8a51321a177c) to fail being uploaded to S3 (particularly, when calling `Export::Report#rows`. We'd like to make sure this doesn't happen on Production.

## Next steps

- [ ] Run the report upload job on staging and make sure all succeeds

### DIFF
--- a/spec/models/export/activity_actuals_columns_spec.rb
+++ b/spec/models/export/activity_actuals_columns_spec.rb
@@ -128,6 +128,21 @@ RSpec.describe Export::ActivityActualsColumns do
         expect(value_for_activity).to eq BigDecimal("300")
         expect(last_column_data.count).to eq 5
       end
+
+      context "if, unexpectedly, there is no value for that report quarter" do
+        let(:activity) { create(:project_activity) }
+        subject {
+          described_class.new(activities: [activity], include_breakdown: breakdown, report: report)
+        }
+
+        it "returns zero and not nil" do
+          last_column_data = subject.rows_for_last_financial_quarter
+          value_for_activity = last_column_data.fetch(activity.id)
+
+          expect(value_for_activity).to eq 0
+          expect(value_for_activity).not_to be_nil
+        end
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


Gweaton:
https://trello.com/c/QKD13zKT/2786-investigate-possible-edge-case-with-report-failing-to-upload

spec/models/export/activity_actuals_columns_spec.rb
```diff
@@ -128,6 +128,21 @@
...
+        let(:activity) { create(:project_activity) }
+        subject {
+          described_class.new(activities: [activity], include_breakdown: breakdown, report: report)
+        }
+
+        it "returns zero and not nil" do
+          last_column_data = subject.rows_for_last_financial_quarter
+          value_for_activity = last_column_data.fetch(activity.id)
+
+          expect(value_for_activity).to eq 0
```
yndajas, line 142: Looks like there's a [be_zero matcher](https://relishapp.com/rspec/rspec-expectations/v/2-2/docs/matchers/predicate-matchers), but I don't have a particular preference for that

yndajas (approved):
Looks good. Do we need add any tests to the place(s) this was causing errors, or is this unit test sufficient, given it's a fairly edge case?

spec/models/export/activity_actuals_columns_spec.rb
Gweaton: I've taken this from the existing example test in the PR above, but good to know about this matcher!

PR #1997, opened 2023-02-03
Add migration to add Chile as a benefitting country to old activities

jkempster34:
We have received a support query regarding two old activities that should have Chile as a benefitting country, but don't.

Chile is no longer available in the UI as it is no longer an eligible benefitting country, however, it was at the time the activities were active.

## Changes in this PR
- Add migration to add Chile as a benefitting country to old activities

## Screenshots
image

## Next steps

We will need to run `rails runner db/data/20230203144832_add_chile_as_a_benefitting_country_to_old_activities.rb` after deployment.


Gweaton (approved):
Looks good to me, thanks for this.

PR #1998, opened 2023-02-06
(2800) Don't upload reports with no financial year or financial quarter

Gweaton:
## Changes in this PR

This tweaks the approved Report-uploading Rake task to ignore reports with a `nil` value for `financial_year` or `financial_quarter` as these fail validations when we try to call `#rows` on them when creating the CSV file. 

We don't actually offer a CSV file for these kinds of reports, which were created from historical data from before RODA existed.

I've tested this locally with 2 organisations after fetching prod data, and all seems to be working as expected.

## Next steps

Run this on staging

### DIFF
--- a/lib/tasks/reports.rake
+++ b/lib/tasks/reports.rake
@@ -8,6 +8,9 @@ namespace :reports do
     uploader = User.find(requester_id)
 
     abort "Report #{report.id} has not been approved" unless report.approved?
+    unless report.financial_year.present? && report.financial_quarter.present?
+      abort "Report #{report.id} has no financial year or quarter"
+    end
     abort "Report #{report.id} not found" if report.nil?
     abort "User with id #{requester_id} not found" if uploader.nil?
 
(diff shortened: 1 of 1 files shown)


Gweaton:
https://trello.com/c/2SidB4iz/2800-exclude-historical-reports-with-no-financial-year-from-s3-upload

yndajas (approved):
Looks good. I tested the individual rake task and got the expected error message. Didn't test the other one, but the code makes sense

lib/tasks/reports.rake
```diff
@@ -8,6 +8,9 @@ namespace :reports do
     uploader = User.find(requester_id)
 
     abort "Report #{report.id} has not been approved" unless report.approved?
+    unless report.financial_year.present? && report.financial_quarter.present?
+      abort "Report #{report.id} has no financial year or quarter"
+    end
     abort "Report #{report.id} not found" if report.nil?
```
CristinaRO, line 14: Sorry for not noticing this earlier! I think the `report.nil?` check should be first.

lib/tasks/reports.rake
```diff
@@ -7,8 +7,11 @@ namespace :reports do
     report = Report.find(report_id)
     uploader = User.find(requester_id)
 
-    abort "Report #{report.id} has not been approved" unless report.approved?
     abort "Report #{report.id} not found" if report.nil?
```
CristinaRO, line 10: Me again, it should probably be `abort "Report #{report_id} not found" if report.nil?` or we'll either get `4` (the id of nil), or an unexpected error if Rails got wiser about it 😅

Gweaton: 🤦

PR #2002, opened 2023-02-06
Use single quotes in password reset script

yndajas:
## Changes in this PR

When running the training database sync script, the password reset method was failing silently in the `force_password_reset_for_users` method

In the process of working on dc87872 (#1803), we provided a command to be run remotely using `-c` and single quotes. In the end, the commit used double quotes, but we've now found the single quotes are necessary to make this script run successfully on certain machines

For future reference, we were able to debug this by running the script on the server: we attempted to run the script normally, but with the method call that removes temp files (and all other irrelevant methods) commented out, then ssh-ed onto the server and ran it manually

## Next steps


Gweaton (approved):
Really unfair that something so tiny can cause so many issues

PR #2003, opened 2023-02-07
(2813) Add Sidekiq web UI

Gweaton:
## Changes in this PR

Adds Sidekiq UI

This will allow us to view the result of scheduled jobs in the service more easily. This is particularly useful now we're about to run the report uploading script

This is accessible at `/sidekiq` but only for logged in "service owner" users.

### After

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1190,6 +1190,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Default to 0, rather than `nil` in the rare case where a report has no actual value for an activity
 - Add QA completed report step in between review and approval
 - Optimise codelist-related logic with the aim of fixing IATI export timeout issues and providing speed gains more generally
+- Add Sidekiq web UI route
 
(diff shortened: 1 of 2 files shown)


yndajas (approved):
Looks good. Are we definitely using Sidekiq for the report uploads?

Gweaton:
> Looks good. Are we definitely using Sidekiq for the report uploads?

@yndajas yep, it's configured to be the `queue_adapter` [here](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/421f5cab51fc547a82d5ed919a9da941cc7c4b84/config/application.rb#L52), meaning we use it for all jobs. [Here's some documentation about how that's used](https://github.com/mperham/sidekiq/wiki/Active-Job#active-job-setup)

There's an ADR about [here](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/architecture/decisions/0009-use-sidekiq-and-redis-to-send-emails.md) if you're interested.

PR #2004, opened 2023-02-07
(2807) Fix organisation reports table for BEIS users

yndajas:
## Changes in this PR

For BEIS users, the organisation reports table included a column heading of "Organisation", but there was no data for this column, meaning all data were offset one column from their headings

This makes it so that the organisation appears only for BEIS users at `/reports`, where the reports table covers multiple organisations. In all other cases - i.e. for PO users at `/reports` and for both PO and BEIS users at the organisation reports path - we leave the organisation out

---

The feature specs for viewing reports only seem to touch `/reports`, not `/organisations/:organisation_id/reports`, and don't appear to test column headings. Is it worth adding some tests for this, or is it too much detail for feature tests?

### Before

image

### After

image

## Next steps

### DIFF
--- a/app/views/shared/reports/_grouped_table.html.haml
+++ b/app/views/shared/reports/_grouped_table.html.haml
@@ -2,7 +2,7 @@
   = render partial: "shared/reports/table_empty", locals: { type: type }
 - else
   %table.govuk-table
-    = render partial: "shared/reports/table_header", locals: { type: type }
+    = render partial: "shared/reports/table_header", locals: { type: type, show_organisation: current_user.service_owner? }
 
     %tbody.govuk-table__body
       - grouped_reports.each do |organisation, reports|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1198,6 +1198,8 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 
 ## [unreleased]
 
+- Remove heading for unpopulated organisation column in organisation reports table for BEIS users
+
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-130...HEAD
 [release-130]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-129...release-130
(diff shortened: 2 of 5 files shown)


app/views/shared/reports/_grouped_table.html.haml
```diff
@@ -2,7 +2,7 @@
   = render partial: "shared/reports/table_empty", locals: { type: type }
 - else
   %table.govuk-table
-    = render partial: "shared/reports/table_header", locals: { type: type }
+    = render partial: "shared/reports/table_header", locals: { type: type, show_organisation: current_user.service_owner? }
```
Gweaton, line 5: I think it'd be good to test this in feature specs, as they're often the only way of testing view logic. Would that be a pain to add here?

CristinaRO (approved):

app/views/shared/reports/_grouped_table.html.haml
yndajas: I've added an additional commit to cover the headings. I toyed with restructuring the whole file into different routes, but it would've messed with some of the existing contexts, and some tests cover multiple routes

Have a couple of concerns:
- I wonder if it's a little complex, although it was already a somewhat abstracted spec
- I've used `HideFromBullet` to avoid an eager load error, but previously we haven't used this in specs. Not sure if it's worth adding it to the controller itself

Would you mind taking a second look?

Gweaton (approved):
I think these tests are good enough as they match the existing spec pattern fairly closely. I've only just realised I don't think this page is linked anywhere in the service for BEIS users though? Or am I missing something? It's accessible at `/organisations/:id/reports` but there's no link to that page. How did you come across it?

yndajas:
> I think these tests are good enough as they match the existing spec pattern fairly closely. I've only just realised I don't think this page is linked anywhere in the service for BEIS users though? Or am I missing something? It's accessible at `/organisations/:id/reports` but there's no link to that page. How did you come across it?

From the homepage, click on any of the "View reports" links in the organisations table

PR #2009, opened 2023-02-09
(2740) Prevent editing `is_oda` via UI

yndajas:
## Changes in this PR

This prevents the activity details tab from including a link to edit the `is_oda` field, while retaining the link to add this for ISPF activities where the field is `nil`

The card also mentioned preventing this via bulk upload, but we don't have an `is_oda` column in the bulk upload template - this is handled by separate templates uploaded via separate forms. We have measures in place to ensure we don't do anything when there's a mismatch between the ODA type of the template and the form, but in any case the importer's activity updater doesn't modify `is_oda`

### Before

image

### After

image

## Next steps

### DIFF
--- a/app/views/shared/activities/_activity.html.haml
+++ b/app/views/shared/activities/_activity.html.haml
@@ -34,7 +34,7 @@
           = t("summary.label.activity.is_oda.#{activity_presenter.is_oda}")
       %dd.govuk-summary-list__actions
         - if policy(activity_presenter).update? && activity_presenter.requires_is_oda?
-          = a11y_action_link(t("default.link.#{activity_presenter.call_to_action(:is_oda)}"), activity_step_path(activity_presenter, :is_oda), t("activerecord.attributes.activity.is_oda"))
+          = a11y_action_link(t("default.link.add"), activity_step_path(activity_presenter, :is_oda), t("activerecord.attributes.activity.is_oda"))
 
   .govuk-summary-list__row.identifier
     %dt.govuk-summary-list__key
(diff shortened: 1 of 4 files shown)


yndajas:
[Prevent all users from editing ODA/Non-ODA type (bulk & UI)](https://trello.com/c/lGYZP14F/2740-prevent-all-users-from-editing-oda-non-oda-type-bulk-ui)

app/views/shared/activities/_activity.html.haml
```diff
@@ -33,8 +33,8 @@
         - unless activity_presenter.is_oda.nil?
           = t("summary.label.activity.is_oda.#{activity_presenter.is_oda}")
       %dd.govuk-summary-list__actions
-        - if policy(activity_presenter).update? && activity_presenter.requires_is_oda?
-          = a11y_action_link(t("default.link.#{activity_presenter.call_to_action(:is_oda)}"), activity_step_path(activity_presenter, :is_oda), t("activerecord.attributes.activity.is_oda"))
+        - if policy(activity_presenter).update? && activity_presenter.requires_is_oda? && activity_presenter.is_oda.blank?
```
CristinaRO, line 36: I wonder if we could achieve the same result by redefining what `requires_is_oda?` means. Currently it checks that the activity is a programme, and that it's ISPF funded. We _could_ add `is_oda.nil?` to those checks, and then it will also apply to the form step (`skip_step unless @activity.requires_is_oda?` in activity_forms_controller).

CristinaRO (approved):

PR #2010, opened 2023-02-09
(2809) Prefix non-ODA programmes with NODA in import

yndajas:
## Changes in this PR

The changes in 46e5f29 meant that non-ODA activities added via the UI are prefixed "NODA-". This change was implemented in the `ActivityUpdater` service, which is not touched when creating activities via the importer. Projects and third-party projects created via the importer would be prefixed "NODA-" through including their parent's identifier in their own, however new level B imported activities would not

This rectifies that by amending the RODA identifier generator to include "NODA-" when passed `is_non_oda: false` via
`ActivityDefaults#generate_roda_identifier` (or the activity factory)

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1195,6 +1195,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 ## [release-130] 2023-02-13
 
 - Update Ruby to 3.0.5
+- Prefix imported non-ODA programmes "NODA-"
 
 ## [unreleased]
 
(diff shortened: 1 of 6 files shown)


yndajas:
[Uploading level B non-ODA activities doesn't create NODA prefix](https://trello.com/c/VRaTCN62/2809-uploading-level-b-non-oda-activities-doesnt-create-noda-prefix)

Gweaton (approved):
This looks good.

It might be an idea to split out the refactor of `roda_identifier_generator_spec` into the first commit, to keep commits as small as possible, as I think it'd reduce cognitive load on the reviewer with the level of detail in one commit message. Even though the issue was surfaced by adding `compact`, that's something you could detail in the commit message as something being fixed in future, if necessary.

PR #2011, opened 2023-02-09
Upgrade parser to fix warnings

yndajas:
## Changes in this PR

In def189d, we fixed some warnings by upgrading `parser`. `parser` warnings have returned since the Ruby 3 upgrade, so this upgrades `parser` once more to remove the warnings

Example of warning:
```
warning: parser/current is loading parser/ruby30, which recognizes3.0.4-compliant syntax, but you are running 3.0.5.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```

## Next steps

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -292,7 +292,7 @@ GEM
       jwt (>= 1.5, < 3)
     orm_adapter (0.5.0)
     parallel (1.22.1)
-    parser (3.1.2.1)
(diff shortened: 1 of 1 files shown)


CristinaRO:
The [standard gem bump](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2016) also updated parser, and it doesn't raise any warnings currently.

PR #2012, opened 2023-02-10
Fix documentation query

jkempster34:
## Changes in this PR
Tiny change - Fix documentation query

Pluck doesn't work on single records, we can access `id` directly instead.


yndajas (approved):
Might be good to hold off merging until we've done a Ruby 3 production release, just to keep that release as small as possible in case of issues

PR #2014, opened 2023-02-13
(2820) Fix incorrectly-passed CSV arguments

Gweaton:
## Changes in this PR

These are a couple of spots we missed when upgrading to Ruby 3, as the language has now changed the way it handles keyword arguments.

I've updated the feature tests here to use less mocking so we can check the result of the methods that are actually called, so that this will hopefully be caught in future. I've mocked responses to S3 in the test environment here too.

## Next steps

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -93,6 +93,7 @@ Rails.application.configure do
     Bullet.add_safelist type: :unused_eager_loading, class_name: "Activity", association: :implementing_org_participations
     Bullet.add_safelist type: :unused_eager_loading, class_name: "OrgParticipation", association: :organisation
     Bullet.add_safelist type: :unused_eager_loading, class_name: "Comment", association: :owner
+    Aws.config.update(stub_responses: true)
   end
 
   config.hosts = [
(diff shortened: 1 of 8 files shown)


config/environments/test.rb
```diff
@@ -93,6 +93,7 @@
     Bullet.add_safelist type: :unused_eager_loading, class_name: "Activity", association: :implementing_org_participations
     Bullet.add_safelist type: :unused_eager_loading, class_name: "OrgParticipation", association: :organisation
     Bullet.add_safelist type: :unused_eager_loading, class_name: "Comment", association: :owner
+    Aws.config.update(stub_responses: true)
```
CristinaRO, line 96: Nice find! #TIL

CristinaRO (approved):
Great work, thank you!

config/environments/test.rb
Gweaton: Me too!

CristinaRO:
[Fix post ruby 3 upgrade errors](https://trello.com/c/GkHBpiB7/2820-fix-post-ruby-3-upgrade-errors)

PR #2019, opened 2023-02-14
(2811) Prevent opposite ODA type child activities

Gweaton:
## Changes in this PR

Prevent child ODA activities being created on non-ODA parents and vice-versa via the bulk upload.

This is currently not possible when adding activities via the UI, but was possible with the bulk upload.

Users were able to specify an ODA parent activity in the non-ODA activity upload CSV and vice-versa without an error being raised.

The error here isn't the _most_ helpful, as the `is_oda` value is not explicitly entered in the form, and we don't display the parent's `is_oda` value in the `value` column of the CSV, but I think this is something we can iterate on in future with the redesign of the error page and the refactoring of the importer class.

I've added some more tests to the ever-growing and unwieldy `import_spec` to act as a kind of integration test here, just to make extra sure this is working as expected, but I do feel there's a growing need to refactor this!

### After

image

## Next steps

### DIFF
--- a/app/validators/same_parent_oda_type_validator.rb
+++ b/app/validators/same_parent_oda_type_validator.rb
@@ -1,0 +1,11 @@
+class SameParentOdaTypeValidator < ActiveModel::Validator
+  def validate(activity)
+    return unless activity.is_ispf_funded?
+    return if activity.is_oda.nil?
+    return if activity.parent.fund?
+
+    if activity.is_oda != activity.parent.is_oda
+      activity.errors.add(:oda_parent, I18n.t("activerecord.errors.models.activity.attributes.parent.invalid"))
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1199,6 +1199,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 ## [unreleased]
 
 - Remove heading for unpopulated organisation column in organisation reports table for BEIS users
+- Prevent child ODA activities being created on non-ODA parents and vice-versa via the bulk upload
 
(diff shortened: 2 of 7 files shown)


app/validators/same_parent_oda_type_validator.rb
```diff
@@ -0,0 +1,11 @@
+class SameParentOdaTypeValidator < ActiveModel::Validator
+  def validate(activity)
+    return unless activity.is_ispf_funded?
+    return if activity.is_oda.nil?
+    return if activity.parent.fund?
```
CristinaRO, line 5: Could this be `activity.programme?` for clarity?

Gweaton: Could I probe on what you mean by clarity here?

I went back and forth on this a little, and this is going to get a weirdly philosophical, but I'll do my best at trying to explain my thought process!

Although there's not _really_ a difference here in terms of the code logic (a `Programme`'s `parent` should always be a `fund` in the app's current state), I felt there was a distinction between an activity being a "Programme" and its parent being a "Fund". We're validating the parent in other places here, so it feels like, to me, _that's_ the entity with a value we want to check against our current entity (a "Programme", or anything below it).

If for any reason (highly unlikely) the hierarchy of levels were to change, and a "Project" was the top level entity, this would be _slightly_ more future-proof as well. I appreciate there are tons of other places this won't apply, so this may not be a compelling argument...

Hope that made some sort of sense, I've struggled to put my thoughts into words here!

CristinaRO (approved):

CristinaRO:
[Uploading activities via non-ODA link while giving them an ODA parent should raise an error](https://trello.com/c/UyQ7L3Rk/2811-uploading-activities-via-non-oda-link-while-giving-them-an-oda-parent-should-raise-an-error)

PR #2022, opened 2023-02-14
Skip Pundit’s policy scope verification on index

CristinaRO:
## Changes in this PR

In our base controller, we use Pundit’s mechanism for alerting us if we forget to authorize any controller actions. Collective controller actions such as index are checked for the use of `policy_scope`.

The reports index doesn’t call the policy scope, so we see this error in our server logs:
`Pundit::PolicyScopingNotPerformedError (ReportsController)`

We can skip the policy scope check, because permissions are checked against the user’s organisation.

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -7,6 +7,8 @@ class ReportsController < BaseController
   include StreamCsvDownload
   include Reports::Breadcrumbed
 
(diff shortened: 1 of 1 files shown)


CristinaRO:
[Investigate server errors on `/reports` page](https://trello.com/c/n3Y4WlEk/2818-investigate-server-errors-on-reports-page)

PR #2023, opened 2023-02-15
Omit legacy fields for ISPF activities

jkempster34:
### Omit legacy fields for ISPF activities
Currently, we show the `intended_beneficiaries`, `recipient_country`, and
`recipient_region` legacy fields on the "details" page for all activities.

ISPF activities will not have the data associated with these legacy fields, so
we can remove the fields when the activity is of that type.

The XML download is unchanged. When an XML download is generated, the tags for
the legacy data are only created if the data is present. As ISPF activities
will never have any data for these field, there is no need for additional checks.

### Remove legacy fields from ISPF activities order object
The order objects defines the fields that will be included in an exported
report.

ISPF activities will not have the data associated with these legacy fields, so
we can remove them from the order object.

### Include the BEIS identifier in the Actual Export spec
The RODA identifier and BEIS identifier make up the first two columns of actual
exports.

The BEIS identifier is missing from this spec, so this adds it in.

### Remove QuarterlyActualExport service
It looks as though the intention was to rename this to `Actual::Export`, but
instead the file was duplicated, leaving behind the `QuarterlyActualExport`
service.

This is no longer referenced anywhere so it can be removed.

## Next steps


spec/services/actual/export_spec.rb
```diff
@@ -2,77 +2,64 @@
   let(:partner_organisation) { create(:partner_organisation) }
```
jkempster34, line 2: The commit that changed this file can be dropped if we want. I was barking up the wrong tree when I changed this, I thought the fields needed removing from the "actuals" export as well, but it seems that that includes all activities types. However, I think the added test coverage might help someone in future.

app/views/shared/xml/_activity.xml.haml
```diff
@@ -45,10 +45,10 @@
     - benefitting_countries_with_percentages(activity.benefitting_countries).each do |country|
       %recipient-country{"code" => country.code, "percentage" => country.percentage}
         %narrative= country.name
-  - elsif activity.recipient_country?
+  - elsif activity.recipient_country? && !activity.is_ispf_funded?
```
jkempster34, line 48: This is added on each `elsif` so that we don't need to nest the conditionals any deeper.

CristinaRO:
[Remove legacy fields from ISPF UI and exports](https://trello.com/c/Ngcl49HV/2816-remove-legacy-fields-from-ispf-ui-and-exports)

app/views/shared/xml/_activity.xml.haml
CristinaRO: I'm of two minds about this. Because these legacy fields cannot be populated for new activities, neither via the UI nor via bulk import, they'll always be nil for ISPF activities, so the check is not necessary. It's not harmful either, since the first condition will always evaluate to `false`. (As long as ruby keeps doing "short circuit" evaluation, and we'd have bigger problems if this changed...)

spec/views/shared/activities/activity_spec.rb
```diff
@@ -559,6 +559,16 @@
         expect(body.find(".intended_beneficiaries .govuk-summary-list__value")).to have_content("Uganda")
         expect(body.find(".intended_beneficiaries .govuk-summary-list__key")).to have_content("Legacy field: not editable")
       end
+
+      context("when the activity is ISPF funded") do
```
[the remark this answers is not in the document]
jkempster34: No idea why I put brackets there. I don't usually!

app/views/shared/activities/_activity.html.haml
```diff
@@ -274,26 +274,27 @@
         = activity_presenter.benefitting_region
       %dd.govuk-summary-list__actions
 
-  .govuk-summary-list__row.recipient_region
-    %dt.govuk-summary-list__key
-      = t("summary.label.activity.recipient_region_html")
-    %dd.govuk-summary-list__value
-      = activity_presenter.recipient_region
-    %dd.govuk-summary-list__actions
+  - unless activity_presenter.is_ispf_funded?
```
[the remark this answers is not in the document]
jkempster34: Do I just want to straight swap this for `unless source_fund.ispf?` ?

app/views/shared/xml/_activity.xml.haml
jkempster34: Understood. I'll remove these!

app/views/shared/activities/_activity.html.haml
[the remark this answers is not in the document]
jkempster34: Okay, I've made the swap. I think that's right!

CristinaRO (approved):
Needs rebasing - hopefully one last time 😅

PR #2024, opened 2023-02-15
(2819) Spike - imported forecasts post-upload

CristinaRO:
## Changes in this PR
- `ForecastHistory#set_value` returns `nil` when a forecast is deleted
- Imported forecasts list unchanged values if included in the uploaded CSV

## Screenshots of UI changes

An extreme edge case example of uploading a spreadsheet with one forecast with the same value as the existing forecast reported in the same report. Previously there would have been a successful message with an empty table:

### Before
Screenshot 2023-02-15 at 21 19 41

After this PR the success table lists the forecast:

### After
Screenshot 2023-02-15 at 17 40 56

## Next steps

### DIFF
--- a/app/services/forecast/import.rb
+++ b/app/services/forecast/import.rb
@@ -106,8 +106,10 @@ class Forecast
 
     def import_forecast(activity, financial_quarter, value, header:)
       return if value.blank?
+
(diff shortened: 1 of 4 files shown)


Gweaton (commented):
I think this approach looks good - the results table looks a lot more logical to me too.

CristinaRO:
Being reworked into a new PR

PR #2026, opened 2023-02-16
(2787) (2819) Successfully uploaded summary table fixes

CristinaRO:
## Changes in this PR
- Fix the display of successfully uploaded forecasts for the cases when:
  - the original forecast is deleted
  - the imported values were already in the report
 
#### Reuploading identical values resulted in an empty summary table
Screenshot 2023-02-15 at 21 19 41

#### Reuploading identical values results in them being listed and counted as part of the total uploaded
Screenshot 2023-02-15 at 17 40 56

## Next steps

### DIFF
--- a/spec/services/forecast_history_spec.rb
+++ b/spec/services/forecast_history_spec.rb
@@ -202,18 +202,24 @@ RSpec.describe ForecastHistory do
         ])
       end
 
-      it "deletes an original entry with a zero value" do
-        history.set_value(0)
-
-        expect(history_entries).to eq([])
-      end
-
       it "does not create a historical event when the value is first set" do
         expect { history.set_value(20) }.to not_create_a_historical_event
       end
 
       it "returns a _Forecast_" do
         expect(history.set_value(20)).to be_a(Forecast)
+      end
+
+      context "when an original entry is set to zero" do
+        it "deletes the original entry" do
+          history.set_value(0)
+
+          expect(history_entries).to eq([])
+        end
+
+        it "returns nil" do
+          expect(history.set_value(0)).to be_nil
+        end
       end
     end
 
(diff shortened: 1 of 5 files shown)


CristinaRO:
[Uploading forecasts: "successfully uploaded" summary table does not include forecasts that already existed in the system and have not changed value](https://trello.com/c/vwV5GSQY/2787-uploading-forecasts-successfully-uploaded-summary-table-does-not-include-forecasts-that-already-existed-in-the-system-and-have-n)

CristinaRO:
[Uploading forecasts: "successfully uploaded" summary table does not correctly reflect forecasts that have been replaced with 0 (zero)](https://trello.com/c/DThEVJmG/2819-uploading-forecasts-successfully-uploaded-summary-table-does-not-correctly-reflect-forecasts-that-have-been-replaced-with-0-zero)

spec/services/forecast_history_spec.rb
```diff
@@ -232,13 +238,19 @@ def history_entries
...
-          ["original", 1, 2015, 10],
-          ["revised", 2, 2015, 0]
-        ])
+          expect(history_entries).to eq([
+            ["original", 1, 2015, 10],
+            ["revised", 2, 2015, 0]
+          ])
+        end
+
+        it "returns a _Forecast_" do
```
Gweaton, line 251: again, no biggie but any reason for the `_` around `Forecast` here? Is that a convention we use for class name?

Gweaton (approved):
Looks good to me, nice work. Just a couple of optional comments

spec/services/forecast_history_spec.rb
CristinaRO: If we were producing documentation with rdoc, the underscore would be formatted as `<em>`tags.

That being said, I don't think we do this anywhere else in our specs, nor have I ever used the specs to generate docs with rdoc.

Is it worth removing it everywhere in the file though?

Gweaton: Ah interesting. Let's not worry about it here, I think we're fine to leave it in if we've used it elsewhere in this file.

PR #2028, opened 2023-02-20
Only generate IATI identifiers for ODA activities

jkempster34:
## Changes in this PR
Currently, we generate an IATI identifier for all activities.

However, we have had some feedback that this is confusing, and the IATI
identifier should be absent for non-ODA activities in the CSV download.

Only generate IATI identifiers for ODA activities. Add some coverage for the
`is_oda` and `transparency_identifier` attributes of the activity defaults.

We don't have any ISPF activities in production, so there aren't any existing
identifiers that need to be hidden.

## Next steps

### DIFF
--- a/spec/services/activity_defaults_spec.rb
+++ b/spec/services/activity_defaults_spec.rb
@@ -3,6 +3,7 @@ require "rails_helper"
 RSpec.describe ActivityDefaults do
   let(:beis) { create(:beis_organisation) }
   let(:partner_organisation) { create(:partner_organisation) }
+  let(:is_oda) { nil }
 
   let(:fund) { create(:fund_activity, :gcrf) }
   let(:programme) { create(:programme_activity, :gcrf_funded, parent: fund) }
@@ -15,6 +16,7 @@ RSpec.describe ActivityDefaults do
   let(:ispf_non_oda_project) { create(:project_activity, parent: ispf_non_oda_programme, is_oda: false) }
 
   let!(:current_report) { create(:report, :active, organisation: partner_organisation, fund: fund) }
+  let(:expected_transparency_identifier) { "#{Organisation::SERVICE_OWNER_IATI_REFERENCE}-#{subject[:roda_identifier]}" }
 
   before do
     # some reports which we don't expect to be returned as 'originating_report'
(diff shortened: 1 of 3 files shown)


jkempster34:
[To remove IATI Identifiers for non-ODA activities in CSV downloads](https://trello.com/c/zaOS1hvy/2804-to-remove-iati-identifiers-for-non-oda-activities-in-csv-downloads)

spec/services/activity_defaults_spec.rb
```diff
@@ -28,12 +29,50 @@
...
     subject { activity_defaults.call }
 
+    describe "`is_oda` and `transparency_identifier`" do
+      let(:parent_activity) { create(:programme_activity, :ispf_funded) }
+      let(:expected_transparency_identifier) { "#{Organisation::SERVICE_OWNER_IATI_REFERENCE}-#{subject[:roda_identifier]}" }
+
+      context "when activity is ODA" do
+        let(:is_oda) { true }
+
+        it { expect(subject[:is_oda]).to be true }
```
Gweaton, line 46: In terms of consistency within this file, we've got a bit of a mix of `it { expect }` and `it "a human-readable test assertion"`. I personally prefer the latter, but am not too precious about it. I do think it'd be good to pick one style and be consistent with it throughout the file, though.

If you've got a preference for `it { expect }`, how would you feel about adding a separate commit to this PR to tidy up the rest of the `it`s for consistency where possible?

Otherwise, I think it would be good to stick with the other pattern. What do you reckon?

spec/services/activity_defaults_spec.rb
```diff
@@ -75,7 +114,7 @@
       end
 
       it "sets the transparency identifier" do
```
Gweaton, line 116: If we're going to keep this style of `it`, might be worth updating this test description to `it "sets the transparency identifier to nil"` here (and below) so it's a bit more obvious it's a special case.

Gweaton (approved):
This looks good to me.

I've got a couple of test-related comments, but nothing too major. Let me know what you think, but otherwise this is good to merge.

spec/services/activity_defaults_spec.rb
jkempster34: Yep, no problem! Updated

Gweaton (approved):

PR #2031, opened 2023-02-21
(2801) Hide SID truncation note on non-ODA activities

yndajas:
## Changes in this PR

The purpose step of the activity form includes a note on both the title and description fields about truncation of data when published to Statistics in International Development. Non-ODA activities are not published to this, so the note is irrelevant and confusing in that context. This therefore hides the note when accessing the step on a non-ODA activity

### Before

image

### After

image

## Next steps

### DIFF
--- a/app/controllers/activity_forms_controller.rb
+++ b/app/controllers/activity_forms_controller.rb
@@ -22,6 +22,8 @@ class ActivityFormsController < BaseController
     when :linked_activity
       skip_step unless policy(@activity).update_linked_activity?
       @options = linkable_activities_options(@activity)
+    when :purpose
+      @activity_presenter = ActivityPresenter.new(@activity)
     when :objectives
       skip_step unless @activity.requires_objectives?
     when :call_present
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1208,6 +1208,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Omit legacy fields for ISPF activities on the "details" page.
 - Remove legacy fields from ISPF reports.
 - Only generate IATI identifiers for ODA activities
+- Remove reference to truncating data published to Statistics in International Development for non-ODA activities
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-130...HEAD
 [release-130]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-129...release-130
--- a/app/presenters/activity_presenter.rb
+++ b/app/presenters/activity_presenter.rb
@@ -5,6 +5,18 @@ class ActivityPresenter < SimpleDelegator
   include ActivityHelper
   include CountryHelper
 
+  def title_hint
+    hint = I18n.t("form.hint.activity.title", level: I18n.t("page_content.activity.level.#{to_model.level}"))
+    hint += " #{I18n.t("form.hint.activity.title_sid_note")}" unless is_non_oda?
+    hint
+  end
+
+  def description_hint
+    hint = I18n.t("form.hint.activity.description")
+    hint += " #{I18n.t("form.hint.activity.description_sid_note")}" unless is_non_oda?
+    hint
+  end
+
   def aid_type
     return if super.blank?
     translate("activity.aid_type.#{super.downcase}")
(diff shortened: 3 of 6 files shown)


yndajas:
[To remove references to Statistics for International Development (non-ODA) (RODA UI)](https://trello.com/c/MtNU7W6t/2801-to-remove-references-to-statistics-for-international-development-non-oda-roda-ui)

spec/controllers/activity_forms_controller_spec.rb
```diff
@@ -27,6 +27,31 @@
...
+          expect(response.body).to include(t("form.hint.activity.title_sid_note"))
+          expect(response.body).to include(t("form.hint.activity.description"))
+          expect(response.body).to include(t("form.hint.activity.description_sid_note"))
+        end
+
+        context "when it's a non-ODA activity" do
+          let(:activity) { create(:programme_activity, :ispf_funded, is_oda: false) }
+
+          it "includes character limit notes about the title and description fields, but not SID information" do
+            expect(response.body).to include(t("form.hint.activity.title", level: t("page_content.activity.level.#{activity.level}")))
```
jkempster34, line 46: Personal rant, but I find using the translations in the specs obfuscates what's actually going on (though I understand it keeps things DRY and the rest of the project is like this!)

app/controllers/activity_forms_controller.rb
```diff
@@ -22,6 +22,12 @@ def show
     when :linked_activity
       skip_step unless policy(@activity).update_linked_activity?
       @options = linkable_activities_options(@activity)
+    when :purpose
+      @title_hint = t("form.hint.activity.title", level: t("page_content.activity.level.#{@activity.level}"))
```
jkempster34, line 26: I wonder if these belong in some helper or presenter, this controller is particularly hefty, and this logic is very view-based 🤔

jkempster34 (approved):
Hello 👋 I wanted to start helping out with the reviews.

This looks great to me! 

I've left a few comments/thoughts, particularly about the hefty controller

app/controllers/activity_forms_controller.rb
```diff
@@ -22,6 +22,12 @@ def show
...
+      @title_hint += " #{t("form.hint.activity.title_sid_note")}" if !@activity.is_non_oda?
```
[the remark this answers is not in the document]
yndajas: Yes, and we should! 🤦🏼‍♀️

spec/controllers/activity_forms_controller_spec.rb
yndajas: Yeah, I did wonder about this - I know Cristina prefers just writing it out, and I think I do too. I kind of copied the pattern from `spec/controllers/level_b/activities/uploads_controller_spec.rb`, but since this spec doesn't have any other translations, I think let's go with explicitly writing it out to continue setting that precedent

app/controllers/activity_forms_controller.rb
yndajas: Yeah I wondered about this too. We don't seem to have any hint methods in the `ActivityPresenter` or `ActivityHelper`, but we do have `linkable_activity_select_label` in the presenter, so I think it could go in the same place

yndajas: I've pushed this as a fixup commit for now: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2031/commits/dc54e4f16ae59438c8d7bfaeb3c1b2cf9bbe3a74. What do you think of this? If it looks good, I'm squash and merge

jkempster34: Hmm, I think that's better? What do you think?

They should maybe be `linkable_activity_title` and `linkable_activity_hint` though?

yndajas: This is for just the title and description field for activities (it's not related to linking - confusing, I know!)

I think it's better to shift this detail over to the presenter, yeah, and I think the presenter unit tests are more direct than testing the controller response

jkempster34: Oh sorry, I didn't mean linking, I have linking on the brain! I meant the purpose step. Should it be `purpose_title` `purpose_hint`

yndajas: Ah! So this step, confusingly, contains two fields, neither of which are `purpose`. It's a single step that includes fields for `title` and `description`, and what what we're doing here is determining the hint for the `title` field and the hint for the `description` field

jkempster34: 👍 I'm happy with everything if you are!

PR #2032, opened 2023-02-21
(2802) Hide fields irrelevant to non-ODA activities

yndajas:
## Changes in this PR

Non-ODA activities are not reported to IATI, therefore the IATI identifier and "Include in IATI export?" fields are irrelevant and confusing when looking at the activity details tab for such activities. This removes them in this context

### Before

image

image

### After

image

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1209,6 +1209,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Remove legacy fields from ISPF reports.
 - Only generate IATI identifiers for ODA activities
 - Remove reference to truncating data published to Statistics in International Development for non-ODA activities
+- Hide IATI identifier and "Include in IATI XML export?" fields on activity details tab for non-ODA activities
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-130...HEAD
 [release-130]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-129...release-130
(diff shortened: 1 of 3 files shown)


CristinaRO (approved):
It occurred to me that once we hide the toggle, there will be no way to make the field `false`.

`publish_to_iati` is `true` by default, so non-ODA activities will be stuck as theoretically publishable.

Currently the way we exclude them is en masse, via an argument given to `Find{Level}Activities` services, `include_ispf_non_oda_activities`, which is false by default.

It's worth checking from a big picture level whether this is still desirable, if there is ever the case that non-ODA activities would be publishable to IATI, and which mechanism would be the most intuitive and maintainable for the future.

The current situation means that we're generating invalid XMLs when there are only non-ODA activities in the system, which will be the case for a while.

yndajas:
I did wonder about this. Is it worth throwing something into the `ActivityFormsController` and `Activity::Import:ActivityCreator` to set `publish_to_iati` to false on non-ODA activities?

> It's worth checking from a big picture level whether this is still desirable, if there is ever the case that non-ODA activities would be publishable to IATI, and which mechanism would be the most intuitive and maintainable for the future.
>
> The current situation means that we're generating invalid XMLs when there are only non-ODA activities in the system, which will be the case for a while.

Is there an immediate action here, or is this more thoughts for the future?

CristinaRO:
> Is there an immediate action here, or is this more thoughts for the future?

Thoughts for the (near) future.

CristinaRO:
[To remove ODA fields from non-ODA activity details (RODA UI)](https://trello.com/c/shwcOfkN/2802-to-remove-oda-fields-from-non-oda-activity-details-roda-ui)

PR #2033, opened 2023-02-21
(2805) Add global feature flag for linked activities

jkempster34:
### Put the wizard linked activities behind a feature flag
We want to disable the linked activities feature for now, to be re-enabled in
future. We have had some feedback that linking is confusing to POs while
there's no ODA funds being spent.

This hides the linked activities feature in the wizard unless the feature flag
is active.

### Put describe page and summary linked activities behind a feature flag
We want to disable the linked activities feature for now, to be re-enabled in
future. We have had some feedback that linking is confusing to POs while
there's no ODA funds being spent.

This hides the linked activities row on the describe page and summary unless
the feature flag is active.

## Next steps

### DIFF
--- a/app/policies/activity_policy.rb
+++ b/app/policies/activity_policy.rb
@@ -69,7 +69,7 @@ class ActivityPolicy < ApplicationPolicy
   end
 
   def update_linked_activity?
-    return unless record.is_ispf_funded?
+    return unless ROLLOUT.active?(:activity_linking) && record.is_ispf_funded?
 
     if record.programme?
       return beis_user? && record.linked_child_activities.empty?
--- a/spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
+++ b/spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
@@ -272,24 +272,31 @@ RSpec.feature "BEIS users can create a programme level activity" do
       expect(created_activity.policy_marker_nutrition).to be_nil
     end
 
-    scenario "a new non-ODA programme can be linked to an existing ODA programme" do
-      linked_oda_activity = create(:programme_activity, :ispf_funded, is_oda: true, extending_organisation: partner_organisation)
-      non_oda_activity.linked_activity = linked_oda_activity
+    context "when the `activity_linking` feature flag is enabled" do
+      before do
+        allow(ROLLOUT).to receive(:active?).and_call_original
+        allow(ROLLOUT).to receive(:active?).with(:activity_linking).and_return(true)
+      end
 
-      visit organisation_activities_path(partner_organisation)
+      scenario "a new non-ODA programme can be linked to an existing ODA programme" do
+        linked_oda_activity = create(:programme_activity, :ispf_funded, is_oda: true, extending_organisation: partner_organisation)
+        non_oda_activity.linked_activity = linked_oda_activity
 
-      click_on t("form.button.activity.new_child", name: linked_oda_activity.associated_fund.title)
+        visit organisation_activities_path(partner_organisation)
 
-      form = ActivityForm.new(activity: non_oda_activity, level: "programme", fund: "ispf")
-      form.complete!
+        click_on t("form.button.activity.new_child", name: linked_oda_activity.associated_fund.title)
 
-      expect(page).to have_content(t("action.programme.create.success"))
+        form = ActivityForm.new(activity: non_oda_activity, level: "programme", fund: "ispf")
+        form.complete!
 
-      created_activity = form.created_activity
+        expect(page).to have_content(t("action.programme.create.success"))
 
-      expect(created_activity.title).to eq(non_oda_activity.title)
-      expect(created_activity.is_oda).to eq(non_oda_activity.is_oda)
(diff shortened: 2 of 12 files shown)


jkempster34:
[Hiding the linking feature (UI)](https://trello.com/c/j2fMEMH3/2805-hiding-the-linking-feature-ui)

app/policies/activity_policy.rb
```diff
@@ -69,7 +69,7 @@ def redact_from_iati?
   end
 
   def update_linked_activity?
-    return unless record.is_ispf_funded?
+    return unless ROLLOUT.active?(:activity_linking) && record.is_ispf_funded?
```
jkempster34, line 72: I wonder if we want to flip this to match the other feature flags.

`return if ROLLOUT.active?(:hide_activity_linking)`

jkempster34: Worth saying that if the flag is absent for any reason, that the safer option is to keep this as it is `ROLLOUT.active?(:activity_linking)`

Gweaton: I think `active?(FEATURE_NAME)` is more readable too as we're talking about the feature being active, rather than hiding it being active 🤯

spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
```diff
@@ -257,24 +257,31 @@
       expect(created_activity.tags).to eq(non_oda_activity.tags)
     end
 
-    scenario "a new non-ODA programme can be linked to an existing ODA programme" do
-      linked_oda_activity = create(:programme_activity, :ispf_funded, is_oda: true, extending_organisation: partner_organisation)
-      non_oda_activity.linked_activity = linked_oda_activity
+    context "when the `activity_linking` feature flag is enabled" do
```
Gweaton, line 260: Do you think it's worth adding tests here for the page being hidden when the flag is disabled for security here?

Gweaton (commented):
This looks good to me.

I think we also want to hide the section from the Activity "details" tab, and the top section of the activity where "Linked activity" is still shown, see screenshot below.

image

spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
jkempster34: I think this is covered multiple times by the existing specs because of how the helper works

```ruby
def fill_in_ispf_project_activity_form
    fill_in_identifier_step
    fill_in_linked_activity_step if ROLLOUT.active?(:activity_linking)
    fill_in_purpose_step
    fill_in_objectives_step if @activity.is_oda
    fill_in_sector_category_step
    fill_in_sector_step
    fill_in_call_details
    fill_in_call_applications
    fill_in_programme_status
    fill_in_dates
```

jkempster34:
> This looks good to me.
> 
> I think we also want to hide the section from the Activity "details" tab, and the top section of the activity where "Linked activity" is still shown, see screenshot below.
> 
> image

Cristina also picked up on this and the activity summary. I've added a new commit for them.

spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
jkempster34: Will do! Working in this now

jkempster34: These are covered in the existing specs.

jkempster34:
> > This looks good to me.
> > I think we also want to hide the section from the Activity "details" tab, and the top section of the activity where "Linked activity" is still shown, see screenshot below.
> > image
> 
> Cristina also picked up on this and the activity summary. I've added a new commit for them.

Worth mentioning tha users can still upload linked activities (successfully) via CSV. We have another ticket for that, but until we get around to it this may be weird...

spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
Gweaton: Do you mean the activity policy ones?

jkempster34: The linking specs are separate from the original "go through the form specs". So, we still have coverage for the journey without that specific page

I put the feature flag in the ActivityForm helper https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/a3954574841f927ca2798fdfb72debf6d8afb28d/spec/support/activity_form.rb#L132, so in [this spec](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/a3954574841f927ca2798fdfb72debf6d8afb28d/spec/features/beis_users_can_create_a_programme_level_activity_spec.rb#L233) the linking page is skipped.

Hope that makes sense

Gweaton: Ah I see - that's fine then

Gweaton (approved):
The second commit says "Put describe page..." - do you mean "details" page here?

jkempster34:
> The second commit says "Put describe page..." - do you mean "details" page here?

Yep, it should. Thanks!

PR #2034, opened 2023-02-22
(2803) Updates to non-ODA CSV templates and order of columns

jkempster34:
### Remove ODA fields from non-ODA CSV upload template
We have had some feedback that these fields are ODA, and therefore should not
be present on the non-ODA upload template.

### Move the comments column to the end of all the CSV templates
Currently, the comments columns is included in all CSV templates. We have had
some feedback that users would like comments to always come last
for consistency. 

Move the comments column so that it always comes last, after any conditional
columns.


jkempster34:
[To remove ODA fields from non-ODA activity (bulk upload)](https://trello.com/c/pFzpfW7J/2803-to-remove-oda-fields-from-non-oda-activity-bulk-upload)

yndajas (approved):
Looks good. Might be worth double checking that we want to keep SDGs in non-ODA templates at levels C and D, but the card does say to remove them from level B specifically, so hopefully that's correct!

PR #2037, opened 2023-02-22
(2500) Improve actuals and refunds importer error messages

yndajas:
## Changes in this PR

This hopefully fixes most of the issues raised around confusing messaging when bulk uploading actual and refund data. It:
- adds more specific error messages
- makes it so that various situations involving zero values raise errors rather than silently not importing the given row (while importing other valid rows)

It also fixes a small styling error with regard to the alignment of the heading and error table on the failed upload page

We may still have some signposting to do regarding adding comments, as mentioned on the card

### New error messages

image

### With fixed alignment

Screenshot 2023-02-22 at 17 34 26

## Next steps


yndajas:
[Simplify and clarify logic around zeroes and blanks for actuals and refunds uploads](https://trello.com/c/lqezUVuW/2500-simplify-and-clarify-logic-around-zeroes-and-blanks-for-actuals-and-refunds-uploads)

app/services/actual/import.rb
```diff
@@ -41,7 +41,13 @@ def import_row(row, index)
     end
 
     def add_error(row_number, column, cell_value, message)
-      @errors << Error.new(row_number, Converter::FIELDS[column], cell_value, message)
+      column_string = if column == :values
```
Gweaton, line 44: Although I'm usually against ternaries for return values in Ruby, I think I find them more readable in variable assignments, but this might just be me! What do you think?

app/services/actual/import.rb
```diff
@@ -155,12 +160,50 @@ def to_h
...
+          @row["Refund Value"] || "blank"
+        ].join(", ")
+
+        types = {
+          actual: value_type(value: @row["Actual Value"], field: :actual_value),
+          refund: value_type(value: @row["Refund Value"], field: :refund_value)
+        }
+
+        error_message = case types
+        when {actual: :non_zero, refund: :non_zero}
```
Gweaton, line 182: I like this! It's a nice way of avoiding a bunch of nested `if`s and feels easy to extend for future errors.

spec/features/users_can_upload_actuals_spec.rb
```diff
@@ -147,7 +147,7 @@ def expect_to_see_successful_upload_summary_with(count:, total:)
       expect(page).to have_xpath("td[1]", text: "Actual Value")
       expect(page).to have_xpath("td[2]", text: "2")
       expect(page).to have_xpath("td[3]", text: "fish")
-      expect(page).to have_xpath("td[4]", text: t("importer.errors.actual.non_numeric_value"))
+      expect(page).to have_xpath("td[4]", text: "Actual and refund values must be blank or numeric")
```
Gweaton, line 150: Is there a reason you've made assertions on real text values, rather than their translations? I've seen this in a few tests in this PR, which would mean we'd need to update them in both places if the translated text were to change. Not a huge deal, but just wondered if it was a testing preference?

Gweaton (approved):
This looks good to me, nice work. Just a couple of small comments

app/services/actual/import.rb
yndajas: Yeah I was trying for a while to work out a clean way of not nesting too much and making things readable, while also dealing with the complicated stuff like financial data conversion. It was all a little messy

app/services/actual/import.rb
yndajas: Me too, but I wondered whether it was a little complicated/long here:

`column_string = column == :values ? "Actual Value/Refund Value" : Converter::FIELDS[column]`

Maybe it's just the colons and quote marks in the symbol, string and namespacing making it look messy - it's probably fine!

spec/features/users_can_upload_actuals_spec.rb
yndajas: I think it might be a preference of @CristinaRO's (and @jkempster34's) to explicitly include the text in the test to make it more readable. I'm quite keen on DRY code, but I can see that it reads as documentation better when explicit. Open to whichever approach!

app/services/actual/import.rb
yndajas: I think with parentheses (suggested by Standard) it's fine:

`column_string = (column == :values) ? "Actual Value/Refund Value" : Converter::FIELDS[column]`

spec/features/users_can_upload_actuals_spec.rb
jkempster34: So for me, testing against the translations:
1. Hides the actual text making the test harder to read.
2. If the translation is missing, the test will still pass (I think) because both will be nil. Kind of relates [to not using logic in tests](https://canro91.github.io/2021/10/11/DontRepeatLogicInAssertions/)
3. The downside of changing the text in both places isn't a big deal, its just a small annoyance.

But I don't have really strong feelings!

yndajas: Feels like a sea of ambivalence 😅

CristinaRO: I don't mind either way; I think originally I wrote the specs as a demo and didn't want to consider the translation keys and values as firmly decided, and they were left like that 😅

PR #2041, opened 2023-02-24
(2827b) collect original commitment figure form

yndajas:
I've made some slight changes, simplifying things and avoiding the repeated building/creation. Testing in the UI, it looks like all the errors appear as expected

I noticed if I put a `binding.pry` after the `activity.commitment = Commitment.new` lines, it wouldn't actually reach the next part (the `.save!` line) in some invalid data cases, whereas doing it this way (with `.build_commitment`) it does. I guess that method failed (silently?) and that's why in the controller you had to build/create it again (because the first time didn't work at all)

### DIFF
--- a/app/controllers/activity_forms_controller.rb
+++ b/app/controllers/activity_forms_controller.rb
@@ -101,9 +101,6 @@ class ActivityFormsController < BaseController
         @implementing_organisations = Organisation.active.sorted_by_name
         render_step :implementing_organisation
       when :commitment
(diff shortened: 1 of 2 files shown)


Gweaton:
Amazing. I think setting `activity.commitment = Commitment.new(x,y)` implicitly called `save` on the `Commitment`, so it was failing validations while `build_commitment` doesn't seem to, until we actually save the commitment. Thanks for this! I've incorporated your change into my commit so will close this PR but have have added you as a co-author so as not to lose your co-writing credit!

PR #2042, opened 2023-02-27
(2826) Make ODA-only fields blank for non-ODA activities

jkempster34:
> I did a lot of thinking about where to place the nullifying of the attributes. I ended up putting them in a before_save > callback. Happy to hear any better ideas. My thinking:
> 1. They can't go in the ActivityDefaults class because that is only applied to new activities. We don't want to apply this class to existing activities because it has some new activity only logic.
> 2. We could create a new NonOdaDefaults class, but then we have to be careful to apply it everywhere + it won't actually do much, it will only set some attributes to nil.
> 3. Even though the Activity class is large, I think there is stuff that is less-related than this that could be moved into another class potentially.
> 4. I've nullified the non-persisted defaults here, so it's nice that they are in the same place. 

### Make sure the transparency indicator is nil for all non-ODA activities
Currently, the transparency indicator is set as `nil` in the ActivityDefaults
class.

However, these defaults are only applied to activities at the start of the UI
wizard, or through the import function.

This means that non-ODA activities that are created via the UI that do not have
a non-ODA parent, will have a transparency indicator. 

To fix, apply the defaults once non-ODA is selected as part of the wizard
journey as well.

### Update non-persisted defaults when activity is non-ODA
These defaults should be null is the activity is non-ODA.

### Make ODA activity eligibility column nullable
ODA eligibility should be null if the activity is non-ODA.

Make the column nullable. 

### Nullify activity columns when activity is non-ODA
Certain columns should be null when an activity is non-ODA.

An activity can be created as non-ODA if it's parent activity is non-ODA, or if
it created through the import function, but can also change to non-ODA as part
of the UI wizard.

Add a before_save callback which sets any ODA-only attributes to nil when
is_oda is false, and the is_oda value has changed.

This also moves the nullification of transparency identifier to the model along
with the other null attributes.

### Update presenters for when non-ODA activities have nil attributes
The presenters don't currently handle when some of the activity attributes are
nil.

Now that these can be nil for the non-ODA activities, update them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1205,15 +1205,16 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Fix the display of successfully uploaded forecasts for the cases when:
   - the original forecast is deleted
   - the imported values were already in the report
-- Omit legacy fields for ISPF activities on the "details" page.
-- Remove legacy fields from ISPF reports.
+- Omit legacy fields for ISPF activities on the "details" page
+- Remove legacy fields from ISPF reports
 - Only generate IATI identifiers for ODA activities
 - Remove reference to truncating data published to Statistics in International Development for non-ODA activities
 - Hide IATI identifier and "Include in IATI XML export?" fields on activity details tab for non-ODA activities
-- Remove ODA fields from the non-ODA CSV upload template.
-- Move the comments column to the end of all the CSV templates.
+- Remove ODA fields from the non-ODA CSV upload template
+- Move the comments column to the end of all the CSV templates
 - Make error messages clearer when importing actuals and refunds
 - Fix login issues after running the training database sync script by clearing remember tokens during password reset step and clearing sessions as an additional step
+- ODA only attributes are blank on non-ODA activities
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-130...HEAD
 [release-130]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-129...release-130
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -372,11 +372,11 @@ class Activity < ApplicationRecord
   end
 
   def finance
-    STANDARD_GRANT_FINANCE_CODE
+    STANDARD_GRANT_FINANCE_CODE unless is_non_oda?
   end
 
   def tied_status
-    UNTIED_TIED_STATUS_CODE
+    UNTIED_TIED_STATUS_CODE unless is_non_oda?
   end
 
   def capital_spend
@@ -384,7 +384,7 @@ class Activity < ApplicationRecord
   end
 
   def flow
-    DEFAULT_FLOW_TYPE
+    DEFAULT_FLOW_TYPE unless is_non_oda?
   end
 
   def default_currency
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -9,6 +9,21 @@ class Activity < ApplicationRecord
   POLICY_MARKER_CODES = Codelist.new(type: "policy_significance", source: "beis").hash_of_integer_coded_names
   DESERTIFICATION_POLICY_MARKER_CODES = Codelist.new(type: "policy_significance_desertification", source: "beis").hash_of_integer_coded_names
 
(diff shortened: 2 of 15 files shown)


jkempster34:
[In Quarterly CSV reports for Non-ODA ISPF activities non-ODA specific fields should be blank](https://trello.com/c/epnEejfQ/2826-in-quarterly-csv-reports-for-non-oda-ispf-activities-non-oda-specific-fields-should-be-blank)

app/models/activity.rb
```diff
@@ -81,6 +96,7 @@ class Activity < ApplicationRecord
   FORM_STATE_VALIDATION_LIST = FORM_STEPS.map(&:to_s).push("complete")
 
   before_validation :strip_control_characters_from_fields!
+  before_save :set_non_oda_defaults, if: %i[is_oda_changed? is_non_oda?]
```
jkempster34, line 99: Could this go wrong? I know one of the reasons some people avoid ActiveRecord callbacks is because they can cause unintended effects.

jkempster34, line 99: I'll probably change this to `if: -> { is_oda_changed? && is_non_oda? }`. I think that's clearer.

db/schema.rb
```diff
@@ -49,7 +49,7 @@
     t.integer "total_applications"
     t.integer "total_awards"
     t.string "collaboration_type"
-    t.integer "oda_eligibility", default: 1, null: false
+    t.integer "oda_eligibility", default: 1
```
Gweaton, line 52: Are there any places we want to move this not-null validation to? e.g. if it's an ODA/non-ISPF activity, do we want to make `oda_eligibility` required somewhere like the model?

CHANGELOG.md
```diff
@@ -1213,6 +1213,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Remove ODA fields from the non-ODA CSV upload template.
 - Move the comments column to the end of all the CSV templates.
 - Update Ruby to 3.1.3
+- ODA only attributes are blank on non-ODA activities.
```
Gweaton, line 1216: Sorry, most annoying comment I've ever written, but let's remove `.`s at the end of the Changelog lines here

Gweaton (approved):
Nice work, I think this is the most sensible approach for now. Just a couple of small things

db/schema.rb
jkempster34: Apparently this is already in place! 

```ruby
validates :oda_eligibility, presence: true, on: :oda_eligibility_step, if: :requires_oda_eligibility?

def requires_oda_eligibility?
  !is_non_oda?
end
```

I think the `on: :oda_eligibility_step` means it won't get validated until the form_state is complete, but I think that's fine, because that happens in the importer too.

Gweaton: Great, thanks for checking that

CHANGELOG.md
yndajas: Ha, that was on my mind too @Gweaton

jkempster34: Is this a rule I don't know about, or just to keep them matching? I thought a changelog was a list of sentences, not titles (if that makes sense?).

Also while we are on the subject, I think the changelog here is upside down https://keepachangelog.com/en/1.0.0/#:~:text=The%20latest%20version%20comes%20first not sure if it's worth flipping it.

Gweaton: More to keep them matching the rest of the file. Yeah the upside-downness is a bit annoying, but we haven't really been too bothered by it to prioritise changing it just yet.

yndajas: Agree that it's more consistency than anything else, but I guess given each point is written as a present tense declarative, they're quite similar to the first line of a commit. They are legit sentences though

I used to be a big adherent of consistently using full stops everywhere, but eventually I [dropped it at the end of paragraphs](https://forum.wordreference.com/threads/omitting-a-full-stop-at-end-of-sentence-in-paragraphs.2362088) in all but the most formal/publication-y writing because of the connotations it has in internet/messaging contexts of a serious tone

jkempster34: Interesting! 

I've got into the habit of using an exclamation mark every other sentence for the same reason!

yndajas: Same!

PR #2030, opened 2023-02-20
[2831b] Refactor Commitment model and importer

Gweaton:
## Changes in this PR

*Part 2 of 2*: Updates references to Financial Year & Financial Quarter to use new `transaction_date` field once it's been populated by the script introduced in #2047.

This PR refactors a few aspects of the Commitment model and the script used to import them:

- Updates the Commitment importer so no date fields are required. The `transaction_date` will be inferred from an Activity's `planned_start_date` if it exists, then the `actual_start_date`. If neither is present for any reason (as appears to be the case with some of our data despite validations), we fall back on the date the activity was created.
- Uses the new `transaction_date` field to display the Commitment figure on an Activity's financial details tab (see screenshot)
- Removes `financial_year` and `financial_quarter` from commitments

### Before

Screenshot 2023-02-20 at 15 40 20

### After

image

## Next steps

- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [x] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.
- [ ] Do any environment variables need amending or adding?
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/xml-validation.md)

PR #2047, opened 2023-03-01
[2831a] Add transaction date to commitments

Gweaton:
## Changes in this PR

*Part 1 of 2*: We'll be running the migration script before merging the second PR #2030 which will update the codebase to reference the new field

This adds a `transaction_date` field to commitments that will be used instead of the Financial Year and Quarter, as well as a script that will populate this value from the current Financial Year and Quarter for existing Commitments.

New activities will, in future, use the planned start date or actual start date for their transaction date.

## Next steps

- [ ] Run `db/data/20230220095332_set_transaction_date_on_commitments.rb` on deployed environment

### DIFF
--- a/db/data/20230220095332_set_transaction_date_on_commitments.rb
+++ b/db/data/20230220095332_set_transaction_date_on_commitments.rb
@@ -1,0 +1,14 @@
+# Run me with `rails runner db/data/20230220095332_set_transaction_date_on_commitments.rb`
+
+commitments_to_update = Commitment.where(transaction_date: nil)
+
+puts "Setting transaction date on #{commitments_to_update.count} Commitments"
+
+commitments_to_update.each do |commitment|
+  commitment.update(transaction_date: commitment.first_day_of_financial_period)
+  print "."
+end
+
+commitments_to_update.reload
+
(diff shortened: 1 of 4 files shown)


db/data/20230220095332_set_transaction_date_on_commitments.rb
```diff
@@ -0,0 +1,14 @@
+# Run me with `rails runner db/data/20230220095332_set_transaction_date_on_commitments.rb`
+
+commitments_to_update = Commitment.where(transaction_date: nil)
+
+puts "Setting transaction date on #{commitments_to_update.count} Commitments"
+
+commitments_to_update.each do |commitment|
+  print "."
```
yndajas, line 8: Should the `.` go after the `update` just in case it's unsuccessful and you want to count how many dots there are as indicators of successful transactions? 🤓

yndajas (approved):
Looks fine and it was already approved, but just one suggestion!

db/data/20230220095332_set_transaction_date_on_commitments.rb
Gweaton: fair point!

PR #2049, opened 2023-03-02
(2835) Feature flag the activity linking for the import CSVs

jkempster34:
### Decouple Import class from `activity_csv_columns.csv`
We want to be able to feature flag one of the CSV columns (activity linking).
However, this is difficult currently. Including the feature flag in the
`activity_csv_columns` initializer means that the effects of the feature flag
won't apply until the app has been restarted. Including the feature flag in the
current large Import class will be messy.

This decouples the rest of the `Import` class from the
`activity_csv_columns.csv` file, and provides an interface for common
operations.

Note: the `invalid_for_level_b_ispf_non_oda` and
`invalid_for_level_c_d_ispf_non_oda` are left as they were for now, but are
candidates for further refactoring.

### Feature flag the activity linking feature in the import CSVs
We have put the activity linking feature behind a feature flag in the UI. We
have had some feedback that linking is confusing to POs while
there's no ODA funds being spent.

This does the same for the import CSVs.

## Next steps


app/models/activity/import/field.rb
```diff
@@ -0,0 +1,80 @@
+class Activity::Import::Field
```
jkempster34, line 1: How do we feel about the interface for this class? I feel I got a bit carried away

jkempster34:
[Feature flag the linking feature in the various CSVs](https://trello.com/c/cCSmtFsP/2835-feature-flag-the-linking-feature-in-the-various-csvs)

jkempster34 (commented):
Uploading a CSV with a linked_activity_id when the feature is turned off will silently fail, i.e., the activity will not be linked, and there will be no error message.

I reckon that's fine?

spec/models/activity/import/field_spec.rb
```diff
@@ -3,8 +3,23 @@
     subject { described_class.all }
 
     it "returns an array of all Field objects" do
-      expect(subject.count).to eq 50
+      expect(subject.count).to eq 49
       expect(subject).to all(be_a(described_class))
+
+      linked_activity_id_field = subject.select { |field| field.attribute_name == :linked_activity_id }
+      expect(linked_activity_id_field).to be_empty
```
yndajas, line 10: `be_empty` or `be_nil`?

spec/models/activity/import/field_spec.rb
```diff
@@ -75,6 +89,14 @@
         it "returns the expected headings" do
           expect(subject).to eq expected_fields
         end
+
+        context "when the `activity_linking` feature flag is active" do
+          before { allow(ROLLOUT).to receive(:active?).with(:activity_linking).and_return(true) }
+
+          it "includes linked activity" do
+            expect(subject).to match_array expected_fields << "Linked activity RODA ID"
```
yndajas, line 97: Hmm I can see how `expected_fields` is useful in this case

yndajas (commented):
This looks good! I've made some comments/suggestions on the details, but in general it seems like a useful improvement in how we handle the import fields

Might be good to get some more eyes on it

yndajas:
> Uploading a CSV with a linked_activity_id when the feature is turned off will silently fail, i.e., the activity will not be linked, and there will be no error message.
> 
> I reckon that's fine?

They would have to manually add the `linked_activity_id` in to the template to do so, right?

app/models/activity/import/field.rb
```diff
@@ -0,0 +1,78 @@
...
```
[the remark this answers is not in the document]
jkempster34: Thanks! I thought they were in `models`

spec/models/activity/import/field_spec.rb
```diff
@@ -0,0 +1,358 @@
...
+
+    subject { described_class.find_all_by(headings: headings) }
+
+    it "returns an instance of Field" do
+      expected = [
+        have_attributes(attribute_name: :roda_identifier, heading: "RODA ID", exclude_from_converter: true),
+        have_attributes(attribute_name: :parent_id, heading: "Parent RODA ID", exclude_from_converter: true)
+      ]
+
+      expect(subject).to match_array(expected)
```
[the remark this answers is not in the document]
jkempster34: It's not unfortunately!

spec/models/activity/import/field_spec.rb
```diff
@@ -0,0 +1,358 @@
...
+    end
+  end
+
+  describe ".where" do
+    subject { described_class.where(level: level, type: type).map(&:heading) }
```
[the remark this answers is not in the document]
jkempster34: Agreed, I've moved it into a `let(:field_headings) { subject.map(&:heading) }` which may still be contentious, but less so

spec/models/activity/import/field_spec.rb
```diff
@@ -0,0 +1,358 @@
...
+            "Sector",
+            "Aid type",
+            "Aims/Objectives",
+            "ISPF themes",
+            "ISPF ODA partner countries",
+            "ISPF non-ODA partner countries",
+            "Tags",
+            "Comments"
+          ]
+        end
```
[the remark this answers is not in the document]
jkempster34: You have mentioned below, but it's so I can add to it for the linked activity specs!

spec/models/activity/import/field_spec.rb
jkempster34: Select returns an empty array when not found, there may be another tidier way of doing this though

app/models/activity/import/field.rb
```diff
@@ -0,0 +1,78 @@
...
+      fields = source.select { |_key, value| headings.include?(value[:heading]) }
+
+      from_source(fields)
+    end
+
+    def where(level:, type:)
+      fields = source.select { |_key, value| value.dig(:inclusion, level, type) }
+
+      from_source(fields)
+    end
```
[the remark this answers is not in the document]
jkempster34: Thanks for the review! I've made most of the changes.

So, part of this is me projecting my desire for it to look like ActiveRecord, but because they are keyword arguments, they have to be included, so really the signature is `find_by(attribute_name:)`.

However, I totally get your point, and I did wonder what would happen when somebody wanted to find by heading because `find_by(heading:)` wouldn't be allowed.

jkempster34:
> > Uploading a CSV with a linked_activity_id when the feature is turned off will silently fail, i.e., the activity will not be linked, and there will be no error message.
> > I reckon that's fine?
> 
> They would have to manually add the `linked_activity_id` into the template to do so, right?

I was thinking if they were using a template they had downloaded previously. Not sure if it's a massive problem though

app/models/activity/import/field.rb
Gweaton: I like this too - that `Activity::Import` class and its tests have been getting more and more unwieldy, so breaking them down into smaller chunks is much appreciated, and this feels like a good place to split out some of this functionality.

spec/services/activity/import/field_spec.rb
```diff
@@ -0,0 +1,377 @@
+RSpec.describe Activity::Import::Field do
+  describe ".all" do
+    subject { described_class.all }
+
+    it "returns an array of all Field objects" do
+      expect(subject.count).to eq 50
+      expect(subject).to all(be_a(described_class))
```
Gweaton, line 7: I have _opinions_ on the readability of `described_class` in general, but for this type-checking tests i.e. `to be_a(Field)` I prefer explicitly asserting on the name of the class as I find it more readable in the context of an individual test. Is there something `described_class` gives us over the use of the class name specifically, or is more of a convention?

Gweaton (approved):
This is a really nice refactor, thanks for your hard work. I'll approve but might be worth waiting for Ynda's approval before we merge.

spec/services/activity/import/field_spec.rb
jkempster34: I used to work on a project with a very strict Rubocop, so I've picked up some habits from that. No particular benefit, happy to change it.

app/models/activity/import/field.rb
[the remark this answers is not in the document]
yndajas: Good point re: signature. I guess I typically try to make the method name descriptive enough in itself, but you're right that the keyword arguments are part of the signature. But as it stands, couldn't you switch the method names of `find_all_by` and `where`? That feels a little strange to me

spec/models/activity/import/field_spec.rb
yndajas: Ah true. Given you're just looking for one field, would be make more sense to use `.find` here and then assert on `nil` (in both tests)?

yndajas:
> I was thinking if they were using a template they had downloaded previously. Not sure if it's a massive problem though

I think since ISPF isn't live in RODA yet, it should be fine (to ignore)

PR #2050, opened 2023-03-02
(2844) Fix missing presenter when step is invalid

jkempster34:
### Fix missing presenter when step is invalid
Currently, when the purpose step is invalid, the view is re-rendered without
an instance of ActivityPresenter, leading to an error.

This moves the instantiation of the presenter to the view so that it is
available for both steps.

### DIFF
--- a/app/views/activity_forms/purpose.html.haml
+++ b/app/views/activity_forms/purpose.html.haml
@@ -1,10 +1,11 @@
 = render layout: "wrapper" do |f|
+  - activity_presenter = ActivityPresenter.new(@activity)
   = f.govuk_fieldset legend: { text: @page_title, tag: :h1, size: "xl" } do
     = f.govuk_text_field :title,
       label: { text: t("form.label.activity.title", level: custom_capitalisation(t("page_content.activity.level.#{f.object.level}"))) },
-      hint: { text: @activity_presenter.title_hint }
+      hint: { text: activity_presenter.title_hint }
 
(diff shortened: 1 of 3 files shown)


app/views/activity_forms/purpose.html.haml
```diff
@@ -1,10 +1,11 @@
 = render layout: "wrapper" do |f|
+  - activity_presenter = ActivityPresenter.new(@activity)
```
jkempster34, line 2: I don't usually like creating variables like this in the view, but I think in this case it's fine and keeps the controller tidier 🤷

jkempster34 (commented):
It will be hard to add nice spec for this without testing an "invalid" submission, which I don't think is worth it in this case.

jkempster34:
[Bug: attempting to delete the title field on an activity results in a server error](https://trello.com/c/hrn0TRy6/2844-bug-attempting-to-delete-the-title-field-on-an-activity-results-in-a-server-error)

app/views/activity_forms/purpose.html.haml
yndajas: It's a little different to how we do it with `:implementing_organisation`, which sets an instance variable in both `show` and `update`, but this way is DRYer

Has it been done in other activity form views? If so, I think it's probably okay as long as we don't set a bunch of variables directly in views

yndajas (approved):

PR #2051, opened 2023-03-03
[2828] Collect original commitment figure in bulk Activity upload

Gweaton:
For reviewing, start at commit [63b6cfa](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2051/commits/63b6cfa13de8d80207c63965002e10858449d73c) as this is based off branches already approved in previous PR #2041 

## Changes in this PR

This adds a new "Original commitment figure" column to the activity CSV template and allows the figure to be set when uploading the CSV. This is required at level D but not at B and C.

This is only possible when creating a new Activity, as only BEIS users should be able to edit this figure and have confirmed they can do so in the UI, and don't need to be able to do this in bulk.

#### Error with invalid commitment value

image

#### Error creating level D activity without commitment

image

#### Trying to update the commitment figure via bulk upload

image

## Next steps


Gweaton:
https://trello.com/c/t6yENcKH/2828-allow-original-commitment-figure-to-be-set-in-the-bulk-upload-csv

spec/services/activity/import_spec.rb
```diff
@@ -1166,6 +1173,38 @@
       expect(subject.errors.first.message).to eq(I18n.t("importer.errors.activity.parent_not_found"))
     end
 
+    context "original commitment figure" do
```
yndajas, line 1193: "Original"?

spec/models/activity_spec.rb
```diff
@@ -652,6 +652,44 @@
...
+      context "when the activity has no planned start or end date" do
+        it "does not validate the commitment's transaction date, so as not to display a misleading error in the activity upload errors" do
+          activity = build(:project_activity, actual_start_date: nil, planned_start_date: nil)
+          activity.build_commitment(value: 1000)
+          activity.save
+
+          expect(activity.errors.full_messages).not_to include(
+            "Original commitment figure Original commitment figure is invalid"
+          )
+        end
```
yndajas, line 690: Is it worth adding in an assertion that the commitment is not persisted?

app/services/activity/import.rb
```diff
@@ -287,6 +287,10 @@ def create
           return
         end
 
+        if @activity.third_party_project? && row["Original commitment figure"].blank?
+          @errors[:commitment] = ["", I18n.t("importer.errors.activity.commitment_required")]
+        end
```
yndajas, line 294: Again I wonder if this is a converter thing, although the previous lines are doing a similar thing

spec/services/activity/import_spec.rb
```diff
@@ -1203,6 +1203,46 @@
...
+          expect(subject.errors.first.csv_column).to eq("Original commitment figure")
+          expect(subject.errors.first.column).to eq(:commitment)
+          expect(subject.errors.first.message).to eq(
+            "An original commitment figure is required for level D activities"
+          )
+        end
+      end
+
+      context "when not present and creating a non-level D activity" do
+        it "doesn't set a commitment,nor raise an error" do
```
yndajas, line 1252: `commitment, nor`

yndajas (approved):
Brain power is a little too divided to think holistically about this, but generally it makes sense. Just commented on a few details

app/services/activity/import.rb
```diff
@@ -389,6 +389,10 @@ def convert_to_attributes
           attributes[:call_present] = (@row["Call open date"] && @row["Call close date"]).present?
         end
 
+        if attributes[:commitment]
+          attributes[:commitment].transaction_date = attributes[:planned_start_date] || attributes[:actual_start_date]
+        end
```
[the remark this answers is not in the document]
Gweaton: I wanted to be able to do this, but we don't have access to other rows (`planned_start_date` and `actual_start_date`) there, unfortunately, so we need to do it outside that method. This felt cleaner than figuring out a way of passing those in to those methods due to the way we `send` them.

app/models/activity.rb
```diff
@@ -176,6 +176,7 @@ class Activity < ApplicationRecord
   has_many :historical_events, dependent: :destroy
 
   has_one :commitment, dependent: :destroy
+  validates_associated :commitment
```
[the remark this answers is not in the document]
Gweaton: Yeah, this is only exercised in the bulk upload when we create a new activity, not in the UI form, it's what allows us to populate the errors against the Activity when saving the Commitment on it.

spec/features/users_can_upload_activities_spec.rb
```diff
@@ -375,6 +383,9 @@
       # Then I should see the tags
       expect(page).to have_content("Ayrton Fund")
       expect(page).to have_content("ICF Funded")
+
+      # And I should see the original commitment figure
+      expect(page).to have_content("Original commitment figure £1,000.00")
```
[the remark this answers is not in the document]
Gweaton: I wanted to test this functionality as it's code involved in creating a new nested entity on an Activity, which we have much of elsewhere in the code (everything else is mostly text values), do you think it's overdoing it?

spec/fixtures/csv/valid_ispf_non_oda_activities_upload.csv
```diff
@@ -1,2 +1,2 @@
-RODA ID,Parent RODA ID,Linked activity RODA ID,Title,Description,Partner organisation identifier,ODA Eligibility,Activity Status,Call open date,Call close date,Total applications,Total awards,Planned start date,Planned end date,Actual start date,Actual end date,Sector,UK PO Named Contact,ISPF themes,ISPF non-ODA partner countries,Comments,Implementing organisation names,Tags
-,ISPF-NON-ODA-PROGRAMME-ID,,A title,A description,example-id-1,1,7,11/01/2019,11/05/2019,14,5,10/10/2020,10/10/2021,,,12182,Someone Somebody,1|2,CA|CN,A comment,,2|4
+RODA ID,Parent RODA ID,Linked activity RODA ID,Title,Description,Partner organisation identifier,ODA Eligibility,Activity Status,Call open date,Call close date,Total applications,Total awards,Planned start date,Planned end date,Actual start date,Actual end date,Sector,UK PO Named Contact,ISPF themes,ISPF non-ODA partner countries,Comments,Implementing organisation names,Tags,Original commitment figure
```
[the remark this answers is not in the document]
Gweaton: Yeah I think this was the case before. It'd be nice for them to be in the same order but the order doesn't really matter in terms of functionality. I'll leave this as is for now as there's a bit of a time limit on this.

app/services/activity/import.rb
Gweaton: We don't have access to `@activity` in the converter, which is why we do this outside.

PR #2053, opened 2023-03-07
(2834) Set `publish_to_iati` to `false` on non-ODA activities

yndajas:
## Changes in this PR

- Set `publish_to_iati` to `false` on non-ODA activities
- Add missing spec
- Remove redundant functionality in activity finder classes

## Next steps

### DIFF
--- a/spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
+++ b/spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
@@ -68,6 +68,8 @@ RSpec.feature "BEIS users can create a programme level activity" do
         organisation: partner_organisation
       )
       expect(created_activity.commitment.value).to eq(activity.commitment.value)
+
+      expect(created_activity.publish_to_iati).to eq(true)
     end
   end
 
@@ -119,6 +121,8 @@ RSpec.feature "BEIS users can create a programme level activity" do
 
       expect(created_activity.transparency_identifier).to eql("GB-GOV-13-#{created_activity.roda_identifier}")
       expect(created_activity.commitment.value).to eq(activity.commitment.value)
+
+      expect(created_activity.publish_to_iati).to eq(true)
     end
   end
 
@@ -168,6 +172,8 @@ RSpec.feature "BEIS users can create a programme level activity" do
 
       expect(created_activity.transparency_identifier).to eql("GB-GOV-13-#{created_activity.roda_identifier}")
       expect(created_activity.commitment.value).to eq(activity.commitment.value)
+
+      expect(created_activity.publish_to_iati).to eq(true)
     end
   end
 
@@ -234,6 +240,8 @@ RSpec.feature "BEIS users can create a programme level activity" do
       expect(created_activity.oda_eligibility).to eq(oda_activity.oda_eligibility)
       expect(created_activity.tags).to eq(oda_activity.tags)
       expect(created_activity.commitment.value).to eq(oda_activity.commitment.value)
+
+      expect(created_activity.publish_to_iati).to eq(true)
     end
 
     scenario "a non-ODA activity can be created" do
@@ -277,6 +285,8 @@ RSpec.feature "BEIS users can create a programme level activity" do
       expect(created_activity.policy_marker_disaster_risk_reduction).to be_nil
       expect(created_activity.policy_marker_nutrition).to be_nil
       expect(created_activity.commitment.value).to eq(non_oda_activity.commitment.value)
+
+      expect(created_activity.publish_to_iati).to eq(false)
     end
 
     context "when the `activity_linking` feature flag is enabled" do
--- a/spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
+++ b/spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
@@ -69,7 +69,7 @@ RSpec.feature "BEIS users can create a programme level activity" do
       )
       expect(created_activity.commitment.value).to eq(activity.commitment.value)
 
-      expect(created_activity.publish_to_iati).to eq(true)
+      expect(created_activity.publish_to_iati).to be(true)
     end
   end
 
@@ -122,7 +122,7 @@ RSpec.feature "BEIS users can create a programme level activity" do
       expect(created_activity.transparency_identifier).to eql("GB-GOV-13-#{created_activity.roda_identifier}")
       expect(created_activity.commitment.value).to eq(activity.commitment.value)
 
-      expect(created_activity.publish_to_iati).to eq(true)
+      expect(created_activity.publish_to_iati).to be(true)
     end
   end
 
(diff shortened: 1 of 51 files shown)


yndajas:
[Non-ODA activities should default to publish_to_iati: false](https://trello.com/c/Ugp9vnnt/2834-non-oda-activities-should-default-to-publishtoiati-false)

app/services/activity_defaults.rb
```diff
@@ -35,6 +36,10 @@ def is_oda?
     @is_oda.nil? ? @parent_activity.is_oda : @is_oda
   end
 
+  def publish_to_iati?
```
jkempster34, line 39: Is this needed. Can we just default `publish_to_iati` to true, then set it to false in `set_non_oda_defaults`?

spec/services/activity/import_spec.rb
```diff
@@ -320,9 +320,9 @@
       expect(existing_activity.objectives).to eq(existing_activity_attributes["Aims/Objectives"])
       expect(existing_activity.beis_identifier).to eq(existing_activity_attributes["BEIS ID"])
       expect(existing_activity.uk_po_named_contact).to eq(existing_activity_attributes["UK PO Named Contact"])
-      expect(existing_activity.sdgs_apply).to eql(true)
+      expect(existing_activity.sdgs_apply).to eq(true)
```
jkempster34, line 323: I think we should use `be(true)` or `be_true`, because Ruby Booleans are singletons

spec/services/activity/import_spec.rb
```diff
@@ -1276,6 +1276,25 @@
         expect(new_activity.roda_identifier).to start_with("NODA-")
       end
 
+      it "sets ODA-only attributes to nil" do
```
jkempster34, line 1279: Thank you for this!

jkempster34 (commented):
Just adding some quick thoughts. Looks good!

app/services/activity_defaults.rb
yndajas: Good question - I wasn't sure whether `is_oda_changed?` would be true when creating projects/third-party projects that inherit `is_oda` during instantiation, and when creating activities via bulk upload. Does `_changed?` return `true` on creation or just when something is changed from one value to another?

Can easily test it by removing this and seeing if the tests pass!

spec/services/activity/import_spec.rb
yndajas: This is old but I think `be_true` is equivalent to `be_truthy`: https://bengribaudo.com/blog/2014/01/17/2674/rspec-be-true-be_true-are-different

We use `eq(true)` or `eq true` more than `be true` by the looks of it. Re: singletons, is the distinction that `be` suggests it is the single value `true` whereas `eq` suggests equivalence between two `true`s?

Happy to change it if `be` makes more sense!

app/services/activity_defaults.rb
yndajas: I think you're right - the only tests that fail when I comment out this stuff are in the `ActivityDefaults` spec, and the values are tested elsewhere

spec/services/activity/import_spec.rb
yndajas: Will update later!

spec/services/activity/import_spec.rb
```diff
@@ -1266,13 +1264,25 @@
...
+      it "prefixes the RODA identifier with 'NODA'" do
         expect { subject.import([new_non_oda_programme_attributes]) }.to change { Activity.count }.by(1)
 
         new_activity = Activity.order(:created_at).last
 
         expect(new_activity.roda_identifier).to start_with("NODA-")
       end
+
+      it "sets publish_to_iati to false" do
+        expect { subject.import([new_non_oda_programme_attributes]) }.to change { Activity.count }.by(1)
```
Gweaton, line 1299: I know it's in the test above for some reason, but is there a reason we need to check that the import creates a new activity in this test?

spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
```diff
@@ -69,7 +69,7 @@
       )
       expect(created_activity.commitment.value).to eq(activity.commitment.value)
 
-      expect(created_activity.publish_to_iati).to eq(true)
```
Gweaton: Ideally in future, I think it'd be good for refactors like this that are kind of separate from the main feature and touch the whole codebase to be part of their own PR, as much of a faff as that is. This'd make it quicker to review an individual PR, easier to track down any merges or PRs that might have introduced issues (even though a lot of that _is_ possible with commits)

Gweaton (approved):
Nice - looks good to me, just a couple of small comments.

spec/services/activity/import_spec.rb
yndajas: Good question. I suppose it gives us confidence that when we set the `new_activity` variable, we're actually checking a new activity rather than an existing one?

spec/features/beis_users_can_create_a_programme_level_activity_spec.rb
yndajas: Yeah, this ballooned a little. Originally I'd only changed three `eql` in one file to `eq`, but then I decided to do it all when changing to `be`. Should've separated it out!

spec/services/activity/import_spec.rb
Gweaton: I'm actually now wondering if this test is necessary in addition to the other tests we have. We currently have:
1. Unit test in the activity class - this feels like the most important one, as it's where the actual behaviour is written.
2. Feature tests that test it's set after the activity form
3. Service (semi-integration-style) tests here that test that it's created with the field in the importer. There's not actually any specific behaviour in the importer that sets this field value, unless I've missed something, so to me testing this value here potentially feels outside the responsibility of this class.

I wonder if it's worth testing 1 and 2 (potentially adding a spec to the `spec/features/users_can_upload_activities_spec.rb` and the BEIS alternative if we feel it needs more coverage).

I'm probably over thinking things, but what do you think?

yndajas: Yeah, Joe and I were talking about this before. We noticed that there's no equivalent to 2 for the importer in `spec/features/users_can_upload_activities_spec.rb` and the BEIS version, so I added it in here. The upload/import feature specs don't seem to test that all the values are right in the same way as the form/UI feature specs, but maybe they should (or we could at least move this and the previous two tests over there?)?

The `prefixes the RODA identifier with 'NODA` test is one you wrote actually, but I think the same applies to that: both the testing that an activity is created and that it's not testing functionality of the importer itself, just that functionality also applies when using the importer

Gweaton: Yeah, I think it would be good to try and reduce the scope of the importer's tests if we can move this kind of thing to the feature tests, although I appreciate this feels like a bit of scope creep in terms of what this PR is supposed to be doing.

I think both of those tests could be moved!

Gweaton: I think to keep scope down, we can park that stuff where it is for now and add it as something to refactor/consider when we get the time to refactor the importer spec?

yndajas: Sounds good - shall I merge for now and add a note on the import spec refactor card then?

yndajas: Or did you mean just the testing of all fields should be parked?

Gweaton: Let's merge now and update that card with this new plan

PR #2054, opened 2023-03-07
List items in tab list links are presentational

CristinaRO:
## Changes in this PR

The `ul` element has the role `tablist`. The specification says that it must “own” `tab` elements.[1] We’ve chosen to give the links inside `li` the role `tab`, rather than give `li` elements `tab` role.

Even though an “owned element” is defined as “any DOM descendant of the element”, Axe tools raises a critical issue: “Element has children which are not allowed” for the `ul` with role `tablist`, and “Required ARIA parent role not present: tablist“ for the `a` elements with role `tab`, suggesting that it was expecting direct child/parent elements rather than going through the intermediate `li` element.

Elsewhere we fixed this by giving the `li` elements presentational roles

This also fixes 4 serious issues for the `li` elements: "List item does not have a \<ul\>, \<ol\> parent element without a role, or a role="list”".

[1] https://www.w3.org/TR/wai-aria-1.1/#tablist

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1226,6 +1226,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Update Commitment-importing script to infer `transaction_date` value from Activity values
 - Collect original commitment figure in new activity form and bulk activity upload
 - Remove password reset field from new user form
+- Fix an accessibility issue on the Organisations page
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-131...HEAD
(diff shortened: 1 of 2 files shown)


CristinaRO:
[Organisations page (logged in as BEIS user): tablist and tab items (critical/serious)](https://trello.com/c/s1ZOJOz5/2850-organisations-page-logged-in-as-beis-user-tablist-and-tab-items-critical-serious)

yndajas (approved):

Gweaton (approved):

PR #2056, opened 2023-03-08
(2799) Budget revisions

jkempster34:
> Still a bit of work to get this perfect with the prototype, but it will require some further refactoring and changes outside of revisions, and this was getting quite large. Will open as a separate PR, card is here: https://trello.com/c/rNe9mCLh

### Install Audited for budget revisions
We want to add "revisions" to budgets.

This installs the Audited Gem which will enable us to record any changes to
budgets, along with who made them, and user provided comments.

Because we are using a UUID primary key for our Budget model, I have had to
update the column type of `auditable_id` in the generated migration.

### Audit the Budget model
This will create an audit of the Budget value when it is created or the value
amount is updated.

### Add script to create an audit for all existing Budgets
We are using the Auditable Gem to create revisions for Budgets.

The "Original" Budget will be the audit that is created when the Budget is
first created. However, while this works for new Budgets, we must create an
audit for all existing Budgets as well.

Auditable does not provide an easy way to do this. This uses a private method
of the library, then updates the created_at attribute to match the created_at
attributes of the Budget itself.

This script should be run following release before any new Budgets are
created/updated.

### Only allow users to edit budget amount
Currently, the create Budget form is used for both the create and update
actions.

Instead, we only want to allow users to update the Budget amount.

This also splits out the params so users can't hack the form and update
attributes they shouldn't.

### Add budget revisions page
Add a budget revisions page that lists the history of a budget.

This includes the value of the budget, the date of the change, and the
difference between a the budget and the previous budget.

### Allow users to add comments to budget revisions
Users want to be able to add comments to budget revisions.

This uses the comments function of Audited to enable that.

## Screenshots of UI changes
image
image
image
image

## Next steps


spec/features/users_can_edit_a_budget_spec.rb
```diff
@@ -84,7 +84,72 @@
       expect(historical_event.report).to eq(report)
     end
 
+    scenario "the revision history of a budget can be viewed" do
```
jkempster34, line 87: This is quite long, I don't think it's too hard to follow, but happy to try to split it out somehow if needed.

jkempster34:
[Build functionality for adding revisions to a budget](https://trello.com/c/mOXjoFty/2799-build-functionality-for-adding-revisions-to-a-budget)

app/views/budgets/revisions.html.haml
```diff
@@ -0,0 +1,31 @@
...
+          %tr.govuk-table__row
+            %th.govuk-table__header= t("revision_table.header.revision")
+            %th.govuk-table__header= t("revision_table.header.value")
+            %th.govuk-table__header= t("revision_table.header.difference")
+            %th.govuk-table__header= t("revision_table.header.changed_at")
+            %th.govuk-table__header= t("revision_table.header.audit_comment")
+
+        %tbody.govuk-table__body
+          - @audits.reverse.each do |audit|
+            - budget = BudgetPresenter.new(audit.revision)
```
jkempster34, line 20: Hoping to remove this Presenter in the follow up PR

spec/policies/budget_policy_spec.rb
```diff
@@ -29,14 +30,16 @@
...
         is_expected.to permit_action(:destroy)
+        is_expected.to permit_action(:revisions)
       end
     end
 
     context "when the activity is a project" do
       let(:activity) { create(:project_activity) }
 
-      it "only permits show" do
+      it "only permits show and revisisons" do
```
jkempster34, line 40: Reminder to self: fix typo

app/views/budgets/revisions.html.haml
```diff
@@ -0,0 +1,28 @@
...
+            %tr.govuk-table__row
+              %th.govuk-table__header{scope: "row"}
+                = row_header(audit)
+              %td.govuk-table__cell
+                = budget.value
+              %td.govuk-table__cell
+                = difference(earlier_audit: @audits.find_by(version: audit.version - 1), later_audit: audit)
+              %td.govuk-table__cell
+                = audit.created_at.strftime("%-d %B %Y")
```
CristinaRO: Could we use the localisation helper here, to maintain consistency if ever this changes?

I.e. `= I18n.l(audit.created_at.to_date)`

CristinaRO (commented):
Thank you for this massive chunk of work! A few comments, nothing major.

CristinaRO:
I forgot to mention: it would be good to have a breadcrumb on the Revisions page similar to how we implement breadcrumbs for the Edit page.

app/helpers/budget_revisions_helper.rb
```diff
@@ -0,0 +1,22 @@
+module BudgetRevisionsHelper
+  def link_to_revisions(budget)
+    update_audits_count = budget.audits.count { |audit| audit.action == "update" }
+    return "None" if update_audits_count.zero?
+
+    name = "#{update_audits_count} #{"revision".pluralize(update_audits_count)}"
+    link_to(name, activity_budget_revisions_path(budget.parent_activity_id, budget))
```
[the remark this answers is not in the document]
jkempster34: I added `govuk-link`, but didn't use `a11y_action_link`.

I _think_ that's only needed if the link is really non-descriptive like "Edit", whereas "1 revision" makes sense on it's own?

I'm not 100% sure though, and I can't find much about it on the guidance, unless I'm looking in the wrong place!

jkempster34:
> Thank you for this massive chunk of work! A few comments, nothing major.

Thanks for the review! Made all of those changes except the a11y link.

db/data/20230307105727_create_audits_for_existing_budgets.rb
```diff
@@ -0,0 +1,35 @@
...
+  def backfill_audit
+    write_audit(action: "create", audited_changes: audited_attributes,
+                comment: audit_comment)
+
+    Audited::Audit.find_by(auditable_type: "Budget", auditable_id: id).update!(created_at: created_at)
+  end
+end
+
+def budgets_without_create_audits
+  Budget
```
jkempster34, line 23: Does this look okay? Is there a better way of doing it. I'm fairly sure it works as expected after poking it a few times.

jkempster34: Hmmm, this is still imperfect. If someone creates a revision before this is run, we end up creating the "create" audit with the revised budget (+ the revision numbers are wrong). Can we assume that we can safely run this before any revisions are made, or do I need to do some more work to make this more robust?

image

db/data/20230307105727_create_audits_for_existing_budgets.rb
```diff
@@ -0,0 +1,42 @@
...
+# Audited does not provide an easy way to do this. This uses a private
+# method of the library, then updates the created_at attribute to match the
+# created_at attributes of the Budget itself.
+
+Budget.class_eval do
+  def backfill_audit
+    write_audit(action: "create", audited_changes: audited_attributes,
+      comment: audit_comment)
+
+    Audited::Audit.find_by(auditable_type: "Budget", auditable_id: id).update!(created_at: created_at)
```
jkempster34, line 18: Just had another thought @CristinaRO Maybe this should be `.update!(created_at: updated_at)` (instead of created_at), so that the "Original" budget is the most recent update of the budget 😕

CristinaRO: I'm reasonably comfortable with running the migration during a quiet time, but it's up to you!

CristinaRO (approved):
Stellar work, thank you!

PR #2057, opened 2023-03-08
(2500b) Signpost adding actual comments on their own

yndajas:
### Before

image

image

### After

image

image

## Next steps

### DIFF
--- a/app/controllers/actuals/uploads_controller.rb
+++ b/app/controllers/actuals/uploads_controller.rb
@@ -54,6 +54,8 @@ class Actuals::UploadsController < BaseController
 
         @success = true
         flash.now[:notice] = t("action.actual.upload.success")
+      else
+        @invalid_with_comment = importer.invalid_with_comment
       end
     else
       @errors = []
--- a/app/services/actual/import.rb
+++ b/app/services/actual/import.rb
@@ -7,7 +7,7 @@ class Actual
       end
     }
 
-    attr_reader :errors, :imported_actuals
+    attr_reader :errors, :imported_actuals, :invalid_with_comment
 
     def self.column_headings
       Converter::FIELDS.values
@@ -17,6 +17,7 @@ class Actual
       @report = report
       @uploader = uploader
       @errors = []
+      @invalid_with_comment = false
     end
 
     def import(actuals)
@@ -35,6 +36,8 @@ class Actual
 
       importer.errors.each do |attr_name, (value, message)|
         add_error(index, attr_name, value, message)
+
+        @invalid_with_comment ||= row["Comment"].present?
       end
 
       importer.actual
--- a/app/views/actuals/uploads/new.html.haml
+++ b/app/views/actuals/uploads/new.html.haml
@@ -8,7 +8,13 @@
 
       = t("page_content.actuals.upload.copy_html", report_actuals_template_path: report_actuals_upload_path(@report_presenter, format: :csv))
 
-      = t("page_content.actuals.upload.warning_html")
+      .govuk-warning-text
+        %span.govuk-warning-text__icon{ "aria-hidden" => "true" }
+          = "!"
+        %strong.govuk-warning-text__text
+          %span.govuk-warning-text__assistive
+            = t("default.warning")
+          = t("page_content.actuals.upload.warning")
 
   .govuk-grid-row
     .govuk-grid-column-two-thirds
--- a/app/views/actuals/uploads/new.html.haml
+++ b/app/views/actuals/uploads/new.html.haml
@@ -8,13 +8,7 @@
 
       = t("page_content.actuals.upload.copy_html", report_actuals_template_path: report_actuals_upload_path(@report_presenter, format: :csv))
 
-      .govuk-warning-text
-        %span.govuk-warning-text__icon{ "aria-hidden" => "true" }
-          = "!"
-        %strong.govuk-warning-text__text
-          %span.govuk-warning-text__assistive
-            = t("default.warning")
-          = t("page_content.actuals.upload.warning")
+      = render partial: "shared/warning_text", locals: { warning_text: t("page_content.actuals.upload.warning") }
 
   .govuk-grid-row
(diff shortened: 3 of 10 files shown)


yndajas:
[Simplify and clarify logic around zeroes and blanks for actuals and refunds uploads](https://trello.com/c/lqezUVuW/2500-simplify-and-clarify-logic-around-zeroes-and-blanks-for-actuals-and-refunds-uploads)

CristinaRO (approved):
Great work!

app/controllers/actuals/uploads_controller.rb
```diff
@@ -55,6 +55,8 @@ def update
         @success = true
         flash.now[:notice] = t("action.actual.upload.success")
       end
+
+      @has_zero_and_blank_error = importer.errors.find { |error| error.value == "0, blank" }.present?
```
jkempster34, line 59: Does this _have_ to be conditional on this specific error?

It seems like it's a helpful message if there are _any_ errors.

If it absolutely does have to be conditional on this, I'm not sure if I feel good about searching for that specific string

yndajas: I did wonder, but I think it makes sense in that those are the default values when you download the template. I.e. if they changed the values from the default of zero + blank, it's likely they aren't _solely_ attempting to add a comment (without adding/changing values). In that case, they'd fix the value errors and then the comment would be added as usual, when the values are valid. I can go back to Ming though if you think it should appear in other cases?

jkempster34: Ah, I see! 

My thinking was that showing it on any would would remove a decent amount of complexity, and that possibly brittle string comparison

Reading the message again, seems like it's always useful to me. Maybe pushing back might be worth it.
```
If you want to upload a comment without a financial amount, please do not provide invalid values via bulk upload. You can go to the Comments section of the relevant activities and create a comment there.
```

Not sure what @CristinaRO thinks, I was a bit slower reviewing this!

CristinaRO: If it was for me, I'd add the signposting unconditionally. (I don't think it would be too confusing for users who are encountering other errors.) Is it worth to check if BEIS agrees?

yndajas: Yep, and I'll let Ming know

yndajas:
Ming will iterate the design following feedback from the design collab, so I've switched this back to draft for now

app/views/actuals/uploads/new.html.haml
```diff
@@ -8,7 +8,13 @@
 
       = t("page_content.actuals.upload.copy_html", report_actuals_template_path: report_actuals_upload_path(@report_presenter, format: :csv))
 
-      = t("page_content.actuals.upload.warning_html")
+      .govuk-warning-text
```
jkempster34, line 11: Should this use the partial?

config/locales/models/actual.en.yml
```diff
@@ -61,21 +61,20 @@ en:
...
-          <li>Complete the template with your additional actuals data</li>
-          <li>Upload the template</li>
-          </ul>
-      upload:
-        heading: Upload data
-        copy_html:
-          <p class="govuk-body">Large numbers of actuals and refunds can be added via the actuals upload.</p>
-          <p class="govuk-body">For guidance on uploading actuals and refunds, see the <a class="govuk-link" target="_blank" rel="noreferrer noopener" href="https://beisodahelp.zendesk.com/hc/en-gb/articles/1500005601882-Downloading-the-Actuals-Template-in-order-to-Bulk-Upload">guidance in the help centre (opens in new tab)</a>.</p>
-        warning: Ensure you use the correct template (available below) when uploading the actuals and refunds.
+        pre_list_text: "To add more actuals on this page:"
```
jkempster34, line 66: Nice!

app/services/actual/import.rb
```diff
@@ -35,6 +36,8 @@ def import_row(row, index)
 
       importer.errors.each do |attr_name, (value, message)|
         add_error(index, attr_name, value, message)
+
+        @invalid_with_comment = true if row["Comment"].present?
```
jkempster34, line 40: Can this be simplified by removing the `@invalid_with_comment = false` and changing this to `@invalid_with_comment = row["Comment"].present?`

app/views/actuals/uploads/new.html.haml
jkempster34: And should it be conditional?

app/services/actual/import.rb
jkempster34: Or will this throw an error if the instance variable is absent and the attr_accessor is called 🤔

CristinaRO (approved):
Approved in order to not block you when you get to this PR again. Joe has some comments it would be good to look at.

app/views/actuals/uploads/new.html.haml
yndajas: This is a different warning and it's on the `new` view rather than the `update` (error) one

yndajas: But I guess we could abstract the partial that gets added later into a generic warning component with the translation passed in, since that's the only thing that differs?

app/services/actual/import.rb
yndajas: So the idea here is that we set it to `false` when the importer is instantiated (so it would never be absent when we get here), then the first row that has a comment sets it to `true` and then it never gets set to `false`. If we set it on every row, we'll forget that an earlier row had a comment in. But I think we can actually do `@invalid_with_comment ||= row["Comment"].present?`, which simplified the syntax but doesn't overwrite an earlier `true` row with a later `false`

app/views/actuals/uploads/new.html.haml
yndajas: Done the above!

jkempster34 (approved):
LGTM!

PR #2059, opened 2023-03-08
(2851) Make activity upload form IDs unique

yndajas:
## Changes in this PR

Oof. This was painful

This overwrites the default IDs for the `input` and `div.govuk-hint` elements, the `for` attribute of the `label` element, and the `aria-describedby` attribute of the `input` element, in order to address critical accessibility issues highlighted by axe DevTools

Would welcome any insight on better ways to do this, but I've already spent a few hours investigating/trial and erroring to get to this point

## Next steps

### DIFF
--- a/app/controllers/activities/uploads_controller.rb
+++ b/app/controllers/activities/uploads_controller.rb
@@ -35,7 +35,7 @@ class Activities::UploadsController < BaseController
     @report_presenter = ReportPresenter.new(report)
     upload = CsvFileUpload.new(params[:report], :activity_csv)
     @success = false
-    @type = params[:type].to_sym
+    @type = CGI.escapeHTML(params[:type]).to_sym
     is_oda = Activity::Import.is_oda_by_type(type: @type)
 
     prepare_default_report_trail(report)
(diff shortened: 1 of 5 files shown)


yndajas:
[Activities upload: hint div id is not unique (Critical)](https://trello.com/c/x5JtIx9V/2851-activities-upload-hint-div-id-is-not-unique-critical)

jkempster34 (commented):
Not tested fully sorry, but just wondering if we can we do something like this?
image

yndajas:
> Not tested fully sorry, but just wondering if we can we do something like this? image

OMG! Needed to update a bunch of tests, but I think it's working, and that's so much cleaner! 🤦🏼‍♀️

jkempster34:
> > Not tested fully sorry, but just wondering if we can we do something like this?
> 
> OMG! Needed to update a bunch of tests, but I think it's working, and that's so much cleaner! 🤦🏼‍♀️

Glad to be of assistance!

yndajas:
Replaced by #2061

PR #2061, opened 2023-03-09
(2851) Make activity upload form IDs unique

yndajas:
## Changes in this PR

This adds the type to the IDs for the `input` and `div.govuk-hint` elements, the `for` attribute of the `label` element, and the `aria-describedby` attribute of the `input` element, in order to address critical accessibility issues highlighted by axe DevTools

Replaces #2059

## Next steps

### DIFF
--- a/app/controllers/activities/uploads_controller.rb
+++ b/app/controllers/activities/uploads_controller.rb
@@ -33,9 +33,9 @@ class Activities::UploadsController < BaseController
     authorize report, :upload?
 
     @report_presenter = ReportPresenter.new(report)
-    upload = CsvFileUpload.new(params[:report], :activity_csv)
-    @success = false
     @type = params[:type].to_sym
+    upload = CsvFileUpload.new(params[:report], :"activity_csv_#{@type}")
+    @success = false
     is_oda = Activity::Import.is_oda_by_type(type: @type)
 
     prepare_default_report_trail(report)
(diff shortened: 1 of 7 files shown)


yndajas:
[Activities upload: hint div id is not unique (Critical)](https://trello.com/c/x5JtIx9V/2851-activities-upload-hint-div-id-is-not-unique-critical)

app/controllers/activities/uploads_controller.rb
```diff
@@ -33,9 +33,9 @@ def update
     authorize report, :upload?
 
     @report_presenter = ReportPresenter.new(report)
-    upload = CsvFileUpload.new(params[:report], :activity_csv)
-    @success = false
     @type = params[:type].to_sym
+    upload = CsvFileUpload.new(params[:report], :"activity_csv_#{@type}")
```
jkempster34, line 37: You can symbolise a string like this? How did I not know that!

jkempster34 (approved):
Makes sense to me!

app/controllers/activities/uploads_controller.rb
yndajas: Yep. I don't know if I prefer it to `to_sym`, but sometimes I feel like using it when interpolating

jkempster34 (commented):
Perhaps not one for this PR, but it's a bit odd that that the CsvFileUpload model receives params. I don't think models should know about controllers

yndajas:
> Perhaps not one for this PR, but it's a bit odd that that the CsvFileUpload model receives params. I don't think models should know about controllers

Yeah, might be worth adding a tech debt card for it, so that we pass the value rather than a symbol for the model to find in the params

PR #2066, opened 2023-03-09
(2833) Only show IATI export with publishable activities

yndajas:
## Changes in this PR

The link to download an IATI export for an organisationfund/level combination of activities is shown if any activities exist for that combination. However, when preparing the XML, activities are only included if publishable to IATI (based on an activity's `publish_to_iati` attribute). This means links can be generated for exports that will contain no activities, which is invalid. This makes it so that a link is only shown if publishable activities exist for the given combination

## Screenshots of UI changes

These screenshots are taken from AMS local data, which may be out of sync with production

### Before

image

### After

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1231,6 +1231,7 @@ activity and on its child transactions (which can be actuals, refunds, and adjus
 - Fix an accessibility issue on the Organisations page
 - Set `publish_to_iati` to `false` for non-ODA activities
 - Add `govuk-link` class to links where missing
+- Only provide an IATI export link if there are publishable activities
 
 [unreleased]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-131...HEAD
 [release-131]: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/compare/release-130...release-131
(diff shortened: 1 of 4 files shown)


yndajas:
[There should be no IATI XML links for ISPF when there are only non-ODA activities](https://trello.com/c/c3BcV4xt/2833-there-should-be-no-iati-xml-links-for-ispf-when-there-are-only-non-oda-activities)

CristinaRO:
A few specs are failing, likely due to the mocking:
```
Failed examples:

rspec ./spec/models/iati/xml_download_spec.rb:103 # Iati::XmlDownload.all_for_organisation when the organisation has activities for some levels and funds only returns XML download paths for levels and funds that have activities
rspec ./spec/models/iati/xml_download_spec.rb:46 # Iati::XmlDownload.all_for_organisation when the organisation has activities for each level and fund returns all XML downloads ordered by levels, then funds
rspec ./spec/models/iati/xml_download_spec.rb:71 # Iati::XmlDownload.all_for_organisation when the organisation has activities for each level and fund and the feature flag hiding ISPF is enabled does not return XML downloads for ISPF
```

yndajas:
> A few specs are failing, likely due to the mocking:

Yup, I tried a few different approaches to fix it, but the mocks were running amok, so I simplified it a little. We were mocking calls within a private method, so I've stripped it back just to mock the return value of the private method, which gets around the chaos of combining mocking of chained calls with stubbed Active Record relations 😬

CristinaRO (approved):

PR #2067, opened 2023-03-10
(2857) Provide text for untitled activity link

yndajas:
## Changes in this PR

In the UI form for adding activities, before the activity has a title, there's a link in the breadcrumbs without any text in it, which makes it unnoticeable to most users and also an accessibility issue

In `Activities::Breadcrumbed` we already have a method for providing fallback text of "Untitled activity" where an activity doesn't have a title, but we weren't using this under particular conditions. This makes it so that we use it in all conditions, which fixes the issue

### Before

image

### After

image

## Next steps


yndajas:
[New activity form: Empty link in breadcrumb on identifier step (minor)](https://trello.com/c/fxByONUY/2857-new-activity-form-empty-link-in-breadcrumb-on-identifier-step-minor)

CristinaRO (approved):

PR #2068, opened 2023-03-10
(2858) Remove duplicate hidden fields on radio collections

yndajas:
## Changes in this PR

On various pages that used `f.govuk_collection_radio_buttons`, we were also using `f.hidden_field`, but this is included by default with `f.govuk_collection_radio_buttons`[1]. This led to a redundant hidden field being created with the same ID as the radio button collection, which is an accessibility issue. This removes the duplicate hidden fields from relevant pages, fixing the issue

[1] https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder

## Next steps


yndajas:
[New activity form: same ID for all radio inputs on sector category step (minor)](https://trello.com/c/YmfNPETh/2858-new-activity-form-same-id-for-all-radio-inputs-on-sector-category-step-minor)

PR #2069, opened 2023-03-10
(2862) Constrain upload forms to 2 thirds layout

CristinaRO:
## Changes in this PR
- Ensure upload forms are constrained to a 2 thirds layout for accessibility

### After
TBA

## Next steps


yndajas (commented):
I think there are a few inconsistencies, as shown in the table below. One path has 2/3 applied twice, one has a redundant `full` applied and another applies it in the partial where all others apply it pre-partial

Activities have separate forms for programmes, but they follow the same pattern as the others

| entity     | context | 2/3: location                 | 2/3: count | notes                                               |
| ---------- | ------- | ----------------------------- | ---------- | --------------------------------------------------- |
| activities | new     | new                           | 1          |                                                     |
| activities | update  | update                        | 1          |                                                     |
| actuals    | new     | new                           | 1          | doesn't use partial - form embedded                 |
| actuals    | update  | **upload form partial**           | 1          |                                                     |
| budgets    | new     | new                           | 1          | doesn't use partial - form embedded                 |
| budgets    | create  | create                        | 1          | **has probably redundant \`full\` in \`_upload_form\`** |
| forecasts  | new     | new                           | 1          | doesn't use partial - form embedded                 |
| forecasts  | update  | update **+ upload form partial** | **2**          |                                                     |

---

I'm not really sure why I did [this](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/1875/commits/bd11e15db64a86bad98664884636384907705967#diff-c56badda200c0bdfd9e117997cdca6aca97969fc2021adb291f6fee725d44153L1-R2), but maybe it can be reverted if it helps standardise things?

CristinaRO:
[Constrain upload forms](https://trello.com/c/nVpo098t/2862-constrain-upload-forms)

app/views/forecasts/uploads/update.html.haml
```diff
@@ -1,20 +1,28 @@
...
-    - if @success
-      .govuk-grid-column-full
-        %h1.govuk-heading-xl
-          = t("importer.success.heading")
-        = render partial: "shared/forecasts/forecasts_by_activity"
-        = link_to t("importer.success.back_link"), report_forecasts_path(@report_presenter), class: "govuk-button"
-
-    - else
+  - unless @success
+    .govuk-grid-row
```
yndajas, line 5: Is there a reason for structuring it this way versus putting `if @success` first?

yndajas (approved):
Looks good. Left one query on one bit, but it's nothing major

Also spotted that within these files, three use `unless @errors.empty?` and two user `if @errors.any?` - might be worth making those consistent, but it's no big deal:
- `views/activities/uploads/update.html.haml`
- `views/actuals/uploads/update.html.haml`
- `views/forecasts/uploads/update.html.haml`
- `views/level_b/activities/uploads/update.html.haml`
- `views/level_b/budgets/uploads/create.html.haml`

app/views/forecasts/uploads/update.html.haml
CristinaRO: I think we wanted to deal with the "unhappy" path first. If it's consistent across the various pages, it's worth the cognitive discomfort of the logic pretzel. If it's not consistent, you can make a case for easier to read logic. :)

yndajas: Seems like it's not 😅

image

CristinaRO: If we start looking at inconsistencies, we also notice that some pages have a different page title for success vs failure, which contributes to the meandering conditionals 😬

PR #2071, opened 2023-03-10
(2778) Use $GITHUB_OUTPUT instead of set-output command

CristinaRO:
## Changes in this PR

The `set-output` command is deprecated, and causes warnings to be thrown when used. Github plan to fully disable it on 31st May 2023 [1].

The suggested replacement approach is to use the `$GITHUB_OUTPUT` environment file, which is available by default.

(h/t @RobjS for the fix)

[1] https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

## Screenshots of UI changes
N/A

## Next steps

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -27,7 +27,7 @@ jobs:
         run: script/docker-push
       - name: Set Docker tag as an output
         id: set_output
-        run: echo "::set-output name=docker_tag::${{ env.DOCKER_TAG }}"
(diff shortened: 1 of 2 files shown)


CristinaRO:
[Rewrite GitHub Actions to remove soon-to-be-deprecated commands](https://trello.com/c/mLkW9DCY/2778-rewrite-github-actions-to-remove-soon-to-be-deprecated-commands)

PR #2076, opened 2023-03-13
(2871) Allow setting new commitment on existing activity via bulk upload

Gweaton:
## Changes in this PR

Before this, any inclusion of the Commitment in the upload template would cause the bulk upload to fail. We believe all users should be able to at least initially set a Commitment on an existing activity if it doesn't exist so they can upload historical data all in one go.

If there's already a commitment, we render an error message to let users know they'll need to edit this via the user interface.

[Trello card](https://trello.com/c/6oiOPlMW/2871-allow-ocf-to-be-set-when-updating-existing-activities-via-bulk-upload-as-well-as-creating)

## Next steps


CristinaRO:
[Allow OCF to be set when updating existing activities via bulk upload, as well as creating](https://trello.com/c/6oiOPlMW/2871-allow-ocf-to-be-set-when-updating-existing-activities-via-bulk-upload-as-well-as-creating)

CristinaRO (approved):

PR #2079, opened 2023-03-15
(2873) Prepopulate OCF when editing & prevent overwriting with `nil` when saving fails

Gweaton:
## Changes in this PR

- Only builds a new Commitment on the Activity when rendering the "edit" page if one has not already been set on the Activity, showing the previous value to be edited.
- Prevents OCF being overwritten as `nil` when failing to save after a validation error by rolling back when `save!` raises an error.

### After (clicking on edit)

image

## Next steps

### DIFF
--- a/spec/controllers/activity_forms_controller_spec.rb
+++ b/spec/controllers/activity_forms_controller_spec.rb
@@ -308,7 +308,8 @@ RSpec.describe ActivityFormsController do
 
     describe "commitment" do
       let(:policy) { double(:policy, show?: true) }
-      let(:activity) { create(:third_party_project_activity, commitment: create(:commitment, value: 1000)) }
+      let(:commitment) { create(:commitment, value: 1000) }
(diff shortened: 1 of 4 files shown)


Gweaton:
https://trello.com/c/1WcVFb1Y/2873-prepopulate-original-ocf-value-in-new-activity-form-when-editing-as-a-beis-user

spec/controllers/activity_forms_controller_spec.rb
```diff
@@ -323,6 +324,27 @@
...
+
+        context "when there is already a commitment" do
+          it "does not build a new commitment when rendering the edit page" do
+            subject { get_step :commitment }
+
+            expect(activity.commitment).to eq(commitment)
+          end
+        end
+
+        context "when there no commitment" do
```
CristinaRO, line 336: "when there *is* no commitment"

CristinaRO (approved):

PR #2080, opened 2023-03-15
(2703) XML budget revisions

jkempster34:
### Determine the IATI type of a Budget
We need to know whether a Budget is "original" or "revised".

This can be awkward, as a Budget can either be an ActiveRecord object, or a
"revision" which is returned by the Audited library.

This adds methods to determine the IATI type of a Budget. It also adds some
additional helper methods, removes an unnecessary presenter method, and tides
up some existing specs with a factory trait.

### Show revised budgets on the XML exports
Now that Budgets have a revision history, we want to show both the "original"
Budget, and the "revised" Budget on the XML exports.

This renders the "original" Budget always, as well the "revised" Budget if
present. Only the most recent revision is shown as per the guidance.

https://iatistandard.org/en/guidance/standard-guidance/activity-budgets/

## Screenshots
A budget with an original value of 50 and a revised value of 150 looks like this:

image

## Next steps

- [] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.


app/models/budget.rb
```diff
@@ -47,13 +47,32 @@ def period_end_date
...
 
+  def original_revision
+    audits.find_by(action: "create").revision
+  end
+
+  def any_update_audits?
+    audits.updates.any?
+  end
+
+  private def original?
```
jkempster34, line 67: I'm not sure whether I'm particularly happy with this..... it does work though!

It will raise and error if you did something like `Budget.new.iati_type`, which I think is fine?

jkempster34: Okay, I still think the modeling here is weird.

If a Budget has revisions, I don't think a Budget should have an `iati_type`, I only think a revision should have an `iati_type`

I've attempted to address this [here](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/f4c88521b8585fd5101bffe07794c2dfc2fd087b) 

The result is ultimately the same, it's whatever makes most sense to everyone.

jkempster34:
[Capture original and revised budget in XML downloads](https://trello.com/c/MxdkLFTB/2703-capture-original-and-revised-budget-in-xml-downloads)

app/models/budget.rb
jkempster34: okay, I've taken inspiration from `paper_trail` which has a `live?` method to check whether a record is live or a revision. I'm fairly happy now

CristinaRO (approved):

PR #2081, opened 2023-03-16
(2874) Prevent CSV file injection

CristinaRO:
## Changes in this PR
- Use `csv-safe` as a drop-in replacement for `csv` to generate safer CSVs

## Screenshots of UI changes
N/A

## Next steps


CristinaRO:
[Prevent CSV File Injection (moderate) by modifying how we export CSVs](https://trello.com/c/SDRqZIoN/2874-prevent-csv-file-injection-moderate-by-modifying-how-we-export-csvs)

jkempster34 (changes requested):
Ah, I just tried this and it still prepends `'` to negative numbers (so they show like `'-1000.00`), you can see an example if you export "External income for Newton Fund" with the current prod data.

Digging into `csv_safe`, [it doesn't prepend `'` to numbers](https://github.com/zvory/csv-safe/blob/70723ee09d14945fa91223065e5e77191373e316/lib/csv-safe.rb#L44), but we [convert our numbers to strings so that they have 2 decimal places](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/3cef9073978c643ae77241acaa7acc4fdaff4a84/app/services/external_income/export.rb#L66)

Gweaton (approved):

config/initializers/csv_safe.rb
```diff
@@ -0,0 +1,12 @@
+require "csv-safe"
+
+class CSVSafe
+  def sanitize_field(field)
+    # we may represent monetary numerics as strings to format them, so we don't want CSVSafe to prefix them
+    if field.nil? || field.is_a?(Numeric) || field.is_a?(String) && Regexp.new(/\A[-\d.,]+\z/).match?(field)
```
yndajas, line 6: Could we label the regex in a variable to clarify what it's matching (more specifically than in the comment above)?

doc/architecture/decisions/0036-use-csv-safe-to-process-csv-files.md
```diff
@@ -0,0 +1,21 @@
...
+
+Date: 2023-03-16
+
+## Status
+
+Accepted
+
+## Context
+
+RODA generates CSV files containing user input. If this user input contains malicious code, it could compromise the machines of the users opening CSV files in applications such as MS Excel.
```
yndajas, line 11: Very minor, but maybe "Microsoft" instead of the "MS" abbreviation?

yndajas (approved):
Couple of minor comments but generally looks good

config/initializers/csv_safe.rb
jkempster34, line 6: Because this conditional is quite long, could we extract the third part into a variable?

```ruby
field_is_currency = field.is_a?(String) && Regexp.new(/\A[-\d.,]+\z/).match?(field)
if field.nil? || field.is_a?(Numeric) || field_is_currency
```

I also think the operator precedence is quite confusing here, so extracting into a variable fixes that too!

doc/architecture/decisions/0036-use-csv-safe-to-process-csv-files.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Decision
+
+We will use the gem `csv-safe` as a replacement for the `csv` gem when generating CSV files.
+
+## Consequences
+
```
jkempster34, line 18: Maybe this file should say that you had to monkey patch the library to handle currencies?

config/initializers/csv_safe.rb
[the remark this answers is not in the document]
CristinaRO: I'm _somewhat_ comfortable now that I've added a refund (with a negative value) to the valid actuals/refunds scenario. I'd still like to have a couple of additions to feature specs where we explicitly mention numeric fields with negative values, and perhaps even a unit test for a "malicious" text field.

CristinaRO: Is it better now that I've extracted the conditional into its own method?

jkempster34 (approved):

config/initializers/csv_safe.rb
yndajas: Looks good!

PR #2085, opened 2023-03-17
 Fix flaky time travel spec

jkempster34:
## Changes in this PR

This spec is flaky in CI.

I'm not sure why, but sometimes when time travelling, the user gets signed
out, and the next assertion fails.

This re-authenticates the user after time travelling. It also freezes the time
in the correct before block, as previously it was accidentally part of a different
context.

## Next steps

### DIFF
--- a/spec/features/users_can_edit_a_budget_spec.rb
+++ b/spec/features/users_can_edit_a_budget_spec.rb
@@ -1,5 +1,9 @@
 RSpec.describe "Users can edit a budget" do
-  before { authenticate!(user: user) }
+  before do
+    authenticate!(user: user)
+    freeze_time
+  end
+
   after { logout }
 
   context "when signed in as a beis user" do
@@ -12,8 +16,6 @@ RSpec.describe "Users can edit a budget" do
       within("##{budget.id}") do
         click_on t("default.link.edit")
       end
-
-      freeze_time
     end
 
     scenario "a budget can be successfully edited and a history event added" do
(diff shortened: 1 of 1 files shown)


spec/features/users_can_edit_a_budget_spec.rb
```diff
@@ -92,29 +94,30 @@
...
+      travel_to(Date.tomorrow)
+      authenticate!(user: user)
 
-        expect(page).to have_content(t("action.budget.update.success"))
-        within("##{budget.id}") do
-          expect(page).to have_content("£20.00")
-          expect(page).to have_link("1 revision", href: activity_budget_revisions_path(budget.parent_activity_id, budget))
-          click_on t("default.link.edit")
-        end
+      click_on t("default.link.edit")
```
yndajas, line 102: Does this need to be in `within("##{budget.id}") do` like before, or is there only one budget?

spec/features/users_can_edit_a_budget_spec.rb
```diff
@@ -127,21 +130,21 @@
       within(original_row) do
         expect(page).to have_content("£10.00")
         expect(page.all("td", exact_text: "").count).to be 2
-        expect(page).to have_content(I18n.l(Date.today))
+        expect(page).to have_content(I18n.l(Date.yesterday))
```
yndajas, line 133: This makes sense to me now, but I don't understand (without putting much thought into it) why it was `Date.today` before, given the time travelling 🤔

yndajas (approved):
One question and one confusion, but looks like this is working in any case

spec/features/users_can_edit_a_budget_spec.rb
jkempster34: Before, the time travelling was in a block, so it was returning to the current time outside of the block 🤯

PR #2088, opened 2023-03-20
(2889) Tidy changelog

yndajas:
## Changes in this PR

- Fix changelog order (convention; dev experience)
- Tidy up (errors)
- Reformat headings and links (style)

You can view the updated changelog in GitHub style here: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/tidy-changelog/CHANGELOG.md

### Before

image

[...]

image

### After

image

## Next steps

- [heh] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Change Log
+# Changelog
 
 ## [unreleased]
 
@@ -10,7 +10,7 @@
   - Show "original" and "revised" budgets on the XML exports
 - Feature flag ISPF ODA bulk upload
 
-## [release-132] 2023-03-16
+## [release-132] - 2023-03-16
 
 - Update Commitment-importing script to infer `transaction_date` value from Activity values
 - Collect original commitment figure in new activity form and bulk activity upload, only editable by BEIS users
@@ -26,7 +26,7 @@
 - Signpost how to add a comment after a failed actual/refund upload containing a comment
 - Add "Back to report" links at all stages of the actuals and forecasts upload journeys
 
-## [release-131] 2023-03-07
+## [release-131] - 2023-03-07
 
 - Remove heading for unpopulated organisation column in organisation reports table for BEIS users
 - Prevent child ODA activities being created on non-ODA parents and vice-versa via the bulk upload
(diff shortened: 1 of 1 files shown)


yndajas:
[Tidy changelog](https://trello.com/c/JWojnnRM/2889-tidy-changelog)

yndajas:
> [...] not sure how I'm feeling about the titles though.
>
> I think the original creator probably got the title format from here https://keepachangelog.com/en/0.3.0/ (not the they are necessarily the arbiter of changelog formatting).

Yeah I think you're right, given that's linked in the Rails template changelog. I think even "[132](example.com) - 2023-03-04" would be cleaner than "[release-132](example.com) - 2023-03-04" (and still in line with the reference format). I probably slightly prefer the Rspec style, but it's a weakly-held view, and it might be best to stick to the format of that source (even if we deviate in some ways, e.g. not grouping our lists)

CHANGELOG.md
```diff
@@ -1,660 +1,753 @@
...
+[Full changelog][unreleased]
+
+- Add budget revisions:
+  - Only budget value can be edited
+  - Add a budget revisions page
+  - Add comments when editing budgets to be shown alongside the revisions
+  - Improve design/content for the budgets area
+  - Show "original" and "revised" budgets on the XML exports
+
+## 132 / 2023-03-16
```
Gweaton, line 14: I'd prefer to see "Release" here, either as "release-132" or "Release 132" if we think that's neater. It _should_ be obvious what the numbers refer to here but that might not always be the case, particularly if this is a repository that's open to the public. I also find it a little easier to scan 👀

Gweaton (approved):
Thanks for tidying this up - have just added one comment re: titles

CHANGELOG.md
yndajas: I did wonder about putting them all under a "Released" `h2` but that felt a little off. Maybe this?

Any thoughts @jkempster34? I wasn't sure if your comment on titles actually included the shifting of the changelog link actually, or just the format (Keep a Changelog includes the link in the heading)

---

## Release 132 - 2023-03-16

[Full changelog](example.com)

jkempster34: Um, honestly I just didn't like the slash because it looked funny to me!

I don't have any preferences beyond that

CristinaRO (approved):

PR #2089, opened 2023-03-20
(2867) Allow users to enter commas when submitting an original commitment figure

Gweaton:
## Changes in this PR

This PR refactors a couple of things before making it possible for users to enter commas in the Original Commitment Figure form field:
1. Validation errors are now handled at the `Commitment` level, rather than its parent `Activity` which allow us to validate the Commitment itself _before_ relying on ActiveRecord's `save!` populating errors from the model level. This has required a slight change in the `ActivityFormsController` to catch these.
1. We no longer use the `build_commitment` method to create a new Commitment in the `Updater` class, but create the Commitment in isolation before linking it to an Activity to line up more closely with the approach taken for creating a `Budget`. It didn't feel necessary to extract this logic into its own `*Creator`-type class just yet as we only call this in one place (in the `Activity::Updater`) at the moment. This has made it easier to catch the errors thrown by `ConvertFinancialValue` when validations fail.

### Before

image

### After

(not a thing)

## Next steps

### DIFF
--- a/app/services/activity/updater.rb
+++ b/app/services/activity/updater.rb
@@ -171,15 +171,12 @@ class Activity
 
       return if !activity.third_party_project? && value.blank?
 
-      ActiveRecord::Base.transaction do
-        activity.build_commitment(
-          value: value, transaction_date: infer_transaction_date_from_activity_attributes(activity)
-        )
+      commitment = activity.commitment || Commitment.new
+      commitment.activity = activity
+      commitment.value = value
+      commitment.transaction_date = infer_transaction_date_from_activity_attributes(activity)
 
-        activity.commitment.save!
-      end
-    rescue => error
-      activity.errors.add(:value, error)
+      commitment.save
     end
 
     def assign_attributes_for_step(step)
--- a/app/services/activity/updater.rb
+++ b/app/services/activity/updater.rb
@@ -171,12 +171,24 @@ class Activity
 
       return if !activity.third_party_project? && value.blank?
 
-      commitment = activity.commitment || Commitment.new
-      commitment.activity = activity
-      commitment.value = value
-      commitment.transaction_date = infer_transaction_date_from_activity_attributes(activity)
+      commitment = activity.commitment || Commitment.new(
+        activity: activity,
+        transaction_date: infer_transaction_date_from_activity_attributes(activity)
+      )
+      convert_and_assign_commitment_value(commitment: commitment, value: value) if value.present?
+
+      return unless commitment.errors.empty?
 
       commitment.save
+    end
+
+    def convert_and_assign_commitment_value(commitment:, value:)
+      commitment.value = ConvertFinancialValue.new.convert(value)
+    rescue ConvertFinancialValue::Error
+      # Setting the value here allows us to display the erroneous value in the form input.
+      # Without it, the previously-valid number value for the existing commitment is displayed.
+      commitment.value = value
+      commitment.errors.add(:value, I18n.t("activerecord.errors.models.activity.attributes.commitment.value.not_a_number"))
     end
 
     def assign_attributes_for_step(step)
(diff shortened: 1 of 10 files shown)


CristinaRO (approved):
I think you've already addressed my pending comment, but GH isn't showing it to me so I can't cancel it 🥲

spec/services/activity/updater_spec.rb
```diff
@@ -42,19 +42,16 @@
...
-            activity.build_commitment
           end
 
-          it "swallows the error and adds it to the activity's errors" do
-            expect(activity.errors.full_messages.first).to eq(
-              "Value Validation failed: Value Value is not a number"
+          it "adds the error to the commitment's errors" do
+            expect(activity.commitment.errors.full_messages.first).to include(
+              "Value is not a number"
             )
```
yndajas: Is this only part of the error message? (Or more specifically, why the shift to `include` vs `eq`?

app/services/activity/updater.rb
```diff
@@ -173,12 +173,23 @@ def set_commitment
...
       commitment.transaction_date = infer_transaction_date_from_activity_attributes(activity)
+      convert_and_assign_commitment_value(commitment: commitment, value: value) if value.present?
+
+      return unless commitment.errors.empty?
 
       commitment.save
     end
 
+    def convert_and_assign_commitment_value(commitment:, value:)
+      commitment.value = ConvertFinancialValue.new.convert(value.to_s)
```
yndajas, line 185: Is the `value` not always already a string (from the params)?

yndajas (approved):
Looks good. Just some minor things in the comments

spec/features/beis_users_can_upload_level_b_activities_spec.rb
```diff
@@ -100,17 +100,10 @@
     end
 
     within "//tbody/tr[2]" do
-      expect(page).to have_xpath("td[1]", text: "Free Standing Technical Cooperation")
-      expect(page).to have_xpath("td[2]", text: "3")
-      expect(page).to have_xpath("td[3]", text: "")
-      expect(page).to have_xpath("td[4]", text: t("activerecord.errors.models.activity.attributes.fstc_applies.inclusion"))
-    end
```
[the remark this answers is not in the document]
CristinaRO: Yes, this is due to the two layers of error catching: if the converter catches something, the process never reaches the model validations.

Also, if a field is nil, it doesn't get converted, so the absence of a required attribute (in this case FSTC) is only caught by the model validation. 🥲

spec/services/activity/updater_spec.rb
Gweaton: The model error generated for us by active record actually comes out as "`Value Value must be a valid number`" but I believe it's formatted in the view so it doesn't look like this. I thought writing that out felt a bit gross so changed this to `include` to assert on the thing the user sees. Do you think it'd be better to be explicit and assert on the exact error?

app/services/activity/updater.rb
Gweaton: Yeah I think it is, I took this line of code from the other places we use `ConvertFinancialValue` but I think those cover the CSV upload as well, which might be why they do that conversion. I'll remove this as it's not necessary.

app/services/activity/updater.rb
```diff
@@ -171,15 +171,12 @@ def set_commitment
 
       return if !activity.third_party_project? && value.blank?
 
-      ActiveRecord::Base.transaction do
-        activity.build_commitment(
-          value: value, transaction_date: infer_transaction_date_from_activity_attributes(activity)
-        )
+      commitment = activity.commitment || Commitment.new
+      commitment.activity = activity
```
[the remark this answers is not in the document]
Gweaton: Good point. Have fixed this up

spec/services/activity/updater_spec.rb
yndajas: I think it's better to make an assertion that the human readable stuff is there!

PR #2092, opened 2023-03-21
(2855) Prevent updating of funds via the bulk upload

Gweaton:
## Changes in this PR

The bulk upload now presents the user with an error when they try to upload a CSV with a fund activity's RODA identifier in the "RODA ID" column.

This PR also fixes a small bug where the wrong Activity column was being used for an activity that wasn't found.

#### Incorrect column

image

#### Corrected error column

image

#### New error

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Feature flag ISPF ODA bulk upload
 - Allow users to enter commas in Original Commitment Figure values
 - Prevent users from editing fund activities via the bulk upload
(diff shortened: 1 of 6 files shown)


Gweaton:
https://trello.com/c/4wTGr80E/2855-prevent-beis-users-from-accidentally-updating-a-level-a-parent-by-bulk-upload

CristinaRO (approved):
Thank you for catching that error display bug too!

PR #2094, opened 2023-03-22
(2859) Prevent overflow of tables when inside tabs

jkempster34:
## Changes in this PR
The results of a recent accessibility audit raised that the content of some tables overflows the viewport width without the ability to scroll right or adapt the content to fit inside the available space.

This takes the approach found here
https://github.com/DFE-Digital/register-trainee-teachers/pull/2217, which in turn references the table on the GOV.UK Cookies page https://www.gov.uk/help/cookie-details. 

Instead of applying the styles to the table itself, we apply them to a container as suggested by the accessibility audit.

>If you want to test locally, you will need to run `yarn build:css`

|Before|After|
|-|-|
|image|image|

## Next steps


CristinaRO:
The audit also mentioned the Users table, if you have a chance to replicate the fix over there.

jkempster34:
> The audit also mentioned the Users table, if you have a chance to replicate the fix over there.

I can't replicate the problem on the users table. I have wrapped it in the same container though

CristinaRO:
[Content overflow in certain conditions (moderate)](https://trello.com/c/ufWGzVKZ/2859-content-overflow-in-certain-conditions-moderate)

CristinaRO (approved):

PR #2096, opened 2023-03-23
(2516) Untitled activity deletion

yndajas:
## Changes in this PR

- Allow BEIS users to delete untitled activities
- Signpost PO users on how to get untitled activities deleted

### Before

image

### After

https://user-images.githubusercontent.com/40244233/227240995-dbe4a35a-b81b-46f6-ac55-764c4ea78b03.mov

## Next steps

### DIFF
--- a/app/policies/activity_policy.rb
+++ b/app/policies/activity_policy.rb
@@ -93,6 +93,8 @@ class ActivityPolicy < ApplicationPolicy
   end
 
   def destroy?
+    return false if record.fund?
+    return true if record.title.blank? && beis_user?
     false
   end
 
--- a/app/views/shared/activities/_activity_summary.html.haml
+++ b/app/views/shared/activities/_activity_summary.html.haml
@@ -59,7 +59,7 @@
               = t("action.activity.delete.details.answer.authorised")
 
             = link_to t("action.activity.delete.details.answer.authorised_action"),
-                "example.com",
+                organisation_activity_confirm_destroy_path(activity_presenter.organisation, activity_presenter),
                 class: "govuk-button govuk-button--warning",
                 "data-module": "govuk-button",
                 role: "button"
--- a/app/views/shared/activities/_activity_summary.html.haml
+++ b/app/views/shared/activities/_activity_summary.html.haml
@@ -45,3 +45,25 @@
       %dd.govuk-summary-list__actions
         - if policy(activity_presenter).update_linked_activity?
           = a11y_action_link(t("default.link.#{activity_presenter.call_to_action(:linked_activity)}"), activity_step_path(activity_presenter, :linked_activity), t("activerecord.attributes.activity.linked_activity"))
+
+- if activity_presenter.can_be_deleted?
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %details.govuk-details{"data-module" => "govuk-details"}
+        %summary.govuk-details__summary
+          %span.govuk-details__summary-text
+            = t("action.activity.delete.details.question")
+        .govuk-details__text
+          - if policy(activity_presenter).destroy?
+            %p
+              = t("action.activity.delete.details.answer.authorised")
+
+            = link_to t("action.activity.delete.details.answer.authorised_action"),
+                "example.com",
+                class: "govuk-button govuk-button--warning",
+                "data-module": "govuk-button",
+                role: "button"
+
+          - else
+            %p
+              = t("action.activity.delete.details.answer.unauthorised_html")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Add "Previously reported under Newton Fund" ISPF tag
 - Rename Commitment date in Activity "Financials" tab
 - Change the colour of success messages to black
+- Allow BEIS users to delete untitled activities and signpost PO users on the process
 
 ## Release 132 - 2023-03-16
 
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -10,6 +10,9 @@ class ActivitiesController < BaseController
 
   def index
     @organisation = Organisation.find(organisation_id)
+
+    @deleted_activity_roda_identifier = params[:deleted_activity_roda_identifier]
+
     if @organisation.service_owner?
       @partner_organisations = Organisation.partner_organisations
 
@@ -71,6 +74,19 @@ class ActivitiesController < BaseController
     authorize @activity, :destroy?
 
     prepare_default_activity_trail(@activity)
+  end
+
+  def destroy
+    @activity = Activity.find(id)
+
+    authorize @activity, :destroy?
+
+    @activity.destroy
+
+    redirect_to organisation_activities_path(
+      @activity.organisation,
+      deleted_activity_roda_identifier: @activity.roda_identifier
+    )
   end
 
   def historic
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -65,6 +65,14 @@ class ActivitiesController < BaseController
     end
   end
 
+  def confirm_destroy
+    @activity = Activity.find(id)
+
+    authorize @activity, :destroy?
+
+    prepare_default_activity_trail(@activity)
+  end
+
   def historic
     @organisation = Organisation.find(organisation_id)
     if @organisation.service_owner?
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -685,6 +685,10 @@ class Activity < ApplicationRecord
     Time.parse(spending_breakdown_filename[-18..-5])
   end
 
+  def can_be_deleted?
+    title.blank?
+  end
+
   def self.hierarchically_grouped_projects
     activities = all.to_a
     projects = activities.select(&:project?).sort_by { |a| a.roda_identifier.to_s }
--- a/app/views/activities/_deleted_activity_notification_banner.html.haml
+++ b/app/views/activities/_deleted_activity_notification_banner.html.haml
@@ -1,0 +1,9 @@
+.govuk-notification-banner.govuk-notification-banner--success{role: "alert", "aria-labelledby": "govuk-notification-banner-title", "data-module": "govuk-notification-banner"}
+  .govuk-notification-banner__header
+    %h2.govuk-notification-banner__title#govuk-notification-banner-title
+      = t("default.success")
+  .govuk-notification-banner__content
+    %h3.govuk-notification-banner__heading
+      = t("action.activity.delete.success.title")
+    %p.govuk-body
+      = t("action.activity.delete.success.body_html", roda_identifier: roda_identifier)
--- a/app/views/activities/confirm_destroy.html.haml
+++ b/app/views/activities/confirm_destroy.html.haml
@@ -1,0 +1,27 @@
+=content_for :page_title_prefix, t("document_title.activity.delete", roda_identifier: @activity.roda_identifier)
+
+%main.govuk-main-wrapper#main-content{ role: "main" }
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %span.govuk-caption-xl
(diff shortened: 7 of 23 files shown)


yndajas:
[Make "empty" activities easier to deal with for BEIS and PO users](https://trello.com/c/xHtG3hO0/2516-make-empty-activities-easier-to-deal-with-for-beis-and-po-users)

spec/controllers/activity_deletions_controller_spec.rb
```diff
@@ -0,0 +1,64 @@
+require "rails_helper"
+
+RSpec.describe ActivityDeletionsController do
```
jkempster34, line 3: The commit message around this confused me a bit, I think it should say something like this instead.

```
The policy dependency of the controller is mocked so that we are only testing 
the controller itself.
```

Rather than:

```
For the controller spec, I wanted to follow the single responsibility
principle, using contexts of "when authorised" and "when unauthorised",
rather than setting up the various combinations of activity level, user
organisation and titled-ness, which should be the concern of the policy.
It took a while to work out how to make the controller not care about
the user, so I considered reverting to a complex context structure, but
in the end I worked out a way
```

For two reasons:
1. I think the details about the journey of how it was implemented are irrelevant unless they may be important in future (sadly!)
2. I read it as the `single responsibility` of the spec itself, which doesn't make sense. Whereas, I think its saying "For the controller spec, I wanted to test only the responsibility of the class (controller)"


Hope that made sense!

spec/controllers/activity_deletions_controller_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  describe "#new" do
+    context "when authorised" do
+      it "responds with a 200" do
+        allow(policy).to receive(:destroy?).and_return(true)
+
+        get :new, params: {organisation_id: activity.organisation, activity_id: activity.id}
+
+        expect(response.status).to eq(200)
```
jkempster34, line 19: Are there some helpers that could be used here? 

`expect(response.status).to eq(:success)`

app/controllers/activity_deletions_controller.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class ActivityDeletionsController < BaseController
```
jkempster34, line 3: I think I, in part, disagree with the group here in it's current implementation for a few reasons:
1. I'm not sure the actions make sense. An action is a user interaction, but they aren't creating a deletion.
2. It sends off the wrong request type `POST` instead of `DELETE`. I'm not sure if this necessarily breaks any rules, but it's definitely _less_ correct.
3. It goes against Rails conventions of having a `destroy` method for a resource.

I think point 2 and 3 could be fixed by making this a destroy, but then does ActivityDeletionsController#destroy make sense? Maybe? I'm not sure!

jkempster34 (commented):
This is all really tidy!

I've left a few initial thoughts, hope they make sense!

spec/controllers/activity_deletions_controller_spec.rb
jkempster34: Found them, they are `assert_response :success` ``assert_response :not_found` etc, but I don't think they are used anywhere in this project

app/policies/activity_policy.rb
```diff
@@ -89,6 +89,8 @@ def set_commitment?
   end
 
   def destroy?
+    return false if record.fund?
```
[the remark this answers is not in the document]
yndajas: It could be. I did think about doing it that way, but we often break down the policy methods quite a bit, e.g. this similar pattern on `set_commitment?` right above:

```ruby
    return false if record.fund?
    return true if beis_user?

    record.commitment.nil? && editable_report? && record.organisation == user.organisation
```

I don't have strong feelings either way!

spec/controllers/activity_deletions_controller_spec.rb
yndajas: Another thing I did wonder about 😅 Are those helpers RSpec? I think you can do something like `have_http_status(:success)` or possibly just `eq(:success)`. The other controller specs have used status codes (which is why I went down this route), but we could update them

spec/controllers/activity_deletions_controller_spec.rb
jkempster34: I think _technically_ single responsibility only relates the the responsibility of a class, which probably led to my confusion. Though I've definitely heard it being used to describe stuff like APIs 

I'm not sure if this is picky? If it's leading to confusion, then maybe not?

app/policies/activity_policy.rb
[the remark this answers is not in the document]
jkempster34: Fair! I think this might be preference

app/controllers/activity_deletions_controller.rb
yndajas: Yeah this is an interesting one. I was initially thinking about just adding `confirm_destroy` and `destroy` to the `ActivityController`, but `confirm_destroy` doesn't feel very RESTful, which is why Ed suggested a deletions controller

I think you _could_ conceptualise the creation of an action, and in this case that action being deletion. It does seem a little strange given that `DELETE` is a method and destroy is a RESTful route, but we do have other controller where we're creating or updating _actions_ or doing other stuff that doesn't map neatly onto method/action + resource: `uploads#update` actually creating and updating activities; `redactions#update` actually updating an activity attribute; `forms#update` updating an activity

I agree it would be nice to use `DELETE` though, which could go back into the `ActivityController` as a `destroy` action, but I'm not sure what to do with the confirm deletion route unless we route `GET` and `DELETE` to separate actions or deal within it one action (but the latter feels like a single responsibility principle violation) 🤔

I suppose we have plenty of custom actions in other controllers, so `confirm_destroy` isn't too unconventional for RODA?

spec/controllers/activity_deletions_controller_spec.rb
yndajas: 😮 interesting. I thought it was just a general principle in all code (including test code) to be followed to varying degrees. In any case, I think I can clarify the wording here!

jkempster34: Don't take my word for it!

[This](https://stackify.com/solid-design-principles/) article starts by defining it as a class thing, then goes on to mention microservices. Maybe it's original meaning was in relation to classes, not it's a more wooly term 🤷

yndajas: Nice! What book is this?

jkempster34: https://www.google.co.uk/books/edition/Spring_5_0_Microservices/CedDDwAAQBAJ?hl=en&gbpv=1&dq=single+responsibility+principle+%22microservices%22&pg=PA128&printsec=frontcover 

I just went to books as a trustable resource. 

If we see Uncle Bob as the ultimate truth on the matter, then I think that second quote he is elaborating on the first.
Found [this tweet](https://twitter.com/unclebobmartin/status/1544655247305490432?s=20) that suggests that too

yndajas: I've reworded this, incorporating your suggestion while making the detail more of an explanation of the implementation instead of recounting the journey 😅

app/controllers/activity_deletions_controller.rb
jkempster34: Considering there are a few people who think this is the right direction then I'm sure this will be okay!

I do think `confirm_destroy` is what [most people land on](https://github.com/search?q=confirm_destroy+lang%3ARuby++path%3A%2F%5Eapp%5C%2Fcontrollers%5C%2F%2F&type=code)

jkempster34 (commented):
Tiny thing that Cristina raised on my PR the other day.

Does the delete page need either the correct breadcrumbs, or none at all? or are we happy with just `Home`?

app/controllers/activity_deletions_controller.rb
yndajas: Did something go wrong with that link? 🤔

image

yndajas:
> Does the delete page need either the correct breadcrumbs, or none at all? or are we happy with just `Home`?

I think it should be the full activity breadcrumbs. I thought about this and forgot, so thanks for the reminder!

app/controllers/activity_deletions_controller.rb
jkempster34: Works for me https://github.com/search?q=confirm_destroy+lang%3ARuby++path%3A%2F%5Eapp%5C%2Fcontrollers%5C%2F%2F&type=code 
image

yndajas: Bizarre - I get nothing 🤔

yndajas: Ah, this works! Looks like it's pretty common then. Whether it's right or right-er is another matter 😅 I'm still open to persuasion

app/views/shared/activities/_activity_summary.html.haml
```diff
@@ -45,3 +45,25 @@
...
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      %details.govuk-details{"data-module" => "govuk-details"}
+        %summary.govuk-details__summary
+          %span.govuk-details__summary-text
+            = t("action.activity.delete.details.summary")
+        .govuk-details__text
+          - if policy(activity_presenter).destroy?
+            %p
+              = t("action.activity.delete.details.details.authorised.explanation")
```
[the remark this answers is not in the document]
yndajas: Yeah, I spotted that when tinkering yesterday. Going by the class names `govuk-details__summary` and `govuk-details__text`, it would be `details.text`, but that feels too generic. What about this?

```yaml
details:
  question: This activity has no title. Can we delete it?
  answer:
    authorised: You can delete this activity as it has no title.
    authorised_action: Delete this activity
    unauthorised_html: Only BEIS can delete this activity...
```

app/controllers/activity_deletions_controller.rb
yndajas: It was good enough for RailsCasts in 2007: http://railscasts.com/episodes/77-destroy-without-javascript

CristinaRO: I suppose the fact that it's hard to decide is proof enough that there's no "one right way".

Some questions I'd ask myself are things like: is one or the other choice more likely to be hit by future changes in Rails (for example if we were relying on an HTTP method being used in a way that might not be supported in future), are they equally maintainable and understandable for future teams etc.

app/views/shared/activities/_activity_summary.html.haml
```diff
@@ -45,3 +45,25 @@
       %dd.govuk-summary-list__actions
         - if policy(activity_presenter).update_linked_activity?
           = a11y_action_link(t("default.link.#{activity_presenter.call_to_action(:linked_activity)}"), activity_step_path(activity_presenter, :linked_activity), t("activerecord.attributes.activity.linked_activity"))
+
+- if activity_presenter.title.blank?
```
CristinaRO, line 49: What do you think about making this a more generic method, e.g. `activity_presenter.can_be_deleted?` and containing the deletion conditions in one place, to be more resilient to the possibility of deletion conditions changing in future?

app/controllers/activity_deletions_controller.rb
yndajas: Maybe better to stick with `destroy` and `DELETE` for destroying in that case!

app/views/shared/activities/_activity_summary.html.haml
yndajas: I like the idea of a method like that. By `containing the deletion conditions in one place` do you mean moving them out of the policy?

CristinaRO: Ah, no, in the policy it would be fine - I only meant that the condition shouldn't live in the view template.

CristinaRO (approved):
Great work, thanks for persevering through all the comments and requests!

jkempster34 (approved):
💯

PR #2099, opened 2023-03-23
Tidy up HTTP status assertions

yndajas:
## Changes in this PR

Previously, there was a lack of consistent syntax/helper method usage when testing the HTTP status of a controller response or page. This cleans things up by using the `have_http_status` matcher in combination with a status symbol[1]. It also makes the `it` descriptions more consistent when referencing status codes

[1] https://github.com/rails/rails/blob/v4.0.0/guides/source/layouts_and_rendering.md#the-status-option

## Next steps


jkempster34 (approved):
These are nice and neat, thank you!

PR #2100, opened 2023-03-24
(2744) Add task to generate a CSV of Activities for annual fund impact metrics

jkempster34:
## Changes in this PR
We want to be able to generate a CSV of activities that will be used to
pre-populate an annual collaborative spreadsheet of fund impact metrics.

This task will be used to capture a snapshot of activities once a year in
July.

The CSV will be created in the `tmp` folder. We have documentation describing
how to download files when connected remotely to an instance of the application.

### DIFF
--- a/lib/tasks/annual_fund_impact_metrics_activities.rake
+++ b/lib/tasks/annual_fund_impact_metrics_activities.rake
@@ -1,0 +1,41 @@
+require "csv-safe"
+
+namespace :activities do
+  task :annual_fund_impact_metrics, [:output_csv_path] => :environment do |_task, args|
+    desc "Generates a CSV file of Activities that will be used to pre-populate an annual
+          collaborative spreadsheet of fund impact metrics"
+
+    csv_path = args[:output_csv_path] || "tmp/annual_fund_impact_metrics.csv"
+
+    CSVSafe.open(csv_path, "wb") do |csv|
+      activities = Activity
+        .joins(:organisation)
+        .includes(:organisation, :actuals)
+        .where.not(programme_status: ["delivery", "agreement_in_place", "open_for_applications", "stopped"])
+        .order("organisations.name, programme_status")
+
+      # We want to exclude Activities that were completed more than two years ago. We have discussed
+      # this and determined that the best way to do this is to exclude those that have no Actuals
+      # reported in the last two years (including if there are no Actuals at all).
+      active_activities = activities.reject do |activity|
+        next unless activity.programme_status == "completed"
+
+        activity.actuals.none? { |actual| actual.date >= 2.years.ago.to_date }
+      end
+
+      headers = ["Partner Organisation name", "Activity name", "RODA ID", "Partner Organisation ID", "Status"]
+
+      csv << headers
+
+      active_activities.each do |activity|
+        partner_organisation_name = activity.organisation.name
+        activity_title = activity.title
+        roda_identifier = activity.roda_identifier
+        partner_organisation_identifier = activity.partner_organisation_identifier
+        status = activity.programme_status
+
+        csv << [partner_organisation_name, activity_title, roda_identifier, partner_organisation_identifier, status]
+      end
+    end
+  end
+end
--- a/spec/lib/tasks/annual_fund_impact_metrics_activities_spec.rb
+++ b/spec/lib/tasks/annual_fund_impact_metrics_activities_spec.rb
@@ -1,0 +1,95 @@
+require "tempfile"
+
+RSpec.describe "rake activities:annual_fund_impact_metrics", type: :task do
+  let!(:completed_activity) { create(:fund_activity, programme_status: "completed") }
+  let!(:actual_2_years_ago) { create(:actual, date: 2.years.ago, parent_activity: completed_activity, report: nil) }
+  let!(:beis_organisation) { completed_activity.organisation }
+  let!(:completed_activity_no_actuals) { create(:fund_activity, programme_status: "completed") }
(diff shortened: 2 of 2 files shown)


spec/lib/tasks/annual_fund_impact_metrics_activities_spec.rb
```diff
@@ -0,0 +1,80 @@
+require "tempfile"
+
+RSpec.describe "rake activities:annual_fund_impact_metrics", type: :task do
+  let!(:completed_activity) { create(:fund_activity, programme_status: "completed") }
+  let!(:actual_2_years_ago) { create(:actual, date: 2.years.ago, parent_activity: completed_activity, report: nil) }
+  let!(:completed_activity_no_actuals) { create(:fund_activity, programme_status: "completed") }
+  let!(:beis_organisation) { completed_activity.organisation }
+  let!(:decided_activity) { create(:fund_activity, organisation: aardvark_organisation, programme_status: "decided") }
+  let!(:review_activity) { create(:fund_activity, organisation: aardvark_organisation, programme_status: "review") }
+  let!(:aardvark_organisation) { create(:beis_organisation, name: "Department for Aadvarks", iati_reference: "CZH-COH-111") }
```
CristinaRO, line 10: Shoutout to testers' favourite animal :D

jkempster34:
[DMEL can download of activities per PO so that they can initiate annual fund metrics reporting](https://trello.com/c/efA0GnEX/2744-dmel-can-download-of-activities-per-po-so-that-they-can-initiate-annual-fund-metrics-reporting)

lib/tasks/annual_fund_impact_metrics_activities.rake
```diff
@@ -0,0 +1,43 @@
...
+
+      # We have decided that the most accurate way to determine whether an Activity
+      # was completed or stopped more than 2 years ago, is to exclude those that have not had
+      # any Actuals reported in the last two years.
+      active_activities = activities.reject do |activity|
+        completed_or_stopped = ["completed", "stopped"].any?(activity.programme_status)
+        has_actuals = activity.actuals.any?
+        actuals_reported_in_last_two_years = activity.actuals.any? { |actual| actual.date >= 2.years.ago.to_date }
+
+        completed_or_stopped && has_actuals && !actuals_reported_in_last_two_years
```
CristinaRO, line 25: To make sure I understand the logic: we're ok with not rejecting complete or stopped activities with no actuals? (Because we can't determine with certainty how long ago the activity was completed or stopped.)

jkempster34: That was my own assumption. BEIS said that we should reject any Activities with no Actuals in the last two years. 

But wouldn't a brand-new Activity have no Actuals at all, and we don't want that rejected?

jkempster34: There seems to be over 2000 Activities like that `Activity.count { |a| a.actuals.count == 0 && (a.programme_status == "completed" || a.programme_status == "stopped") }`

lib/tasks/annual_fund_impact_metrics_activities.rake
```diff
@@ -0,0 +1,43 @@
+require "csv-safe"
+
+namespace :activities do
+  task :annual_fund_impact_metrics, [:output_csv_path] => :environment do |_task, args|
+    desc "Generates a CSV file of Activities that will be used to pre-populate an annual
+          collaborative spreadsheet of fund impact metrics"
+
+    csv_path = args[:output_csv_path] || "tmp/activities.csv"
```
CristinaRO, line 8: Is it worth giving the filename a more descriptive default? Such as `activities_for_impact_metrics.csv` for example.

lib/tasks/annual_fund_impact_metrics_activities.rake
jkempster34: Thanks for spotting this, the assumption I made was definitely wrong!

lib/tasks/annual_fund_impact_metrics_activities.rake
jkempster34: Yep, I really struggled with naming the task as well. What do you think of that `annual_fund_impact_metrics`?

lib/tasks/annual_fund_impact_metrics_activities.rake
```diff
@@ -0,0 +1,42 @@
...
+        .joins(:organisation)
+        .includes(:organisation, :actuals)
+        .where.not(programme_status: ["delivery", "agreement_in_place", "open_for_applications", "stopped"])
+        .order("organisations.name, programme_status")
+
+      # We want to exclude Activities that were completed more than two years ago. We have discussed
+      # this and determined that the best way to do this is to exclude those that have no Actuals
+      # reported in the last two years (including if there are no Actuals at all).
+      active_activities = activities.reject do |activity|
+        completed = activity.programme_status == "completed"
```
CristinaRO, line 21: Could we exit early from this block if the activity is not completed?

Maybe if we extract the block into a method that returns true or false, and it can return false quickly if the activity isn't completed, without needing to fetch and examine the actuals.

Not sure if this is premature optimisation, but worth checking.

CristinaRO (approved):
Thank you for bearing with all the iterations!

lib/tasks/annual_fund_impact_metrics_activities.rake
yndajas: TIL you can pass a block to `.count`

PR #2103, opened 2023-03-27
[FIX] Use `Time.current` instead of `Time.now` in test setup

Gweaton:
## Changes in this PR

`spec/presenters/report_presenter_spec.rb` had a failing test after the switch to British Summer Time.

`Time.now` uses the system time (which is BST) while `Time.current` uses the time in the current zone (defaulting to UTC in Rails).

An alternative to this would have been to set the `config.time_zone` in `application.rb`. Although we don't have users outside the UK who might be confused by this, it's still best practice to avoid setting an application-wide time zone.

For more information, [this blog
post](https://thoughtbot.com/blog/its-about-time-zones) was helpful at explaining how to work best with timezones in Rails.

We may want to revisit timezones in future, but I didn't want to update them everywhere now, as it's just a single test that's failing.

## Next steps


Gweaton:
https://trello.com/c/AjfW6zr7/2906-fix-timezone-related-failing-test

CristinaRO (approved):

jkempster34 (changes requested):
Doesn't this mean the time is still displayed incorrectly?

Gweaton:
@jkempster34 It may be, but this PR is about getting the local environment fixed, rather than fixing that issue. We should look into that as a separate task

PR #2107, opened 2023-03-27
[2907] Set local time zone of app to London

Gweaton:
## Changes in this PR

Set TimeZone to London

This change was motivated by one of our tests using `Time.now` (BST) clashing with ActiveRecord storing times as UTC and failing.

Although [this article](https://thoughtbot.com/blog/its-about-time-zones) advises against setting a global timezone and recommends allowing users to set their own, this is a larger change than we have capacity (and need?) for at the moment.

We believe that all users should be accessing the site from a UK timezone, so this shouldn't change what they see.

Noticeable places this will change will be visible are:

    - The approval time on a Report
    - The time an export was uploaded to S3

Before this change, these would be one hour behind BST as they call `Time.current` which uses the app's time zone which defaulted to UTC, rather than the running machine's system timezone.

This also updates places we call `Time.now` to use `Time.current` to use the app's timezone instead, as recommended in the article linked above.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Change the colour of success messages to black
 - Allow BEIS users to delete untitled activities and signpost PO users on the process
 - Add more information to the form guidance on the original commitment figure step
+- Set app's local time zone to "London"
 
 ## Release 132 - 2023-03-16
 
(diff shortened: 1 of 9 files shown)


CristinaRO:
[Check whether change to BST is supported correctly in the service](https://trello.com/c/Atp1uqZA/2907-check-whether-change-to-bst-is-supported-correctly-in-the-service)

jkempster34 (approved):
✅

PR #2111, opened 2023-03-28
(2912) Fix regression: funds and programmes should not have Download as XML links

CristinaRO:
## Changes in this PR
- Ensure BEIS users can't download individual fund and programme activities as XML

### Before
Screenshot 2023-03-28 at 17 02 41

### After
Screenshot 2023-03-28 at 17 02 35

## Next steps


CristinaRO:
[Download as XML inconsistencies: some more tests needed](https://trello.com/c/jAVhaHpu/2180-download-as-xml-inconsistencies-some-more-tests-needed)

app/policies/activity_policy.rb
```diff
@@ -61,8 +61,10 @@ def update?
     false
   end
 
-  def download?
-    beis_user?
+  def show_xml?
+    return false if record.fund? || record.programme?
```
yndajas, line 65: Can this use the `is_project?` method?

```ruby
return false unless record.is_project?
```

app/policies/activity_policy.rb
```diff
@@ -61,8 +61,10 @@ def update?
...
+    return true if beis_user?
+    false
```
yndajas, line 67: I think the last two lines can just be `beis_user?`, no?

spec/features/users_can_view_fund_level_activities_spec.rb
```diff
@@ -17,6 +18,12 @@
       page_displays_an_activity(activity_presenter: ActivityPresenter.new(fund_activity))
     end
 
+    it "does not show a link to download as XML" do
+      visit organisation_activity_path(fund_activity.organisation, fund_activity)
+
+      expect(page).to_not have_content t("default.button.download_as_xml")
+    end
```
yndajas, line 25: Follow-up to activity summary comment - this would only check if it appears on the financials tab, right? Whereas if it was in the summary, it wouldn't matter which tab it is

spec/features/users_can_view_programme_level_activities_spec.rb
```diff
@@ -1,20 +1,26 @@
 RSpec.feature "Users can view programme level activities" do
   context "when the user belongs to BEIS" do
     let(:user) { create(:beis_user) }
+    let(:fund_activity) { create(:fund_activity, organisation: user.organisation) }
```
yndajas, line 4: This was already in there, but do we need to create this fund here? It doesn't seem to be used beyond setting the programme's parent, so I'd have thought we could just leave it as the default random parent. Same in the other context

spec/features/users_can_view_programme_level_activities_spec.rb
```diff
@@ -39,6 +45,7 @@
       end
 
       page_displays_an_activity(activity_presenter: ActivityPresenter.new(programme_activity))
+      expect(page).to_not have_content t("default.button.download_as_xml")
```
yndajas, line 48: Is there a reason for not separating this into its own `it` block like for BEIS users?

spec/features/users_can_view_project_level_activities_spec.rb
```diff
@@ -6,14 +6,12 @@
 
     scenario "can view a project" do
       fund = create(:fund_activity)
-      programme = create(:programme_activity)
-      fund.child_activities << programme
-      project = create(:project_activity, organisation: user.organisation)
-      programme.child_activities << project
+      programme = create(:programme_activity, parent: fund)
+      project = create(:project_activity, organisation: user.organisation, parent: programme)
```
yndajas, line 10: Again not sure if we really need all the explicit ancestor mocking in the test

yndajas (approved):
Again mostly superficial comments - generally looks good!

app/views/activities/_download_as_xml.haml
```diff
@@ -0,0 +1,4 @@
+- if policy(activity).download?
+  .govuk-grid-row
+    .govuk-grid-column-full
+      = link_to t("default.button.download_as_xml"), organisation_activity_path(activity.organisation, activity, format: :xml), class: "govuk-button"
```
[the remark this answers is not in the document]
CristinaRO: We've had this as a "tech debt" for ages, but it never came up as a need from BEIS users 🤷🏻 It's rarely used anyway, since the most XML downloads happen from the Exports page.

spec/features/users_can_view_programme_level_activities_spec.rb
CristinaRO: Test suite speed 😬 Building up the context twice didn't seem worth it. Of course, then comes the question of why was it worth for the other cases...

spec/features/users_can_view_fund_level_activities_spec.rb
CristinaRO: It's true, and if we had more time to ask for design opinions I'd like to revise this!

app/views/shared/xml/_activity.xml.haml
```diff
@@ -5,7 +5,7 @@
   %title
     %narrative= activity.title
   %description{"type" => "1"}
-    %narrative= activity.covid19_related.zero? ? activity.description : "#{activity.description} COVID-19"
```
jkempster34, line 8: Are all of the other `nil` attributes for non-ODA activities handled?

CristinaRO: Anecdotally yes, since this is the only one that caused errors when viewing a NODA activity as XML.

jkempster34 (approved):
Makes sense to me 🚢

yndajas (approved):

PR #2112, opened 2023-03-28
(2903) Make sure users can sign out if JavaScript is disabled

jkempster34:
### Make sure users can sign out if JavaScript is disabled
Rails uses JavaScript to intercept requests from links that are given a
`method`, which allows us to use links for methods including DELETE. When
JavaScript is disabled, this falls back to a GET request.

Currently, this means that users see an error page when attempting to sign out
with JavaScript disabled, as Devise is expecting a DELETE request.

Devise allows us to specify the method used to sign out, so we can update this
to GET.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -273,7 +273,7 @@ Devise.setup do |config|
   # config.navigational_formats = ['*/*', :html]
 
   # The default HTTP method used to sign out a resource. Default is :delete.
-  config.sign_out_via = :delete
+  config.sign_out_via = :get
 
   # ==> OmniAuth
   # Add a new OmniAuth provider. Check the wiki for more information on setting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Allow BEIS users to delete untitled activities and signpost PO users on the process
 - Add more information to the form guidance on the original commitment figure step
 - Set app's local time zone to "London"
+- Fix bug where users could not sign out if JavaScript was disabled
 
 ## Release 132 - 2023-03-16
 
--- a/app/views/shared/_navigation.html.haml
+++ b/app/views/shared/_navigation.html.haml
@@ -32,4 +32,4 @@
           = link_to t("page_title.users.index"), users_path, class: "govuk-header__link"
 
       %li{ class: navigation_item_class(destroy_user_session_path) }
-        = link_to t("header.link.sign_out"), destroy_user_session_path, method: :delete, class: "govuk-header__link"
+        = link_to t("header.link.sign_out"), destroy_user_session_path, class: "govuk-header__link"
--- a/spec/features/users_can_sign_out_spec.rb
+++ b/spec/features/users_can_sign_out_spec.rb
@@ -3,7 +3,7 @@ require "rails_helper"
 RSpec.feature "Users can sign out" do
   after { logout }
 
-  scenario "success", js: true do
+  scenario "success" do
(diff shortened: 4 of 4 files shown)


config/initializers/devise.rb
```diff
@@ -273,7 +273,7 @@
   # config.navigational_formats = ['*/*', :html]
 
   # The default HTTP method used to sign out a resource. Default is :delete.
-  config.sign_out_via = :delete
+  config.sign_out_via = :get
```
yndajas, line 276: We decided against changing the config to fix the sign out in #1954, although mainly out of an assumption (unresearched) about conventions. I'm not opposed to doing it now, but we could get around this by making it a form like in `app/views/activities/confirm_destroy.html.haml:20` and `app/views/shared/implementing_organisations/_table.html.haml:23`

yndajas (commented):
I've left a comment about the approach on the config file, but if we're going down this route, we can probably remove the package `@rails/ujs` and get rid of `js: true` from the spec. See the commit message on #1954 for details

Also, I creeped the scope a little on the card, and it might make sense to shift this into a separate card/PR, but we should address the other places where we rely on JavaScript to intercept delete links

> There are other instances of link_to with method: :delete that are preventing users from deleting without JavaScript (e.g. deleting budgets)

config/initializers/devise.rb
jkempster34: Urgh, didn’t realise this had a history. I’ll revisit tomorrow.

thanks!

yndajas: Yeah, I should've dug around and referenced it on the card! I don't have much of an opinion whether one approach is better than the other though - as I say, it was just based on an assumption about conventions more than anything

yndajas: Oh there are more instances than I realised! Looks like they're mostly `delete` but there's one `post`

Still, I'm not seeing anything we can't replicate with forms (and keep the same result in the UI, so no design needed). Obviously we can set the method in forms, but for the `confirm` functionality we can just add `html: { onsubmit: "return confirm('Are you sure?');" }`

If we do keep the hacked links, we should probably check that the tests are failable - without `js: true`, the sign out test was producing a false positive, hence why we didn't catch the bug for a while!

yndajas: Yeah, I was more thinking about how we can get rid of UJS. `onsubmit` would do the same thing, so it's not a regression in that sense, but if we make it a form users would still be able to take those actions without JS, just minus the confirmation dialog (the `onsubmit` is just ignored in that case). Arguably a risk to allow those actions without confirmation though?

jkempster34: Do we want to get rid off UJS? 

Are you proposing that we:
- Remove UJS
- Replace all `link_to method:` with forms
- Replace all `link_to: confirm` with `onsubmit`

🤔

jkempster34: Thought about this over lunch. Going to be stubborn and say:
- The guidance says we should have a separate page for confirmations (I think), which would be a larger piece of work for later
- Removing UJS and replacing with other JS doesn't seem worth it
- Seems like a bad time to do this in case we break something
- Replacing links/buttons with forms is going to give us styling problems
- This is low priority because internal users should probably just have JS enabled
- We might as well fix the sign out link while we are here because Devise give us the option to

But happy to follow the general agreement reach by the group!

yndajas (approved):
Let's merge for now and save the progressive enhancement and HTTP methods discussion for later 😊

PR #2113, opened 2023-03-29
(2909) Allow valid actual/refund without deleting invalid rows

yndajas:
### What and why

The actuals/refunds upload template is prepopulated with rows for all the activities to which you can add values. Previously, if you didn't want to provide values for all activities (e.g. if you just wanted to add values for one activity), you would have to remove all the remaining rows, since the default values are invalid

This makes it so that if there's at least one valid row and the only errors are default value errors, the valid row is imported and the invalid rows are ignored/discarded. This saves users the effort of manually deleting rows they don't care about

Rows with other errors will still prevent the import from completing, and the actual: zero, refund: blank error will still show if there are other errors and/or no valid rows

### How

The `has_only_default_value_errors?` method relies on the error column being the string "Actual Value/Refund Value" and the message being a specific translation. This doesn't feel as resilient as I'd like, but hopefully the tests are sufficient enough to catch a breaking change

I would have preferred to have access to the type hash created in `Converter#validate_values`, however getting this all the way up to the base `Import` class would likely mean a more significant refactor, modifying both the `Error` `Struct` and the iteration over `importer.errors` in `import_row`. If we added a `value_type` attribute to `Error`, this wouldn't be relevant to errors from non-value columns, which would make things a little messy

### Before

image

(and so on)

### After

image

## Next steps

### DIFF
--- a/app/services/actual/import.rb
+++ b/app/services/actual/import.rb
@@ -55,6 +55,8 @@ class Actual
     end
 
     def has_only_default_value_errors?
+      return false if invalid_with_comment
+
       @errors.all? do |error|
         return false unless error.column == "Actual Value/Refund Value"
         error.message == I18n.t("importer.errors.actual.cannot_be_zero_when_refund_blank")
--- a/app/services/actual/import.rb
+++ b/app/services/actual/import.rb
@@ -22,13 +22,18 @@ class Actual
 
     def import(actuals)
       ActiveRecord::Base.transaction do
-        @imported_actuals = actuals.map.with_index { |row, index| import_row(row, index) }
+        @imported_actuals = actuals.map.with_index { |row, index| import_row(row, index) }.compact
+
+        @errors = [] if @imported_actuals.present? && has_only_default_value_errors?
+
         unless @errors.empty?
           @imported_actuals = []
           raise ActiveRecord::Rollback
         end
       end
     end
+
+    private
 
     def import_row(row, index)
       importer = RowImporter.new(@report, @uploader, row)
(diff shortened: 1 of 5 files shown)


yndajas:
[Allow users to upload file with at least one valid actual/refund without having to delete the zero actuals pre-populated by the template](https://trello.com/c/fsXs60xH/2909-allow-users-to-upload-file-with-at-least-one-valid-actual-refund-without-having-to-delete-the-zero-actuals-pre-populated-by-the)

yndajas:
Hmm, but what if they add a comment in one of the default value rows? I think it'll be ignored in this case (not imported but with no error), because we've been relying on value errors to prevent comments with invalid values and flag an error when there are default values. So the user might not notice it's been ignored...

app/services/actual/import.rb
```diff
@@ -22,14 +22,19 @@ def initialize(report:, uploader:)
 
     def import(actuals)
       ActiveRecord::Base.transaction do
-        @imported_actuals = actuals.map.with_index { |row, index| import_row(row, index) }
+        @imported_actuals = actuals.map.with_index { |row, index| import_row(row, index) }.compact
+
+        @errors = [] if @imported_actuals.present? && has_only_default_value_errors?
```
jkempster34, line 27: I wondered if this reassignment of `@errors` could be avoided, but I'm having problems figuring out if it's possible without too much refactoring. 

I think it might be nicer if we could only assign an error in this scenario if comment is present. However, it seems like it would be a pain to assign errors if comment is not present and other errors are present

jkempster34 (approved):
I think this looks good, you might have some insight into my comment on reassigning `@errors`, but maybe not worth it at this stage

app/services/actual/import.rb
yndajas: Yeah, it's that latter thing that makes it messy, if I understand you correctly! We still want the error if there are no comments and no valid rows

I think it might be nice to do a further refactor/design to draw out the comments scenarios in the error messaging itself though

PR #2116, opened 2023-03-29
Release 133

yndajas:
- Add budget revisions:
  - Only budget value can be edited
  - Add a budget revisions page
  - Add comments when editing budgets to be shown alongside the revisions
  - Improve design/content for the budgets area
  - Show "original" and "revised" budgets on the XML exports
- Feature flag ISPF ODA bulk upload
- Allow users to enter commas in Original Commitment Figure values
- Prevent users from editing fund activities via the bulk upload
- Use correct column name in bulk upload results when trying to update an activity that cannot be found
- Fix accessibility issue of overflowing contents on the reports table and users table when zoomed in
- Ensure non-BEIS users cannot download activities as XML
- Use a safer method to display accessible links that can contain user input
- Add inclusion validation for Activity attributes
- Add "Previously reported under Newton Fund" ISPF tag
- Rename Commitment date in Activity "Financials" tab
- Change the colour of success messages to black
- Allow BEIS users to delete untitled activities and signpost PO users on the process
- Add more information to the form guidance on the original commitment figure step
- Set app's local time zone to "London"
- Fix bug where users could not sign out if JavaScript was disabled
- Allow users to upload a single actual/refund without deleting the (invalid) default values in other rows of the template. Rows with (invalid) default values will be ignored in this case. Any other invalid values will still throw an error
- Ensure BEIS users can't download individual fund and programme activities as XML

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 [Full changelog][unreleased]
 
+## Release 133 - 2023-03-30
+
+[Full changelog][133]
+
 - Add budget revisions:
   - Only budget value can be edited
   - Add a budget revisions page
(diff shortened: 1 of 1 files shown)


yndajas:
[Release 133](https://trello.com/c/zsitgvOo/2908-release-133)

PR #2118, opened 2023-03-30
2916 editing budget errors

jkempster34:
### Validate that value has changed when a Budget is edited with a comment
Currently, when a user edits a Budget and does not change the value, no
revision is created. We think this is the desired behaviour.

However, we want to show a validation error if the user does not change the
value, but does attempt to add a comment.

### Make sure current budget shows correctly when value is invalid
Currently, when a user enters an invalid value for a Budget, the displayed
"current value" won't be correct. It will represent the invalid value, rather
than the previous valid value.

Assign to an instance variable, so that when the page is re-rendered the value
is shown correctly.

image

## Next steps


app/models/budget.rb
```diff
@@ -76,6 +77,10 @@ def any_update_audits?
     audit_version.nil?
   end
 
+  private def value_changed
```
Gweaton, line 80: maybe we should call this `validate_value_changed` to distinguish it from the ActiveRecord-generated `value_changed?` method?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Exclude planned activities from annual fund impact metrics CSV
 - Show a validation error when a Budget is edited with a comment but without changing the value
+- Fix bug where current value was incorrect when editing a Bidget with an invalid value
```
Gweaton, line 9: Typo in `Bidget` here

app/controllers/budgets_controller.rb
```diff
@@ -32,6 +32,8 @@ def create
   def edit
     @budget = Budget.find(id)
     @budget_presenter = BudgetPresenter.new(@budget)
+    @current_value = @budget_presenter.value
```
Gweaton, line 35: I'm not sure about this - do we not want to show the user the invalid value they tried to enter so they can correct it here?

app/models/budget.rb
Gweaton: `ensure` rather than `validate`?

app/controllers/budgets_controller.rb
Gweaton: Ah sorry, I've just tested this and see what you mean! Yep, this is definitely an improvement over:

image

Gweaton (approved):
Approving - good to merge once that typo and rename are in!

PR #2123, opened 2023-03-31
Metrics CSV - level updates 

jkempster34:
### Exclude funds from annual fund impact metrics CSV
We have confirmation that funds should not be included in the annual fund
impact metrics CSV.

### Add a "level" column to annual fund impact metrics CSV
We want to add a "level" column to the annual fund impact metrics CSV, to that
the activities can be easily sorted.

## Next steps

### DIFF
--- a/lib/tasks/annual_fund_impact_metrics_activities.rake
+++ b/lib/tasks/annual_fund_impact_metrics_activities.rake
@@ -12,6 +12,7 @@ namespace :activities do
         .joins(:organisation)
         .includes(:organisation, :actuals)
         .where.not(programme_status: ["delivery", "agreement_in_place", "open_for_applications", "stopped", "planned"])
+        .where.not(level: "fund")
         .order("organisations.name, programme_status")
 
       # We want to exclude Activities that were completed more than two years ago. We have discussed
--- a/lib/tasks/annual_fund_impact_metrics_activities.rake
+++ b/lib/tasks/annual_fund_impact_metrics_activities.rake
@@ -24,7 +24,7 @@ namespace :activities do
         activity.actuals.none? { |actual| actual.date >= 2.years.ago.to_date }
       end
 
-      headers = ["Partner Organisation name", "Activity name", "RODA ID", "Partner Organisation ID", "Status"]
+      headers = ["Partner Organisation name", "Activity name", "RODA ID", "Partner Organisation ID", "Status", "Level"]
 
       csv << headers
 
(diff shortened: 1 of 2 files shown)


lib/tasks/annual_fund_impact_metrics_activities.rake
```diff
@@ -12,6 +12,7 @@ namespace :activities do
         .joins(:organisation)
         .includes(:organisation, :actuals)
         .where.not(programme_status: ["delivery", "agreement_in_place", "open_for_applications", "stopped", "planned"])
+        .where.not(level: "fund")
```
yndajas, line 15: Is it worth mentioning in the commit that you haven't combined these two `where.not` as that would result in funds outside of those programme statuses being included and those programme statuses being included where they aren't funds (presumably)? Or is that obvious 😅

spec/lib/tasks/annual_fund_impact_metrics_activities_spec.rb
```diff
@@ -1,21 +1,21 @@
 require "tempfile"
 
 RSpec.describe "rake activities:annual_fund_impact_metrics", type: :task do
```
yndajas, line 3: Don't want to make work on a non-project day, but is it worth adding a fund exclusion test, like you do for the `programme_status` exclusions?

lib/tasks/annual_fund_impact_metrics_activities.rake
```diff
@@ -24,7 +24,7 @@ namespace :activities do
         activity.actuals.none? { |actual| actual.date >= 2.years.ago.to_date }
```
yndajas, line 24: Typo in the commit message: "to that" -> "so that"

yndajas (approved):
I think it's fine if you want to merge this and get it over and done (given we're technically finished on the project!), but one minor suggestion re: testing and a couple of commit message things in case you feel like spending any more time on it!

spec/lib/tasks/annual_fund_impact_metrics_activities_spec.rb
jkempster34: This is already captured in the existing specs, the factories create a load of funds, so I thought that would be sufficient

lib/tasks/annual_fund_impact_metrics_activities.rake
jkempster34: This is also captured in the specs by happy coincidence! I think that should also be sufficient.

PR #2165, opened 2023-06-15
Adds script to change partner org identifiers

shuldt:
There are two activities with incorrect partner org identifiers that need amending, as reported by the client in [this ticket.](https://dxw.zendesk.com/agent/tickets/17993) 

This script finds each activity by its existing (incorrect) partner org identifier, and replaces the partner org identifier on the activity with the correct one.

## Next steps
- [ ] Run the added script in production


Gweaton (approved):
Looks good to me

PR #2173, opened 2023-07-19
Adds script to upload multiple level D budgets. 

shuldt:
A large number of budgets for level D activities need to be added - it is not possible to do this currently via the bulk upload interface. 

## Changes in this PR
This PR adds a re-usable script to upload budgets that cannot be handled by the client via bulk upload.

## Next steps

When this PR has been merged, we will need to make a deployment to production. Once the release has been made, the csv with the required budgets needs to be uploaded to GPaas; we can then run the script in production.

### DIFF
--- a/db/data/20230718131711_import_ukri_level_d_budgets.rb
+++ b/db/data/20230718131711_import_ukri_level_d_budgets.rb
@@ -1,0 +1,86 @@
+# Run me with `rails runner db/data/20230718131711_import_ukri_level_d_budgets.rb`
+
+# Script to add budgets to activities when those budgets cannot be added via bulk upload (for example they are for a
+# level not currently handled by bulk upload).
+
+# Re-usable if you replace `file_name` with the path to your chosen csv.
+# Note that the script expects column headers of the following form: "Activity RODA ID"; "Budget amount"; "Type";
+# "Financial year" The order of the columns is not sensitive.
+
+# This script bypasses the usual budget validation required by the Budget model. This is to allow us to upload these
+# budgets without the need for a report attached to each.
+
+# Details of budgets that cannot be successfully created are logged to the terminal at the end of the script run.
+
+require "csv"
+
+def get_financial_year(financial_year_range)
+  financial_year_range.split("-")[0]
+end
+
+def validate_budget(budget)
+  budget.valid?
+  report_error = budget.errors.messages[:report]
+  budget.errors.messages.count == 1 && report_error.count == 1
+end
+
+file_name = "ukri_level_d_budget_upload.csv"
+
+puts "Loading csv data from #{file_name}"
+
+file = File.open(file_name)
+budget_data = CSV.parse(file.read, headers: true)
+
+initial_budgets_count = Budget.count
+
+puts "There are #{initial_budgets_count} existing budgets."
+puts "There are #{budget_data.count} budgets to create."
+
+puts "Creating budgets..."
+
+logged_unsuccessful = {}
+
+budget_data.each do |row|
+  roda_identifier = row["Activity RODA ID"]
+  parent_activity = Activity.find_by(roda_identifier: roda_identifier)
+  value = row["Budget amount"]
+  budget_type = row["Type"]
+  financial_year_range = row["Financial year"]
+
+  budget_attributes = {
+    parent_activity_id: parent_activity&.id,
+    currency: parent_activity&.default_currency,
+    value: value,
+    budget_type: Budget.budget_types.key(Integer(budget_type, exception: false)),
+    financial_year: get_financial_year(financial_year_range)
+  }
+
+  begin
+    budget = Budget.new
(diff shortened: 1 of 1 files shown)


yndajas (commented):
I've tested this locally after syncing my local environment with production data and unfortunately it didn't work. I've done some digging and fixed some stuff in a separate PR, but there's an issue with the data that needs resolving before we can get this working fully. See the linked PR for details

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2174

yndajas:
Before I forget, one more minor thing I'm unsure about is whether this should be a data migration or a Rake task, which Cristina is probably better placed to answer

shuldt:
> Before I forget, one more minor thing I'm unsure about is whether this should be a data migration or a Rake task, which Cristina is probably better placed to answer

I'll ask Cristina about this

yndajas (commented):
Looking good! Mostly just a few small suggestions, but there's some validation and error rescuing stuff that I think might need tightening

Also, is it worth adding some count logging to double check everything persisted (even if there are no errors), e.g.

```ruby
# before
budget_start_count = Budget.count
puts "Number of budgets in database before import: #{budget_start_count}"
puts "Number of budgets in CSV: #{budget_data.count}"

# after
puts "Number of budgets in database before import: #{budget_start_count}"
puts "Number of budgets in database now: #{Budget.count}"
puts "Number of budgets imported: #{Budget.count - budget_start_count}"
puts "Number of budgets in CSV: #{budget_data.count}"
```

I'm still not sure whether this should be a data migration or a Rake task, but I'm happy to approve in the absence of other input once the validation stuff is sorted

db/data/20230718131711_import_ukri_level_d_budgets.rb
```diff
@@ -0,0 +1,81 @@
...
+
+def get_financial_year(financial_year_range)
+  financial_year_range.split("-")[0]
+end
+
+def validate_budget(budget)
+  budget.valid?
+  value_error = budget.errors.messages[:value]
+  report_error = budget.errors.messages[:report]
+  value_error.count.zero? && report_error.count == 1
```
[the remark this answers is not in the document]
shuldt: I feel very strongly that we had a reason, but I can't remember what it was and I can't make a good argument for it now, so I think you're right

db/data/20230718131711_import_ukri_level_d_budgets.rb
```diff
@@ -0,0 +1,81 @@
+# Run me with `rails runner db/data/20230718131711_import_ukri_level_d_budgets.rb`
+
+# Script to add budgets to activities when those budgets cannot be added via bulk upload (for example they are for a
+# level not currently handled by bulk upload).
+
+# Re-usable if you replace `file_name` with the path to your chosen csv.
+# Note that the script expects column headers of the following form: "Activity RODA ID"; "Budget amount"; "Type";
+# "Financial year" The order of the columns is not sensitive.
+
+# This script bypasses the usual budget validation required by CreateBudget. This is to allow us to upload these
```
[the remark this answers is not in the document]
shuldt: Good catch!

db/data/20230718131711_import_ukri_level_d_budgets.rb
```diff
@@ -0,0 +1,85 @@
...
+require "csv"
+
+def get_financial_year(financial_year_range)
+  financial_year_range.split("-")[0]
+end
+
+def validate_budget(budget)
+  budget.valid?
+  report_error = budget.errors.messages[:report]
+  budget.errors.messages.count == 1 && report_error.count == 1
```
yndajas, line 24: This works, but you can also simplify `budget.errors.messages.count` to `budget.errors.count`

db/data/20230718131711_import_ukri_level_d_budgets.rb
```diff
@@ -0,0 +1,85 @@
...
+    budget.assign_attributes(budget_attributes)
+    unless validate_budget(budget)
+      logged_unsuccessful[roda_identifier] = budget
+      next
+    end
+
+    budget.save(validate: false)
+  rescue
+    logged_unsuccessful[roda_identifier] = budget
+  end
```
yndajas, line 69: If there's a failure on `.assign_attributes` line for any reason it won't have reached `budget.valid?` in `validate_budget`, which is what creates/populates the `.errors` property, so you might need `budget.valid?` here as well to be able to access `budget.errors` in the `logged_unsuccessful` loop later in this scenario

yndajas (approved):
Looks good! You might need one more call to `budget.valid?` just to cover all bases, but this should work with the data we have in any case

Let me know if you need a pair on production

db/data/20230718131711_import_ukri_level_d_budgets.rb
shuldt: Yup you're right

PR #2177, opened 2023-10-03
Change error handling in CmdRunner to handle stderr errors

shuldt:
## Changes in this PR

We currently use a script to periodically [sync production data to the training environment](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/support_tasks/training_db_sync.md) for DSIT testing. 

This script does not break when errors are thrown while running commands with Ruby module Open3.

This PR changes the error handling in the script to check for errors present in stderr. This allows us to capture any such errors, and raise an exception if any are present. If errors are present in stderr while capturing data from source; copying data; or loading data to the destination, the script will delete sessions and tempfiles before exiting. 

### DIFF
--- a/script/training_db_sync.rb
+++ b/script/training_db_sync.rb
@@ -63,14 +63,26 @@ class TrainingDbSync
               --no-owner
     CMD
     Kernel.puts "Running pg_dump on #{source} ==> #{cmd}"
-    CmdRunner.run(cmd)
+    begin
+      CmdRunner.run(cmd)
+    rescue => e
+      clear_sessions
+      remove_temp_files
+      raise e
+    end
   end
 
   def copy_data_to_destination
     authenticate_to(destination)
     cmd = %(cat #{data_filename} | cf ssh beis-roda-#{destination} -c "cat > #{data_filename}")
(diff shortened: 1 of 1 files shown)


script/training_db_sync.rb
```diff
@@ -138,13 +156,11 @@ class CmdRunner
...
     def self.run(command)
-      begin
-        stdout, _stderr, _status = Open3.capture3(command)
-      rescue => error
-        raise "'#{command}' failed (#{error})"
-      end
+      _stdout, stderr, _status = Open3.capture3(command)
 
-      Kernel.puts stdout.chomp
+      if stderr
```
Gweaton, line 161: I'm not sure how `capture3` works - does it rescue errors and store them in `stderr`, or do we need to keep that `rescue` in here too (rather than removing it) to be safe?

Gweaton (approved):
Nice, this looks a fair bit more robust! Just one question

script/training_db_sync.rb
shuldt: So, as far as I can tell from docs, the debug process I went through, and some manual testing, the error is always stored in stderr - I don't think this was ever throwing the the rescue block in the first place.

PR #2189, opened 2023-10-16
(2938) Change how the app accesses s3 storage

CristinaRO:
## Changes in this PR

- Support s3 access via ECS credentials, with the bucket name set as an environment variable

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/app/services/export/s3_uploader.rb
+++ b/app/services/export/s3_uploader.rb
@@ -4,14 +4,22 @@ module Export
   class S3Uploader
     def initialize(file:, filename:, use_public_bucket:)
       @use_public_bucket = use_public_bucket
+      @bucket_name_from_env_var = ENV.fetch("EXPORT_DOWNLOAD_S3_BUCKET", false)
       @config = S3UploaderConfig.new(use_public_bucket: use_public_bucket)
-      @client = Aws::S3::Client.new(
-        region: config.region,
-        credentials: Aws::Credentials.new(
-          config.key_id,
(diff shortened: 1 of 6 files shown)


CristinaRO:
[Change how the app accesses the S3 bucket](https://trello.com/c/TT3JtsvH/2938-change-how-the-app-accesses-the-s3-bucket)

app/services/export/s3_uploader.rb
```diff
@@ -4,14 +4,22 @@ class S3UploadError < StandardError; end
...
+      @bucket_name_from_env_var = ENV.fetch("EXPORT_DOWNLOAD_S3_BUCKET", false)
       @config = S3UploaderConfig.new(use_public_bucket: use_public_bucket)
-      @client = Aws::S3::Client.new(
-        region: config.region,
-        credentials: Aws::Credentials.new(
-          config.key_id,
-          config.secret_key
+      @client = if @bucket_name_from_env_var
+        Aws::S3::Client.new(
+          region: "eu-west-2",
```
rjw1, line 11: I wonder if we want to put a comment here saying we have hardcoded the region since we dotn expect it to change. Or make it the default and use config.region if it exists. Setting config.region with an env var elsewhere.
This shouldnt block deployment though.

rjw1 (approved):
Looks good to me. I have a small comment about future proofing but that shouldnt stop this from being merged.

PR #2197, opened 2023-11-01
Fix the mismatch between headers and rows

CristinaRO:
## Changes in this PR
- Fix the logic that generates actuals and variance rows for reports, to avoid inserting a spurious column for reports that have forecasts but no actuals

In certain circumstances, such as when a report has forecasts but no actual spend, there is an extra column for the data that is not matched to a header.

The reason the extra column shows up is due to the data structures used for the actuals and variance, which will always return a hash with keys and values, even if the values are not meaningful (empty arrays or zero values).

We cannot simply filter out empty arrays or zero values, because they can be meaningful in other contexts, such as when there are some nil or zero values in some quarters, but they are surrounded by quarters with non-zero values, so we have to include them.

The simplest solution is to trust the logic of the headers generation, and not attempt to compute actuals or variance for reports where the app has already determined there are no actuals or variance to display.

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Amend the check for a pre-existing later report to take ODA type into account
 - Error message shown when attempting to create a report while an unapproved one for the same fund, org, and ODA type exists includes the ODA type where relevant
 - Show ODA/non-ODA alongside the fund short name on the reports table for ISPF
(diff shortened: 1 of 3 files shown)


shuldt (approved):
Yeah this makes sense to me, I think this is a good solution

CristinaRO:
[Support request: an extra column in the report CSV](https://trello.com/c/e1Nrax94/2962-support-request-an-extra-column-in-the-report-csv)

PR #2198, opened 2023-11-01
Amend editable report finders acknowledge ODA-ness

shuldt:
We need to be able to distinguish between two different open reports on ISPF - ODA or non-ODA. Currently the code finds an editable report with a finder system that matches reports to activities by organisation, and by fund. 

## Changes in this PR
This PR changes the finders for editable reports to now use the 'ODA status' (is_oda; is_non_oda) of the report and activity as well - when an editable report is fetched, it must have the same ODA status as the activity. 

Where necessary, tests have been updated to ensure that reports and activities can be matched by ODA status. 

## Next steps

### DIFF
--- a/spec/services/activity/projects_for_report_finder_spec.rb
+++ b/spec/services/activity/projects_for_report_finder_spec.rb
@@ -1,14 +1,41 @@
 require "rails_helper"
 
 RSpec.describe Activity::ProjectsForReportFinder do
-  it "only returns projects and third party projects that are for the report's organisation and fund" do
+  it "only returns projects and third party projects that are for the report's organisation, fund and oda-ness" do
     organisation = create(:partner_organisation)
     newton_fund = create(:fund_activity, :newton)
-    report = create(:report, organisation: organisation, fund: newton_fund)
+    report = create(:report, organisation: organisation, fund: newton_fund, is_non_oda: true)
 
     programme = create(:programme_activity, :newton_funded, parent: newton_fund)
-    project = create(:project_activity, :newton_funded, organisation: organisation, parent: programme)
-    third_party_project = create(:third_party_project_activity, :newton_funded, organisation: organisation, parent: project)
+    project = create(:project_activity, :newton_funded, organisation: organisation, parent: programme, is_oda: false)
+    third_party_project = create(:third_party_project_activity, :newton_funded, organisation: organisation, parent: project, is_oda: false)
+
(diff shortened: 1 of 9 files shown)


spec/services/activity/projects_for_report_finder_spec.rb
```diff
@@ -1,14 +1,41 @@
...
+  it "only returns projects and third party projects that are for the report's organisation, fund and oda-ness" do
     organisation = create(:partner_organisation)
     newton_fund = create(:fund_activity, :newton)
-    report = create(:report, organisation: organisation, fund: newton_fund)
+    report = create(:report, organisation: organisation, fund: newton_fund, is_non_oda: true)
 
     programme = create(:programme_activity, :newton_funded, parent: newton_fund)
-    project = create(:project_activity, :newton_funded, organisation: organisation, parent: programme)
-    third_party_project = create(:third_party_project_activity, :newton_funded, organisation: organisation, parent: project)
+    project = create(:project_activity, :newton_funded, organisation: organisation, parent: programme, is_oda: false)
```
CristinaRO, line 10: The thing I've realised while going back and forth on the decision to use an `is_oda` versus `is_non_oda` field on reports, and on the decision to have a default or not, is that we are pretty much constrained forever by the fact that all non-ISPF funds (GCRF, Newton, and OODA at the time writing) are _implicitly_ ODA, _but_ a lot of code relies on the fact that their `is_oda` value is `nil`.[1]

There are no non-ISPF activities with `is_oda` false, because those funds only handle ODA, so this is not a realistic scenario to test, but the bigger issue is that the finder won't be able to match activities where `is_oda` is `nil` to corresponding reports by negating the value of `report.is_non_oda`, because the database query works on exact matching, not falsey matching. 🤦🏻 

I hadn't explored far enough before I proposed `is_non_oda` as the fix-all solution. 😅 

[1] One of the good reasons for this uncomfortable choice of having a boolean allowed to be nil is that we wanted to validate that the user has explicitly set a value of `is_oda` for ISPF activities, so we didn't want to provide a default value. In practice it turned out that the only level of activity where it's required and allowed for a user to set ODA/non-ODA is at programme level. All projects and 3rd party projects take on the ODA type of their parent programme, and the attribute gets set behind the scenes.

shuldt:
PR closed after we re-thought our approach to the is_oda boolean

PR #2210, opened 2023-11-09
Show environment banner and email prefix for the newly added dev domain

CristinaRO:
## Changes in this PR
- Show environment banner and email prefix on the `dev` domain

### After
The banner will look similar to the development banner, but saying "dev" instead of "development".
Screenshot 2023-11-09 at 18 50 24

## Next steps

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -76,23 +76,15 @@ RSpec.describe ApplicationHelper, type: :helper do
       end
 
       context "when the domain is 'https://training'" do
-        it "returns 'training" do
+        it "returns 'training'" do
           ClimateControl.modify DOMAIN: "https://training.report-official-development-assistance.service.gov.uk" do
             expect(helper.environment_name).to eql("training")
           end
         end
       end
 
-      context "when the domain is 'https://pentest'" do
-        it "returns 'training" do
-          ClimateControl.modify DOMAIN: "https://pentest.report-official-development-assistance.service.gov.uk" do
-            expect(helper.environment_name).to eql("training")
-          end
-        end
-      end
-
       context "when the domain is 'https://sandbox'" do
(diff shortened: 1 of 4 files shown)


CristinaRO:
[Show the environment banner on the dev environment](https://trello.com/c/mCYKXneW/2970-show-the-environment-banner-on-the-dev-environment)

spec/helpers/application_helper_spec.rb
```diff
@@ -76,37 +76,45 @@
...
+        it "returns 'training'" do
           ClimateControl.modify DOMAIN: "https://training.report-official-development-assistance.service.gov.uk" do
             expect(helper.environment_name).to eql("training")
           end
         end
       end
 
       context "when the domain is 'https://pentest'" do
-        it "returns 'training" do
+        it "returns 'training'" do
```
shuldt, line 87: Are there supposed to be two 'training's here? I thought pentest was always training

CristinaRO: I think this was defensive development - when we introduced the banners, there was a pentest domain that was used for training, and we wanted to display training, but I was anticipating that the domain might one day change to actually say training, and we'd still want it to be recognised.

If we're sure that there no longer exists a pentest domain, we could remove that branch and simplify a bunch of code.

shuldt (approved):
Happy to approve this - please also see my reply to your comment

spec/helpers/application_helper_spec.rb
CristinaRO: I'm not in a huge rush, and I'm keen to not accumulate tech debt that we already see coming, so I'll check that the pentest domain is not used anymore!

CristinaRO: @shuldt I've confirmed that the pentest domain is not in use anymore, and updated the code accordingly.

PR #2212, opened 2023-11-13
Apply ODA eligibility criteria within report

CristinaRO:
## Changes in this PR
- Remove the `oda_eligibility` criterion from `reportable` scope

Applying the ODA eligibility criteria globally, by including it in the definition of `reportable` activities, is not correct in the global context: ISPF non-ODA activities are reportable, as long as they have the relevant `programme_status`.

Instead, we will apply the ODA eligibility criteria on a report basis, as the report defines what it means for activities to be reportable within its purview.

For example, only projects and third-party projects are currently relevant for reporting, and it is the report's responsiblity to invoke the service object that applies the activity level criteria.

(This is not to say that we couldn't rethink these responsibilities, but in the current state, it's not much of a stretch to move this criterion under the report.)

### Before
Screenshot 2023-11-14 at 14 02 24

### After
Screenshot 2023-11-14 at 14 02 54

## Next steps

### DIFF
--- a/spec/features/users_can_upload_actuals_spec.rb
+++ b/spec/features/users_can_upload_actuals_spec.rb
@@ -43,40 +43,92 @@ RSpec.feature "users can upload actuals" do
     expect(page.html).to include t("action.actual.upload.back_link")
   end
 
-  scenario "downloading a CSV template with activities for the current report" do
-    visit report_actuals_upload_path(report, format: :csv)
+  context "when the report is for ODA activities" do
+    let!(:sibling_project_no_longer_eligible) { create(:project_activity, :newton_funded, oda_eligibility: "no_longer_eligible", organisation: organisation, parent: project.parent) }
 
-    csv_data = page.body.delete_prefix("\uFEFF")
-    rows = CSV.parse(csv_data, headers: true).map(&:to_h)
+    scenario "the actuals template includes all the reportable activities for the current report" do
+      visit report_actuals_upload_path(report, format: :csv)
 
-    expect(rows).to match_array([
-      {
-        "Activity Name" => project.title,
-        "Activity Partner Organisation Identifier" => project.partner_organisation_identifier,
-        "Activity RODA Identifier" => project.roda_identifier,
-        "Financial Quarter" => report.financial_quarter.to_s,
-        "Financial Year" => report.financial_year.to_s,
-        "Actual Value" => "0.00",
-        "Refund Value" => nil,
-        "Receiving Organisation Name" => nil,
-        "Receiving Organisation Type" => nil,
-        "Receiving Organisation IATI Reference" => nil,
-        "Comment" => nil
-      },
-      {
-        "Activity Name" => sibling_project.title,
-        "Activity Partner Organisation Identifier" => sibling_project.partner_organisation_identifier,
-        "Activity RODA Identifier" => sibling_project.roda_identifier,
-        "Financial Quarter" => report.financial_quarter.to_s,
-        "Financial Year" => report.financial_year.to_s,
-        "Actual Value" => "0.00",
-        "Refund Value" => nil,
-        "Receiving Organisation Name" => nil,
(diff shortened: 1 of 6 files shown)


CristinaRO:
[Support request: actuals template empty for ISPF non-ODA report](https://trello.com/c/Vzw56NUn/2957-support-request-actuals-template-empty-for-ispf-non-oda-report)

spec/features/users_can_upload_actuals_spec.rb
```diff
@@ -43,40 +43,92 @@ def expect_to_see_successful_upload_summary_with(count:, total:)
...
+  context "when the report is for ODA activities" do
+    let!(:sibling_project_no_longer_eligible) { create(:project_activity, :newton_funded, oda_eligibility: "no_longer_eligible", organisation: organisation, parent: project.parent) }
+
+    scenario "the actuals template includes all the reportable activities for the current report" do
+      visit report_actuals_upload_path(report, format: :csv)
+
+      csv_data = page.body.delete_prefix("\uFEFF")
+      rows = CSV.parse(csv_data, headers: true).map(&:to_h)
+
+      expect(rows).to match_array([
```
mec, line 55: If I understand this right we are only asserting that the correct activities are present - if so comparing the entire contents of the csv row feels like too much and makes the test hard for me to parse, an alternative could be something like:

```ruby
    scenario "the actuals template includes all the reportable activities for the current report" do
      visit report_actuals_upload_path(report, format: :csv)

      csv_data = page.body.delete_prefix("\uFEFF")
      activity_identifiers = CSV.parse(csv_data, headers: true).values_at("Activity Partner Organisation Identifier").flatten

      expect(activity_identifiers).to include(project.partner_organisation_identifier)
      expect(activity_identifiers).to include(sibling_project.partner_organisation_identifier)

      expect(activity_identifiers).not_to include(sibling_project_no_longer_eligible.partner_organisation_identifier)
    end
```

Which has a bit of overhead, but is explicit about what we expect to see and not see.

spec/features/users_can_upload_actuals_spec.rb
```diff
@@ -43,40 +43,92 @@ def expect_to_see_successful_upload_summary_with(count:, total:)
...
+        organisation: organisation)
+    end
+
+    scenario "the actuals template includes all the reportable activities for the current report" do
+      visit report_actuals_upload_path(report, format: :csv)
+
+      csv_data = page.body.delete_prefix("\uFEFF")
+      rows = CSV.parse(csv_data, headers: true).map(&:to_h)
+
+      expect(rows).to match_array([
```
mec, line 103: Same down here, reading this makes me feel like I am missing something.

mec (approved):
I think I have this and I can follow the decision we have made! 👌 

I found the specs a little hard to read and left a suggestion, but it is likely just a taste thing so I am approving and leaving the choice yours.

spec/features/users_can_upload_actuals_spec.rb
CristinaRO: Hmm, good point. I think we want to test at least once the expected contents of the file, for example that the actuals get pre-populated with zero values, while the refunds do not get populated.

The description of the spec would get quite unwieldy though, any suggestions?

In theory we wouldn't need to test the actuals and refunds pre-population in the other tests, but then, the entire reason we're adding these new specs is that it turned out there were things that only happened for the ISPF non-ODA pathway. 😅

mec: I guess if what we want to test is the acutal column values, it might be best to have specs for that - I don't think it should hold us back merging this though - I would hope that is covered elsewhere, but I am fresh back the the code, so don't have that level of confidence yet.

CristinaRO: Thank you! Let's put a pin in it for now, if only because if I start looking at what tests could be refactored, it's going to be hard to stop 🙈

PR #2213, opened 2023-11-14
Release 139

CristinaRO:
## Changes in this PR
- Remove ispf_fund_in_stealth_mode that hides ISPF funds from users
- Add `is_oda` attribute to reports and populate it to `false` for all existing ISPF reports
- Amend report validation to permit two editable reports per fund and organisation, as long as they have different ODA types
- Amend the check for a pre-existing later report to take ODA type into account
- Error message shown when attempting to create a report while an unapproved one for the same fund, org, and ODA type exists includes the ODA type where relevant
- Show ODA/non-ODA alongside the fund short name on the reports table for ISPF
- Fix the logic that generates actuals and variance rows for reports, to avoid inserting a spurious column for reports that have forecasts but no actuals
- Show environment banner and email prefix on the `dev` domain

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 [Full changelog][unreleased]
 
+## Release 139 - 2023-11-14
+
+[Full changelog][139]
+
 - Remove ispf_fund_in_stealth_mode that hides ISPF funds from users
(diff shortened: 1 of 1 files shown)


CristinaRO:
[Prepare a release](https://trello.com/c/mdvsBFRG/2972-prepare-a-release)

shuldt (approved):

PR #2215, opened 2023-11-14
Upgrade Rails to 6.1.7.6

CristinaRO:
## Changes in this PR
- Upgrade Rails to 6.1.7.6

Actionpack has a possible cross-site scripting vulnerability via User Supplied Values to redirect_to. 6.1.7.4 is the first patched version.

We don't think the app is vulnerable, since we don't have user supplied values in redirect destinations, and we sanitise user input, but it's good practice to upgrade to the latest 6.x version anyway.

## Screenshots of UI changes

N/A

## Next steps


mec (approved):
🚢

PR #2223, opened 2023-11-16
(2952) New ISPF reports are created as ODA or non-ODA

CristinaRO:
## Changes in this PR
- Allow ISPF reports to be created as ODA or non-ODA
- Ensure that non-ISPF reports do not get a value set for `is_oda` (for historical reasons, the app relies on non-ISPF reports and activities not having an `is_oda` value set)
- Validate that the value of `is_oda` is set for ISPF reports
- Reorder the fields on the new report form per design

### Before
Screenshot 2023-11-16 at 17 46 19

### After
Screenshot 2023-11-16 at 17 45 46

## Next steps

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -114,6 +114,7 @@ class ReportsController < BaseController
       :financial_quarter,
       :financial_year,
       :fund_id,
+      :is_oda,
       :organisation_id,
       :deadline,
       :description
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Update how Node and Yarn are installed inside the application container
 - Allow ISPF reports to be created as ODA or non-ODA
 - Ensure that non-ISPF reports do not get a value set for `is_oda` (for historical reasons, the app relies on non-ISPF reports and activities not having an `is_oda` value set)
+- Validate that the value of `is_oda` is set for ISPF reports
 
 ## Release 139 - 2023-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Remove the `oda_eligibility` criterion from `reportable` scope, and apply it on a report by report basis, so we don't exclude ISPF non-ODA activities from ISPF non-ODA reports
 - Update how Node and Yarn are installed inside the application container
+- Allow ISPF reports to be created as ODA or non-ODA
+- Ensure that non-ISPF reports do not get a value set for `is_oda` (for historical reasons, the app relies on non-ISPF reports and activities not having an `is_oda` value set)
 
 ## Release 139 - 2023-11-14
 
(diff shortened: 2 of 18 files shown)


CristinaRO:
[Update the new report form to offer/enforce a choice of ISPF ODA and ISPF non-ODA when creating an ISPF report](https://trello.com/c/O0XfA5fc/2952-update-the-new-report-form-to-offer-enforce-a-choice-of-ispf-oda-and-ispf-non-oda-when-creating-an-ispf-report)

app/controllers/reports_controller.rb
```diff
@@ -114,10 +114,12 @@ def report_creatable_params
       :financial_quarter,
       :financial_year,
       :fund_id,
+      :is_oda,
       :organisation_id,
       :deadline,
       :description
-    )
+    ).reject! { |param| param.to_s == "is_oda" && params.dig("report", "fund_id") != Activity.by_roda_identifier("ISPF")&.id }
```
mec, line 121: This feels more like the models responsibility to me? If I've understood this right when a report is not ISPF `is_oda` needs to be `nil` could we model that with something like this on `Report`:

```ruby
def is_oda
  return nil unless for_ispf?

  super
end
```
My assumption here is that all other funded activities are inherently ODA.  There is evey likelyhood more to it than this, let me know.

mec (commented):
Left a comment, let me know your thoughts.

app/controllers/reports_controller.rb
mec: Okay, makes sense with the db queries, could we overide the setter then, something like:

```ruby
def is_oda=(value)
  value = nil unless for_ispf?

  super(value)
end
```

As this issue arrises when adding a new `Report` we are contolling how the models get added to the DB? Maybe even (I can't think what the retun value is for setters):

```ruby
def is_oda=(value)
  return true unless for_ispf?

  super(value)
end
```

I think my rationale is that without a defence on the model it is possible to create a Report in an unwanted state, as this only protects this controller action by removing the parameter before assignment.

I would be interested to understand the scope of the queries that rely on `is_oda` being nil if there is any time!

Thanks for bearing with me, I am sure this is frustrating and old ground, it is my first day back on the project so still catching up! 🙏

mec: Just back from dog walk and so time to mull - maybe all this needs is validation? We are adding the inclusion validator here, we could extend that out to include 'it must be nil unless the fund is ISPF, in which case it must be true or false'?

CristinaRO: The primary thing we're trying to prevent is the situation where a user:
- selects ISPF
- selects ODA or non-ODA
- then changes their mind and selects GCRF.

When selecting a non-ISPF fund, the nested radio boxes get hidden, but underneath, in the form, the ODA/non-ODA parameter is not nullified, so it gets sent to the controller without the user realising it, and without being able to reset the choice. In that context, throwing a validation error is not user friendly. We shouldn't make them reset the choice manually, since we already know what the outcome needs to be.

We could add more JavaScript to nullify `is_oda` when the user selects a different fund, but we still need a way to reject that attribute when the user has JS disabled.

I'm more inclined to accept your suggestion to override the setter, to avoid anyone else accidentally setting an invalid `is_oda` value.

mec: I will feel better when we have something on the model that begins to prevent the 'undesirable' state. 👍 

Whilst overiding the setter is a bit opqaue, so is dropping the param, but has the advantage of:

- working everywhere
- being closer to the effected model

mec (approved):
🚢

PR #2228, opened 2023-11-20
Remove VCAP services

CristinaRO:
## Changes in this PR
- Remove the option to use a public bucket (no longer applicable)
- Remove all code having to do with VCAP services, since this has not been used since moving to the new infrastructure

## Screenshots of UI changes

N/A

## Next steps


CristinaRO:
[Remove the VCAP_SERVICES code post-replatforming](https://trello.com/c/lfU10qwy/2939-remove-the-vcapservices-code-post-replatforming)

app/services/export/s3_uploader.rb
```diff
@@ -34,14 +34,7 @@ def upload
...
 
-      if use_public_bucket
-        OpenStruct.new(
-          url: bucket.object(filename).public_url,
-          timestamped_filename: filename
-        )
-      else
-        OpenStruct.new(timestamped_filename: filename)
-      end
+      OpenStruct.new(timestamped_filename: filename)
```
mec, line 25: I would really like to see a clearer commit message for this, I feel like it is less about the expectations and more about removing the need for `use_public_bucket` why are we doing that?

app/jobs/report_export_uploader_job.rb
```diff
@@ -26,7 +26,7 @@ def save_tempfile(export)
   end
 
   def upload_csv_to_s3(file:, filename:)
-    Export::S3Uploader.new(file: file, filename: filename, use_public_bucket: false).upload
+    Export::S3Uploader.new(file: file, filename: filename).upload
```
mec, line 29: Here again I would like to see more in the commit message, I appreciate this may be a taste thing.

mec (approved):
I would like to see more detail/why in the commit messages, I especially appreciate the extra context when stripping out code.

Everything looks solid though so I will leave that choice up to you! 👍

app/services/export/s3_uploader.rb
CristinaRO: It's a fair question, I was a bit lax in retrieving the history, but now I have. (And found a comment from Ynda suggesting we could remove `use_public_bucket` already, back then 🙈 )

PR #2230, opened 2023-11-21
(2974) Show ODA state on ISPF reports

mec:
## Changes in this PR

The introduction of the International Science Partnership Fund led to a state where reports contain either ODA activities or non-ODA activities - an important distinction for users.

This works displays the 'ODA state' on the report show view and report edit form.

The decisions was made for two versions of this:

- on the show page, for the ISPF fund reports, the ODA state is shown, for all other funds the row is not shown in the summary
- on the edit page, for the ISPF fund reports, the ODA state is appended to the fund name, for all other funds nothing is shown

(See screenshots below)

As we already have a way to present the fund short name with the ODA state appended, we followed this pattern and then renamed the original method to be clear about the behaviour of each.

We are currently in the position where the application will assume that for all funds except ISPF, `is_oda` MUST be nil, for this reason we guard against a nil value and take the appropriate action to try to prevent the nil value being interpreted as falsey when that would be the exact opposite of the desired state.

https://trello.com/c/q3CLtXRQ

## Screenshots of UI changes

image

image

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Validate that the value of `is_oda` is set for ISPF reports
 - Reorder the fields on the new report form per design
 - Show if a report for ISPF is ODA or non-ODA on the report show view
+- Show if a report for ISPF is ODA or non-ODA on the report edit form
 
 ## Release 139 - 2023-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Ensure that non-ISPF reports do not get a value set for `is_oda` (for historical reasons, the app relies on non-ISPF reports and activities not having an `is_oda` value set)
 - Validate that the value of `is_oda` is set for ISPF reports
 - Reorder the fields on the new report form per design
+- Show if a report for ISPF is ODA or non-ODA on the report show view
 
 ## Release 139 - 2023-11-14
 
--- a/app/presenters/report_presenter.rb
+++ b/app/presenters/report_presenter.rb
@@ -16,6 +16,12 @@ class ReportPresenter < SimpleDelegator
     I18n.l(super)
   end
 
+  def oda_type_summary
+    return if is_oda.nil?
+
+    I18n.t("is_oda_summary.#{is_oda}")
+  end
+
   def fund_and_oda_type
     return fund.source_fund.short_name if is_oda.nil?
 
--- a/app/presenters/report_presenter.rb
+++ b/app/presenters/report_presenter.rb
@@ -28,7 +28,7 @@ class ReportPresenter < SimpleDelegator
     is_oda ? "#{fund.source_fund.name} (ODA)" : "#{fund.source_fund.name} (non-ODA)"
   end
 
-  def fund_and_oda_type
+  def short_fund_name_and_oda_type
     return fund.source_fund.short_name if is_oda.nil?
 
     is_oda ? "#{fund.source_fund.short_name} (ODA)" : "#{fund.source_fund.short_name} (non-ODA)"
--- a/app/presenters/report_presenter.rb
+++ b/app/presenters/report_presenter.rb
@@ -22,6 +22,12 @@ class ReportPresenter < SimpleDelegator
     I18n.t("is_oda_summary.#{is_oda}")
   end
 
+  def fund_name_and_oda_type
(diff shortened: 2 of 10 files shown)


app/presenters/report_presenter.rb
```diff
@@ -16,7 +16,19 @@ def deadline
     I18n.l(super)
   end
 
-  def fund_and_oda_type
+  def oda_type_summary
+    return I18n.t("is_oda_summary.true") if is_oda.nil?
```
CristinaRO, line 20: Ah, sorry, this must be a misunderstanding of omission - we don't want to show ODA for non-ISPF funds. See for example the Reports index view (e.g https://staging.report-official-development-assistance.service.gov.uk/reports) where only ISPF gets the ODA / non-ODA disambiguation.

The designs for 2974 only showed ISPF because ISPF is the only fund for which we want to make any changes. Sorry if it wasn't clear.

CristinaRO (changes requested):
Sorry if it wasn't clear, but we didn't want to show ODA for any non-ISPF funds!

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 - Ensure that non-ISPF reports do not get a value set for `is_oda` (for historical reasons, the app relies on non-ISPF reports and activities not having an `is_oda` value set)
 - Validate that the value of `is_oda` is set for ISPF reports
 - Reorder the fields on the new report form per design
+- show the if a report contains ODA or non-ODA on the report view
```
CristinaRO, line 13: Would you mind removing the stray "the" and adding that this only affects ISPF reports, for consistency with the line below?

spec/features/users_can_view_reports_spec.rb
```diff
@@ -132,25 +132,43 @@ def expect_to_see_grouped_rows_of_reports_for_an_organisation(organisation:, exp
...
+          expect(page).to have_content t("page_content.tab_content.summary.refunds_total")
+        end
       end
 
-      expect(page).to have_content report.description
-      expect(page).to have_content l(report.deadline)
-      expect(page).not_to have_content t("page_content.report.summary.editable.#{report.editable?}")
-      expect(page).to have_content report.organisation.name
+      context "when the fund is any other" do
+        scenario "the report summary does not include the ODA state" do
```
CristinaRO, line 163: This might be a preference thing, but it doesn't feel right that we only test the negatives for a generic, non-ISPF fund.

Usually the way we test is the generic case first, with all the trimmings, and then add a test for the special cases such as ISPF. What we want to communicate through our tests is "the reports always have a summary with all these things, and then there's also ISPF which has an extra field".

In this case we do want to test that non-ISPF funds don't accidentally show the ODA row, but negative-only tests have a higher risk of passing by accident. (For example, we break displays for non-ISPF funds, and the page doesn't show a summary at all. The test would still pass.)

app/presenters/report_presenter.rb
CristinaRO: Returning "ODA" for cases where `is_oda` is nil makes this the only place in the application where we break the convention that only `is_oda: true` means "ODA". Where do you see this being used in future?

Given that the only place we use this summary is the report view, and we add a guard so we don't show the summary when `is_oda` is nil, won't this add more confusion? Developers will have to remember to always add a guard against nil.

CristinaRO (commented):
Thank you, almost there!

I still have concerns about muddying the waters with potentially showing "ODA" for non-ISPF funds. The users know they're implicitly ODA, but the users may also have associated the explicit showing of "ODA" with ISPF, and therefore be confused if in the future the summary pops up on non-ISPF funds.

Good thinking on renaming the existing method to better reflect its name!

app/presenters/report_presenter.rb
mec: As we no longer show the value of `oda_type_summary` for non ISPF reports, this is redundant now.

The logic for `is_oda: nil` means "ODA" is based on my understanding that all funds except ISPF are in fact ODA, so this followed when I thought we had to show it.

That said I think this highlights that the code model is wrong, as for all funds that are not ISPF `is_oda` feels (to me) like it should be true - as that is the case in reality, I appreciate that we are not in a position to resolve that though and of course, I could just be wrong!

CristinaRO (approved):
Thank you!

PR #2238, opened 2023-11-22
(2982) ISPF reports templates are split between ODA and non-ODA

CristinaRO:
## Changes in this PR
- Upload templates for ISPF reports use the ODA type of the report

### Before
Screenshot 2023-11-22 at 14 10 09

### After
Screenshot 2023-11-22 at 14 09 20

## Next steps

### DIFF
--- a/app/controllers/activities/uploads_controller.rb
+++ b/app/controllers/activities/uploads_controller.rb
@@ -11,7 +11,6 @@ class Activities::UploadsController < BaseController
     authorize report, :show?
 
     @report_presenter = ReportPresenter.new(report)
-    @is_ispf = report.for_ispf?
 
     prepare_default_report_trail(report)
     add_breadcrumb t("breadcrumb.report.upload_activities"), new_report_activities_upload_path(report)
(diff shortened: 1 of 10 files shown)


CristinaRO:
[ISPF reports need to only show the template corresponding to their ODA type](https://trello.com/c/hcSMGLgf/2982-ispf-reports-need-to-only-show-the-template-corresponding-to-their-oda-type)

app/controllers/activities/uploads_controller.rb
```diff
@@ -30,7 +30,7 @@ def update
     authorize report, :upload?
 
     @report_presenter = ReportPresenter.new(report)
-    @type = params[:type].to_sym
+    @type = @report_presenter.oda_type
```
mec, line 33: Do we even need `@type`?

mec (approved):
One thing, but not worried. 🚢

app/controllers/activities/uploads_controller.rb
CristinaRO: You're right, we don't! 🎉

PR #2245, opened 2023-11-22
Upgrade to ruby 3.0.6

CristinaRO:
## Changes in this PR
- Upgrade ruby version to 3.0.6

This is the last 3.0.x ruby version, containing security fixes.

After this, ruby 3.0 will be in security maintenance mode, and we should already be planning to upgrade to 3.1.

## Screenshots of UI changes
N/A

## Next steps


mec (approved):
Nice one! 🚢

PR #2246, opened 2023-11-23
(2927) Add total actuals to report export

mec:
## Changes in this PR

The total actuals is the sum of the 'actuals net' columns in the report,
of which there can be multiple..

Here we keep a total of the actual net values as it is calculated for each
financial quarter and add it to the columns for the current row if
appropriate.

Because the ActivityActualsColumns class also provides data for the
'spending breakdown' we have to check that a report is being generated
and only include the total actuals column if that is the case.

Most changes here relate to the addition of the extra column, the
ActivityActualsColumns adds `n` number of financial quarters and then
adds the total actuals to the end, some of this code has to rely on the
position of the values i.e the column number which has been altered by
this work.

We could not see another approach for this other than a major refactor
of how the exports are put together, so we settled on making this work.

https://trello.com/c/VE5g5sJi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 - Show if a report for ISPF is ODA or non-ODA on the report show view
 - Show if a report for ISPF is ODA or non-ODA on the report edit form
 - Upload templates for ISPF reports use the ODA type of the report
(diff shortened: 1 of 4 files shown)


CristinaRO (approved):
Great work, thank you!

PR #2203, opened 2023-11-07
(2947) (2966) Update report - activity finders

CristinaRO:
## Changes in this PR
- Update the query that retrieves the report to take into account the ODA type (`is_oda`) of the activity.
- Update projects for report finder to filter activities by ODA type.

## Screenshots of UI changes

N/A

## Next steps

- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [ ] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.
- [ ] Do any environment variables need amending or adding?
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/xml-validation.md)

PR #2247, opened 2023-11-23
(2991) Identify originating report by using is_oda value

CristinaRO:
## Changes in this PR
- Fix originating report finder in ActivityDefaults to take into account the ODA type of the activity

ActivityDefaults is responsible for setting a number of attributes for an activity that can be inferred from data the app already has.

One of these is the originating report, which is the currently active report for the fund and organisation of the activity.

Now that ISPF reports are split between ODA and non-ODA, the originating report finder needs to take into account the ODA type of the activity, and correctly find the corresponding report.

We forgot to update this finder in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2203

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Upload templates for ISPF reports use the ODA type of the report
 - The report export now includes a column for Total Actuals, the sum on the
   Actuals net values for all financial quarters in the report
+- Fix originating report finder in ActivityDefaults to take into account the ODA type of the activity
 
 ## Release 139 - 2023-11-14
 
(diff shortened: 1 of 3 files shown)


CristinaRO:
[Bug: Originating report finder in ActivityDefaults does not take into account the ODA type of the activity](https://trello.com/c/b4A1BxsG/2991-bug-originating-report-finder-in-activitydefaults-does-not-take-into-account-the-oda-type-of-the-activity)

mec (approved):
🚢

PR #2248, opened 2023-11-23
(2928) Outgoing commitment value in report csv export

mec:
## Changes in this PR

We want to show the original commitment value for an activity in the report csv export.

The value is entered by a user and is optional, when there is no value we show a blank value as zero could imply there was explicitally £0 of commitment, which may be missleading.

After spending time in the spec for the report export, we refactored the helpers to make the spec more approachable.

We also came across a redundant method, which we removed.

https://trello.com/c/7TbEQrNO

### DIFF
--- a/app/services/export/report.rb
+++ b/app/services/export/report.rb
@@ -108,10 +108,6 @@ class Export::Report
     @_forecast_rows ||= @forecast_columns.rows
   end
 
-  def has_forecast_rows?
-    forecast_rows.values.flatten.any?
-  end
-
   def variance_rows
     @_variance_rows ||= @variance_column.rows
   end
--- a/app/services/export/report.rb
+++ b/app/services/export/report.rb
@@ -13,6 +13,8 @@ class Export::Report
       Export::ActivityPartnerOrganisationColumn.new(activities_relation: activities)
     @change_state_column =
       Export::ActivityChangeStateColumn.new(activities: activities, report: @report)
+    @commitment_column =
+      Export::ActivityCommitmentColumn.new(activities: activities.includes([:commitment]))
     @actuals_columns =
       Export::ActivityActualsColumns.new(activities: activities, report: @report)
     @forecast_columns =
@@ -39,6 +41,7 @@ class Export::Report
     headers << @implementing_organisations.headers
     headers << @partner_organisations.headers
     headers << @change_state_column.headers
+    headers << @commitment_column.headers
     headers << @actuals_columns.headers if @actuals_columns.headers.any?
     headers << @variance_column.headers if @actuals_columns.headers.any? && @forecast_columns.headers.any?
     headers << @forecast_columns.headers if @forecast_columns.headers.any?
@@ -55,6 +58,7 @@ class Export::Report
       row << implementing_organisations_rows.fetch(activity.id, nil)
       row << partner_organisation_rows.fetch(activity.id, nil)
       row << change_state_rows.fetch(activity.id, nil)
+      row << commitment_rows.fetch(activity.id, nil)
       row << actuals_rows.fetch(activity.id, nil) if @actuals_columns.headers.any?
       row << variance_rows.fetch(activity.id, nil) if @actuals_columns.headers.any? && @forecast_columns.headers.any?
       row << forecast_rows.fetch(activity.id, nil) if @forecast_columns.headers.any?
@@ -90,6 +94,10 @@ class Export::Report
 
   def change_state_rows
     @_change_state_rows ||= @change_state_column.rows
+  end
+
+  def commitment_rows
+    @_commitment_rows ||= @commitment_column.rows
   end
 
   def actuals_rows
--- a/spec/models/export/activity_commitment_column_spec.rb
+++ b/spec/models/export/activity_commitment_column_spec.rb
@@ -1,0 +1,52 @@
+RSpec.describe Export::ActivityCommitmentColumn do
+  describe "#headers" do
+    it "returns an array that contains the header name of the column" do
+      activities = []
+
+      column = described_class.new(activities: activities)
+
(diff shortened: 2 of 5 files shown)


app/services/export/report.rb
```diff
@@ -55,6 +58,7 @@ def rows
       row << implementing_organisations_rows.fetch(activity.id, nil)
       row << partner_organisation_rows.fetch(activity.id, nil)
       row << change_state_rows.fetch(activity.id, nil)
+      row << commitment_rows.fetch(activity.id, nil)
```
CristinaRO, line 61: I think the way we've coded it now, there will always be a value (which could be the empty string), but if there wasn't, do we want it to return nil, or the empty string?

(I'm not sure, that's why I'm asking!)

spec/models/export/activity_commitment_column_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      activities = []
+
+      column = described_class.new(activities: activities)
+
+      expect(column.headers).to eql ["Original Commitment"]
+    end
+  end
+
+  describe "#rows" do
+    context "when there are no acitvities" do
```
CristinaRO, line 13:
```suggestion
    context "when there are no activities" do
```

spec/models/export/activity_commitment_column_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      it "returns an empty hash" do
+        activities = []
+
+        column = described_class.new(activities: activities)
+
+        expect(column.rows).to be {}
+      end
+    end
+
+    context "when there are acitivites" do
```
CristinaRO, line 23:
```suggestion
    context "when there are activities" do
```

spec/models/export/activity_commitment_column_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      context "when the activity has a commitment" do
+        it "returns the activity ID and the value of the commitment" do
+          commitment = double(Commitment, value: 100000.00)
+          acitivity = double(Activity, commitment: commitment, id: "ACTIVITY_ID")
```
CristinaRO, line 27:
```suggestion
          activity = double(Activity, commitment: commitment, id: "ACTIVITY_ID")
```

spec/models/export/activity_commitment_column_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+          activities = [acitivity]
```
CristinaRO, line 28:
```suggestion
          activities = [activity]
```

spec/models/export/activity_commitment_column_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+          column = described_class.new(activities: activities)
+
+          expect(column.rows.count).to eql 1
+          expect(column.rows.first[0]).to eql "ACTIVITY_ID"
+          expect(column.rows.first[1]).to eql 100000.00
+        end
+      end
+
+      context "when the activity has no commitment" do
+        it "returns the value as 0" do
```
CristinaRO, line 39: Should the spec description say "returns the value as an empty string", or am I missing something?

spec/models/export/activity_commitment_column_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+          acitivity = double(Activity, commitment: nil, id: "ACTIVITY_ID")
```
CristinaRO, line 40:
```suggestion
          activity = double(Activity, commitment: nil, id: "ACTIVITY_ID")
```

spec/models/export/activity_commitment_column_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+          activities = [acitivity]
```
CristinaRO, line 41:
```suggestion
          activities = [activity]
```

spec/services/export/report_spec.rb
```diff
@@ -194,8 +206,10 @@
         expect(headers).to include("Implementing organisations")
         expect(headers).to include("Partner organisation")
         expect(headers).to include("Change state")
+        expect(headers).to include("Original Commitment")
         expect(headers.to_s).to_not include("Actual net")
         expect(headers.to_s).to_not include("Variance")
+        expect(headers).not_to include("Total Acutals")
```
CristinaRO, line 212:
```suggestion
        expect(headers).not_to include("Total Actuals")
```

spec/services/export/report_spec.rb
```diff
@@ -325,7 +337,7 @@
 
         it "returns the rows with the correct tags" do
           first_row = subject.rows.first.to_a
-          tags_index = 55
+          tags_index = 56
```
CristinaRO, line 340: Could you apply the awesome refactoring to the Tags column as well?

CristinaRO (approved):
Great work, particularly thankful for the refactoring!

There are a few things I'd like addressed, especially the typo that means a spec could pass by coincidence rather than by design.

I do think we should test that the order of the columns is as expected, if we don't do that already, but it shouldn't be a blocker for this PR.

spec/services/export/report_spec.rb
mec: Yep.

app/services/export/report.rb
mec: Yeah, I am not sure - I guess it depends on if when output to csv it handles the nil, in which case I prefer the nil - I will have to look.

mec: It seems to be fine with nil, so going with that.

spec/models/export/activity_commitment_column_spec.rb
mec: This is 'returns nil' now 👍

CristinaRO (approved):
Great work, thank you!

PR #2259, opened 2023-11-27
(2929) Published to IATI value in report csv

mec:
## Changes in this PR

This work adds the `publish_to_iati` from `Activity` value in the report csv export.

- The value is true by default 
- The value is shown as one of 'Yes' or 'No'
- The column comes at the very end of the columns in the csv table

https://trello.com/c/c4GFG0UW

### DIFF
--- a/spec/models/export/activity_iati_column_spec.rb
+++ b/spec/models/export/activity_iati_column_spec.rb
@@ -1,0 +1,51 @@
+RSpec.describe Export::ActivityIatiColumn do
+  describe "#headers" do
+    it "returns an array that contains the header name of the column" do
+      activities = []
+
+      column = described_class.new(activities: activities)
+
+      expect(column.headers).to eql ["Published on IATI"]
+    end
+  end
+
+  describe "#rows" do
+    context "when there are no activities" do
+      it "returns an empty hash" do
+        activities = []
+
+        column = described_class.new(activities: activities)
+
+        expect(column.rows).to be {}
+      end
+    end
+
+    context "when there are activities" do
+      context "when the activity is publishable to IATI" do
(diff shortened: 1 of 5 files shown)


spec/models/export/activity_iati_column_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        activities = []
+
+        column = described_class.new(activities: activities)
+
+        expect(column.rows).to be {}
+      end
+    end
+
+    context "when there are activities" do
+      context "when the activity has been published to IATI" do
```
CristinaRO, line 24: It would be more accurate to describe the spec as "when the activity is publishable to IATI" - that reflects the purpose of the field, and it might ease your unease about the past tense.

The past tense will still be present in the column header, but we've had confirmation that this is what works and makes sense for the client, so it's ok.

spec/models/export/activity_iati_column_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+          column = described_class.new(activities: activities)
+
+          expect(column.rows.count).to eql 1
+          expect(column.rows.first[0]).to eql "ACTIVITY_ID"
+          expect(column.rows.first[1]).to eql "Yes"
+        end
+      end
+
+      context "when the activity has not been published to IATI" do
```
CristinaRO, line 37:
```suggestion
      context "when the activity is publishable to IATI" do
```

CristinaRO (approved):
Had another thought to help with the past tense unease, if you agree! Wouldn't make you fiddle with the spec description so much, but in this case it feels like an accurate spec description might help everyone!

PR #2263, opened 2023-11-28
(2932) Add IATI Scope to xml

mec:
## Changes in this PR

The 'scope' of the activity is an IATI metric that we want to include for publishing. Like many IATI metrics, it is up to DSIT to interpret as they see. Here the logic is:

- national when the acitivity benefits a single country
- regional when the activity benefits multiple countries
- mulit-national when the activity benefits multiple countries across multiple regions

No other scope value were deemed necessary for DSIT.

With the business logic clear, we went ahead and added a service object to calculate it - the decision is covered in the commit.

Once we can calculate the scope we then show it in the IATI xml download that is published.

We have run a sample xml file through the IATI validator which raised no new errors, although there are errors, they are pre-existing.

https://trello.com/c/pAXIO0Ce

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 [Full changelog][unreleased]
 
+- The IATI activity scope is calculated based on the benefitting countries and
+  included in the IATI XML download as applicable
+
 ## Release 140 - 2023-11-28
 
 [Full changelog][140]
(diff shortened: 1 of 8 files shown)


shuldt (approved):
Looks good, and very happy with the detailed notes in the commit messages. Worth having a chat about these "pre-existing errors" you mention in the PR message maybe?

mec:
> Looks good, and very happy with the detailed notes in the commit messages. Worth having a chat about these "pre-existing errors" you mention in the PR message maybe?

@shuldt can do, wrote it up here: https://trello.com/c/IcuVvsCV

PR #2267, opened 2023-11-30
(2995) Fix: flaky sign in spec

mec:
There are two quite different changes here:

The first tightens the time control around the user signing in and the spec storing the OTP for comparing later.

The second makes the OTP always correct, decoupling the spec from the OTP entirely.

I've documented what I learnt along wth the changes to the acutal specs.

I found that the OTP seems to last 5 mintutes and 30 seconds, even though it is configured to 5 miniutes. I've added a spec that demonstrates the odd extra 30 seconds and encourage you to have a play with the specs and see what you think and let me know.

I am confident the end result here will stop the flaking, but other folks tastes may vary so I am leaving this draft for now.

Appreciate any other views on this spec flaking, there is an example CI run linked in the commit, but here it is again:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/actions/runs/7020362563/job/19099874908#step:6:9200

https://trello.com/c/AmrnslpW

### DIFF
--- a/spec/features/users_can_sign_in_spec.rb
+++ b/spec/features/users_can_sign_in_spec.rb
@@ -45,37 +45,62 @@ RSpec.feature "Users can sign in" do
     end
 
     context "user has a confirmed mobile number" do
-      scenario "successful sign in via header link" do
-        # Given a user with 2FA enabled and a confirmed mobile number exists
-        user = create(:administrator, :mfa_enabled, mobile_number_confirmed_at: DateTime.current)
+      context "and the OTP is correct" do
+        scenario "successful sign in via header link" do
+          # Given a user with 2FA enabled and a confirmed mobile number exists
+          user = create(:administrator, :mfa_enabled, mobile_number_confirmed_at: DateTime.current)
+          otp_at_time_of_login = nil
 
-        # When I log in with that user's email and password
-        visit root_path
-        log_in_via_form(user)
+          # When I log in with that user's email and password
+          travel_to Time.new(2023, 11, 29, 10, 0o0, 0o0) do
+            visit root_path
+            log_in_via_form(user)
+            otp_at_time_of_login = user.current_otp
 
-        # Then there should be no link to check my mobile number
-        expect(page).not_to have_link "Check your mobile number is correct"
+            # Then there should be no link to check my mobile number
+            expect(page).not_to have_link "Check your mobile number is correct"
 
-        otp_at_time_of_login = user.current_otp
-        # And I should receive an OTP to my mobile number
-        expect(notify_client).to have_received(:send_sms).with({
-          phone_number: user.mobile_number,
-          template_id: ENV["NOTIFY_OTP_VERIFICATION_TEMPLATE"],
-          personalisation: {otp: otp_at_time_of_login}
-        })
+            # And I should receive an OTP to my mobile number
+            expect(notify_client).to have_received(:send_sms).with({
+              phone_number: user.mobile_number,
+              template_id: ENV["NOTIFY_OTP_VERIFICATION_TEMPLATE"],
+              personalisation: {otp: otp_at_time_of_login}
+            })
+          end
 
-        # When I enter the one-time password before it expires
-        travel 3.minutes do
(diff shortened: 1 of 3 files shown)


doc/user-and-authenitcation.md
```diff
@@ -0,0 +1,29 @@
+# Users and authentication
+
+The applications faciliates user management, authentication and
```
CristinaRO, line 3:
```suggestion
The applications facilitates user management, authentication and
```

doc/user-and-authenitcation.md
CristinaRO, line 1: Typo in filename! ;) `users-and-authentication` maybe? Or just `authentication`?

doc/user-and-authenitcation.md
```diff
@@ -0,0 +1,29 @@
...
+authorisation itself.
+
+## Two factor authentication
+
+We rely on the following gems to buld out the two factor authenication:
```
CristinaRO, line 8:
```suggestion
We rely on the following gems for our implementation of two factor authentication:
```

doc/user-and-authenitcation.md
```diff
@@ -0,0 +1,29 @@
...
+- [Devise-Two-Factor
+  Authentication](https://github.com/devise-two-factor/devise-two-factor)
+
+As there are a number of Devise 2FA gems, make sure you are referencing the
+correct documentation!
+
+Our one time passwords are time based. Generally speaking OTP have a concept of
+'drift' to allow for differences in the time on both the service and client.
+
+Whilst many 2FA Devise gems have seperate setting for the expiry and drift,
```
CristinaRO, line 20:
```suggestion
Whilst many 2FA Devise gems have separate setting for the expiry and drift,
```

doc/user-and-authenitcation.md
```diff
@@ -0,0 +1,29 @@
...
+Devise-Two-Factor does not appear to, combining them into a single
+`[otp_allowed_drift](https://github.com/devise-two-factor/devise-two-factor#enabling-two-factor-authentication)` that we have set to 5 mintues in
```
CristinaRO, line 22:
```suggestion
`[otp_allowed_drift](https://github.com/devise-two-factor/devise-two-factor#enabling-two-factor-authentication)` that we have set to 5 minutes in
```

doc/user-and-authenitcation.md
```diff
@@ -0,0 +1,29 @@
...
+`[config/initializers/devise.rb](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/config/initializers/devise.rb)`.
+
+Something appears to introduce an addtitional 30 seconds on top of this which is
```
CristinaRO, line 25:
```suggestion
Something appears to introduce an additional 30 seconds on top of this, which is
```

spec/features/users_can_sign_in_spec.rb
```diff
@@ -45,37 +45,62 @@ def log_in_via_form(user, remember_me: false)
...
+      end
 
-        otp_at_time_of_login = user.current_otp
-        # And I should receive an OTP to my mobile number
-        expect(notify_client).to have_received(:send_sms).with({
-          phone_number: user.mobile_number,
-          template_id: ENV["NOTIFY_OTP_VERIFICATION_TEMPLATE"],
-          personalisation: {otp: otp_at_time_of_login}
-        })
+      context "and the OTP has expired (Configured as 5 minutes, but appears to be 5 minutes 30 seconds?)" do
```
CristinaRO, line 86: I would remove the question mark - we are already expressing doubt with "appears".

Also not sure I'd put this in the spec description or a comment within the spec, but that may be a matter of preference.

CristinaRO (commented):
Thank you for tackling this tricky issue! I'm still ruminating on it, mainly on the topics of:
- do we still have an end-to-end feature spec that is not too mocked/stubbed?
- do we want such a spec, given that what is under test is not the third-party gems, but our implementation?

mec:
@CristinaRO I've updated the subbed OTP to a string as we discussed. This is ready to have another look. 👍

CristinaRO (approved):

PR #2270, opened 2023-11-30
(2985) Order activity comments exported in budget CSV

CristinaRO:
## Changes in this PR

PostgreSQL doesn't guarantee an order for records unless we specify an ordering criterion ourselves. It is *likely* that the records will be returned with the most recently updated last, which is why the spec only fails occasionally.

This is potentially the cause of the budget spec failing sometimes.

Specifying the order in which the comments are returned will fix the intermittent failures, if they were caused by the unpredictable ordering coming from the database.

If, on the other hand, the intermittent errors are caused by some randomness in how the test objects are created[1], we may see failures again sometimes.

[1] `programme_activity_comment_1` should be created before `programme_activity_comment_2`. If *this* ordering is not guaranteed, and `programme_activity_comment_2` ends up with a `created_at` older than `programme_activity_comment_1`'s created_at, then the fix may not make a difference.

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/app/services/budget/export.rb
+++ b/app/services/budget/export.rb
@@ -56,7 +56,7 @@ class Budget
         activity.extending_organisation&.name,
         I18n.t("table.body.activity.level.#{activity.level}"),
         activity.title,
-        activity.programme? ? comments_formatted_for_csv(activity.comments) : ""
+        activity.programme? ? comments_formatted_for_csv(activity.comments.order(:created_at)) : ""
       ]
     end
 


CristinaRO:
[Flaky spec: level B budget comments do not specify any ordering criteria](https://trello.com/c/9tHeLR5v/2985-flaky-spec-level-b-budget-comments-do-not-specify-any-ordering-criteria)

mec (approved):
I always thought the default order was by the record ID and becuase we use UUID, it looks random?

PR #2271, opened 2023-11-30
Alias health_check so it responds to health-check

CristinaRO:
## Changes in this PR

The AWS health check endpoint has been configured to look for the route `health-check` instead of `health_check`.

Rather than asking them to reconfigure it, we will accept that both versions (with dash and with underscore) are common, and provide an alias for the route, so that it responds to both requests.

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -21,6 +21,7 @@ Rails.application.routes.draw do
 
   scope module: "public" do
     get "health_check" => "base#health_check"
+    match "health-check" => "base#health_check", :via => :get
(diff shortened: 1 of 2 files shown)


CristinaRO:
[Application health check should respond to health-check as well as health_check](https://trello.com/c/LmEcnssW/3000-application-health-check-should-respond-to-health-check-as-well-as-healthcheck)

mec (approved):
🚢

PR #2273, opened 2023-12-04
Release 141

CristinaRO:
- The IATI activity scope is calculated based on the benefitting countries and included in the IATI XML download as applicable
- The error summary is now shown correctly when adding new matched effort in the application
- The filenames for ISPF report CSVs now include the ODA type


CristinaRO:
[Release 141](https://trello.com/c/npxWktSM/3001-release-141)

PR #2281, opened 2023-12-13
(2931) Actual, Refund and Comment csv import row

mec:
- Add Actial and Refund csv import row
- WIP: Use new csv import row

## Next steps

### DIFF
--- a/app/models/import/transactions/actual_and_refund_csv_row.rb
+++ b/app/models/import/transactions/actual_and_refund_csv_row.rb
@@ -1,0 +1,216 @@
+class Import::Transactions::ActualAndRefundCsvRow
+  attr_reader :errors
+
+  def initialize(csv_row)
+    @row = csv_row
+    @errors = {}
+  end
+
+  def valid?
+    validate_roda_identifier_cannot_be_blank
+    validate_financial_quarter
+    validate_financial_year
+    validate_receiving_organisation_type
+
+    if validate_financial_value_for("Actual Value") && validate_financial_value_for("Refund Value")
+      validate_actual_and_refund_separate
+      validate_activity_comment
+      validate_refund_require_comment
+    end
+
+    @errors.empty?
+  end
+
+  def invalid?
+    !valid?
+  end
+
+  def description
+    return nil unless validate_financial_quarter && validate_financial_year
+
+    quarter = financial_quarter
+    year = financial_year
+
+    "#{FinancialQuarter.new(quarter, year)} spend on #{activity.title}"
+  end
+
+  def activity
+    @_activity ||= Activity.find_by_roda_identifier(roda_identifier)
+    if @_activity.nil?
+      @errors["Activity RODA Identifier"] = [roda_identifier, "Activity cannot be found"]
+    end
+    @_activity
+  end
+
+  def roda_identifier
+    @row.field("Activity RODA Identifier") || :blank
+  end
+
+  def actual_value
+    fetch_financial_value_for_column("Actual Value")
+  end
+
+  def refund_value
+    fetch_financial_value_for_column("Refund Value")
+  end
+
+  def comment
+    @row.field("Comment") || :blank
+  end
+
+  def providing_organisation
+    return nil if activity.nil?
+
+    activity.providing_organisation
+  end
+
+  def financial_quarter
+    @row.field("Financial Quarter") || :blank
+  end
+
+  def financial_year
+    @row.field("Financial Year") || :blank
+  end
+
+  def receiving_organisation_name
+    @row.field("Receiving Organisation Name") || :blank
+  end
+
+  def receiving_organisation_type
+    @row.field("Receiving Organisation Type") || :blank
+  end
+
+  def receiving_organisation_iati_reference
+    @row.field("Receiving Organisation IATI Reference") || :blank
+  end
+
+  def transaction_type
+    return nil unless validate_financial_value_for("Actual Value") && validate_financial_value_for("Refund Value")
+
+    return :blank if blank?(actual_value) && blank?(refund_value) && blank?(comment)
+    return :comment if not_blank?(comment) && exactly_zero?(actual_value) && exactly_zero?(refund_value)
+    return :actual if not_blank_or_zero?(actual_value) && blank_or_zero?(refund_value)
+    return :refund if not_blank_or_zero?(refund_value) && blank_or_zero?(actual_value)
+  end
+
+  private def validate_receiving_organisation_type
+    return true if receiving_organisation_type.eql?(:blank)
+
+    unless value_in_code_list?("organisation_type", receiving_organisation_type)
+      @errors["Receiving Organisation Type"] = [
+        receiving_organisation_type, I18n.t("importer.errors.actual.invalid_iati_organisation_type")
+      ]
+      return false
+    end
+    true
+  end
+
+  private def validate_activity_comment
+    if blank?(comment) && exactly_zero?(actual_value) && exactly_zero?(refund_value)
+      @errors["Comment"] = [comment, "Activities with 0 actual and 0 refund require a comment"]
+    end
+
+    if not_blank?(comment) && blank?(actual_value) && blank?(refund_value)
+      @errors["Comment"] = [comment, "Actual and refund must both be zero to provide a comment"]
+    end
+  end
+
+  private def validate_actual_and_refund_separate
+    if not_blank_or_zero?(actual_value) && not_blank_or_zero?(refund_value)
+      @errors["Actual Value"] = [actual_value, "You cannot report actual and refunds together"]
+      @errors["Refund Value"] = [refund_value, "You cannot report actual and refunds together"]
+    end
+  end
+
+  private def validate_financial_quarter
+    if blank?(financial_quarter)
+      @errors["Financial Quarter"] = [financial_quarter, "Cannot be blank"]
+      return false
+    elsif ["1", "2", "3", "4"].none?(financial_quarter)
+      @errors["Financial Quarter"] = [financial_quarter, "Must be 1, 2, 3 or 4"]
+      return false
+    end
+    true
+  end
+
+  private def validate_financial_year
+    if blank?(financial_year)
+      @errors["Financial Year"] = [financial_year, "Cannot be blank"]
+      return false
+    elsif financial_year.length > 4
+      @errors["Financial Year"] = [financial_year, "Must be a four digit year"]
+      return false
+    end
+    true
+  end
+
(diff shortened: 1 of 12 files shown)


spec/support/shared_examples/behaves_like_financial_value.rb
```diff
@@ -0,0 +1,61 @@
+RSpec.shared_examples "a financial value" do |header|
```
CristinaRO, line 1: Could we have an explicit spec for "when value is 0", so we're clear what happens?

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+        allow(csv_row).to receive(:field).with("Activity RODA Identifier").and_return(nil)
+        row = described_class.new(csv_row)
+
+        expect(row.roda_identifier).to eql(:blank)
+      end
+    end
+  end
+
+  describe "#actual_value" do
+    it "returns the value from the csv row" do
```
CristinaRO, line 54: The description of this spec feels incomplete. It's not just returning the value, it's converting it into a BigDecimal.

It feels like there's an implicit context here too, "when the value is numeric".

What happens when the value is zero?

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+        allow(csv_row).to receive(:field).with("Actual Value").and_return(nil)
+        row = described_class.new(csv_row)
+
+        expect(row.actual_value).to eql(:blank)
+      end
+    end
+  end
+
+  describe "#refund_value" do
+    it "returns the value from the csv row" do
```
CristinaRO, line 84: Same observation as for the actual_value spec.

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+  end
+
+  describe "#comment" do
+    it "returns the value from the csv row" do
+      row = described_class.new(valid_csv_row)
+
+      expect(row.comment).to eql("This is a comment.")
+    end
+
+    context "when the comment is empty" do
```
CristinaRO, line 120: What happens if the comment is an empty string instead of nil?

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+        allow(csv_row).to receive(:field).with("Receiving Organisation Type").and_return(nil)
+        row = described_class.new(csv_row)
+
+        expect(row.receiving_organisation_type).to eql(:blank)
+      end
+    end
+  end
+
+  describe "#transaction_type" do
+    context "when the actual value is not zero and the refund value is zero" do
```
CristinaRO, line 287: What does it return when the actual value is not zero and the refund value is an empty string or nil?

I would expect it to return `:actual`.

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+        allow(csv_row).to receive(:field).with("Actual Value").and_return("10000.50")
+        allow(csv_row).to receive(:field).with("Refund Value").and_return("0")
+        allow(csv_row).to receive(:field).with("Comment").and_return("This is a comment on an actual")
+        row = described_class.new(csv_row)
+
+        expect(row.transaction_type).to eql :actual
+      end
+    end
+
+    context "when the actual value is zero and the refund value is not zero" do
```
CristinaRO, line 299: It should also return `:refund` if the actual value is an empty string or nil, just in case the user has deleted the pre-populated zero.

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+      it "returns refund" do
+        csv_row = double(CSV::Row)
+        allow(csv_row).to receive(:field).with("Actual Value").and_return("0")
+        allow(csv_row).to receive(:field).with("Refund Value").and_return("10000.32")
```
CristinaRO, line 303: If the refund value is not numeric, does it return `:refund` and let the validation deal with the non-numeric elsewhere?

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+        allow(csv_row).to receive(:field).with("Comment").and_return("This is a comment on an refund")
+        row = described_class.new(csv_row)
+
+        expect(row.transaction_type).to eql :refund
+      end
+    end
+
+    context "when the actual value is zero and the refund value is zero" do
```
CristinaRO, line 311: What happens when the actual value is zero and the refund value is an empty string or nil, and there is a comment? I would expect it to also be `:comment`.

Or when the actual value and refund value are both empty strings/nil, and there is a comment?

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+          end
+        end
+      end
+    end
+
+    describe "refunds" do
+      it_behaves_like "a financial value", "Refund Value"
+
+      context "when there is a comment" do
+        it "is invalid" do
```
CristinaRO, line 423: Did you mean "is valid"? The expectation on row 430 asserts that it is valid.

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+        it "is invalid with an appropriate message" do
+            csv_row = valid_csv_row
+            allow(csv_row).to receive(:field).and_call_original
+            allow(csv_row).to receive(:field).with("Actual Value").and_return(nil)
+            allow(csv_row).to receive(:field).with("Refund Value").and_return(nil)
+            allow(csv_row).to receive(:field).with("Comment").and_return("This is a valid activity comment")
+            row = described_class.new(csv_row)
+
+            expect(row).to be_invalid
+            expect(row.errors["Comment"][1]).to include("must both be zero to provide a comment")
```
CristinaRO, line 530: Ah, this is an assumption that I would like to validate with Dan & co.

For the PO users' convenience, I would expect that they would prefer to not have to add a zero in the Refund value column manually. Is there a strong need for ease of development to require this?

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+            allow(csv_row).to receive(:field).with("Comment").and_return(nil)
+            row = described_class.new(csv_row)
+
+            expect(row).to be_invalid
+            expect(row.errors["Comment"][1]).to include("with 0 actual and 0 refund require a comment")
+        end
+      end
+
+      context "when all of the required values are blank" do
+        it "is invalid with an appropriate message" do
```
CristinaRO, line 549: The description says invalid, but the assertion on row 557 says valid, so I'm confused!

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
mec: I went with the explicit '0' to make it clear the intention is to say:

> I have 0 actual value and 0 refund value and a comment to leave

But I don't mind what we do, it just needs deciding?

app/models/import/transactions/actual_and_refund_csv_row.rb
```diff
@@ -0,0 +1,216 @@
...
+    begin
+      ConvertFinancialValue.new.convert(value)
+    rescue ConvertFinancialValue::Error
+      @errors[header] = [value, I18n.t("importer.errors.actuals_and_refunds.non_numeric")]
+      valid = false
+    end
+    valid
+  end
+
+  private def blank?(value)
```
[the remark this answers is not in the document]
mec: Interesting, I went with `blank` as that is the language used in the existing implementation for 'empty' when informing the user:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/0378e81390bf956432e9d457175c0220ebcdf383/app/services/actual/import.rb#L183

I think I preferred the idea of 'empty' but that is also a Rails method!

It was original nil, but this became tricky and messy in parts, in the end I decided to stick to the path that was already trodden.

Happy to do something else though.

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
mec: 👍 Fair point, I'll update the spec.

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
mec: My understanding here is that `CSV::Row`, which is the expected passed in object returns nil for empty values, you would have to do something else to get an empty string?

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
mec: `refund_value` cannot be an empty string or `nil` by design - we try to represent that as :blank right now.

See note about about " " (space) though.

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
mec: The first state cannot happen as `refund_value` returns a string or :blank along with " " (space)!

The second state is explicitally considered invalid at the moment by desgin, I made the decision to expect explicit zeros when adding a 'activity comment' - we can change that decision, I just need to know and Ideally, know all the states first.

spec/models/import/transactions/actual_and_refund_csv_row_spec.rb
mec: I've added a guard that represents:

> we cannot infer the type unless the actual and refund value are numeric

which would equte to an invalid row anyway that would not be imported - but now we cover the case be returning `nil`

mec: Wow, moving the responsibility up creates a bunch of duplication I am not super happy with! I'll leave it witht the guard for now.

mec:
Thanks for all the feedback, this work has been superceded by #2284, closing now.

PR #2291, opened 2024-01-03
(3015) Update specs for Factory Bot Rails > 6.4.3

mec:
The change introduced in factory bot rails 6.4.3 impacts this spec.

https://github.com/thoughtbot/factory_bot_rails/commit/0040292890541736d0065c30ab2f798e8608787d#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R168

Updating to 6.4.3 results in our test suite failing. This work fixes the resulting isses.

The majority can be done by stubbing calls instead of trying to `update` a `build_stubbed` object.

The last two commits were tricky. The policies fetch items from the database which no longer works for `build_stubbed` objects. The most straightforward fix is to call `create` instead of `build_stubbed` this is has small performance implications, but feels like a good enough step.  

https://trello.com/c/S7YBjRm1


CristinaRO:
[Fix test suite for Factory bot rails > 6.4.3](https://trello.com/c/S7YBjRm1/3015-fix-test-suite-for-factory-bot-rails-643)

CristinaRO (approved):
Thanks for doing this work!

PR #2292, opened 2024-01-04
Fix organisation currency seed value

mec:
The currency codes are all capitals, this incorrect seed value tripped
me up, so let's fix it.

https://trello.com/c/IcuVvsCV


shuldt (approved):
Seems reasonable!

CristinaRO:
[IATI XML: Currency values must be all capitals](https://trello.com/c/IcuVvsCV/2996-iati-xml-currency-values-must-be-all-capitals)

PR #2300, opened 2024-01-09
(2931/3) Actual, Refund or Comment row import service

mec:
In #2293 we introduced a model of a row that contains Actual, Refund or Activity comment data for importing.

Here we add a service that can turn one of these row objects into actual objects in the application database.

Given a appropriate `CSV::Row` this service will with return the resulting object that was added or false, a row can be skipped, see below.

This is the point at which we will attempt to locate the `Activity` to associate the record with. We also check that the activity is authorised for the report and user.

In the current import implementation (that this work intends to replace) a row cannot be skipped i.e. a row cannot result in no data being created, this was seen by users as a drawback, they want to include rows that have no result. To support this change, we introduce a simple `SkippedRow` class that models just enough attributes to show the users which rows in their data were skipped - whilst there is no interface for this, we want to be able to support it later.

This work is the third in a series, next step: add a service that calls this service to import an entire file of rows.

If you want more context about the longer plan - just ask! :)

https://trello.com/c/lBZkYq5c

### DIFF
--- a/app/services/import/csv/activity_actual_refund_comment/row_service.rb
+++ b/app/services/import/csv/activity_actual_refund_comment/row_service.rb
@@ -1,0 +1,135 @@
+class Import::Csv::ActivityActualRefundComment::RowService
+  attr_reader :errors
+
+  def initialize(report, user, row)
+    @report = report
+    @user = user
+    @activity = nil
+    @errors = {}
+    @row = Import::Csv::ActivityActualRefundComment::Row.new(row)
+  end
+
+  def import!
+    if @row.valid?
+      @activity = find_activity(@row.roda_identifier)
+
+      return false if @activity.nil?
+
+      return false if @row.empty?
+
+      return false unless authorise_activity(@activity)
+
+      create_record
+    else
+      @errors.update(@row.errors)
+      false
+    end
+  end
+
+  private def create_record
+    case type_of_row(@row)
+    when :actual
+      result = create_actual(@row, @activity)
+    when :refund
+      result = create_refund(@row, @activity)
+    when :comment
+      result = create_activity_comment(@row, @activity)
+    end
+
+    return false unless result
+
+    if result.failure?
+      result.object.errors.each do |error|
+        @errors[error.attribute] = [error.attribute, error.message]
+      end
+      false
+    else
+      result.object
+    end
+  end
+
+  private def find_activity(roda_identifier)
+    activity = Activity.find_by_roda_identifier(roda_identifier)
+
+    unless activity
+      @errors.update({"Activity RODA Identifier" => [
+        roda_identifier, I18n.t("import.csv.activity_actual_refund_comment.errors.activity_roda_identifier.not_found")
+      ]})
+    end
+
+    activity
+  end
+
+  private def authorise_activity(activity)
+    policy = ActivityPolicy.new(@user, activity)
+    if !policy.create?
+      @errors.update({"Activity RODA Identifier" => [
+        activity.roda_identifier,
+        I18n.t("import.csv.activity_actual_refund_comment.errors.activity_roda_identifier.not_authorised")
+      ]})
+      return false
+    elsif !reportable_activity?(activity)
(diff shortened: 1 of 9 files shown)


app/services/import/csv/activity_actual_refund_comment/row_service.rb
```diff
@@ -0,0 +1,135 @@
...
+    end
+    true
+  end
+
+  private def reportable_activity?(activity)
+    activity.organisation == @report.organisation && activity.associated_fund == @report.fund
+  end
+
+  private def type_of_row(row)
+    return :actual if row.actual_value.positive? && row.refund_value.zero?
```
CristinaRO, line 86: This might be a preference, but this scenario leaves the situation where the actual_value is negative and the refund_value is zero uncovered.

If a user tries to input a negative value in the new actual form, they get an error message, "Value cannot be negative".

Here this would result in a row whose type cannot be identified, and is therefore ignored without an error message. My preference would be to show an error message if possible, rather than skipping it quietly.

mec: Nice one! 

Yep, I've missed that :( I would like the responsibility for that to belong to `Import::Csv::ActivityActualRefundComment::Row` so I will make a PR for that change to rebase this onto.

mec: @CristinaRO negative actual value validation PR #2303

app/services/import/csv/activity_actual_refund_comment/row_service.rb
```diff
@@ -15,10 +15,10 @@ def import!
 
       return false if @activity.nil?
 
-      return false if @row.empty?
-
       return false unless authorise_activity(@activity)
 
+      return Import::Csv::ActivityActualRefundComment::SkippedRow.new(@row) if @row.empty?
```
CristinaRO, line 20: Would it make sense to use the SkippedRow for unrecognised rows? (Where `type_of_row` is not one of the three expected types.)

Are unrecognised rows surfaced to the user in any way?

mec: We definately could, currently the guard on line 20 means the `create_record` method will not get called if the row considers itself empty and so skipped, this would mean refacotring and renaming - any suggestions?

No, the skipped rows are not surfaced at this point, nothing is because this class is not actuall in use yet, we'll be looking to update the UI later once we and DSIT are happy with the new import and the changes to expectation on users - the loose plan right now is to surface all the types of rows in seperate tables upon successful import, so:

- actuals
- refunds
- activity comments
- skipped rows

But this all come later.

app/services/import/csv/activity_actual_refund_comment/row_service.rb
```diff
@@ -0,0 +1,135 @@
...
+    when :refund
+      result = create_refund(@row, @activity)
+    when :comment
+      result = create_activity_comment(@row, @activity)
+    end
+
+    return false unless result
+
+    if result.failure?
+      result.object.errors.each do |error|
```
CristinaRO, line 42: I'm a little confused about what's happening here, with copying the errors in a slightly different format.

CristinaRO (approved):
I don't want to block this work on working out the details, and I'm keen to get DSIT looking at it on staging so we can get as much early feedback as possible.

app/services/import/csv/activity_actual_refund_comment/row_service.rb
mec: These are the actual model errors which have a slightly different format - it is unlikely there will be any as we have already validate everything, but if at a later date the underlying model validation changes, this would surface the errors.

This is how the current implementation works:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/4f0a9d46d4a457206c29292caaf1b3a26e28f609/app/services/actual/import.rb#L42

mec:
@CristinaRO agree! We can always refine the importer once we know the expectation is met and it is doing everything we want.

PR #2301, opened 2024-01-09
WIP (3006) Produce a list of activities that will need the new transparency ID

CristinaRO:
### After
Screenshot 2024-01-08 at 19 33 47

## Next steps

- [ ] Is an ADR required? An ADR should be added if this PR introduces a change to the architecture.
- [ ] Is a changelog entry required? An entry should always be made in `CHANGELOG.md`, unless this PR is a small tweak which has no impact outside the development team.
- [ ] Do any environment variables need amending or adding?
- [ ] Have any changes to the XML been checked with the IATI validator? See [XML Validation](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/xml-validation.md)

PR #2302, opened 2024-01-09
Export a CSV of all continuing activities

CristinaRO:
## Changes in this PR

The first step in updating activities with the new transparency identifier: producing a list of activities that fulfill the conditions to continue, from the old GB-GOV-13 identifier corresponding to BEIS, to the new GB-GOV-26 identifier corresponding to DSIT.

The list will need to be signed off by DSIT.

If the synchronous request proves to be too much with production amounts of data, we'll go back to the more complex async option drafted in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2301

### After
Screenshot 2024-01-09 at 18 43 31

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   comments
 - the new actual, refund and activity importer no longer accepts negative actual
   values
+- Provide a CSV download of all activities that are likely to continue under DSIT and will need the new DSIT transparency identifier
 
 ## Release 141 - 2023-12-04
 
--- a/app/controllers/exports_controller.rb
+++ b/app/controllers/exports_controller.rb
@@ -74,4 +74,18 @@ class ExportsController < BaseController
     response.stream.write(spending_breakdown_csv)
     response.stream.close
   end
+
+  def continuing_activities
+    authorize :export, :show_continuing_activities?
+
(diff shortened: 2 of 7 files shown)


CristinaRO:
[New DSIT org ID: change the transparency identifier for activities started under BEIS and continuing under DSIT](https://trello.com/c/sPejdG6n/3006-new-dsit-org-id-change-the-transparency-identifier-for-activities-started-under-beis-and-continuing-under-dsit)

mec:
Can we be clear about why we have to do this in the commit message - we cannot be given console access so our only option is to work around like this?

I know why there are no tests, but it makes me uneasy. I think I would like to see something, but if you feel like it is too much effort given the context, we should at least explain why in the commit message?

CristinaRO:
> Can we be clear about why we have to do this in the commit message - we cannot be given console access so our only option is to work around like this?
> 
> I know why there are no tests, but it makes me uneasy. I think I would like to see something, but if you feel like it is too much effort given the context, we should at least explain why in the commit message?

Tests added for one status of "definitely active" activity, one status of "potentially active", and for the "paused" status, which has different criteria.

mec (approved):
🚢

PR #2304, opened 2024-01-10
Feature/2931 actual refund comment csv file import service

shuldt:
## Changes in this PR

Activity actual, refund and comment file import
    
This PR adds a class responsible for taking the uploaded csv data and running through the rows to import the data.
    
It relies on the classes created in previous work to do so,
    
It is concerned with:
    
    - checking the required column headers are present
    - collating the objects that the rows become on import
    - collating any errors from the import process
    - rolling back the database if there are errors

## Next steps


spec/models/import/row_error_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+  describe "#value" do
+    it "returns the supplier value" do
+      expect(subject.value).to eql "The value"
+    end
+  end
+
+  describe "#csv_row" do
```
CristinaRO, line 31: I'd prefer if this method was named `csv_row_number`, to avoid confusion that it might refer to the CSV row being imported.

CristinaRO (approved):

spec/models/import/row_error_spec.rb
mec: Oh, this was named to be compatible with the existing importer and views - I've had to add it back in later! :(

PR #2310, opened 2024-01-15
Download a list of non-continuing ODA activities

CristinaRO:
## Changes in this PR

To help DSIT with confirming the lists of activities that will and won't need to be updated with the new transparency identifier, allow them to export a list of the activities that are assumed to not continue under GB-GOV-26.

The reason we have used the criteria of elimination rather than a new standalone query is that this is how the migration is likely to look like in practice: activities that fulfill the continuation criteria will be updated, and all other activities will be left as they are.

Non-ODA activities are excluded from both, because they are not concerned with transparency reporting to begin with.

### After
Screenshot 2024-01-15 at 15 58 41

## Next steps

### DIFF
--- a/app/controllers/exports_controller.rb
+++ b/app/controllers/exports_controller.rb
@@ -88,4 +88,18 @@ class ExportsController < BaseController
       end
     end
   end
+
+  def non_continuing_activities
+    authorize :export, :show_continuing_activities?
+
+    respond_to do |format|
+      format.csv do
+        export = Export::ContinuingActivities.new
+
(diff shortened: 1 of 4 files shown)


CristinaRO:
[New DSIT org ID: change the transparency identifier for activities started under BEIS and continuing under DSIT](https://trello.com/c/sPejdG6n/3006-new-dsit-org-id-change-the-transparency-identifier-for-activities-started-under-beis-and-continuing-under-dsit)

PR #2313, opened 2024-01-15
(2999) Replace the deprecated call to `Redis.current`

CristinaRO:
## Next steps


CristinaRO:
I think the problem here is configuring the CI env properly, to use a Redis image when testing with the `use_original_rollout` flag, instead of mocking out all real calls to Redis, as we seem to have been doing so far.

However, given the time we've left, I don't to risk spending too much time on this, nor releasing a change without at least one real CI test, so I'm closing this and we'll be locking the gem version instead until we can deal with the deprecation properly.

PR #2315, opened 2024-01-16
(3026) Lock Redis version

CristinaRO:
## Changes in this PR

Redis 5 will deprecate the call to `Redis.current`. We haven't been able to replace this call as of now, so we're locking the gem version to prevent upgrading to an incompatible version.

## Screenshots of UI changes

N/A

## Next steps


CristinaRO:
[Lock Redis version to less than 5](https://trello.com/c/S1eORoGM/3026-lock-redis-version-to-less-than-5)

mec (approved):
🚢

PR #2316, opened 2024-01-16
(3027) Update the factory to use current rules

CristinaRO:
## Changes in this PR

The purpose of this change is to make the factory more realistic, to avoid it being a source of confusion as to how the transparency identifier is generated in RODA.

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/spec/factories/activity.rb
+++ b/spec/factories/activity.rb
@@ -341,8 +341,7 @@ FactoryBot.define do
 
   trait :with_transparency_identifier do
     after(:create) do |activity|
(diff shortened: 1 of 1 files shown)


CristinaRO:
[Update the activity factory to use the current rules for the transparency identifier generation](https://trello.com/c/jysrDBlS/3027-update-the-activity-factory-to-use-the-current-rules-for-the-transparency-identifier-generation)

mec (approved):
🚢

PR #2317, opened 2024-01-16
Release 142

CristinaRO:
- Display the budgets headings as "Activity budgets" for level C and D activities on the Financials tab
- Model a financial value from a CSV file
- Add a new Activity tag, Strategic Allocation Pot (SAP), code 8
- Model a single row of CSV data that contains actual, refund and activity comments
- The new actual, refund and activity importer no longer accepts negative actual values
- Provide a CSV download of all activities that are likely to continue under DSIT and will need the new DSIT transparency identifier
- Add a service that can import a single row of CSV that contains either a Actual, Refund or Activity Comment
- When importing an Activity, Refund or Activity comment, the result can now be a row that was skipped
- Add a service that can import a single row of CSV that contains either a Actual, Refund or Activity Comment
- Update Import Row Error to expose csv row number as expected by the user
- Add a service that can import a file of multiple rows of CSV that contain either an Actual, Refund, or Activity Comment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 [Full changelog][unreleased]
 
+## Release 142 - 2024-01-16
+
+[Full changelog][142]
+
 - Display the budgets headings as "Activity budgets" for level C and D activities on the Financials tab
 - model a financial value from a csv file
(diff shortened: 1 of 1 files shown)


mec (approved):
🚀

PR #2319, opened 2024-01-16
(3021) Exclude health check requests from host auth

CristinaRO:
## Changes in this PR

Exclude health check requests from host authorisation, to allow the AWS deploy pipeline to proceed faster. Currently health check requests are denied by the host auth process, because the IPs are not in the DOMAIN or ADDITIONAL_HOSTNAMES env vars.

We don't have a canonical list of the IPs, nor do we have easy access to the env vars (these particular env vars live in task definitions, not in ParameterStore).

https://api.rubyonrails.org/classes/ActionDispatch/HostAuthorization.html

## Screenshots of UI changes

N/A

## Next steps

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -74,5 +74,7 @@ module Roda
 
     # Default headers
     config.action_dispatch.default_headers["X-XSS-Protection"] = "0"
+
+    config.host_authorization = {exclude: ->(request) { request.path =~ /health(_|-)check/ }}
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 [Full changelog][unreleased]
 
+- Exclude health check requests from host authorisation
+
 ## Release 142 - 2024-01-16
 
 [Full changelog][142]


CristinaRO:
[Get rid of unauthorised host error for AWS health check requests](https://trello.com/c/wGtEwO3O/3021-get-rid-of-unauthorised-host-error-for-aws-health-check-requests)

config/application.rb
```diff
@@ -74,5 +74,7 @@ class Application < Rails::Application
 
     # Default headers
     config.action_dispatch.default_headers["X-XSS-Protection"] = "0"
+
+    config.host_authorization = {exclude: ->(request) { request.path =~ /health_check/ }}
```
mec, line 78: Do we need to include the `/health-check/` path as well?

mec (changes requested):

config/application.rb
CristinaRO: I was counting on the fact that the app responds to both health-check and health_check requests, but we might, not sure.

CristinaRO: I think you're right, the request needs to match, even if it's processed by the same controller. PR updated.

mec (approved):
🚢

PR #2324, opened 2024-01-18
(3023) View updates for new importer

mec:
## Changes in this PR

Now that we have done some testing on the new importer with the DSIT team, we can be more confident to update the view. Whilst the existing view shows a combination of imported Actuals and Refunds, the new view split each type into it's own table:

- Actuals
- Refunds
- Activity comments
- Skipped rows

Skipped rows have quite a different shape to the other types, to help users identify which rows in their file were skipped we introduce the concept of a `ImportedRow`. This lets us store the resulting object along with the row number that sourced it, this is very much the same approach we take for any errrors.

https://trello.com/c/a6CvTA01

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Change the transparency identifier and names for the DSIT organisations (DSIT and DSIT Finance)
 - Add a rake task to change the transparency identifier for activities that will continue under DSIT, and the providing org reference for their actual spend and forecasts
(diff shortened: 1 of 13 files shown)


CristinaRO (approved):

CristinaRO:
Needs rebasing, but otherwise I'd be happy for this to go in, so DSIT can test the latest improvements, and then released to production where it will be behind the feature flag anyway, until they can train users to make the best of the new features.

PR #2327, opened 2024-01-19
Allow service owner to be identified by both IDs

CristinaRO:
## Changes in this PR

To facilitate the transition between the service owner being identified by the old IATI reference `GB-GOV-13` and being identified by the new IATI reference `GB-GOV-26`, allow both identifiers to be used.

This will prevent user permissions from being lost during the short interval between the code being deployed and the migration being run.

## Screenshots of UI changes

N/A
## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Exclude health check requests from host authorisation
 - Remove the feature flag for ODA bulk upload
+- Allow service owner to be identified by both IDs during the transition between BEIS and DSIT
 
 ## Release 142 - 2024-01-16
 
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -101,6 +101,6 @@ class Organisation < ApplicationRecord
   end
 
   def self.service_owner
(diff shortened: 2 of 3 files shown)


CristinaRO:
[Allow the service owner organisation to be identified by either GB-GOV-13 or GB-GOV-26](https://trello.com/c/A80jQoat/3030-allow-the-service-owner-organisation-to-be-identified-by-either-gb-gov-13-or-gb-gov-26)

mec (commented):
I might be missing context here, could we refactor `Organisation.service_owner` to use the `role` attribute and not worry about the IATI ID being invloved at all?

shuldt (approved):

CristinaRO:
> I might be missing context here, could we refactor `Organisation.service_owner` to use the `role` attribute and not worry about the IATI ID being invloved at all?

Good observation, it didn't occur to me!

Definitely something to look at in future, but right now we are very pressed for time. I would want to look at the data and convince myself that the role is used consistently before making something that feels like a bigger refactoring/reworking.

Would you like to add a card to the tech debt backlog?

PR #2335, opened 2024-01-23
Release 143

CristinaRO:
- Support both the new and the old actual importer, allowing switching between
  them with a feature flag
- Exclude health check requests from host authorisation
- Remove the feature flag for ODA bulk upload
- Allow service owner to be identified by both IDs during the transition between BEIS and DSIT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 [Full changelog][unreleased]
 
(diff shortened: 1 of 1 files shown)


CristinaRO:
[Release 143](https://trello.com/c/UxNvj3aq/3032-release-143)

mec (approved):
🚀

PR #2338, opened 2024-01-24
Release 144

mec:
- Change the transparency identifier and names for the DSIT organisations (DSIT and DSIT Finance)
- Add a rake task to change the transparency identifier for activities that will continue under DSIT, and the providing org reference for their actual spend and forecasts
- model a simple imported row so that users can see which row in the csv import
  were skipped
- the Activity actual, refund and comment upload success view now shows the
  imported actuals, refunds, activity comments and skipped rows
- Use an activity's `transparency_identifier` as `iati-identifier` in XML exports, and the `previous_identifier`, if it exists, as `other-identifier`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 [Full changelog][unreleased]
 
+## Release 144 - 2024-01-24
+
+[Full changelog][144]
+
 - Change the transparency identifier and names for the DSIT organisations (DSIT and DSIT Finance)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## Unreleased
```
CristinaRO, line 3: I would prefer having an "unreleased full changelog" as a placeholder, so people don't forget to add it when they start adding things to the changelog.

CristinaRO (approved):
Approving with comment.

CristinaRO:
[Release 144](https://trello.com/c/IyNTnS7i/3035-release-144)

PR #2363, opened 2024-07-18
bumps ruby from 3.0.6 to 3.1.2

shuldt:
## Changes in this PR

Ruby 3.0 is now end-of-life. This PR bumps us to 3.1.2, which is currently maintained. 


app/services/forecast/import.rb
```diff
@@ -1,5 +1,4 @@
 require "csv"
-require "set"
```
yndajas, line 2: Are these no longer required because of the Ruby upgrade?

spec/views/layouts/application_spec.rb
```diff
@@ -1,27 +0,0 @@
...
-  end
-
-  it "does not show the meta tags when ROBOT_NOINDEX is set to false" do
-    ClimateControl.modify ROBOT_NOINDEX: "false" do
-      render
-      expect(response).to_not include("<meta content='noindex' name='robots'>")
-      expect(response).to_not include("<meta content='noindex' name='googlebot'>")
-    end
-  end
-end
```
yndajas, line 27: And what's the reason for this removal?

app/services/forecast/import.rb
shuldt: Yes that's right, this is deleted when setup runs with the new version

spec/views/layouts/application_spec.rb
shuldt: Good question. I will take another look, I'd somehow missed this

shuldt: Hi Ynda - I've now added this file back in. It looks from the git logs like it was deleted when the new ruby version was installed, however I've not been able to nail down exactly why that was. It was last amended two years ago to stop a previous Ruby update from breaking tests with some new style requirements. 

Tests pass both with and without it, so it appears to make no difference, however I would rather not delete a spec file for no good reason.

yndajas (approved):

PR #2382, opened 2024-11-11
Update to Rails 6.1.7.10

mec:
This is the last of the 6.x branch versions. We'll pin to this for now
as we are hoping to update to 7.x later.

This should close a number of security alerts: 
https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/security/dependabot


benshimmin (approved):
LGTM! 👍

PR #2417, opened 2024-11-13
Refactor DownloadLinkMailer

mec:
When we attempted to update to a new version of Ruby, this Mailer
stopped working.

We are not sure why, but switching to positional arguments resolves the
issue, UserMailer uses positional arguments so we assume that this is
acceptable.

Making this refactor opens the door to update Ruby.


benshimmin (approved):
LGTM 👍

PR #2421, opened 2024-11-14
Data migration for IATI and other identifiers

mec:
This data migration fixes the data in the application that was
previously updated in error.

We previously updated the `transparency_identifier` which should be
immutable. This data migration reverts that change.

We then set the new `hybrid_beis_dsit_activity` which makes the
application display 'previous reporting organisation' info in the IATI
xml.

A 'hybrid activity' is one that was started under BEIS but continued
under DSIT and so has to show:

- the previous reporting organisation
- retain the old IATI identifier

There can never be new 'hybrid activities'.

Instructions for running this data migration are in the file.


benshimmin (approved):
This looks great. Very comprehensive and thorough 👍

PR #2423, opened 2024-11-19
Separate /users page into active and inactive users

benshimmin:
## Changes in this PR

This PR separates out the /users page into /users/active and /users/inactive with a tabbed view.

### Before

users-before

### After

users-after

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 [Full changelog][149]
 
 - Rails version updated to 6.1.7.10
+- Separate out users into tabbed view of active and inactive
 
 ## Release 148 - 2024-09-30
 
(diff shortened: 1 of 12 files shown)


mec (changes requested):
Looks solid! I would have gone for `active` and `inactive` actions to keep logic out of the views, but this works. 👍 

I left a couple of comments.

Would you mind making the commit message have a title and be < 50 characters - it's a bit of a dxw thing, something like:

```
Split User index view into active and inactive

We have a potentially large number of users to display at /users. This commit refactors the /users page into a tabbed view at /users/active and /users/inactive which both reduces page load and affords an end user the ability to see a filtered view with either active or inactive users.

By default we display active users by redirecting to /users/active.
```

mec (approved):
Thanks for getting this together! 🚢

PR #2424, opened 2024-11-19
Update CI

mec:
The intention here is to make CI a little faster across workflow runs by utilising the Github action cache.

We start off by moving the test install earlier in the Dockerfile so it can be cached, anything after the application code will not end up being cached.

We rename and update the docker compose file as it is only used in CI at this point.

We create a new workflow that allows for parallelisation of the jobs to save a little time.

Finally we make some changes to make the spellcheck linter run a little bit simpler.

The result is a potential saving of ~2 minutes per CI run.


benshimmin (approved):
This is great work @mec - well done! 👍

PR #2427, opened 2024-11-21
Send coverage to Coveralls

mec:
When updating CI we missed this step. We are not sure anyone actually
uses the Coveralls site, but we where sending it so should continue to.

https://coveralls.io/github/UKGovernmentBEIS/beis-report-official-development-assistance


benshimmin (approved):
LGTM 👍

PR #2428, opened 2024-11-21
Stop compiling assets outside production

mec:
If we use `bin/rails spec` to run our specs in CI we will not need to
  compile the assets - this should save some time building the image for
  CI.
    
  The reason is that the `spec` task includes `test:prepare` which builds
  the assets in test on the fly.


benshimmin (approved):
LGTM 👍

PR #2430, opened 2024-11-22
Update Redis and related

mec:
We are running an old version of the Redis gem that is pinned, this also locks us to an older version of Sidekiq.

This work clears the deprecation warning, moves tests off a fake Redis onto a real instance and then updates both Redis and Sidekiq gems.


benshimmin (approved):
LGTM! 👍

PR #2440, opened 2024-11-27
Fix: Asset compilation

mec:
We recently switched off asset compilation in environments other than
production.

In production, we thought the environment variables required for asset
compilation would be present, when we deployed the change to the
development environment, it turned out that was not the case (we have
no direct control over how the infrastructure is configured).

Returning these stand in variables allows the assets to be compiled, the
variables are not used in the running container and this is how we
approached this issue before the change.


benshimmin (approved):
Env vars look good 👍

PR #2442, opened 2024-11-28
Remove unused Github action workflows

mec:
We cannot see the last time anyone used this feature, so in the spirit
of keeping things simple, we are going to remove them.


benshimmin (approved):
LGTM! 👍

PR #2450, opened 2024-12-04
Fix planned disbursement providing organisation

mec:
The name of the providing organisation does not match the reference
number in a lot of cases, which we assume happened when BEIS became
DSIT.

This migration fixes that.

Let me explain why this looks 'unusual':

The Forecast model is protected against direct access, for background
see our docs:

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/doc/forecasts-and-versioning.md

In this instance the data is simply incorrect, the reference 'GB-GOV-26'
should be named 'DEPARTMENT FOR SCIENCE, INNOVATION AND TECHNOLOGY' and
nothing else. We are not creating new or deleting old Forecast and so
should not upset the versioning.

As we are correcting data we have to touch a lot of Forecasts and the
protection makes this interesting!

For the actual data migration, things are pretty straight forward, we
can use `update_column` to set the name and nothing else.

Testing is where things get more difficult. Due to the protections,
creating a set of Forecasts for testing is extremely convoluted (by
design) we need to put Forecasts into a specific incorrect state.

Instead we've opted to test closer to the database. We've provided a
method to create Forecast in the right state, they would be considered
invalid as models, but that is not under test here, we simply want to
verify that:

- only records in the appropriate state are processed
- only the column we want to change is altered

As it is harder to get the data into the right state to test, we've
included some assertions for the initial state to try and show what is
happening.

I think it is worth remembering that this is a one off fix and is not
really 'production' code that will run day in, day out!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 [Full changelog][unreleased]
 
 - Migration (and logic) to allow users to belong to multiple organisations
+- Data migration to fix forecasts with incorrect providing organisation name
 
 ## Release 154 - 2024-12-05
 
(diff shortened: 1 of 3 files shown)


benshimmin (approved):
Looks great, well done @mec, very thorough!

PR #2451, opened 2024-12-04
Remove Spring and switch to Zeitwerk

mec:
For an application of this size and the kind of hardware developers have
access to, Spring is not required, ahead of an update to Rails 7, which
does not include Spring by default, let's remove it here.

We also want to run the Zeitwerk autoloader ahead of Rails 7 and we only
appear to NOT be using it in the test environment, the reasoning for
which also appears to relate to Spring?!

The only Zeitwerk issue was in the GuidanceUrl spec, which was trying to
load it from a non-existent module - not sure how this was ever working,
but it is now!


benshimmin (approved):
I heartily approve of removing Spring! 👍

PR #2457, opened 2024-12-05
Upgrade to Rails 7.0

mec:
This work updates Rails to 7.0.

Our first step is to fix the custom form builder, moving it and changing how it is configured.

We apply the update and fix a bug with an enum that is revealed in the process.

With 7.0 running, we move on to start migrating the config over to the new defaults, some of the new config will be good to ship now, some we may wait and ship over time, we have added a card to cover this work https://trello.com/c/4uBgh4J0


benshimmin (approved):
This is a great effort @mec - well done! Approved. 👍

PR #2459, opened 2024-12-06
Allow users to have additional organisations

benshimmin:
This creates a `has_and_belongs_to_many` join table relationship between organisations and users, allowing a user to belong to `additional_organisations` and paves the way to allowing a user to set their `organisation` to be a `Current.user_organisation` and thus view data for that organisation instead of their own "real" organisation.

This should change nothing about the existing functionality of the site and will be entirely invisible to end users. (The next phase will include front-end changes which *will* allow selected users to switch their current organisation.)

Helper methods on the `User` model to navigate this relationship include `primary_organisation` (to retrieve the user's actual `organisation`, which is necessarily overwritten), `all_organisations` (the union of `primary_organisation` and `additional_organisations`) and `additional_organisations?` (to determine whether a user has a non-zero number of `additional_organisations`).

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 [Full changelog][unreleased]
 
+- Migration (and logic) to allow users to belong to multiple organisations
+
 ## Release 154 - 2024-12-05
 
 [Full changelog][154]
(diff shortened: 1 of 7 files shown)


mec (approved):
Really nice piece of work @benshimmin!

Would you mind mentioning the `ActiveSupport::CurrentAttributes` and linking the docs in the commit message, then go ahead and merge this!

🚢

PR #2461, opened 2024-12-10
Release 155

mec:
- Migration (and logic) to allow users to belong to multiple organisations
 - Data migration to fix forecasts with incorrect providing organisation name


benshimmin (approved):
Perfect 👍

PR #2466, opened 2024-12-12
Fix welcome email

mec:
The welcome email was set as a Mail Notify gem `view_email` when it is a
`template_email` - we can only assume this worked in old versions of the
gem or we have not been adding new users?

Either way, we started to see Rollbar errors from the Notifications
client and found this.

Switching to a `template_mail` and updating a couple of specs for the
changes fixes things.

We take the time to swap instances of BEIS to DIST here as well.


benshimmin (approved):
LGTM! 👍

PR #2469, opened 2024-12-12
Prepare user model for anonymisation

mec:
This is the work to start to allow us to anonymise users - our version of 'deletion'. We decided not to delete users as we want to retain the history of 'who did what when'. DSIT have decided  that a user who has been deactivated for 5 years can be anonymised, which will remove all personally identifiable information (PII) from a user in the application.

To do so we will need to store when a user was deactivated so we can either anonymise ourselves or let admin users do so.

This first step introduces two date and time fields to user:

- deactivated_at
- anonymised_at

We are taking the `updated_at` timestamp as our best guess at the deactivated_at date and time of existing users as we assume a deactivated user last activity will be the deactivation of their account.

As the state of the user is an important part of the application, I've made this work in three separate commits to keep it clear and have good commit messages, this means the application and test suite are broken until the final commit - we can squash these down if that is preferable (I like it this way).

First we alter the table and migrate the data, we've made the migration reversible just in case.

We then change the User model to use the deactivated_at date to determine is the user is active or not along with updates to spec setups.

Finally we temporarily change how the flow for admin users to deactivate and activate users works - this is temporary as we plan to change the UI for this next, that work is in scope and on our Trello board.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -29,7 +29,7 @@ class UsersController < BaseController
   end
 
   def create
-    @user = User.new(user_params)
+    @user = User.new(user_params.except(:active))
     authorize @user
     @service_owner = service_owner
     @partner_organisations = partner_organisations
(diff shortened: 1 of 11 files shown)


app/controllers/users_controller.rb
```diff
@@ -60,10 +60,11 @@ def update
     @partner_organisations = partner_organisations
 
     reset_mfa = user_params.delete(:reset_mfa)
-    @user.assign_attributes(user_params.except(:reset_mfa))
+    active = user_params[:active] === "true"
```
benshimmin, line 63: @mec I don't _think_ this needs to be triple equals, does it?

mec: No it really shouldn't be! 🙏

benshimmin (approved):
LGTM! 👍

PR #2470, opened 2024-12-13
Set theme colour

mec:
This meta tag is used to set the colour that you see when the browser
'bounces':

https://webkit.org/blog/11989/new-webkit-features-in-safari-15/

it is currently set to `blue` which is pretty distracting!

In the current GOV.UK front end it is set to `#0b0c0c` so we do the
same:

https://design-system.service.gov.uk/styles/page-template/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 [Full changelog][unreleased]
 
+- Theme colour has changed from blue to a black
+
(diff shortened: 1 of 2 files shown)


benshimmin (changes requested):
Let's go for `Theme` rather than `theme` to be consistent with the rest of the CHANGELOG.

benshimmin (approved):

PR #2473, opened 2024-12-13
Front-end work for allowing users to switch organisations

benshimmin:
## Changes in this PR

We have a new front-end component, a dropdown menu organisation switcher, which will be displayed for a user with a non-zero number of `addtional_organisations` and will allow them to switch their organisation. (There is no functionality currently to populate a given user's `additional_organisations` - this would need to be done manually. By default a user will have zero `additional_organisations` and this component will be hidden.)

Once switched, the user will see reports, exports, and activities for that organisation. They can switch back to their original organisation in the same way.

We store the user's switched organisation in the session as `current_user_organisation`, and we also write this into `Current.user_organisation` so we can easily access it in the `User` model where we override the `#organisation` method in order to retrieve the data.

### Before

before1

### After

after1

## Next steps

### DIFF
--- a/app/views/shared/_organisation_switcher.html.haml
+++ b/app/views/shared/_organisation_switcher.html.haml
@@ -1,0 +1,7 @@
+- if authenticated? && current_user.additional_organisations?
+  .govuk-grid-row
+    .govuk-grid-column-full{:class => "govuk-!-padding-top-4"}
+      =form_with url: users_session_organisation_path, method: :patch do |form|
+        =form.label t("organisation_switcher.label"), class: "govuk-label", for: "current_user_organisation"
+        =form.select :current_user_organisation, options_for_select(@organisation_list.pluck(:name, :id), current_user.current_organisation_id), {}, class: "govuk-select"
+        =form.submit t("organisation_switcher.submit"), class: "govuk-button govuk-!-margin-bottom-1", "data-module": "govuk-button"
--- a/config/locales/views/organisation_switcher.en.yml
+++ b/config/locales/views/organisation_switcher.en.yml
@@ -1,0 +1,5 @@
+---
+en:
+  organisation_switcher:
+    label: "Available organisations"
+    submit: "Set organisation" 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -32,6 +32,10 @@ Rails.application.routes.draw do
   resources :users, except: [:index]
   resources :activities, only: [:index]
 
+  namespace :users do
+    patch "session/organisation" => "organisation_session#update", :as => :session_organisation
+  end
+
   roles = %w[implementing_organisations partner_organisations matched_effort_providers external_income_providers]
   constraints role: /#{roles.join("|")}/ do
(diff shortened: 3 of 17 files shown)


spec/features/users_can_edit_a_comment_spec.rb
```diff
@@ -1,4 +1,6 @@
 RSpec.describe "Users can edit a comment" do
+  include HideFromBullet
```
mec, line 2: It would be great to explain why this had to be added in the commit message or even inline in the instance.

app/controllers/organisation_session_controller.rb
```diff
@@ -0,0 +1,15 @@
+class OrganisationSessionController < ApplicationController
```
mec, line 1: Do you think this should live inside the `Users` namespace along with `SessionsController`?

app/controllers/organisation_session_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  include Secured
+
+  def update
+    desired_organisation_id = params[:current_user_organisation]
+
+    if desired_organisation_id
+      if current_user.additional_organisations.pluck(:id).include?(desired_organisation_id) ||
```
mec, line 8: Could this be expressed as?

```
if current_user.all_organisations.pluck(:id).include?(desired_organisation_id)
```

Or extracted to something like `desired_organisation_is_in_additional_organisations?`

config/locales/views/organisation_switcher.en.yml
```diff
@@ -0,0 +1,5 @@
+---
```
mec, line 1: Errant line break here?

app/views/shared/_organisation_switcher.html.haml
```diff
@@ -0,0 +1,7 @@
+- if authenticated? && current_user.additional_organisations?
+  .govuk-grid-row
+    .govuk-grid-column-full{:class => "govuk-!-padding-top-4"}
+      =form_with url: organisation_session_path, method: :patch do |form|
+        =form.label t("organisation_switcher.label"), class: "govuk-label", for: "current_user_organisation"
+        =form.select :current_user_organisation, options_for_select(@organisation_list.pluck(:name, :id), Current.user_organisation || current_user.organisation.id), {}, class: "govuk-select"
```
mec, line 6: It's a shame we have to have this or logic for the selected organisation, I wonder if there is a way to move it somewhere else?

config/routes.rb
```diff
@@ -32,6 +32,8 @@
   resources :users, except: [:index]
   resources :activities, only: [:index]
 
+  patch "/organisation_session" => "organisation_session#update"
```
mec, line 35: How would you feel about giving this a bit of name spacing, maybe `/users/session/organisation` (only because that is where all the Devise routes are)?

mec (changes requested):
It's so cool to see this! 👍  Couple of things that I was wondering about.

mec (approved):
🚢

PR #2477, opened 2024-12-17
Show deactivated time

mec:
This work adds the length of time in words that a deactivated user has been
deactivated for on the users > deactivated index and user show views.

We also change the tab title to make it clear these are:

> Users that have been deactivated by an administrator and not idle users


benshimmin (approved):
LGTM 👍

PR #2478, opened 2024-12-18
Refactor user scopes

mec:
We recently added the length of time a user has been deactivated for on
the deactivated user index, the ordering on this did not work as
expected, this work fixes that and takes the time to refactor how we
approach this.

We create two scopes for use in the views, we've gone for scopes as we
can test those in isolation.

As we have the ordering tested on the scopes, we deleted the old feature
spec that tested the order of the user list.

We kept the call to `policy_scope(User)` although we think it is just
equivalent to `all` as there is no scope so it is inherited.


benshimmin (approved):
LGTM @mec! 👍

PR #2485, opened 2024-12-20
Administrative interface for user additional organisations

benshimmin:
We now have the ability to add additional organisations from the app itself via the create/edit user forms.

This replaces the radio list of BEIS + partner organisations with a dropdown, for the primary organisation, and also adds a checkbox list of partner organisations for the additional organisations. Appropriate explanatory hint text is also provided.

We also have a sprinkling of JavaScript to enhance the form such that a user's additional organisation selections cannot include their primary organisation. As a safeguard for users who do not have JavaScript enabled, we have validation in the model for this too.

### Before

before

### After

after

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -28,12 +28,12 @@ class UsersController < BaseController
   end
 
   def create
-    @user = User.new(user_params.except(:active))
+    @user = User.new(user_params.except(:active, :additional_organisations))
     authorize @user
     @service_owner = service_owner
     @partner_organisations = partner_organisations
 
-    result = CreateUser.new(user: @user, organisation: organisation).call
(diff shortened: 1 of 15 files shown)


app/controllers/users_controller.rb
```diff
@@ -93,6 +94,10 @@ def organisation
     Organisation.find_by(id: organisation_id)
   end
 
+  def additional_organisations
+    user_params[:additional_organisations].reject(&:blank?).map { |id| Organisation.find_by(id:) }
```
rgarner, line 98:
```suggestion
    user_params[:additional_organisations].compact.map { |id| Organisation.find_by(id:) }
```

rgarner (approved):
LGTM, run locally, no issues, structurally fine, one instance of pedantry just to look like I did something 👍🏻

PR #2488, opened 2025-01-06
Fix flaky healthcheck spec

rgarner:
## Changes in this PR

Mild fix for a spec that would fail in dev if you had any unprocessed Sidekiq detritus from other projects. Benefit for devs only, no need for CHANGELOG stuff

## Screenshots of UI changes

None

### Before

Sometimes specs would fail if you had unprocessed enqueued sidekiq stuffs

### After

They no longer fail

## Next steps


benshimmin (approved):
LGTM! 👍

PR #2490, opened 2025-01-08
Release 160

benshimmin:
- Admin interface to enable additional organisations


rgarner (approved):
Could have rhymed better but I'll let it slide

PR #2491, opened 2025-01-09
Upgrade to rails 7.0 defaults from 6.1

rgarner:
## Changes in this PR

Upgrades to Rails defaults; whittle away at `config/initializers/new_framework_defaults_7_0.rb` comment by comment, with a commit for each explaining the change and the likely effect on the application. Have pushed for (nearly) every commit here and monitored the builds; see below.

Finally, delete `config/initializers/new_framework_defaults_7_0.rb` and set `config.load_defaults = 7.0`

Possibly a CHANGELOG.md entry is justified here such that in the unlikely event that any behavioural changes are found (perhaps a spot check of dates in UI and exports would be wise due to https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2491/commits/2f727714be2d4c9051f033db119819f49a211a85) we can tie it to the release with notes.

## Screenshots of actions demonstrating no adverse effect at each commit

image


benshimmin (approved):
Great work @rgarner! LGTM 👍

PR #2493, opened 2025-01-10
Change deactivate/reactivate interactions

benshimmin:
Here we move the deactivate/reactivate a user functionality into a separate pair of pages, with explanatory and confirmatory text on each page. We continue using the same `update` method on the Users controller and just pass the active state through.

### Before

activate_before

### After

activate_after

## Next steps

### DIFF
--- a/app/views/users/_form.html.haml
+++ b/app/views/users/_form.html.haml
@@ -26,10 +26,10 @@
     - @partner_organisations.each do |dp|
       = f.govuk_check_box :additional_organisations, dp.id, label: { text: dp.name }, checked: @user.additional_organisations.include?(dp)
 
-  = f.govuk_collection_radio_buttons :active,
-      user_active_options,
-      :id,
-      :name,
-      legend: { tag: :h2 }
-
-  = f.govuk_submit t("form.button.user.submit")
+  .govuk-button-group
+    = f.govuk_submit t("form.button.user.submit")
+    - if !@user.new_record?
+      - if @user.active?
+        = link_to t("form.button.user.deactivate"), deactivate_user_path, class: "govuk-button govuk-button--secondary"
+      - else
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -29,6 +29,8 @@
 
   get "/users", to: redirect("/users/active")
   get "/users/:user_state" => "users#index", :constraints => {user_state: /(in)?active/}, :as => "users_index"
+  get "/users/:id/deactivate" => "users#deactivate", :as => "deactivate_user"
+  get "/users/:id/reactivate" => "users#reactivate", :as => "reactivate_user"
   resources :users, except: [:index]
```
rgarner, line 34: You might get away with slightly less of the `:as` and `=>` here:

```suggestion
  resources :users, except: [:index] do
    member do
      get :deactivate
      get :reactivate
    end
  end
```

You might *also* be able to do something with `/users/:user_state` to simplify, but that existed before this PR so 😴 🐶 and all that...

rgarner (commented):
Looks good! I think we need to move some things to translations, but everything else here is optional.

Also this led me to look at `UpdateUser` a little bit, and it's not related to this PR, but what happens when `user.save` fails? I'd be tempted maybe to drop a bang in there so at least we find out about it noisily, but that's out of scope here.

app/views/users/_form.html.haml
```diff
@@ -26,10 +26,10 @@
...
-  = f.govuk_collection_radio_buttons :active,
-      user_active_options,
-      :id,
-      :name,
-      legend: { tag: :h2 }
-
-  = f.govuk_submit t("form.button.user.submit")
+  .govuk-button-group
+    = f.govuk_submit t("form.button.user.submit")
+    - if !@user.new_record?
```
rgarner, line 31: Oh, darn, forgot to say you could have used the antonym `if @user.persisted?` here

rgarner: L'esprit d'escalier

PR #2497, opened 2025-01-14
3076 Clear Rails 7.1 deprecation warnings

rgarner:
## Changes in this PR

Errors such as

```
DEPRECATION WARNING: Passing instance variables to `render` is
deprecated.
In Rails 7.1, @activities, @activity, @current_user, @implementing_organisation_presenters, and @tab_name will be ignored.
 (called from block (2 levels) in show at /Users/russ/dev/roda/app/controllers/activities_controller.rb:52)
```

block us from going to Rails 7.1. 

Our offences are confined to places where we've used `Activity::Tab`.
This class sends a message to itself in order to set a list of
`@instance_variables` for the given tab. The `ActivitiesController` then
uses `Activity::Tab#locals` to copy a hash of all the instance variables
in the instance of the `Tab` with keys like

`{ :@tab_name => 'financials', :https://github.com/actuals => [...] }`

to the locals of the template that is about to be rendered. The most
surprising behaviour here is that the `@activity` that is set in the
controller ends up being overwritten by the `@activity` in the
`Tab#locals` method, which is actually an `ActivityPresenter` – a
decorator that shares some identical method names like `total_spend`
which have view-specific behaviours in the Presenter version. We need to
keep defaulting to Presenter.

Make `Tab#locals` pass vanilla keys like `:activity`, and stop using
`@instance_variable`s in `app/view/activities/|tab|.html.haml` templates.
e.g. `activity.total_spend` instead of `@activity.total_spend`

## Next steps

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -8,6 +8,8 @@ class ActivitiesController < BaseController
   after_action :skip_policy_scope, only: [:index, :historic]
   skip_before_action :clear_breadcrumb_context, only: [:show]
 
+  helper_method :current_tab
+
   def index
     @organisation = Organisation.find(organisation_id)
 
--- a/app/services/activity/tab.rb
+++ b/app/services/activity/tab.rb
@@ -23,9 +23,10 @@ class Activity
     end
 
     def locals
-      instance_variables.map { |attribute|
-        [attribute, instance_variable_get(attribute)]
-      }.to_h
+      instance_variables.each_with_object({}) do |attr, hash|
+        key = attr.to_s.delete_prefix("@").to_sym
+        hash[key] = instance_variable_get(attr)
+      end
     end
 
     def template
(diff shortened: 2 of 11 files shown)


rgarner:
Parking this here @benshimmin – I think I might have changed my mind about `@activity` and want to change it to the presenter everywhere, largely because if we don't, `@activity` is an `Activity` and `activity` would be an `ActivityPresenter` and I think that's a confusion bomb far bigger than @-as-orientation...

rgarner:
(but we might notice there's a Better Way tomorrow and be able to negotiate around it a bit)

app/controllers/activities_controller.rb
```diff
@@ -49,6 +49,7 @@ def show
           tab_name: current_tab
         )
         @commitment = CommitmentPresenter.new(@activity.commitment)
+        @tab_name = current_tab
```
rgarner, line 52: I wonder in fact if I shouldn't just make a `helper_method :current_tab`, drop this line and use `current_tab == tab` in https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/12bba9f0ea1c4748ccc3b3f77500a45baf9d5519/app/views/shared/activities/_tab_nav_item.html.haml#L1

rgarner: Yep, that does indeed feel substantially better

benshimmin (approved):
This looks great @rgarner - good work. 👍

PR #2499, opened 2025-01-16
Functionality to anonymise deactivated users

benshimmin:
Here we add functionality to anonymise an already deactivated user.

An anonymised user has:
- a name which is simply "Deleted User <database id>"
- an email address which is "deleted.user.<database id>@<original domain name>"
- no telephone number

In the frontend, an admin user will see an "Anonymise" button when editing an inactive user; this leads to a confirmation screen which submits a form to UsersController#anonymise_update, which calls the AnonymiseUser service object.

### After

anonymise

## Next steps


app/controllers/users_controller.rb
```diff
@@ -97,6 +97,28 @@ def reactivate
...
+    @user = User.find(id)
+    authorize @user
+
+    result = AnonymiseUser.new(user: @user).call
+
+    if result.success?
+      flash[:notice] = t("action.user.update.success_anonymised")
+      redirect_to user_path(@user)
+    else
+      render :anonymise
```
benshimmin, line 118: @rgarner Perhaps we ought to put an error in `flash` here...

spec/services/anonymise_user_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+RSpec.describe AnonymiseUser do
+  let(:user) { create(:administrator) }
+
+  describe "#call" do
+    it "returns a successful result" do
+      result = described_class.new(user: user).call
+
+      expect(result.success?).to be(true)
+      expect(result.failure?).to be(false)
```
rgarner, line 11: I guess this is local customs. I'd usually write

```suggestion
      expect(result).to be_success
      expect(result).not_to be_failure
```

app/views/users/_form.html.haml
```diff
@@ -33,3 +33,4 @@
         = link_to t("form.button.user.deactivate"), deactivate_user_path, class: "govuk-button govuk-button--secondary"
       - else
         = link_to t("form.button.user.reactivate"), reactivate_user_path, class: "govuk-button govuk-button--secondary"
+        = link_to t("form.button.user.anonymise"), anonymise_user_path, class: "govuk-button govuk-button--warning"
```
rgarner, line 36: From the card, it seems this should only appear for deactivated users.

app/services/anonymise_user.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def initialize(user:)
+    self.user = user
+  end
+
+  def call
+    result = Result.new(true)
+
+    User.transaction do
+      user.anonymised_at = DateTime.now
```
rgarner, line 12: Possibly a guard `raise` needed to protect in the circumstance that the user is not deactivated.

spec/features/beis_users_can_edit_a_user_spec.rb
```diff
@@ -101,6 +101,24 @@
     expect(user.reload.active).to be(true)
   end
 
+  scenario "an inactive user can be anonymised" do
```
rgarner, line 104: Probably needs the counter-spec that an active user doesn't even see the operation?

spec/services/anonymise_user_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+
+RSpec.describe AnonymiseUser do
+  let(:user) { create(:administrator) }
+
+  describe "#call" do
+    it "returns a successful result" do
+      result = described_class.new(user: user).call
```
rgarner, line 8:
```suggestion
      result = described_class.new(user:).call
```

spec/services/anonymise_user_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      expect(result.success?).to be(true)
+      expect(result.failure?).to be(false)
+    end
+
+    it "anonymises the email address" do
+      email = user.email
+      described_class.new(user:).call
```
rgarner, line 16: I tend to use `described_class` in shared examples or when the `ClassName::Is::ExceptionallyLong::AndWouldPushEverythingOffThePage`. I think:

- specs tend to be improved for getting the point home by repeating the `DoThingToThing` class name
- this is not a hill I would die on, `described_class` is perfectly acceptable, and it's a thing I like to discuss in passing

app/controllers/users_controller.rb
rgarner: It wouldn't get here because of the bang in the anonymise class? Review half done!

spec/services/anonymise_user_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      expect(user.email).not_to eql(email)
+
+      # Domain element of email address should be unchanged:
+      expect(user.email.split("@").last).to eql(email.split("@").last)
```
rgarner, line 20: You might like this form for English-style readability, depending on your personal preferences. I'm happy with what's here so this is more of an "ooh what if" from my brain which gets some of its dopamine from writing left-to-right RSpec and hoping other people also like it

```ruby
      the_email_domain = user.email.split("@").last
      expect { AnonymiseUser.new(user:).call }.to change(user, :email).from(user.email).to(
        a_string_ending_with(the_email_domain)
      )
```

app/views/users/_form.html.haml
rgarner: Completely missed the if here because above the fold. Sorry @benshimmin, resolving
image

spec/services/anonymise_user_spec.rb
benshimmin: This is very clever but I honestly find mine a bit easier to read 😝

rgarner: Fair. I wasn't going for clever; I found the `split("@").last` difficult to parse. Which could say something more about me. It's not a deal-breaker.

spec/features/beis_users_can_edit_a_user_spec.rb
benshimmin: We don't have a counter-spec for this, but we do now have code in the service object that stops it happening which is probably enough.

app/models/user.rb
```diff
@@ -15,8 +15,8 @@ class User < ApplicationRecord
     organisation_id: :organisation
   }.freeze
 
-  scope :active, -> { where(deactivated_at: nil) }
-  scope :deactivated, -> { where.not(deactivated_at: nil) }
+  scope :active, -> { where(deactivated_at: nil, anonymised_at: nil) }
+  scope :deactivated, -> { where(anonymised_at: nil).where.not(deactivated_at: nil) }
```
rgarner, line 19: I don't think I'd understood that anonymised users would disappear from both tabs before. Makes sense. Resolve this conversation if this assumption is correct 😃

config/locales/models/user.en.yml
```diff
@@ -95,6 +98,28 @@ en:
...
+      anonymise:
+        title: Anonymise user?
+        content_html: |
+          <p class="govuk-body">
+            You are about to anonymise the user %{email}.
+          </p>
+
+          <p class="govuk-body">
+            Doing so will prevent the user from signing in to the application.
+          </p>
```
rgarner, line 110: I suspect this para is redundant; if we're anonymising, we are by definition doing so from the Deactivated tab and consequently they're already unable to sign in?

rgarner (approved):
One possible redundancy in the anonymise blurb, but otherwise looking fine 🚢

PR #2501, opened 2025-01-16
5.x devise-two-factor upgrade phase 1

rgarner:
## Post-release

Run:

`rails runner db/data/20250117151047_regenerate_otp_secrets.rb` on each environment in turn.

`User#legacy_otp_secret` will cover the period that "new" otp_secrets are NULL.

## Phase 2 Cleanup release

We can remove the old DB columns and `legacy_otp_secret` in a separate release:

https://github.com/devise-two-factor/devise-two-factor/blob/main/UPGRADING.md#phase-2-clean-up

In our case we won't need to copy otp_secret; because we're SMS-only we're able to do that in the Phase 1 release data
migration.

NOTES:

- given that this is the first commit to `schema.rb` since the upgrade, some bits like `precision` go away with the
  Rails 7.0 upgrade

## Next steps

See the Plan checklist on https://trello.com/c/sXsThu5A/3099-upgrade-devise-two-factor-gem-from-4x-to-5x


app/models/user.rb
```diff
@@ -74,4 +74,74 @@ def email_cannot_be_changed_after_create
       errors.add(:email, :cannot_be_changed)
     end
   end
+
```
benshimmin, line 77: I think probably worth including in a comment at the top of this method that we've lifted this code from https://github.com/devise-two-factor/devise-two-factor/blob/main/UPGRADING.md

rgarner: After discussion, we note that this comment was meant for `legacy_otp_secret`

benshimmin (approved):
Great work @rgarner, this looks really tidy. _Jazda!_ 👍

PR #2502, opened 2025-01-17
Fix admin user switching issue

benshimmin:
This fixes an issue where if an admin user switches their organisation, the user list view shows the switched organisation rather than each user's actual organisation. We include a test to reflect that this behaviour is now correct.

## Next steps

### DIFF
--- a/spec/features/users_can_switch_organisation_spec.rb
+++ b/spec/features/users_can_switch_organisation_spec.rb
@@ -101,6 +101,48 @@ RSpec.feature "Users can switch organisation" do
     end
   end
 
+  context "when a BEIS user has switched organisation" do
+    let(:user_with_additional_organisations) do
+      user = create(:beis_user)
+      user.additional_organisations << [
+        create(:partner_organisation),
+        create(:partner_organisation),
+        create(:beis_organisation)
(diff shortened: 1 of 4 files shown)


spec/features/users_can_switch_organisation_spec.rb
```diff
@@ -101,6 +101,47 @@
...
   end
 
+  context "when a BEIS user has switched organisation" do
+    let(:user_with_additional_organisations) do
+      org1 = create(:partner_organisation)
+      org2 = create(:partner_organisation)
+      org3 = create(:beis_organisation)
+      user = create(:beis_user)
+      user.additional_organisations << [org1, org2, org3]
+      user
```
rgarner, line 111:
```suggestion
      user = create(:beis_user)
      user.additional_organisations << [
        create(:partner_organisation), 
        create(:partner_organisation), 
        create(:beis_organisation)
      ]
      user
```

This is about the only thing I'd be tempted to do, just so my brain doesn't try and equate the `org1` and `org2` up here with the `org1` and `org2` down there

rgarner (approved):
Swift 🚀

PR #2506, opened 2025-01-23
Level B exports

rgarner:
## Changes in this PR

Level B exports for DSIT users

## Screenshots of UI changes

image

## Post-deploy tasks

Enable this for specific users only using `rails c` in the relevant environments, e.g.:

```
ROLLOUT.activate_user(:level_b_exports, User.find_by!(email: 'daniel.sparks@odamanagement.org'))
```

(or use the Rollout UI, one email per line)

Sam will also need access, and possibly Martha (Dan to confirm).

Note that `level_b_exports` as a named feature is only currently in `production`, so you may need to visit the [Rollout UI](https://report-official-development-assistance.service.gov.uk/rollout/) for other environments.

## Notes

Linked activities [were assured](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/commit/5ff62be73726bfa66bf7a37d30602ef9e21898e2) but damaged spec readability for the edgiest of edge cases. Given that there's no special logic around presenting linked activities at all (straight from the DB, for one cell), we removed that commit immediately.

## Next steps


app/presenters/activity_csv_presenter.rb
```diff
@@ -113,6 +113,17 @@ def publish_to_iati
     translate("boolean.#{super}")
   end
 
+  def sustainable_development_goals
+    if sdgs_apply == false && step_is_complete_or_next?(activity: self, step: :sustainable_development_goals)
+      "Not applicable"
```
rgarner, line 118: There isn't a translation for this; there wasn't one in the base class either. I could elide this by saying `super`, or I could introduce it Somewhere, but there's already an `N/A` localised string, and didn't want to overload that

app/views/exports/index.html.haml
```diff
@@ -50,6 +50,29 @@
                 - else
                   = a11y_action_link("Request", spending_breakdown_exports_path(fund_id: fund.id), t("table.export.spending_breakdown.name", fund: fund.name), ["govuk-link--no-visited-state"])
 
+      - if ROLLOUT.active?(:level_b_exports, current_user)
```
benshimmin, line 53: This is becoming quite a big old view with a pile of _stuff_ in it. I wonder if we could have a partial for "Level B exports" and a partial for "Ad hoc exports" (at the risk of touching something slightly outside the remit of this PR?). Could we perhaps have the `ROLLOUT.active...` logic in the controller? Then you might end up with:

```
- if @current_user_can_view_level_b_exports
  = render :partial => "level_b_exports"

= render :partial => "ad_hoc_exports"
```

Or some such (with some due caution taken over any locals!). Not a hill I'd die on, just a suggestion.

spec/features/beis_users_can_download_exports_spec.rb
```diff
@@ -291,4 +291,84 @@
...
+    aggregate_failures do
+      # Then I should have downloaded a file named for the fund and those activities with an appropriate timestamp in the filename
+      expect(page.response_headers["Content-Disposition"]).to include(
+        "attachment; filename=LevelB_International_Science_Partnerships_Fund_2025-01-21_11-26-34.csv"
+      )
+
+      document = CSV.parse(page.body.delete_prefix("\uFEFF"), headers: true).map(&:to_h)
+      expect(document.size).to eq(2)
+
+      # And each row should <meet data acceptance criteria>
```
benshimmin, line 326: This reads like we're expecting to receive `<data acceptance criteria>` at some future point - are we? 😁

spec/presenters/activity_csv_presenter_spec.rb
```diff
@@ -439,4 +439,61 @@
...
+      let(:activity) { build(:project_activity, sdgs_apply: true, sdg_1: 5) }
+      it { is_expected.to eql "Gender Equality" }
+    end
+
+    context "there are multiple SDGs, return their names, separated by a semicolon" do
+      let(:activity) { build(:project_activity, sdgs_apply: true, sdg_1: 5, sdg_2: 1) }
+      it { is_expected.to eql "Gender Equality; No Poverty" }
+    end
+
+    context "when there multiple nil SDGs" do
```
benshimmin, line 480: Missing the word "are" I think.

spec/presenters/activity_csv_presenter_spec.rb
```diff
@@ -439,4 +439,61 @@
       expect(result).to eq("Resilient Planet|Transformative Technologies|Healthy People, Animals & Plants")
     end
   end
+
+  describe "#publish_to_iati" do
+    subject { described_class.new(activity).publish_to_iati }
+
+    context "the activity does not publish to iati" do
```
benshimmin, line 446: I feel like "iati" deserves to be uppercase (this may not be very consistent through the rest of the specs though).

spec/presenters/activity_csv_presenter_spec.rb
```diff
@@ -439,4 +439,61 @@
...
+  describe "#tags" do
+    subject { described_class.new(activity).tags }
+
+    context "the activity has no tags" do
+      let(:activity) { build(:project_activity, tags: []) }
+      it { is_expected.to be_nil }
+    end
+
+    context "the activity has some tags from the codelist in tags.yml" do
+      let(:activity) { build(:project_activity, tags: [4, 5]) }
```
benshimmin, line 495: We add some `tags` in both this spec and in a previous one; perhaps that could be a trait in the activity fixture?

app/services/export/activities_level_b.rb
```diff
@@ -0,0 +1,108 @@
...
+      activity.benefitting_countries,           # "Benefitting countries",
+      activity.benefitting_region,              # "Benefitting region",
+      activity.gdi,                             # "Global Development Impact",
+      activity.sustainable_development_goals,   # "Sustainable Development Goals",
+      activity.ispf_themes,                     # "ISPF themes",
+      activity.aid_type,                        # "Aid type",
+      activity.oda_eligibility,                 # "ODA eligibility",
+      activity.publish_to_iati,                 # "Publish to IATI?",
+      activity.tags,                            # "Tags",
+      budgets_by_year[2023]&.value,             # "Budget 2023-2024",
```
benshimmin, line 94: `(2023..2028).map { |y| budgets_by_year[y]&.value }` (with`.flatten` on the end of the array) perhaps!

benshimmin (commented):
This is a great piece of work @rgarner especially given our difficulties accessing the actual data; it should be fun to run this in production and see what comes out! Well done for exploring how previous exports have been done and adapting that - fantastic archeology!

My comments are mostly just nitpicking and stylistic tweaks more than anything - I think the functionality itself looks really solid, though of course the proof of the pudding will be in the eating.

spec/presenters/activity_csv_presenter_spec.rb
rgarner: For me, it needs to be somewhat reusable to introduce a `trait`, which is really part of a dev-only API. I didn't want to introduce a [Mystery Guest](https://thoughtbot.com/blog/mystery-guest#mystery-guest) here and then have it propagate, though recalling the work it took to figure out what a tag *was*, maybe a `with_tags` wouldn't be entirely terrible

app/services/export/activities_level_b.rb
rgarner: I did have that! However, I prefer the repetition. The Gestalt branch of psychology makes grouping this stuff very easy, and the shape of it resembles the output of the CSV itself. There's one line per column at the minute and I find that helpful at 4.30 on a Friday 😉

spec/features/beis_users_can_download_exports_spec.rb
rgarner: That was me waving my hands until I'd understood every field. And before I'd discovered `ActivityCsvPresenter`, which significantly de-risked bits.

app/views/exports/index.html.haml
rgarner: I'm going on prior art; this stuff is already all over views, so I didn't want to make any kind of exception.

app/services/export/activities_level_b.rb
```diff
@@ -0,0 +1,108 @@
...
+  def initialize(fund:)
+    @fund = fund
+  end
+
+  def headers
+    HEADERS
+  end
+
+  def filename
+    "LevelB_#{@fund.name.tr(" ", "_")}_#{Time.zone.now.strftime("%Y-%m-%d_%H-%M-%S")}.csv"
```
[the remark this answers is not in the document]
rgarner: > I'm fine with the lack of the translation (hardly seems necessary in this case!) but I'm struggling more with sdgs_apply == false rather than !sdgs_apply; I can see the same incantation in the super class but I can't immediately understand why we'd write it like this!

I believe it's because we need to leave `nil` alone; `sdgs_apply` is a straight DB field. Nonetheless, I've cargo-culted it because it shouldn't appear to be different – or behave differently from – `ActivityPresenter` and it'd be more time to work out why and fix in a class that's unrelated to this PR.

Gemfile
```diff
@@ -83,6 +83,7 @@ group :test do
   gem "rails-controller-testing"
   gem "shoulda-matchers"
   gem "selenium-webdriver"
+  gem "super_diff"
```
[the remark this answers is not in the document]
rgarner: Does 

> - Add the super_diff gem such that we can see at a glance what fields
  we're getting wrong in `beis_users_can_download_exports_spec.rb`

not do it for you?

app/services/export/activities_level_b.rb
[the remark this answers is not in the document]
rgarner: > Golfing slightly but I think you can say strftime("%F_%H-%M-%S") here (you could do %F_%T if you were content with colons instead of dashes in the time bit).

I have to look up date/time format strings every time I use them, and I wouldn't be as confident reading `%F` as having it spelled out for me...

Gemfile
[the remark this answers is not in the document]
benshimmin: I missed that! If you can include that in the squashed commit message that'd be perfect. 👍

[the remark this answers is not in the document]
rgarner: Got to be honest, I don't typically squash, and I'm not sure local customs here do either? Although this one is a bit front-heavy so I don't mind doing it in this instance.

app/presenters/activity_csv_presenter.rb
rgarner: I believe it's because we need to leave nil alone; sdgs_apply is a straight DB field. Nonetheless, I've cargo-culted it because it shouldn't appear to be different – or behave differently from – ActivityPresenter and it'd be more time to work out why and fix in a class that's unrelated to this PR.

spec/presenters/activity_csv_presenter_spec.rb
rgarner: I'm not sure I want to do this one, because it took me some time to figure out that tags were not some `['folksonomic', 'list']`, so I wanted literal numbers with some accompanying blurb in the context that makes people unfamiliar with tags sit up and go "oh, right they're just numbers, and there's the YML in which they're defined". A layer of `:with_tags` trait symbol indirection in that (which would be fine for me, as I can just CMD-B and RubyMine knows what a trait is and goes there, but would still be indirection) would make that a bit harder for the next person.

spec/features/beis_users_can_download_exports_spec.rb
rgarner: Have made suitably broad

app/views/exports/index.html.haml
rgarner: By which I mean *the `ROLLOUT`* stuff, but wasn't even vaguely clear† on; have extracted four partials, which may be a little out of remit, but...

†oxymoron klaxon

rgarner: Ok, left the `ROLLOUT` stuff in the `index.html.haml` because that's fairly decent and obvious `ALL_CAPS` signal, but extracted the partials as suggested, for a very svelte index.html.haml.

benshimmin (approved):
Superb work! 💪

PR #2508, opened 2025-01-24
Activity import: prevent duplicates

rgarner:
Pipe-delimited values were allows to have duplicate values on import, meaning for example that countries would appear twice in an activity's details, e.g. "Brazil, Egypt and Egypt".

This was the only visible duplication on the surface; however, it applied to any pipe-delimited field on import and they were stored with dupes (even though some views elided that duplication).

Update the converter in all the places required. There *could* have been a method for this, but for convert_country_partner_organisations that requires a `strip`. Kept the duplication. Ironically.

## Next steps


rgarner:
We think this is bobbins and they want effective row-level validation errors. Closing.

PR #2515, opened 2025-01-31
Brakeman is complaining about our EOL version of Rails 7.0.8.7.

benshimmin:
We'll be upgrading Rails to 7.1.x shortly so let's temporarily disable the EOLRails warning in Brakeman, and re-enable it when we're updating to Rails 7.1.x.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,13 +1,32 @@
 {
   "ignored_warnings": [
     {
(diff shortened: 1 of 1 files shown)


rgarner:
I've put a note in https://trello.com/c/dFuUhB3t/3077-upgrade-to-rails-71

rgarner (approved):

PR #2517, opened 2025-02-03
Anonymise users background job

benshimmin:
This gives us a simple background job which anonymises users who have been deactivated for more than 5 years. We introduce the sidekiq-scheduler gem for the purpose of scheduling jobs in a cron-like manner.

(Also worth noting that the sidekiq_scheduler_spec basically just checks for typos in the Sidekiq schedule config, in case we ever need to tweak that again or add additional jobs.)

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 [Full changelog][unreleased]
 
+- Anonymise users background job
+
 ## Release 165 – 2025-02-03
 
 [Full changelog][165]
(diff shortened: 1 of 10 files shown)


rgarner (commented):
Looks good, though I think we need to tick an item on the checklist...

rgarner (approved):
C'est magnifique, mais ce n'est pas la gu– attends, non, excusez moi, c'est la guerre 🔫

PR #2518, opened 2025-02-03
Level B reporting fixes

rgarner:
## Changes in this PR

Note that these are separate changes to the budget window changes, which are opened in a [separate card](https://trello.com/c/FGbAceUU/3116-for-each-remaining-fund-show-historical-budgets-when-they-exist)

- Fix IATI identifier coming through as GB-GOV-26 only
- Separate comments by "|"
- For GCRF, remove "ODA or Non-ODA", "ISPF ODA partner countries",  
  "ISPF themes", "Tags"
- For GCRF, add "GCRF Strategic Area", "GCRF Challenge Area"
- For ISPF, add "ISPF non-ODA partner countries"
- For Newton, remove "ODA or Non-ODA", "ISPF ODA partner countries", 
  "ISPF themes", and "Tags"
- For Newton, add "Newton Fund Country Partner Organisations", 
  "Newton Fund Pillar"
- For Other ODA, remove "ODA or Non-ODA", "ISPF ODA partner countries", 
  "ISPF themes", "Tags"

## Next steps

### DIFF
--- a/app/services/export/activities_level_b.rb
+++ b/app/services/export/activities_level_b.rb
@@ -97,7 +97,7 @@ class Export::ActivitiesLevelB
       budgets_by_year[2026]&.value,             # "Budget 2026-2027",
       budgets_by_year[2027]&.value,             # "Budget 2027-2028",
       budgets_by_year[2028]&.value,             # "Budget 2028-2029",
-      activity.comments.map(&:body).join("\n")  # "Comments"
+      activity.comments.map(&:body).join("|")   # "Comments"
     ]
   end
 
--- a/app/services/export/activities_level_b.rb
+++ b/app/services/export/activities_level_b.rb
@@ -69,7 +69,7 @@ class Export::ActivitiesLevelB
       activity.is_oda,                          # "ODA or Non-ODA",
       activity.partner_organisation_identifier, # "Partner organisation identifier",
       activity.roda_identifier,                 # "RODA identifier", e.g. GCRF-LCXHF
-      activity.organisation.iati_reference,     # "IATI identifier",
+      activity.transparency_identifier,         # "IATI identifier",
       activity.linked_activity_identifier,      # "Linked activity",
       activity.title,                           # "Activity title",
       activity.description,                     # "Activity description",
--- a/app/services/export/activities_level_b.rb
+++ b/app/services/export/activities_level_b.rb
@@ -1,48 +1,91 @@
 class Export::ActivitiesLevelB
-  HEADERS = [
-    "Partner Organisation",
-    "Activity level",
-    "Parent activity",
-    "ODA or Non-ODA",
-    "Partner organisation identifier",
-    "RODA identifier",
-    "IATI identifier",
-    "Linked activity",
-    "Activity title",
-    "Activity description",
-    "Aims or objectives",
-    "Sector",
-    "Original commitment figure",
-    "Activity status",
-    "Planned start date",
-    "Planned end date",
-    "Actual start date",
-    "Actual end date",
-    "ISPF ODA partner countries",
-    "Benefitting countries",
-    "Benefitting region",
-    "Global Development Impact",
-    "Sustainable Development Goals",
-    "ISPF themes",
-    "Aid type",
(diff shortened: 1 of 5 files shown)


spec/services/export/activities_level_b_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+          {
+            "Parent activity" => "International Science Partnerships Fund",
+            "ODA or Non-ODA" => "ODA",
+            "ISPF ODA partner countries" => "India (ODA)",
+            "ISPF non-ODA partner countries" => "India (non-ODA)",
+            "ISPF themes" => "Resilient Planet",
+            "Tags" => "Tactical Fund|Previously reported under OODA"
+          }.reverse_merge(common_expected_values)
+        )
+        expect(first_row.length).to eql(36)
```
benshimmin, line 108: Throughout this file - which is for the most part noble in reason and infinite in faculties - there are a few magic values for the array lengths: 36, 31, 33, etc. Perhaps it would be nicer to have these as constants like `NUMBER_OF_COLUMNS_IN_<REPORT_VARIANT> = 36` which would then be easier to deal with if we ever changed these reports.

spec/services/export/activities_level_b_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    context "fund is ISPF" do
+      let(:fund) { Fund.by_short_name("ISPF") }
+
+      it "has ISPF-only columns" do
+        expect(headers).to include("ODA or Non-ODA")
+        expect(headers).to include("ISPF ODA partner countries")
+        expect(headers).to include("ISPF themes")
+        expect(headers).to include("Tags")
+      end
```
benshimmin, line 21: This is a bit repetitive (we check this list three times variously) but I think the intent is very clear so I probably wouldn't bother turning it into an array.

app/services/export/activities_level_b.rb
```diff
@@ -1,48 +1,91 @@
...
+
+  # A place to:
+  #   - Name all the fields on the left
+  #   - filter applicable fields on a per-fund basis in the middle
+  #   - evaluate a row's values in the context of a fund's activity via a `value` Proc on the right
+  #   - show all this on a line-by-line basis to avoid one tall export per fund, given
+  #     there are many common fields
+  # standard:disable Layout/ExtraSpacing
+  FIELDS = [
+    Field.new("Partner Organisation",                      "ALL",  -> { activity.organisation.name }),
```
benshimmin, line 12: Given we have rows, do you think these are really _columns_ rather than _fields_? Controversial, I know.

benshimmin (commented):
This is a great piece of work @rgarner - just a few minor comments, some more for discussion than action.

app/services/export/activities_level_b.rb
rgarner: I had considered that, and they sort of are, but then the presence of the `value` `Proc` makes them more like `CELLS` at that point...

spec/services/export/activities_level_b_spec.rb
rgarner: Yeah I very briefly made this a `shared_example` so we could say `it_behaves_like "it has no ISPF-specific fields"`, but the use of the `shared_example` block was arguably worse than the repetition...

spec/services/export/activities_level_b_spec.rb
rgarner: I'm not sure I agree with this one, primarily because it's in an ISPF-specific example – they're all in fund-specific examples. The reason the expectation exists is as a checksum because we're using `match a_hash_including` which by definition only compares a subset and doesn't cope with completeness (but you can't use `eq` because then the matchers like `a_string_starting_with` don't work).

I can agree that the *distance* between that number and the title of the example could become smaller, by swapping the order of the expectations?

benshimmin: After discussion: let's just ditch the `length` expectation altogether.

benshimmin (approved):
_Great_ work @rgarner 👍

PR #2527, opened 2025-02-06
Migrate to Rails 7.1 defaults

rgarner:
## ~This is "Don't Merge Yet" because:~

~The [Rails 7.1 update](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2521) needs releasing with Rails 7.0 defaults first.~

## Changes in this PR

Updates to 7.1-compatible config. Most consequential:

- Make sure ActiveRecord Encryption is set up correctly (possible to set it so nobody can log in). We needed `Rails.application.config.active_record.encryption.hash_digest_class = OpenSSL::Digest::SHA256` and `Rails.application.config.active_record.encryption.support_sha1_for_non_deterministic_encryption = true`.
- YAML serialization required for `HistoricalEvent` attrs. Set that by default for the app (although Rails recommendation is we set it per attribute).
- Disable raising on read-only attribute assignment for the test environment. Possibly this is a bad call; we have one spec that causes an exception on a `belongs_to` reassignment: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/b0ba880f7cb5b5830a333340f83ba55795b902fe/spec/policies/forecast_policy_spec.rb#L172 and maybe we could rejig that instead of `config.active_record.raise_on_assign_to_attr_readonly = false` in `config/environments/test.rb`.

## Next steps


benshimmin (commented):
@rgarner This looks great - very thorough.

We have discussed `config.active_record.raise_on_assign_to_attr_readonly = false` and we think rather than setting this in `config/environments/test.rb` so it applies to all specs, let's just set this for the one spec where it's necessary (on the forecast policy); we think either manually setting and unsetting the config variable or mocking it in an `around` block will probably do the trick.

benshimmin (approved):
Clever solution to the `readonly_attribute` issue - really nice! 👍

PR #2528, opened 2025-02-07
Changes to how benefitting countries are displayed in IATI XML

benshimmin:
Due to a change in DSIT policy, activities where multiple benefiting countries have been reported should only display the associated region of those countries. We use the existing `Activity#benefitting_region` method in the IATI XML now rather than iterating over the benefitting countries and assigning percentages.

We also patch up the specs where previously we were expecting multiple countries to be shown in the XML.

## Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 [Full changelog][unreleased]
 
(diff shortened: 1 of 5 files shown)


rgarner (approved):
🔴 🖊️ 
🟢 👍🏻

PR #2532, opened 2025-02-10
Level B Exports: dynamic budget columns

rgarner:
## Next steps

### DIFF
--- a/app/services/export/activities_level_b.rb
+++ b/app/services/export/activities_level_b.rb
@@ -42,66 +42,80 @@ class Export::ActivitiesLevelB
     Field.new("ODA eligibility",                           "ALL",  -> { activity.oda_eligibility }),
     Field.new("Publish to IATI?",                          "ALL",  -> { activity.publish_to_iati }),
     Field.new("Tags",                                      "ISPF", -> { activity.tags }),
-    Field.new("Budget 2023-2024",                          "ALL",  -> { budgets_by_year[2023]&.value }),
-    Field.new("Budget 2024-2025",                          "ALL",  -> { budgets_by_year[2024]&.value }),
-    Field.new("Budget 2025-2026",                          "ALL",  -> { budgets_by_year[2025]&.value }),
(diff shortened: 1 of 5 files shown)


spec/controllers/exports_controller_spec.rb
```diff
@@ -120,8 +121,8 @@
       end
 
       it "returns a CSV of all of the exports" do
-        header = Export::ActivitiesLevelB::Row.new(fund, nil).to_a
-        expect(CSV.parse(response.body.delete_prefix("\uFEFF")).first).to match_array(header)
+        header = CSV.parse(response.body.delete_prefix("\uFEFF")).first
+        expect(header.length).to be > 1
```
rgarner, line 125: This is a smoke test that returns *some* fields. We don't really care what they are here, because caring would cost now the headers are no longer fixed.

app/services/export/activities_level_b.rb
```diff
@@ -42,66 +42,80 @@ class Export::ActivitiesLevelB
...
   def rows
     activities.map do |activity|
-      Row.new(fund: @fund, activity: ActivityCsvPresenter.new(activity)).to_a
+      Row.new(fund:, activity: ActivityCsvPresenter.new(activity), fields:).to_a
     end
   end
 
   private
 
+  BEFORE_COMMENTS = -2
```
benshimmin, line 93: I wonder if this would be _slightly_ clearer with a despicably prolix name like `BEFORE_COMMENTS_INSERTION_INDEX`?

benshimmin (commented):
It is a wonderful piece of work - the code in the export itself is extremely clear and the tests are thorough. Well done @rgarner! Just one minor comment you may or may not wish to address.

app/services/export/activities_level_b.rb
rgarner: I realise this makes me at best an amateurlix.

benshimmin (approved):
🚀

PR #2535, opened 2025-02-10
Fix IATI XML country export where there is only one benefitting country

benshimmin:
This fix slightly undoes the previous work on IATI XML country export where we merged `benefitting_countries` into one region; we were overzealous in that work, in that cases where there was only one benefitting country should have remained as they were previously, ie. rendered in the XML as a single country rather than a region.

## Next steps

### DIFF
--- a/app/views/shared/xml/_activity.xml.haml
+++ b/app/views/shared/xml/_activity.xml.haml
@@ -48,9 +48,14 @@
   - if activity.iati_scope
     %activity-scope{"code" => activity.iati_scope}/
   - if activity.benefitting_countries.present?
-    - region = activity.benefitting_region
-    %recipient-region{code: region.code, percentage: "100.0", vocabulary: "1"}
-      %narrative= region.name
+    - if activity.benefitting_countries.size > 1
+      - region = activity.benefitting_region
+      %recipient-region{code: region.code, percentage: "100.0", vocabulary: "1"}
+        %narrative= region.name
+    - else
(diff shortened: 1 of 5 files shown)


app/views/shared/xml/_activity.xml.haml
```diff
@@ -48,9 +48,14 @@
...
+      - region = activity.benefitting_region
+      %recipient-region{code: region.code, percentage: "100.0", vocabulary: "1"}
+        %narrative= region.name
+    - else
+      - country = first_benefitting_country(activity.benefitting_countries)
+      %recipient-country{"code" => country.code, percentage: "100.0"}
+        %narrative= country.name
   - elsif activity.recipient_country?
     %recipient-country{"code" => activity.recipient_country}
       %narrative= country_name_from_code(activity.recipient_country)
```
rgarner, line 61: What happens when there are 0 benefitting countries? Should we have a spec context for this? The inclusion validator is `if: :requires_benefitting_countries?` and this only seems to apply when

```  
  def requires_benefitting_countries?
    !is_non_oda?
  end
```

benshimmin: We do actually have a spec for just that, see here: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/spec/features/beis_users_can_view_an_activity_as_xml_spec.rb#L98-L106

rgarner (approved):
🦸🏻

PR #2538, opened 2025-02-11
Upgrade to `devise-two-factor` 6.x

rgarner:
## Upgrade to Devise 6.x

We do not call any of `save!`, `consume_otp!` or `invalidate_otp_backup_code!` ourselves, so this upgrade has been checked with the `DeveloperClient` from the first commit and found to present no issues.

We should still manually test on staging following merge to `develop`.

From https://github.com/devise-two-factor/devise-two-factor/blob/main/UPGRADING.md#upgrading-from-5x-to-6x :

> ## save!
>
> `consume_otp!` and `invalidate_otp_backup_code!` now call `save!` instead
> of `save` (or nothing at all in the case of
> `invalidate_otp_backup_code!`). If you manually called `save`/`save!` after
> calling `invalidate_otp_backup_code!` you may be able to remove it.

## Fixes
https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/security/dependabot/104

## Next steps


benshimmin (approved):
Very nice instrumentation here @rgarner! Will be great to see the back of updates on the `devise-two-factor` gem!

PR #2540, opened 2025-02-12
Warn when creating/editing a user where their organisation is set to DSIT

benshimmin:
We add a further sprinkling of JavaScript to the app to raise a `confirm` modal, with appropriate warning messaging, if a new user is being created, or an existing user is being edited, with a non-whitelisted email address (ie. one not ending @odamanagement.org or @dsit.gov.uk), where their organisation is being set to DSIT (or being set *from* something else *to* DSIT in the case of an editing an existing user).

## Next steps


app/javascript/src/warn-dsit-organisation-change.js
```diff
@@ -0,0 +1,40 @@
...
+  Show a standard JavaScript `confirm` modal with a warning message if a new
+  user is being created whose email address is not in the DSIT whitelist and
+  their organisation is being set to DSIT, or when an existing user is being
+  edited whose email address is not in the DIST whitelist and their
+  organisation is being set to DSIT from a different organisation.
+*/
+document.addEventListener("DOMContentLoaded", function() {
+  const VALID_DSIT_DOMAINS = ["dsit.gov.uk", "odamanagement.org"];
+  const DSIT_ORGANISATION_INDEX = 0;
+  const WARNING = "Are you sure you wish to add this user's primary organisation as DSIT?";
```
rgarner, line 11: We should probably say why this is a warning to give a clue as to what to do next. For example
```suggestion
  const WARNING = "This user does not appear to be from dsit.gov.uk or odamanagement.org. Are you sure you wish to add this user's primary organisation as DSIT?";
```

rgarner, line 11: This is the first literal string in JS (🎉); all other user-facing form strings in the UI are [localised in `config/locales/models/user.en.yml`](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/config/locales/models/user.en.yml#L15-L46). To keep our 100% streak of "all user UI strings together", perhaps we can modify the user form declaration to something like
 ```
= form_with model: @user, class: "user-form", data: { :"warn-on-non-dsit" => t("form.user.modal.warn_on_non_dsit") } do |f|
```

```
  form:
    user:
      active: ...
      modal:
        warn_on_non_dsit: "This user does not appear to be from dsit.gov.uk or odamanagement.org. Are you sure you wish to add this user's primary organisation as DSIT?"
```

and pull from the form's `dataset`?

Bonus points (but completely not necessary) for putting the valid DSIT domains in `data` too.

File: app/javascript/src/warn-dsit-organisation-change.js [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+
+  const userForm = document.querySelector("form.user-form");
```
rgarner, line 13: The `class` added for this purpose doesn't have a CSS counterpart. To use the subjunctive, *were* we to move things into translations via a `data: { ... }` hash, perhaps we could also use `form[data=warn-on-non-dsit]` here and remove the class from `user/_form.html.haml` to make it clear that there aren't any visual CSS classes to look for and this is pure behaviour?

app/javascript/src/warn-dsit-organisation-change.js
```diff
@@ -0,0 +1,40 @@
...
+
+  if (!userForm) return;
+
+  const checkEmail = email => VALID_DSIT_DOMAINS.includes(email.split("@").pop());
+  const getEmail = () => document.querySelector("#user-email-field").value;
+  const getOrganisation = () => document.querySelector("#user_organisation_id").selectedIndex;
```
rgarner, line 19: Could this be `getOrganisationIndex`? And could these latter two – which are called once – be inlined to L25:26 below?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 [Full changelog][unreleased]
 
+- Show a warning when a user is being assigned to DSIT
```
rgarner, line 7:
```suggestion
- Show a warning when a user not from DSIT/ODA Management is being assigned to DSIT
```

rgarner (commented):
Nice behaviour, nice specs (particularly the block form, which is lovely and communicative)

Some comments, probably the most important bit of which is a change in the actual user-facing text to make clear why they're being shown the modal. It'd also be good keep the translations together.

app/javascript/src/warn-dsit-organisation-change.js
rgarner: I'm incorrect here, they're called twice.

app/views/users/_form.html.haml
```diff
@@ -1,4 +1,4 @@
-= form_with model: @user do |f|
+= form_with model: @user, class: "user-form", data: { :"warn-on-non-dsit" => t("form.user.modal.warn_on_non_dsit"), :"domains" => t("form.user.modal.domains") } do |f|
```
rgarner, line 1: Are we OK to drop `class: "user-form"` now?

PR #2548, opened 2025-02-13
Upgrade to Ruby 3.4.1

rgarner:
Also upgrade parser, which continues to emit:

```
warning: parser/current is loading parser/ruby34, which recognizes 3.4.0-compliant syntax, 
but you are running 3.4.1. 
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```

## Next steps


benshimmin (approved):
Looking very good @rgarner - nice simple one. Let's stick a `CHANGELOG` entry in for the sake of good form.

PR #2550, opened 2025-02-14
Turn off Rails 7.1 callback checks in dev

rgarner:
We *should* do this in dev, but can't at present because of

```
AbstractController::ActionNotFound at /home
The index action could not be found for the :verify_policy_scoped
callback on HomeController, but it is listed in the controller's
:only option.
```

In order to turn it on in dev, we need to do a bit of review/refactoring around when callbacks are called. At present, we allow them to fail silently when not applicable. This isn't a behaviour Rails 7.1 and up encourages, but we're on our last day of our last sprint and can't fix this now.

## Next steps


benshimmin (approved):
Oops! Approved 👍

PR #2551, opened 2025-02-14
Enable JS sourcemap generation

rgarner:
Started by trying to set `sourcemap: true` in `rollup.config.js`. No dice, just a lot of `(!) Broken sourcemap` warnings. Read a bit about Rollup and plugin chains. Determined (by removing terser) that it – and the 2022 version of it – were the problem.

used `yarn upgrade-interactive --latest` and upgraded only internal build packages (and the low-risk rails-ujs).

Sourcemaps are now being generated.

## Next steps


benshimmin (approved):
LGTM @rgarner 👍

PR #2567, opened 2025-03-19
Alter the maximum age of an activity start date

shuldt:
Ticket [20715](https://dxw.zendesk.com/agent/tickets/20715) flagged two activity start dates that did not pass validation as they were more than 17 years in the past. 

We agreed with the client to add 10 years to this validation range, ie. to allow planned and actual activity start dates up to 27 years in the past. This will allow the resolution of this ticket, and also help to support the service as it matures and is more likely to need historical activity records added that exceed the previous 17 year limit. 

## Next steps

### DIFF
--- a/app/validators/date_within_boundaries_validator.rb
+++ b/app/validators/date_within_boundaries_validator.rb
@@ -1,5 +1,5 @@
 class DateWithinBoundariesValidator < ActiveModel::EachValidator
-  MIN = 17
+  MIN = 27
   MAX = 25
 
   def validate_each(record, attribute, value)
--- a/app/validators/date_within_boundaries_validator.rb
+++ b/app/validators/date_within_boundaries_validator.rb
@@ -1,14 +1,14 @@
 class DateWithinBoundariesValidator < ActiveModel::EachValidator
-  MIN = 27
-  MAX = 25
+  MAX_YEARS_AGO = 27
+  MAX_YEARS_IN_FUTURE = 25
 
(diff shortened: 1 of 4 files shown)


app/validators/date_within_boundaries_validator.rb
```diff
@@ -1,5 +1,5 @@
 class DateWithinBoundariesValidator < ActiveModel::EachValidator
-  MIN = 17
+  MIN = 27
```
edavey, line 2: is this perhaps an opportunity to make these `MIN | MAX` terms a little clearer?

In reality I think they are both maximums -- in different directions -- and that the use of "min" (especially a MIN which is larger than the MAX) is inherently confusing. If we were to change these constants to allow us to say something like:

```rb
unless value.between?(MAX_YEARS_AGO.years.ago, MAX_YEARS_FROM_NOW.years.from_now)
```

would that be an improvement?

Feel free to ignore if you think it's not worth the candle.

edavey (approved):
Thanks for the super-clear narrative. I've left a comment around the naming of the constants but it's very much just a suggestion....

app/validators/date_within_boundaries_validator.rb
shuldt: Thanks for this Ed, and yes I do agree. I'll make this tweak as part of this PR.

edavey (approved):
👍 nice, I think that's clearer now

PR #2581, opened 2025-04-11
Fix rails ujs links not working

jkempster34:
## Fix Rails UJS links not working

@rails/ujs is used to transform links to enable deletes. It creates a hidden
form and submits the request as a POST request with a hidden field
_method=delete.

A recent change means this is not working, users therefore cannot perform
important actions.

Calling `start` isn't supposed to be necessary, but seems to fix the issue.

We may need to do a review of these links to see if we can use an improved
method vs Rails UJS.

## Next steps

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -12,3 +12,5 @@ document.addEventListener("DOMContentLoaded", () => {
   GOVUKFrontend.initAll();
   initTableTreeView();
 });
+
+Rails.start()


rjw1 (approved):

yndajas:
> We may need to do a review of these links to see if we can use an improved method vs Rails UJS.

For reference if this gets picked up at some point:
- https://guides.rubyonrails.org/working_with_javascript_in_rails.html#replacements-for-rails-ujs-functionality
- https://github.com/rails/rails/pull/50535
- (as above, but the merged commit) https://github.com/rails/rails/commit/8397eb24daab954e220003fa4cb1e784de6a471a

jkempster34:
> > We may need to do a review of these links to see if we can use an improved method vs Rails UJS.
> 
> For reference if this gets picked up at some point:
> 
> * https://guides.rubyonrails.org/working_with_javascript_in_rails.html#replacements-for-rails-ujs-functionality
> * [Rails UJS has been deprecated since Rails 7, time to die rails/rails#50535](https://github.com/rails/rails/pull/50535)
> * (as above, but the merged commit) [rails/rails@8397eb2](https://github.com/rails/rails/commit/8397eb24daab954e220003fa4cb1e784de6a471a)

I can't find it unfortunately, but I thought there was an established GOV.UK pattern of having a "confirm deletion" page which might be a bit nicer than the JavaScript confirm dialogs.

PR #2601, opened 2025-06-13
Clean up

yndajas:
## Changes in this PR

- Update schema to Rail 8
- Fix two typos
- Remove duplicate CHANGELOG entry


db/schema.rb
```diff
@@ -10,10 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2025_02_04_092602) do
+ActiveRecord::Schema[8.0].define(version: 2025_02_04_092602) do
```
rjw1, line 13: Schema.rb is just used when bringing up stuff locally now. since migrations would have done this in deployed environments?

PR #2608, opened 2025-06-26
Update Partner Organisation IDs for specified Activities

jkempster34:
> [!NOTE]  
> This must be run after we have deleted the activities in https://dxw.zendesk.com/agent/tickets/20963.
> Confirm that that has been done before merging

We have been provided with a list of Activities where the Partner Organisation ID requires correction.

Refer to [Zendesk ticket #20965](https://dxw.zendesk.com/agent/tickets/20965) for full details.


db/data/20250626105226_change_partner_org_ids.rb
```diff
@@ -0,0 +1,49 @@
...
+  {roda_id: "NODA-ISPF-BBSRC-Y29QRFJ-W6HY3U9-KWNTRA9", new_partner_organisation_id: "BB/Y513799/1"},
+  {roda_id: "NODA-ISPF-BBSRC-Y29QRFJ-W6HY3U9-9LUMKHZ", new_partner_organisation_id: "BB/Y514032/1"},
+  {roda_id: "NODA-ISPF-BBSRC-Y29QRFJ-W6HY3U9-HAE33XP", new_partner_organisation_id: "BB/Y513830/1"},
+  {roda_id: "NODA-ISPF-BBSRC-Y29QRFJ-W6HY3U9-J5BNYC6", new_partner_organisation_id: "BB/Y514056/1"},
+  {roda_id: "NODA-ISPF-BBSRC-Y29QRFJ-W6HY3U9-SPAK5UF", new_partner_organisation_id: "BB/Y513854/1"}
+]
+
+ActiveRecord::Base.transaction do
+  changes.each do |change|
+    activity = Activity.find_by!(partner_organisation_identifier: change.fetch(:existing_partner_organisation_identifier))
```
yndajas, line 40: Do you mean to find by `change[:roda_id]`?

jkempster34:
> This looks good enough to me, but if we wanted to make it really robust and easier in future, we could extract some of [Ed's PR](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/develop/db/data/20240919000000_change_partner_org_ids.rb) to reuse the classes there, we could then pass in the `changes` each time. What do you think? I know time is a little short, so maybe this is something we can tackle in future.

I agree, something reusable might be nice. Some of those checks might be redundant here though, such as verify_that_activity_has_expected_po_org_id as we don't have the current PO ID

PR #2611, opened 2025-07-01
Move non-breaking space outside link

yndajas:
## Changes in this PR

This fixes the script so that the non-breaking space following "change your cookie settings" in the cookie banner (after accepting cookies) is not part of the link

### Before

image

### After

image

## Next steps


jkempster34 (approved):
That's some impressive attention to detail!

PR #2657, opened 2026-02-04
Ensure Actuals / Refunds are compatible with Report & Activity, "ODA-wise"

edavey:
See [Zendesk 21429](https://dxw.zendesk.com/agent/tickets/21429)

In this PR we make 2 fixes:

## 1. Ensure that a report's transactions have compatible Activity and Report associations

An ODA report should only contain ODA actuals and ODA refunds and conversely a Non-ODA report should only contain non-ODA actuals and refunds.

There's a preliminary task for both Activity and Refund models of overriding the built-in `#is_oda?` predicate, which returns false when the attribute is nil. This isn't what we want, as:

- `Report#is_oda` is only set when the fund is ISPF. For reasons which are not clear to me.
- `Activity#is_oda` is not always set

In both cases:

- there may be historic activities and reports which lack this value as they were created before the column was introduced
- the only fund which has Non-ODA activities (and therefore reports) is ISPF, so a `nil` value indicates that the activity/report **is ODA**.

### Uploading an incompatible actual

roda_upload

## 2. Ensure a report's downloaded transactions match those shown in UI

This work originates in ticket [Zendesk 21429](https://dxw.zendesk.com/agent/tickets/21429) in which it appeared that certain actuals had been lost because they were missing from the downloaded version of a report. In this case we had a non-ODA report which included ODA actuals:
  
1. these ODA actuals were **shown** in the UI view of the report  
  
2. these ODA actuals were **omitted** from the CSV download of the report

In this PR we align the mechanisms used for the UI and download so that this would not happen in future. It should now be impossible for this mix of ODA and non-ODA transactions to occur. It's a worthwhile exercise in consistency in any case as this should reduce the chance of similar discrepancies in future.

### Viewing a report with mixed ODA and Non-ODA actuals (BEFORE)

report_ui_before

### Viewing a report with mixed ODA and Non-ODA actuals (AFTER)

report_ui_after

## Next steps

### DIFF
--- a/app/models/refund.rb
+++ b/app/models/refund.rb
@@ -7,6 +7,8 @@ class Refund < Transaction
 
   validates_associated :comment
 
+  validate :compatibility_of_report_and_activity_oda_wise
+
   def value=(amount)
     big_decimal = begin
       BigDecimal(amount)
@@ -14,5 +16,14 @@ class Refund < Transaction
       return
     end
     write_attribute(:value, -big_decimal.abs)
+  end
+
+  private
+
+  def compatibility_of_report_and_activity_oda_wise
+    return unless report && parent_activity
+    return if report.is_oda? == parent_activity.is_oda?
+
+    errors.add(:base, "A non-ODA report can not include ODA refunds, and vice-versa")
   end
 end
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -129,11 +129,15 @@ class Report < ApplicationRecord
   end
 
   def summed_actuals
-    actuals.sum(&:value)
+    actuals.joins(:parent_activity)
+      .where("activities.id": reportable_activities.pluck(:id))
+      .sum(&:value)
   end
 
   def summed_refunds
-    refunds.sum(&:value)
+    refunds.joins(:parent_activity)
+      .where("activities.id": reportable_activities.pluck(:id))
+      .sum(&:value)
   end
 
   def summed_forecasts_for_reportable_activities
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -75,6 +75,10 @@ class Report < ApplicationRecord
     super
   end
 
+  def is_oda?
+    is_oda != false
+  end
+
   def editable?
     state.in?(EDITABLE_STATES)
   end
--- a/spec/features/users_can_upload_actuals_spec.rb
+++ b/spec/features/users_can_upload_actuals_spec.rb
@@ -129,6 +129,47 @@ RSpec.feature "users can upload actuals" do
         }
       ])
     end
+
+    context "and we attempt to upload an actual associated with an ODA activity" do
+      let(:oda_project) do
+        create(
+          :project_activity,
+          :ispf_funded,
+          organisation: report.organisation,
+          is_oda: true,
+          source_fund_code: Fund.by_short_name("ISPF").id,
+          title: "ODA activity"
+        )
+      end
+
+      let(:non_oda_project) do
+        create(
+          :project_activity,
+          :ispf_funded,
+          organisation: report.organisation,
+          is_oda: false,
+          source_fund_code: Fund.by_short_name("ISPF").id,
+          title: "Non-ODA activity"
+        )
+      end
+
+      scenario "see an error explaining that ODA actuals can not be included in a non-ODA report" do
+        oda_project_id = oda_project.roda_identifier
+        non_oda_project_id = non_oda_project.roda_identifier
+
+        upload_csv <<~CSV
+          Activity RODA Identifier | Financial Quarter | Financial Year | Actual Value | Refund Value | Receiving Organisation Name | Receiving Organisation Type | Receiving Organisation IATI Reference | Comment
+          #{oda_project_id}        | 1                 | 2020           | 5            |              |                             |                             |                                       |
+          #{non_oda_project_id}    | 2                 | 2020           | 6            |              |                             |                             |                                       |
+        CSV
+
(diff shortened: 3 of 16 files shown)


spec/models/activity_spec.rb
```diff
@@ -2430,6 +2430,35 @@ def create_report(organisation, fund)
     end
   end
 
+  describe "#is_oda? (it may not be obvious that only non-ODA actuals " \
```
yndajas: **question:** do you mean "only ISPF activities" rather than "only non-ODA actuals"?

spec/models/refund_spec.rb
```diff
@@ -37,6 +37,106 @@
...
+          :refund,
+          report: non_oda_report
+        )
+      end
+
+      context "and the associated ACTIVITY is ODA" do
+        before { refund.parent_activity = oda_activity }
+
+        it "is not valid" do
+          refund.valid?
```
yndajas: **suggestion:** might be nice to include an expectation on this method too (even though the error is caused by it being invalid)?

spec/models/refund_spec.rb
```diff
@@ -37,6 +37,106 @@
...
   end
 
+  describe "compatibility of associated report and activity 'ODA-wise'" do
+    let(:oda_report) do
+      create(
+        :report,
+        :active,
+        :for_gcrf,
+        description: "ODA report",
+        is_oda: true
```
yndajas: **question:** the `is_oda` value would end up as `nil` here regardless of what you pass to the factory, right?

app/models/refund.rb
```diff
@@ -15,4 +17,13 @@ def value=(amount)
...
     write_attribute(:value, -big_decimal.abs)
   end
+
+  private
+
+  def compatibility_of_report_and_activity_oda_wise
+    return unless report && parent_activity
+    return if report.is_oda? == parent_activity.is_oda?
+
+    errors.add(:base, "A non-ODA report can not include ODA refunds, and vice-versa")
```
yndajas, line 27: **suggestion:** "refunds for ODA activities"?

spec/features/users_can_upload_actuals_spec.rb
```diff
@@ -129,6 +129,47 @@ def expect_to_see_successful_upload_summary_with(count:, total:)
         }
       ])
     end
+
+    context "and we attempt to upload an actual associated with an ODA activity" do
```
yndajas, line 133: **question:** is there an equivalent refund feature spec worth updating?

app/models/report.rb
```diff
@@ -129,11 +129,15 @@ def forecasts_for_reportable_activities
   end
 
   def summed_actuals
```
yndajas, line 131: **issue:** in the commit message you mention the divergent approaches to retrieving actuals but you describe both as being for the UI. Presumably you meant CSV for one of them?

yndajas (commented):
Makes sense to me. A few minor comments

spec/models/activity_spec.rb
edavey: yes, thank you!

spec/models/refund_spec.rb
edavey: 🤷 I've added descriptions on the return values of `#valid?` to add/reinforce clarity...

spec/models/refund_spec.rb
edavey: yes... 

thks for drawing attention to this. I'm going to try to improve the clarity by stubbing `Report#is_oda?` and `Activity#is_oda?` here...

app/models/refund.rb
edavey: hmm, I know that "refund for ODA activity" is literally true and more precise  -- and that a refund "is ODA" only through its association with an activity -- but I believe it's helpful to try to reinforce the language of "ODA Actual/Refund" and "Non-ODA Actual/Refund" and in fact we see this language in use now in our support discussions.

spec/features/users_can_upload_actuals_spec.rb
edavey: hmm, we can't actually describe the equivalent tidy behaviour for the refund, without some refactoring.

The `CreateRefund` class works a little differently from the `CreateActual` class. `CreateRefund` uses the `Report::editable_for_activity(activity)` finder to look for reports which are compatible with the given activity. (it asks whether the organisation and fund match, **and whether `Report#is_oda` matches `Activity#is_oda`**) If there's a mismatch and no compatible report is available, then we raise an error. So, in the case of an attempt to upload an ODA refund to a non-ODA report, the user won't see the precise error message. 

I've updating the `CreateRefund` unit test to include a mention of the ODA-mismatch use-case  / scenario.

I think this behaviour is adequate (we have seen zero incidences of the problem involving Refunds, our problem report transactions have all been Actuals).

app/models/report.rb
edavey: thanks for spotting that rather confusing slip!

edavey:
@yndajas thanks a lot for your careful and very helpful review. I think I've now addressed or responded to all points, so ready for another look when you have a chance. 🙇

yndajas (approved):

edavey:
I'm going to close this now and replace it with 2 PRs with the exact same contents, as we actually want to release this work in 2 parts:

1. https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2658
2. https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/pull/2659

